PR #20, opened 2017-09-22
Add flow-typed, resolve flow errors

wolverineks:
"flow-typed" is a central repository of flow types.
It installs type definitions for third party libraries
that use flow and have submitted type definitions.


thehobbit85 (changes requested):
remove our flowed default types

PR #162, opened 2017-10-30
Implement change mining fee UI

koretskiyav:


src/modules/UI/scenes/SendConfirmation/SendConfirmationOptions.js
```diff
@@ -31,7 +33,7 @@ export default class SendConfirmationOptions extends Component {
             </Text>
```
ahujsak: The 'fee' 'help' and 'max' strings should be referencing a constant instead of being stand alone strings. 
Example Constants.FEE.   If they don't already exist , they need to be created.

src/modules/UI/scenes/ChangeMiningFee/ChangeMiningFee.ui.js
```diff
@@ -0,0 +1,68 @@
...
+  hendlePress = (feeSetting) => this.setState({ feeSetting, fee: '' });
+  hendleChange = (fee) => this.setState({ fee });
+
+  render () {
+    const { feeSetting, fee } = this.state
+
+    return (
+      <View style={style.wrapper}>
+        <Text style={style.header} >
+          Warning: Low Fees may cause long delays in transaction confirmation
```
wolverineks: When displaying text, please use 
`import strings from '...locales/default'`
...
`strings.enUS[ {{string_key}} ]`

If the desired text does not exist in the strings file, please create it

wolverineks:
Please style `ChangeMiningFee.ui` so that it does not cover the header

koretskiyav:
@wolverineks I do not quite understand what is meant by

> does not cover the header

Can you describe in more detail what exactly needs to be done?

wolverineks:
@koretskiyav Please, disregard the request to fix the scene covering the header. That was a header bug, not a `Change Mining Fee` scene bug

src/modules/UI/scenes/ChangeMiningFee/ChangeMiningFee.ui.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  componentWillUnmount () {
+    if (this.state.fee !== this.props.fee || this.state.feeSetting !== this.props.feeSetting) {
+      this.props.onSubmit(this.state.feeSetting, this.state.fee)
+    }
+  }
+
+  handlePress = (feeSetting) => this.setState({ feeSetting, fee: '' });
+  handleChange = (fee) => this.setState({ fee });
```
wolverineks: Please filter fees to be numeric strings only, e.g `"123"` = good, `"123[;.sdfv"` = bad
One possible way of doing that:
```
const regex = /\D/g
const filteredFee = fee.replace(/\D/g, '')
this.setState({ fee: filteredFee })
```

wolverineks:
Please add type checking to any files you create
1. add `// @flow` to the top of the file
2. Eliminate all flow errors

wolverineks:
We'd like the `Change Mining Fee` scene to be used to update the state on multiple scenes, `Exchange` and `SendConfirmation`.

If the previous scene was the `Exchange` scene, have the `Change Mining Fee` scene update the `Exchange` state.
If the previous scene was the `Send Confirmation` scene, have the `Change Mining Fee` scene update the `Send Confirmation` state.

One way of doing this could be to pass an `onSubmit` callback action into the scene via `Actions.changeMiningFee({onSubmit: changeFee})` and call that action from the `Change Mining Fee` scene

koretskiyav:
@wolverineks do I understand the flow correctly: the user exhibiting a fee from different scenes receives independent windows? That is, if I set a `custom` fee out of `Exchange` scene. Then I'll go in through the `Send Confirmation` scene, then I should see the `default` fee? Here I will set a `high` fee. And going back through `Exchange` scene there steel will be a `custom` fee?

If all is correct, then as i see, we have different structure at  `src/modules/UI/scenes/SendConfirmation/reducer.js` and `src/reducers/CryptoExchangeReducer.js`. Namely `{ feeSatoshi: number }` and `{ fee: '', feeSetting: Constants.STANDARD_FEE }`. What should be set in `Send Confirmation` reducer when selecting, for example, `hight` fee?

Otherwise, if you set fee once from any scene, and the user sees it from the another scene, then we just need one reducer (or as now different reducers that both can react to one action).

paullinator:
@koretskiyav You will need two reducers, one for each scene. The setting should actually be temporary. it should only allow the fee to be set for that one transaction on that one scene. Once the transaction goes through, the setting should go back to default.

koretskiyav:
I separate these two flows, now there are two new fields in the reducer `src/modules/UI/scenes/SendConfirmation/reducer.js`: `{ fee: string, feeSetting: string }`.

wolverineks:
@koretskiyav 
We've decided to postpone adding a `Custom` option.
Please remove the `Custom` option from the UI.

PR #176, opened 2017-11-05
Fix TouchID popup occurring on login

paullinator:
Only remount the Login.ui scene when logging out. Check if nextProps.account.username is undefined as that will be the case upon logout but not upon login

### DIFF
--- a/src/modules/UI/scenes/Login/Login.ui.js
+++ b/src/modules/UI/scenes/Login/Login.ui.js
@@ -20,8 +20,10 @@ class Login extends Component {
 
   componentWillReceiveProps (nextProps) {
     // If we have logged out, destroy and recreate the login screen:
-    if (this.props.account && nextProps.account !== this.props.account)
-      this.setState({key: this.state.key + 1})
+    if (this.props.account && (nextProps.account !== this.props.account))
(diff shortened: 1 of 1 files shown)


src/modules/UI/scenes/Login/Login.ui.js
```diff
@@ -20,8 +20,10 @@ class Login extends Component {
 
   componentWillReceiveProps (nextProps) {
     // If we have logged out, destroy and recreate the login screen:
-    if (this.props.account && nextProps.account !== this.props.account)
-      this.setState({key: this.state.key + 1})
+    if (this.props.account && (nextProps.account !== this.props.account))
+      if (typeof nextProps.account.username === 'undefined') {
+        this.setState({key: this.state.key + 1})
+      }
   }
```
swansontec: This would be simpler:

```js
if (
  this.props.account &&
  nextProps.account !== this.props.account &&
  nextProps.account.username === undefined
) {
  this.setState({ key: this.state.key + 1 })
}
```

Using `typeof x.y === 'undefined'` is unnecessary complication. Just do `x.y === undefined`.

swansontec:
This can be merged right away, since the logic seems solid. Cleaning up the code is optional.

PR #177, opened 2017-11-05
Minimize debug logging to reduce # of lines spewed on callbacks.

paullinator:
Remove logging of all transactions for onTransactionsChanged and onNewTransactions. Only log number of transactions and the txids.
Minimize debug logs for callbacks to a single line vs outputting an object. Saves many lines of debug spew

### DIFF
--- a/src/modules/Core/Account/callbacks.js
+++ b/src/modules/Core/Account/callbacks.js
@@ -24,40 +24,37 @@ const makeAccountCallbacks = (dispatch: Dispatch): AbcAccountCallbacks => {
     },
 
     onAddressesChecked (walletId: string, progressRatio: number) {
-      const callbackDetails = {
-        walletId,
-        progressRatio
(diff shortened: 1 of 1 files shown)


paullinator:
@wolverineks Can I get a review on this pls?

wolverineks:
Which debugger are you targeting with this PR?
React Native Debugger automatically collapses objects, so currently each callback is a single line
Strings are no collapsed, so with this PR `onNewTransactions` and `onTransactionsChanged`  take up multiple lines
Additionally, we lose syntax highlighting in React Native Debugger if we log strings instead of objects.
I find the logs harder to read and prefer to keep things the way they are.

paullinator:
I'm optimizing for logging to a file which is our  goal. As we discussed in a previous engineering meeting, logging full objects should not be done but only the actual useful values.

PR #178, opened 2017-11-06
fix flash

koretskiyav:


wolverineks:
@koretskiyav why does this PR use `react-native-torch` instead of `react-native-camera`?
was there a bug in that library?

koretskiyav:
@wolverineks the native camera, as far as I understand, can not turn on the flash. It can use flash only at the moment of making photo / video. To enable the flash permanently, you need to use the native RCTTorch module. This library almost no js code, it just connects native module.

koretskiyav:
@wolverineks my bad, did not pay attention to this possibility. Created new PR: https://github.com/Airbitz/edge-react-gui/pull/183/files

PR #179, opened 2017-11-06
Enable torch on Scan scene

wolverineks:


wolverineks:
This PR uses `react-native-camera`'s build in `torchMode`
#179 also enables `torchMode` but using a different library (`react-native-torch`)
please merge one or the other, but not both

paullinator:
Kevin, please verify this works on Android and go ahead and merge the PR.

PR #180, opened 2017-11-07
Wallet modal text fix

ahujsak:
this pull request fixed the incorrect header when selecting a wallet in the exchange scene. Also, some styling work was done in order to have the drop down button down carrot align properly.
https://app.asana.com/0/361770107085503/317312547827506


ahujsak:
Yes, I looked where the component is being used. Right now, even thought it is a common component, it has not replaced the wallet selector button yet. Also, I did test to confirm the messages are consistent with the wallet being chosen

paullinator:
Awesome. merged.

PR #200, opened 2017-11-15
fix double tapping on back button bug (go back to twice)

koretskiyav:


paullinator:
Does this leave the top left button the same as it would normally be on iOS and Android? The two have very distinct and different styles and we want to keep them correct to the platform?

paullinator:
This appears to override the button content with one style that is used on both platforms. Can you send screenshots of how this looks on iOS and Android for the TransactionList, SendConfirmation, and Settings screens.

paullinator:
@koretskiyav Any comment on this?

paullinator:
@koretskiyav Any comment on this?

paullinator:
There are now conflicts to resolve. Please address them

PR #213, opened 2017-11-17
Modify Modal to not cast square shadow on Android.

ahujsak:
This removes the weird box present on some android phones. It is a direct result of layering two round items on top of each other. The fix was to kill the gradient, go with a 2 px border and have the icon center inside that.


kylanhurt:
@ahujsak I will have to check with Paul and Reina to see if they want to kill that gradient border on iOS as well.

PR #219, opened 2017-11-21
add OS level notification

koretskiyav:


paullinator:
This PR appears to remove the original Notification code that was in the app and replaces it with LocalNotification. The goal of this task was to add the OS level notification and not replace the current one. The OS level notification should popup even if the app is in the background and place a system notification in the list of notifications of the user.

This would be a good library to use https://www.npmjs.com/package/react-native-local-notifications

paullinator:
Wrong implementation. Closing PR

PR #221, opened 2017-11-22
fix outdated users list

koretskiyav:


paullinator:
@wolverineks needs to review this. I assumed this fix would be in the Core not the GUI.

PR #232, opened 2017-11-28
Only re-render the currently displayed scene

paullinator:
Track the current scene using a `global.currencyScene` and the `onEnter` parameter of the router.
Return false in shouldComponentUpdate if the scene is not the current scene.

Brings Core callback updates to GUI down from 300ms+ to about 70ms


thehobbit85:
@paullinator  can you follow up on this branch and close it if it's not needed anymore?

thehobbit85:
@wolverineks

PR #239, opened 2017-12-07
Crypto Exchange scene modified

ahujsak:
• User must select their own wallets. No longer done by default
• Excahnge rate will not show until there are two wallets selected.
• localized strings needed


ahujsak:
found a few bugs. costing for now.

PR #241, opened 2017-12-08
Fix: No content in help modal 

vsashyn:
iOS. Tap “Help” in upper left of most screens. The modal should show help content but is blank.


src/modules/UI/components/HelpModal/style.js
```diff
@@ -23,7 +22,7 @@ export default StyleSheet.create({
     padding: 4
   },
   modalMiddleWebView: {
-
+    height: 240
```
paullinator: This shouldn't be hardcoded but instead stretch to nearly fill the full screen. Ideally only about 10dp margin from top and bottom. Same for sides.

PR #245, opened 2017-12-11
Fix Implementation of ManageTokens and AddCustomToken

kylanhurt:
The purpose of this task is to fix the implementation of AddCustomToken and ManageTokens functionality to properly reflect account and wallet-level consistency.

Asana Task: https://app.asana.com/0/361770107085503/489057571180672/f


src/modules/UI/Settings/reducer.js
```diff
@@ -28,6 +28,15 @@ export const settings = (state = initialState, action) => {
       loginStatus
     }
   }
+
+  case ACTION.SET_CUSTOM_TOKENS: {
+    const {customTokens} = data
+    return {
+      ...state,
+      customTokens
```
wolverineks: why are you storing this data in redux?

src/modules/UI/Settings/selectors.js
```diff
@@ -25,7 +25,12 @@ export const getExchangeTimer = (state: State): number => {
 
 export const getCurrencySettings = (state: State, currencyCode: string) => {
   const settings = getSettings(state)
-  const currencySettings = settings[currencyCode] || isoFiatDenominations[currencyCode]
+  let currencySettings
+  if (settings[currencyCode] || isoFiatDenominations[currencyCode]) {
+    currencySettings = settings[currencyCode] || isoFiatDenominations[currencyCode]
+  } else {
+    currencySettings = settings.customTokens.find((token) => token.currencyCode === currencyCode)
```
wolverineks: why are you treating user added tokens different that the default enabled tokens?

src/modules/UI/Settings/selectors.js
```diff
@@ -37,13 +42,20 @@ export const getDenominations = (state: State, currencyCode: string) => {
...
 export const getDisplayDenominationKey = (state: State, currencyCode: string) => {
   const settings = getSettings(state)
-  const currencySettings = settings[currencyCode]
-  const selectedDenominationKey = currencySettings.denomination
+  let currencySettings, selectedDenominationKey
+  if (settings[currencyCode]) {
+    currencySettings = settings[currencyCode]
+    selectedDenominationKey = currencySettings.denomination
+  } else {
+    let tokenInfo = settings.customTokens.find((token) => token.currencyCode === currencyCode)
```
wolverineks: why not just call this currencySettings?

src/modules/UI/Settings/selectors.js
```diff
@@ -37,13 +42,20 @@ export const getDenominations = (state: State, currencyCode: string) => {
...
+    currencySettings = tokenInfo
+    selectedDenominationKey = currencySettings.multiplier
+  }
   return selectedDenominationKey
 }
 
 export const getDisplayDenominationFromSettings = (settings: any, currencyCode: string) => {
-  const currencySettings = settings[currencyCode] || isoFiatDenominations[currencyCode]
+  const currencySettings = settings[currencyCode] || isoFiatDenominations[currencyCode] || settings.customTokens[currencyCode]
```
wolverineks: why are user added tokens different than all the other currency settings?

src/modules/UI/Wallets/action.js
```diff
@@ -74,54 +73,57 @@ export const upsertWallet = (wallet: AbcCurrencyWallet) => (dispatch: any, getSt
...
 
-// setEnabledTokens is specifically for enabling them *within the GUI*, not within the core
-export const setEnabledTokens = (walletId: string, enabledTokens: any) => (dispatch: any, getState: any) => {
+// adds to core and enables in core
+export const addCustomToken = (walletId: string, tokenObj: any) => (dispatch: any, getState: any) => {
+  const state = getState()
+  const wallet = CORE_SELECTORS.getWallet(state, walletId)
+  WALLET_API.addCoreCustomToken(wallet, tokenObj)
+  .then(() => {
+    return
```
wolverineks: is this necessary?

src/modules/UI/Wallets/action.js
```diff
@@ -74,54 +73,57 @@ export const upsertWallet = (wallet: AbcCurrencyWallet) => (dispatch: any, getSt
...
+  // tell Redux that we are updating the enabledTokens list
   dispatch(setTokensStart())
+  // get a snapshot of the state
   const state = getState()
+  // get a copy of the relevant core wallet
   const wallet = CORE_SELECTORS.getWallet(state, walletId)
-  WALLET_API.setSyncedTokens(wallet, enabledTokens)
-  .then((tokens) => {
+  // now actually tell the wallet to enable the token(s) in the core and save to file
+  WALLET_API.setEnabledTokens(wallet, enabledTokens, disableTokens)
```
wolverineks: please use consistent naming, `enabledTokens` is a noun, `disableTokens` is a verb

src/modules/UI/Wallets/action.js
```diff
@@ -74,54 +73,57 @@ export const upsertWallet = (wallet: AbcCurrencyWallet) => (dispatch: any, getSt
...
+  .then(() => {
     dispatch(setTokensSuccess())
     dispatch(refreshWallet(walletId))
-    return tokens
+    wallet.getEnabledTokens()
+    .then((enabledTokens) => {
```
wolverineks: this can be cleaned up
`.then(console.log)`

wolverineks: but if all it does it log the enabled tokens, then please remove it altogether

src/modules/UI/Wallets/action.js
```diff
@@ -74,54 +73,57 @@ export const upsertWallet = (wallet: AbcCurrencyWallet) => (dispatch: any, getSt
...
-      }
-    }
-    WALLET_API.enableTokens(wallet, guiEnabledTokens) // take GUI enabled tokens and enable them in the core
+    let modifiedWallet = wallet
+    modifiedWallet.enabledTokens = tokens
+    WALLET_API.enableTokens(modifiedWallet, tokens) // take GUI enabled tokens and enable them in the core
     .then(() => {
-      dispatch(upsertWallet(wallet)) // now update the wallet in Redux so that the tokensEnabled property can be used by GUI
+      dispatch(upsertWallet(modifiedWallet)) // now update the wallet in Redux so that the tokensEnabled property can be used by GUI
+      wallet.getEnabledTokens()
```
wolverineks: is this being used for anything? ive never used the return value of a dispatched action before.

src/modules/UI/components/WalletListModal/components/WalletListModalBody.ui.js
```diff
@@ -95,10 +95,10 @@ export default class WalletListModalBody extends Component<$FlowFixMeProps> {
 
     // need to crossreference tokensEnabled with nativeBalances
     let enabledNativeBalances = {}
-    const enabledTokens = guiWallet.tokensEnabled
+    const enabledTokens = guiWallet.enabledTokens
 
     for (let prop in guiWallet.nativeBalances) {
-      if ((prop !== currencyCode) && enabledTokens[prop] && enabledTokens[prop].enabled) {
+      if ((prop !== currencyCode) && (enabledTokens.indexOf(prop) >= 0)) {
```
wolverineks: if checking for the existence of a element in an array, please use [Array.prototype.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)
`[1, 2, 3].includes(1) === true`

src/modules/utils.js
```diff
@@ -82,6 +82,21 @@ export const inputBottomPadding = () => {
   }
 }
 
+// will take the metaTokens property on the wallet (that comes from currencyInfo), merge with account-level custom tokens added, and only return if enabled (wallet-specific)
```
wolverineks: there are libraries capable of doing this for you
we already use Lodash, so feel free to use anything that library provides
[_.intersectionWith](https://lodash.com/docs/4.17.4#intersectionWith) would likely do the trick
```
var objects = [{ 'x': 1, 'y': 2 }, { 'x': 2, 'y': 1 }];
var others = [{ 'x': 1, 'y': 1 }, { 'x': 1, 'y': 2 }];
 
_.intersectionWith(objects, others, _.isEqual);
// => [{ 'x': 1, 'y': 2 }]
```

src/modules/Core/Wallets/api.js
```diff
@@ -100,64 +100,52 @@ export const addCoreCustomToken = (wallet: AbcCurrencyWallet, tokenObj: any) =>
...
+      currentTokens.push(token)
+      setEnabledTokens(wallet, currentTokens)
+      return currentTokens
+    }
+    return currentTokens
+  })
+}
+
+export async function setEnabledTokens (wallet: AbcCurrencyWallet, tokens: Array<string>, tokensToDisable?: Array<string>) {  // initialize array for eventual setting of file
+  let finalText = []
```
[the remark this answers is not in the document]
paullinator: I would not *avoid* declaring the type in the variable name. Just make sure it's accurate. The above should therefore be: finalTextArray

src/modules/Login/action.js
```diff
@@ -61,15 +61,20 @@ const loadSettings = () => (dispatch: Dispatch, getState: GetState) => {
     .then((settings) => {
       const syncDefaults = SETTINGS_API.SYNCED_ACCOUNT_DEFAULTS
       const syncFinal = {...syncDefaults, ...settings}
-
+      // const customTokens = settings ? settings.customTokens : []
```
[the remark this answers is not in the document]
kylanhurt: Yup

src/modules/UI/Settings/reducer.js
kylanhurt: They need to be stored in Redux because are account-specific but we do not want to always re-query the disklet in order to obtain them.

src/modules/UI/Settings/selectors.js
kylanhurt: If you look at the data structure for the settings in Redux, the denomination keys (eg `REP`, `WINGS`, etc) are mixed in with the rest of the top-level settings which, in my opinion, is a major issue since it can eventually cause namespace clashes with actual settings. If someone happened to add a token with a currency code like `pinMode` or `PINMODE`, there's a chance something bad could happen. So I am storing them separately and in the format that matches other currencyInfo's on the wallet objects.

src/modules/UI/Settings/selectors.js
kylanhurt: Again, hypothetically we could but I see us eventually moving away from the way the denomination data is stored in Redux's settings:
image

src/modules/UI/Settings/selectors.js
kylanhurt: Again, because a user could hypothetically enter a currency key (currencyCode) that matches one of the other settings in Redux. I realize using an array and doing multiple bitwise operations isn't the most elegant of solutions.

src/modules/utils.js
kylanhurt: I will familiarize myself with that library. I was not aware that we were already using it.

src/modules/UI/Wallets/action.js
paullinator: It shouldn't be. There's an implied `return void` if you don't .then() the Promise

src/modules/UI/Wallets/action.js
```diff
@@ -86,54 +85,56 @@ export const upsertWallet = (wallet: AbcCurrencyWallet) => (dispatch: any, getSt
...
+  WALLET_API.addCoreCustomToken(wallet, tokenObj)
+  .then(() => {
+    return
+  })
+  .catch((e) => console.log(e))
+}
+
+export const setEnabledTokens = (walletId: string, enabledTokens: Array<string>, disabledTokens: Array<string>) => (dispatch: any, getState: any) => {
+  // tell Redux that we are updating the enabledTokens list
   dispatch(setTokensStart())
```
paullinator: Seems like it's overkill to use redux for this. Same for setTokensSuccess(). The spinner is state that's specific to the component and isn't global to the app. Not to be fixed in this PR since it's not a showstopper, but please clean up after this PR is merged

src/modules/UI/scenes/AddToken/action.js
```diff
@@ -4,39 +4,87 @@ export const ADD_TOKEN_SUCCESS = 'ADD_TOKEN_SUCCESS'
...
+import {setCustomTokens} from '../../Settings/action.js'
 import * as SETTINGS_API from '../../../Core/Account/settings.js'
 import * as CORE_SELECTORS from '../../../Core/selectors.js'
 import * as UI_ACTIONS from '../../Wallets/action.js'
+import * as UI_WALLET_SELECTORS from '../../selectors.js'
 
+// adding a token should only be done on the account level, while enabling will occur on the wallet level
 export const addToken = (walletId, tokenObj) => (dispatch, getState) => {
   const state = getState()
+  // create modified object structure to match metaTokens
```
paullinator: This needs a major refactor and much of this action's logic should be in the reducer. Please clean up after this PR is merged.

src/modules/UI/scenes/ManageTokens/ManageTokenRow.ui.js
```diff
@@ -1,19 +1,28 @@
...
 } from 'react-native'
 import Text from '../../components/FormattedText'
 import CheckBox from '../../components/CheckBox'
-import {connect} from 'react-redux'
-import * as CORE_SELECTORS from '../../../Core/selectors.js'
 import styles from './style.js'
-import * as UTILS from '../../../utils.js'
 import THEME from '../../../../theme/variables/airbitz'
 
+export type State = {
```
paullinator: This should be named `ManageTokenRowState`

src/modules/UI/scenes/ManageTokens/ManageTokenRow.ui.js
```diff
@@ -1,19 +1,28 @@
...
+  enabled?: boolean
+}
+
+export type Props = {
```
paullinator: This should be named `ManageTokenRowProps`

src/modules/UI/scenes/ManageTokens/ManageTokens.ui.js
```diff
@@ -13,61 +15,91 @@ import {connect} from 'react-redux'
...
-import * as UTILS from '../../../utils.js'
 import {
   getEnabledTokens,
   setEnabledTokens
 } from '../../Wallets/action.js'
+import type { GuiWallet } from '../../../../types'
+import type {AbcMetaToken} from 'airbitz-core-types'
+
+
+export type Props = {
```
paullinator: This should be `ManageTokenProps`

src/modules/UI/scenes/ManageTokens/ManageTokens.ui.js
```diff
@@ -13,61 +15,91 @@ import {connect} from 'react-redux'
...
+  guiWallet: GuiWallet,
+  manageTokensPending: boolean,
+  accountMetaTokenInfo: Array<AbcMetaToken>
+}
 
+export type DispatchProps = {
```
paullinator: This should be `ManageTokenDispatchProps`

src/modules/UI/scenes/ManageTokens/ManageTokens.ui.js
```diff
@@ -13,61 +15,91 @@ import {connect} from 'react-redux'
...
+  getEnabledTokensList: (string) => void,
+  setEnabledTokensList: (string, Array<string>, Array<string>) => void
+}
 
-class ManageTokens extends Component {
-  constructor (props) {
+export type State = {
```
paullinator: This should be `ManageTokensState`

src/modules/UI/scenes/WalletList/WalletList.ui.js
```diff
@@ -41,14 +41,14 @@ const options = [
   },{
     value: 'sort',
     syntax: strings.enUS['fragment_wallets_sort']
+  },{
```
paullinator: Use new string syntax

src/modules/UI/scenes/WalletList/components/WalletListRow/FullWalletListRow.ui.js
```diff
@@ -14,22 +16,57 @@ import styles, {styles as styleRaw} from '../../style.js'
...
+  getEnabledTokens
 } from '../../../../Wallets/action.js'
 import * as SETTINGS_SELECTORS from '../../../../Settings/selectors'
 import platform from '../../../../../../theme/variables/platform.js'
+import type {GuiDenomination} from '../../../../../../types'
+import type {AbcMetaToken} from 'airbitz-core-types'
 const DIVIDE_PRECISION = 18
 
-class FullWalletRow extends Component {
+export type FullWalletRowProps = {
```
paullinator: Should be `FullWalletListRowProps`

src/modules/UI/scenes/WalletList/components/WalletListRow/FullWalletListRow.ui.js
```diff
@@ -14,22 +16,57 @@ import styles, {styles as styleRaw} from '../../style.js'
...
+  customTokens: Array<AbcMetaToken>,
+  sortHandlers: any
+}
+
+type InternalProps = {
+  displayDenomination: GuiDenomination,
+  exchangeDenomination: GuiDenomination
+}
+
+type DispatchProps = {
```
paullinator: Should be `FullWalletListRowDispatchProps`

src/modules/UI/scenes/WalletList/components/WalletListRow/FullWalletListRow.ui.js
```diff
@@ -14,22 +16,57 @@ import styles, {styles as styleRaw} from '../../style.js'
...
 
-class FullWalletRow extends Component {
+export type FullWalletRowProps = {
+  data: any, // TODO: Need to type this
+  sortableMode: boolean,
+  customTokens: Array<AbcMetaToken>,
+  sortHandlers: any
+}
+
+type InternalProps = {
```
paullinator: Should be `FullWalletListRowOwnProps`

paullinator:
Verified most functionality of adding custom tokens and enabling/disabling of tokens per wallet by running build on iOS simulator. Found small bug whereby custom tokens added will not show a balance when account is logged into a second device. Logout -> PIN login will then show the balance of custom tokens.

I'm approving this PR for now as we need this functionality in Beta 3. @kylanhurt Please look over all the comments posted by me and Kevin and cleanup this code for Beta 4 prior to working on Edit/Delete tokens.

kylanhurt:
@paullinator Understood. I'll go back through these comments today.

PR #247, opened 2017-12-11
Optimize selectwallet and upsertWallet

paullinator:
- Combine selectWalletId and selectCurrencyCode into one action
- Throttle calls to upsertWallet to max once per 3s


paullinator:
Verbally approved by @ahujsak. Was just waiting on Travis checks

PR #248, opened 2017-12-13
Fix autologout

wolverineks:
Move autologout to separate component
Refactor to use timestamp comparison
Unify logout modal


src/modules/MainConnector.js
```diff
@@ -17,24 +14,17 @@ import {enableScan, disableScan} from './UI/scenes/Scan/action'
...
+const mapStateToProps = () => ({})
+
 const mapDispatchToProps = (dispatch: Dispatch) => ({
-  dispatchEnableScan: () => {
-    return dispatch(enableScan())
-  },
-  dispatchDisableScan: () => {
-    return dispatch(disableScan())
-  },
+  dispatchEnableScan: () => dispatch(enableScan()),
```
paullinator: Please don't collapse these constructs if they are already coded as such. Collapsing them removes the ability to set a breakpoint. Please revert these lines

PR #254, opened 2017-12-18
Cleanup/manage tokens

wolverineks:


src/modules/Core/Wallets/api.js
```diff
@@ -36,72 +32,43 @@ const dummyAbcReceiveAddress: AbcReceiveAddress = {
...
+  .then(() => wallet.enableTokens([tokenObj.currencyCode]))
   .catch((e) => console.log(e))
 }
 
 export const getEnabledTokensFromFile = (wallet: AbcCurrencyWallet): Promise<Array<any>> => {
   return getEnabledTokensFile(wallet).getText()
-  .then((text) => {
-    return JSON.parse(text)
-  })
+  .then(JSON.parse)
```
kylanhurt: Will this change not make it harder to debug? No way of checking the returned text before the code attempts to parse it?

wolverineks: yeah, ill undo this one, i just like the way it looks better

PR #255, opened 2017-12-19
Remove magic error strings

swansontec:
The core exports these error types symbolically, so use that instead.


swansontec:
@ahujsak or @wolverineks  I think this might be your code.

PR #257, opened 2017-12-20
Cleanup l10n strings and move unused str separately

vsashyn:
Applied determining unused strings during runtime. I made series of simple interactions with system to collect used strings and left them in `en_US.js`. Unused strings were moved to `en_US.unused.js`. Also with `Proxy` functionality I left ability to use strings from `en_US.unused.js` to not break system. When localization string takes from unused package, `warn` message is shown.


ahujsak:
@vsashyn there are conflicts with this that need to be resolved before merging. Can you fix and re-submit.

vsashyn:
Hi @ahujsak. Conflicts are resolved.

PR #278, opened 2017-12-30
Implementation of Edit and Delete Tokens

kylanhurt:
The purpose of this pull request is to properly implement the Edit Token and Delete Token features, as well as improve upon the Add Token and Manage Tokens functionality.

Asana Tasks:
https://app.asana.com/0/361770107085503/491385731813992/f
https://app.asana.com/0/361770107085503/500391822345163/f
https://app.asana.com/0/361770107085503/502882927154102/f


src/locales/en_US.js
```diff
@@ -26,6 +26,125 @@ const strings = {
   exchange_failed: 'Exchange Failed',
   exchange_succeeded: 'Exchange Succeeded',
   exchanges_may_take_minutes: 'Exchanges may take several minutes to process. Please check your destination wallet after a few minutes',
+  bitid_identity: 'BitID Identity',
```
paullinator: This looks like a merge issue. These strings shouldn't have all been added back. Most were moved to the en_unused.js file

src/modules/Core/Account/settings.js
```diff
@@ -95,18 +95,40 @@ export const setDenominationKeyRequest = (account: AbcAccount, currencyCode: str
...
+    // If Settings.json doesn't exist yet, create it, and return it
+    return setSyncedSettings(account, SYNCED_ACCOUNT_DEFAULTS)
+  }
+}
+
 export const setSyncedSettings = (account: AbcAccount, settings: Object) => {
   const text = JSON.stringify(settings)
   const SettingsFile = getSyncedSettingsFile(account)
+  SettingsFile.setText(text)
+  .then(() => settings)
```
paullinator: What is this line for? 

The `.then(() => settings)` won't do anything. Are you trying to return the settings to the caller of setSyncedSettings?

src/modules/Core/Account/settings.js
```diff
@@ -170,6 +192,13 @@ export const getCoreSettings = (account: AbcAccount): Promise<{otpMode: boolean,
   return Promise.resolve(coreSettings)
 }
 
+export async function getSyncedSettingsFileAsync (account: AbcAccount) {
```
paullinator: Why do you have this routine at all? It does the same thing as getSyncedSettingsFile and none of the operations inside are async in the first place

src/modules/UI/Settings/reducer.js
```diff
@@ -213,7 +322,7 @@ export const settings = (state = initialState, action) => {
       isTouchEnabled: data
     }
   }
-
+*/
```
ahujsak: If you are commenting out this action, is it no longer needed?. If so delete it.

src/modules/Core/Account/settings.js
kylanhurt: I think I just got it mixed up with `getSyncedSettings`. I will remove it.

src/modules/UI/Wallets/action.js
```diff
@@ -187,10 +195,13 @@ export const editCustomToken = (walletId: string, tokenObj: any, oldCurrencyCode
             coreWalletsToUpdate,
             code: tokenObj.currencyCode
           }))
-          Actions.pop()
+          Actions.waletList()
```
paullinator: This typo will likely crash the app.

src/modules/UI/Wallets/action.js
```diff
@@ -161,16 +162,23 @@ export const editCustomToken = (walletId: string, tokenObj: any, oldCurrencyCode
...
+          dispatch(editCustomTokenFailure())
+        })
       } else { // replacing an existing but invisible token CASE 3
         addTokenAsync(walletId, tokenObj, state) // update the receiving token
         .then(() => {
           deleteCustomTokenAsync(walletId, oldCurrencyCode, getState) // delete the sending token
           .then((coreWalletsToUpdate) => {
             dispatch(overwriteThenDeleteTokenSuccess(tokenObj, oldCurrencyCode, coreWalletsToUpdate))
-            Actions.pop()
+            Actions.walletList()
```
paullinator: Why are you not using pop()? If the edit token scene was pushed onto the stack, you need to use pop() or you'll add more scenes to the stack.

src/modules/UI/Wallets/action.js
```diff
@@ -187,10 +195,13 @@ export const editCustomToken = (walletId: string, tokenObj: any, oldCurrencyCode
...
             code: tokenObj.currencyCode
           }))
-          Actions.pop()
+          Actions.waletList()
         })
       })
-      .catch((e) => dispatch(displayErrorAlert(e)))
+      .catch((e) => {
+        dispatch(displayErrorAlert(e.message))
+        dispatch(editCustomTokenFailure())
```
paullinator: This dispatch does not appear to be received by any reducer

kylanhurt: Ah, looks like I missed that one.

src/modules/UI/Settings/selectors.js
```diff
@@ -59,6 +59,15 @@ export const getDisplayDenominationFromSettings = (settings: any, currencyCode:
   return selectedDenomination
 }
 
+export const getDisplayDenominationFull = (state: State, currencyCode: string) => {
```
paullinator: What is the intended difference between this routine and getDisplayDenomination?

Seems that the only diff is that getDisplayDenomination will return fiat denominations if given a fiat currencyCode.

src/modules/Core/Wallets/api.js
```diff
@@ -82,21 +82,33 @@ export const getEnabledTokensFile = (wallet: AbcCurrencyWallet) => {
...
   // now stringify the new tokens
   let stringifiedTokens = JSON.stringify(finalTextArray)
   // grab the enabledTokensFile
+  const tokensFile = getEnabledTokensFile(wallet)
+  await tokensFile.setText(stringifiedTokens)
+  enableTokens(wallet, tokens)
+  if (tokensToDisable && tokensToDisable.length > 0) {
+    disableTokens(wallet, tokensToDisable)
+  }
+  return tokens
```
paullinator: Why even return tokens here as it's the same array that's passed into the function. The caller should already have access to it.

src/modules/Core/Wallets/api.js
```diff
@@ -82,21 +82,33 @@ export const getEnabledTokensFile = (wallet: AbcCurrencyWallet) => {
...
+}
+
+export async function updateEnabledTokens (wallet: AbcCurrencyWallet, tokensToEnable: Array<string>, tokensToDisable: Array<string>) {
```
paullinator: What is the purpose of having both setEnabledTokens and updateEnabledTokens? They seem to do almost exactly the same thing. You only use updateEnabledTokens for deleting a token, so shouldn't it just be deleteTokens() and you can remove the enable functionality

src/modules/UI/Wallets/action.js
```diff
@@ -93,40 +111,247 @@ export const setEnabledTokens = (walletId: string, enabledTokens: Array<string>,
...
     dispatch(setTokensSuccess())
+    dispatch(updateWalletEnabledTokens(walletId, enabledTokens))
     // refresh the wallet in Redux
     dispatch(refreshWallet(walletId))
   })
-  .catch((e) => console.log(e))
+  .catch((e) => dispatch(displayErrorAlert(e.message)))
 }
 
 export const getEnabledTokens = (walletId: string) => (dispatch: Dispatch, getState: GetState) => {
```
paullinator: You are calling getEnabledTokens for all wallets including those that aren't ETH. This will throw an error for BTC, LTC, Dash wallets as they don't support tokens and you are calling addCustomToken() on them.

Worst yet, you are calling addCustomToken on all wallets for all customTokens that you have in the account, even if the wallet doesn't have that token enabled. This will be a big performance and resource hit as each ETH wallet will have to query for every token even if it's not enabled in that wallet.

src/modules/UI/Settings/selectors.js
kylanhurt: Solely for debugging purposes. Having the `getDisplayDenomination` function call another function, which calls another function, etc makes it more of a hassle to debug with breakpoints.

src/modules/Core/Wallets/api.js
kylanhurt: We created the function as a more general way to enable or disable tokens, all in one function. So the idea would be to eventually get rid of `setEnabledTokens`. I can replace the `setEnabledTokens` calls with `updateEnabledTokens` as part of this pull request, if you'd like.

PR #279, opened 2017-12-31
URGENT: Fix crashes due to Contacts permission

paullinator:
URGENT fix: This fixes the bulk of crash reports which are due to accessing phone contacts without permission.


ahujsak:
Looks good waiting on travis to approve.

PR #288, opened 2018-01-02
Flip input animation on android

amanalili:


paullinator:
In testing this change, it looks like when the inputs are flipped, the numpad still enters values into the same textfield. So if USD was at the top, after flipping, the USD field is still being edited. This should refocus the input into the top text field.

paullinator:
The animation does look awesome though! :)

amanalili:
Old branch, will close see new branch with submitted PR (art/flipinput-android)

PR #293, opened 2018-01-04
Fix symbols font

wolverineks:
Add symbols font to theme
Use symbol font for symbols
Break strings up into individually styled components


thehobbit85:
Closing since PR should now be into develop instead

PR #297, opened 2018-01-05
Linting Pt 1 (All but No Unused Vars)

kylanhurt:
The purpose of this task is to apply the first set of eslint rules from core to edge-react-gui.

Asana Task: https://app.asana.com/0/361770107085503/510015864564817/f


.eslintrc.js
```diff
@@ -1,89 +1,14 @@
...
-  'env': {
-    'es6': true,
-    'node': true
-  },
-  'parser': 'babel-eslint',
-  'extends': [
-    'eslint:recommended',
-    'plugin:flowtype/recommended',
-    'plugin:react/recommended',
-    'plugin:react-native/all'
```
swansontec: We might want to add these rules back (react/recommended, react-native/all). The core's eslintrc didn't have them, since the core doesn't use React, but the GUI should keep them.

.eslintrc.js
```diff
@@ -1,89 +1,14 @@
...
-  'parserOptions': {
-    'ecmaFeatures': {
-      'experimentalObjectRestSpread': true,
-      'jsx': true
-    },
-    'sourceType': 'module'
-  },
-  'plugins': [
-    'react',
-    'react-native',
```
swansontec: If we re-enable the react rules, we'll need to re-enable the react and react-native plugins as well.

src/modules/UI/scenes/Settings/action.js
```diff
@@ -200,34 +192,3 @@ const setBitcoinOverrideServerStart = (overrideServer: string) => ({
...
-// dailySpendLimitSatoshi  (integer)
-
-// Currency Settings
-// advancedFeatures          (boolean)
-// bitcoinDenomination       (Value)?
-// exchangeRateSource        (string)
-// language                  (string)
-// numCurrency?              (integer)
-// overrideBitcoinServers    (boolean)
-// overrideBitcoinServerList (string)
```
swansontec: Maybe we should put this block back? It looks like valuable information.

.eslintrc.js
kylanhurt: Should be added back in now

kylanhurt: They have been added back now with the latest commit

.eslintrc.js
kylanhurt: They have been added back now with the latest commit

src/modules/UI/scenes/Settings/action.js
kylanhurt: The block has been added back in

src/modules/UI/scenes/CreateWallet/CreateWallet.ui.js
```diff
@@ -104,7 +104,7 @@ export default class CreateWallet extends Component {
   }
 
   handleSelectWalletType = ({value} = {value: ''}) => {
-    const selectedWalletType = this.props.supportedWalletTypes.find((type) => type.value = value)
+    const selectedWalletType = this.props.supportedWalletTypes.find((type) => type.value === value)
```
thehobbit85: Great Catch

src/modules/UI/scenes/TransactionDetails/TransactionDetails.ui.js
```diff
@@ -99,22 +99,22 @@ export class TransactionDetails extends Component<Props & DispatchProps, State>
       this.guiWallet = props.wallets[props.abcTransaction.wallet.id]
       this.fiatSymbol = UTILS.getFiatSymbol(this.guiWallet.fiatCurrencyCode)
     } else {
-      throw 'No wallet on transaction object'
+      this.props.displayDropdownAlert(s.strings.transaction_detail_no_wallet, s.strings.transaction_detail_unable_to_load_transaction)
```
thehobbit85: another good catch

PR #298, opened 2018-01-05
Otp settings

ahujsak:
Adds OTP and Password Recovery back to the settings menu, needing a confirmed password to change the settings. 
• adds OTP settings screen
• Checks the users OTP status.
• Allows user to activate and de-activate OTP


paullinator:
@ahujsak 

Looks like most of the functionality is actually in js-ui as expected. This change looks good but testing this branch shows that OTP doesn't seem to work. I'm able to enable OTP but entering my backup code on another device upon login does nothing. The QR code EdgeLogin doesn't work either. What is the actual state of OTP as of this change? It doesn't appear to be functional.

PR #299, opened 2018-01-05
Better logs

thehobbit85:
Added a massive dataDump from all wallets when sending the logs
Also cleaned up and reformatted the GUI callback logs so that they all will be in the same format and not too verbose


src/modules/Core/Account/callbacks.js
```diff
@@ -19,42 +19,33 @@ const makeAccountCallbacks = (dispatch: Dispatch): AbcAccountCallbacks => {
...
 
     onTransactionsChanged (walletId: string, transactions: Array<AbcTransaction>) {
-      if (transactions && transactions.length) {
-        console.log('onTransactionsChanged length=' + transactions.length.toString())
-        for (const tx of transactions) {
-          console.log('walletId:' + walletId + ' txid:' + tx.txid)
-        }
-      } else {
-        console.log('onTransactionsChanged: No transactions')
+      if (!transactions || !transactions.length) {
```
paullinator: Looks like it would be useful to also have logging when there actually are transactions that have changed

src/modules/Core/Account/callbacks.js
```diff
@@ -19,42 +19,33 @@ const makeAccountCallbacks = (dispatch: Dispatch): AbcAccountCallbacks => {
...
 
     onNewTransactions (walletId: string, transactions: Array<AbcTransaction>) {
-      if (transactions && transactions.length) {
-        console.log('onNewTransactions length=' + transactions.length.toString())
-        for (const tx of transactions) {
-          console.log('walletId:' + walletId + ' txid:' + tx.txid)
-        }
-      } else {
-        console.log('onNewTransactions: No transactions')
+      if (!transactions || !transactions.length) {
```
paullinator: same here

thehobbit85: I added it back in the next commit

src/modules/Core/Account/callbacks.js
thehobbit85: I added it back in the next commit

thehobbit85:
@paullinator  I added back the logs you requested, you can now merge it.

paullinator:
Looks good. Just waiting for Travis. Merge if successful.

PR #303, opened 2018-01-07
Localize numbers according to user locale settings

vsashyn:
Wrap all the places where we display a number, with a localization library that can be set to a region that properly displays the time in the format of that region.
Used standard JS `Intl` internationalization object with polyfill (intl) for Android platform. 

1. Created singleton wrapper `intl`. Created method to set locale which we can get from `react-native-locale`. Then available as `import` object
2. Some `utils` moved to their internationalized versions (`isValidInput`, `formatNumberInput`, `prettifyNumber`). Backward compatibility saved.
3. `FormatNumber` is method to use for internationalization numbers


src/locales/intl.js
```diff
@@ -0,0 +1,125 @@
+import 'intl'
```
paullinator: Please add Flow to this file.

src/locales/intl.test.js
```diff
@@ -0,0 +1,320 @@
+/* globals describe test expect beforeEach */
+import intl, {setIntlLocale} from './intl'
```
paullinator: We are standardizing against using unnamed exports. Please give `intl` a specific export name and change to something like

`import {intl, setIntLocale} from './intl/`

src/locales/intl.test.js
```diff
@@ -0,0 +1,320 @@
...
+import * as UTILS from '../../src/modules/utils'
```
paullinator: Please import specific routines. We are standardizing against `import *` going forward

src/locales/intl.js
```diff
@@ -0,0 +1,125 @@
...
+  'currencyCode': 'USD',
+  'groupingSeparator': ','
+}
+let locale = EN_US_LOCALE
+
+const intlHandler = {
+  /**
+   * Formats number input according to user locale
+   * Allows decimalSeparator at the end of string
+   * @param number  - Native style (JS) number
```
paullinator: Looks like this can take a number or a string. Be careful as some numbers a big ints and would overflow a JS number.

src/locales/intl.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+}
+
+export const setIntlLocale = (l) => {
+  if (!locale) throw new Error('Please select locale for internationalization')
+  locale = l
+  return intlHandler
+}
+
+export default intlHandler
```
paullinator: Do not use default exports

src/locales/intl.test.js
```diff
@@ -0,0 +1,320 @@
...
+
+const EN_US_LOCALE = {
+  'localeIdentifier': 'en_US',
+  'decimalSeparator': '.',
+  'quotationBeginDelimiterKey': '“',
+  'quotationEndDelimiterKey': '”',
+  'currencySymbol': '$',
```
paullinator: We deal with currency symbols and currency codes in a different module. These should be removed

src/locales/intl.js
vsashyn: Done

src/locales/intl.test.js
vsashyn: Fixed

vsashyn:
Fixed

PR #313, opened 2018-01-09
Change margin of help modal text, fix icon positions of modals.

amanalili:


paullinator:
Art. This PR doesn't seem to address the problem. The help modals need to be made bigger to take up more of the screen real estate. They should take up all but 10px on the left and right as mentioned in the task. Closing

PR #316, opened 2018-01-09
Fix transaction alert

wolverineks:


paullinator:
Approved once Travis build is checked

PR #317, opened 2018-01-10
Fix missing left side menu on the exchange scene in android

amanalili:


src/styles/components/HeaderMenuDropDownStyles.js
```diff
@@ -47,7 +47,16 @@ const MenuDropDownStyle = {
   altIconText: {
     fontSize: 20
   }
+}
 
+const MenuDropDownStyleHeader = {
```
paullinator: I don't understand how this fixes the problem. So you had to remove `flex` and `justifyContent` to get this to work? Does this affect iOS?

amanalili:
Yup it is affected with flex and justifyContent. Nope, it is not affected in iOS because that is what I styled it way before it was affected by other styles. Header in the android has a different style one the iOS, that is what the default theme we are using.

PR #321, opened 2018-01-10
Read etherscan API from env.json and put into global for use by ethereum plugin

paullinator:
William and I discussed and decided this was the easiest and still effective way to get an API key from GUI to a plugin


swansontec:
This seems fine. Where can we find the corresponding key to update our development machines?

paullinator:
The key is optional so the app should still build and run fine. But I will distribute a key in a separate message to Edge devs

PR #323, opened 2018-01-11
Auto capitalize forms/Text input on various places

amanalili:
Some of it are already autocapitalize, just check it.


src/modules/UI/scenes/AddToken/AddToken.ui.js
```diff
@@ -69,7 +69,7 @@ class AddToken extends Component<Props, State> {
                 style={[styles.currencyName]}
                 value={this.state.currencyName}
                 onChangeText={this.onChangeName}
-                autoCapitalize={'words'}
+                autoCapitalize="words"
```
paullinator: `autoCapitalize={'words'}` is the proper syntax and style. Please change other use of `autoCapitalize="words"` to  `autoCapitalize={'words'}`

amanalili:
Done editing to single qoutes

PR #336, opened 2018-01-16
Fix account creation navigation

wolverineks:
* Decouple creating a wallet on account creation from creating a wallet
after login

* Move shared simple actions outside of create wallet scene


src/modules/Login/action.js
```diff
@@ -50,10 +51,10 @@ export const initializeAccount = (account: AbcAccount, touchIdInfo: Object) => a
         dispatch(updateWalletsRequest())
         return
       }
-      dispatch(actions.createCurrencyWallet(
+      dispatch(createCurrencyWallet(
```
paullinator: Does this need to be a dispatch? Why not a function call?

PR #338, opened 2018-01-16
added android and ios local notifications from background task for OTP.

ahujsak:
This gets background task working and adds the ability to do local notification for when a user has had their password 2FA reset on a different device.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -92,13 +92,11 @@ global.pcount = function (label: string) {
 }
 
 BackgroundTask.define(async () => {
-  console.log('background Shit')
   makeCoreContext()
     .then(async (context) => {
       try {
         const result = await context.fetchLoginMessages()
         const date = new Date(Date.now() + 1000)
-        console.log(result)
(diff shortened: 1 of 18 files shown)


src/app.js
```diff
@@ -89,14 +91,40 @@ global.pcount = function (label: string) {
   }
 }
 
-BackgroundTask.define(() => {
-  console.log('Hello from a background task')
+BackgroundTask.define(async () => {
+  console.log('background Shit')
```
wolverineks: Come on dude.....come on...

src/app.js
```diff
@@ -89,14 +91,40 @@ global.pcount = function (label: string) {
...
+                subText: sprintf(s.strings.otp_notif_body, key),
+                date
+              })
+            }
+          }
+        }
+        // return response
+      } catch (error) {
+        console.error('Failed error ')
+        console.error(error)
```
wolverineks: Just a single console.log will do

PR #342, opened 2018-01-17
Allen/deep link fix

ahujsak:
fixes a deep link


src/modules/Main.ui.js
```diff
@@ -215,27 +215,23 @@ export default class Main extends Component<Props, State> {
...
+      }
+      // this.navigate(url);
+    }).catch(err => console.log('error occurred ', err))
+    Linking.addEventListener('url', this.handleOpenURL)
   }
-  handleOpenURL = (event: Object) => {
-    // this.props.urlRecived(event.url)
-    const splitArray = event.url.split('recovery?token=')
+  doDeepLink (url: string) {
+    const splitArray = url.split('recovery?token=')
```
paullinator: This should use a URL handler which parses out the different URI components. There's no guarantee that `token=` will be at the beginning of the URI

ahujsak: @paullinator not sure what you mean by uri handler. ... based on your comments I was thinking that I could split on token= instead of recovery?token= since token = would be the spot where the token var is in the url. .. I could also add a second split on a & if there is another variable after the token.. however for this use case where we are getting the uri, we can guarantee the format since we create the link....

paullinator:
This is a URI handling/parsing library

ahujsak:
@paullinator , made the changes .. ready to when you have time.

src/modules/Main.ui.js
```diff
@@ -215,27 +215,24 @@ export default class Main extends Component<Props, State> {
...
+        this.doDeepLink(url)
+      }
+      // this.navigate(url);
+    }).catch(err => console.log('error occurred ', err))
+    Linking.addEventListener('url', this.handleOpenURL)
+  }
+  doDeepLink (url: string) {
+    const parsedUri = URI.parse(url)
+    const query = parsedUri.query
+    const splitArray = query.split('token=')
```
paullinator: This is missing the point, Allen. `uri-js` will parse all the elements of the URI for you, you are still making the bad assumption that everything after `token=` is the token. If this URI had more parameters this would break. Please use uri-js for all parsing.

ahujsak: Paul, I disagree, 
When I parse this url edge://recovery?token=IamToken&nextVar=nextVar
I get the following object 
```
{
  fragment: undefined,
  host: “recovery”
  path: “”,
  port: undefined,
  query: “token=IamToken&nextVar=nextVar”
  reference: “absolute
  scheme:”edge”
  userinfo: undefined
}
```

My code then takes the query and splits it. 
I then take into consideration any additional items in the query string and trim them. The one thing I did not consider was if ‘token=‘ was even present in the query string, so I am now checking for that.
I have also added a const called token for the value that is finalArray[0]
So running the following code here are my tests and results. 

```
const parsedUri = URI.parse(url)
const query = parsedUri.query
if (!query.includes('token=')) {
 return
}
const splitArray = query.split('token=')
const nextString = splitArray[1]
const finalArray = nextString.split('&')
const token = finalArray[0]
console.log(token)
```

Test url: edge://recovery?token=IamToken&nextVar=nextVar
Result of console log: IamToken

Test url: https://recovery.edgesecure.co/recovery?token=IamToken&nextVar=nextVar
Result of console log: IamToken

Test url: https://recovery.edgesecure.co/recovery?bob=bob&token=IamToken&nextVar=nextVar
Result of console log: IamToken

Test url: https://recovery.edgesecure.co/recovery?bob=bob&nextVar=nextVar
Result of console log: nothing, the function returns before we get the console log 

Based on my results I have covered the all the different use cases. I did not see anything readily apparent in the uri-js library that allows me to just pluck a value out of the query string. If I had seen that I would have done it.

src/modules/Main.ui.js
```diff
@@ -191,27 +191,28 @@ export default class Main extends Component<Props, State> {
...
+    const parsedUri = URI.parse(url)
+    const query = parsedUri.query
+    if (!query.includes('token=')) {
+      return
     }
+    const splitArray = query.split('token=')
+    const nextString = splitArray[1]
+    const finalArray = nextString.split('&')
+    const token = finalArray[0]
+    this.props.urlRecived(token)
```
wolverineks: typo

src/modules/Main.ui.js
```diff
@@ -191,27 +191,28 @@ export default class Main extends Component<Props, State> {
...
+  doDeepLink (url: string) {
+    const parsedUri = URI.parse(url)
+    const query = parsedUri.query
+    if (!query.includes('token=')) {
+      return
     }
+    const splitArray = query.split('token=')
+    const nextString = splitArray[1]
+    const finalArray = nextString.split('&')
+    const token = finalArray[0]
```
paullinator: With the check for `token=`. this appears safe to arbitrary query strings.

paullinator:
Fix merge conflict, test, and go ahead and merge

PR #345, opened 2018-01-17
Enlarge help modals to occupy most of the screen

amanalili:


amanalili:
Did changes without actually changing the main modal style file.

src/lib/isIphoneX.js
```diff
@@ -0,0 +1,8 @@
+import { Dimensions, Platform } from 'react-native'
+
+const isIphoneX = () => {
+  const { height, width } = Dimensions.get('window')
+  return Platform.OS === 'ios' && (height === 812 || width === 812)
+}
+
+export default isIphoneX()
```
paullinator: No default exports. We are standardizing against this.

paullinator:
Also fix conflicts

src/lib/isIphoneX.js
amanalili: What should I use? also are there anywhere where I can read the standards we have?

kylanhurt: @amanalili I think it would be something like `export const isThisIPhoneX = isIphoneX()` then `import {isThisIPhoneX} from 'thisFile.js'` or something along those lines. We are moving away from `export default ...` because it makes it too easy to rename things.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

kylanhurt:
Approved after making Paul's comments.

thehobbit85:
I've added the changes Paul asked for and fixed the conflicts.
I'm merging the PR.

PR #348, opened 2018-01-17
Remove routes reducer

wolverineks:
Routes aren't used in the app


thehobbit85:
What about the "routes: state.routes" prop inside the following files:

https://github.com/Airbitz/edge-react-gui/blob/develop/src/modules/UI/components/Header/Component/BodyConnector.js
https://github.com/Airbitz/edge-react-gui/blob/develop/src/modules/UI/components/Header/HeaderConnector.js
https://github.com/Airbitz/edge-react-gui/blob/develop/src/modules/UI/components/HelpModal/HelpModalConnector.js
https://github.com/Airbitz/edge-react-gui/blob/develop/src/modules/UI/components/TabBar/TabBarConnector.js

As far as I can see they are used there but I'm not sure I fully understand what goes on there.

PR #349, opened 2018-01-18
Kevin/test/reducers

wolverineks:
Smoke Test / Test Template for reducers

Each test imports the reducer, "dispatches" and empty action to initialize the reducer, and compares the initial state to the expected initial state

Each reducer test should include a case for each action that the reducer responds to


paullinator:
Pls resolve conflicts first

PR #364, opened 2018-01-23
sendConfirmation performance optimization and bug fixes

thehobbit85:
The "sendConfirmation" got rendered on EVERY update on EVERYTHING in the entire Redux Store while it is mounted....

Now, render will be called ONLY when actual data that we care about in this scene change.

It solved multiple weird issues such as with the fees and when changing between one wallet and another.


kylanhurt:
Is there an Asana task for this? @thehobbit85 @paullinator

thehobbit85:
No.
It was something I needed to do in order to get the customFee to work without having really weird UI/UX issues and with out hacking the hell out of the APP.

It also fixes several bugs related to changing fees in general.
And as a side bonus, we also gain a significant increase in the overall performance when we are on the sendConfirmation/ChangeFees scenes.

src/modules/UI/scenes/SendConfirmation/SendConfirmation.ui.js
```diff
@@ -20,86 +20,109 @@ import ABSlider from '../../components/Slider/index.js'
...
-      primaryInfo,
-      secondaryInfo,
-      fiatPerCrypto,
-      errorMsg,
-      nativeAmount,
-      currencyConverter,
-      currencyCode
-    } = this.props
-    const color = 'white'
+    this.props.makeSpend({ ...this.props })
```
paullinator: Very very bad. Do not do a dispatch inside a render()

src/modules/UI/scenes/SendConfirmation/action.js
```diff
@@ -125,43 +136,70 @@ export const updateWalletTransfer = (wallet: AbcCurrencyWallet) => (dispatch: an
...
   const abcWallet = CORE_SELECTORS.getWallet(state, walletId)
-  const spendInfo: AbcSpendInfo = makeSpendInfo(parsedUri)
-  spendInfo.networkFeeOption = state.ui.scenes.sendConfirmation.feeSetting
-  spendInfo.customNetworkFee = state.ui.scenes.sendConfirmation.feeSatoshi
-
-  return WALLET_API.makeSpend(abcWallet, spendInfo)
+  return WALLET_API.makeSpend(abcWallet, makeSpendInfo(options))
   .then((abcTransaction: AbcTransaction) => {
-    dispatch(updateTransactionAction(parsedUri, abcTransaction, null))
+    dispatch(updateTransaction(abcTransaction))
```
paullinator: Do not make 2 dispatch calls. These can be one call with two parameters. Save two separate renders happening instead of one

paullinator: Why did this even need to be changed?

src/modules/UI/scenes/SendConfirmation/action.js
```diff
@@ -34,45 +37,54 @@ export const UPDATE_CRYPTO_AMOUNT_REQUEST = PREFIX + 'UPDATE_CRYPTO_AMOUNT_REQUE
...
 })
 
-export const updateTransactionAction = (
-  parsedUri: AbcParsedUri,
-  transaction: AbcTransaction | null,
-  error: Error | null): UpdateTransactionAction => ({
-    type: UPDATE_TRANSACTION,
-    data: {parsedUri, transaction, error}
-  })
+export const updateTransaction = (transaction: ?AbcTransaction): any => ({
```
paullinator: This is exactly what you do NOT want to do in redux. Do not break up one action into two if not necessary. This causes multiple renders.

src/modules/UI/scenes/SendConfirmation/action.js
thehobbit85: It's fixed in the new version I'm currently testing

src/modules/UI/scenes/SendConfirmation/action.js
thehobbit85: It's fixed in the new version I'm currently testing

PR #365, opened 2018-01-23
Separated WalletSelectorConnector into 2 files.

thehobbit85:
One contains the declarative static attributes of the Component and the other the dynamic state-bounded props.

This tiny bit of code removes somewhere in the ballpark of 50-100 unnecessary renders a sec to zero.

### DIFF
--- a/src/modules/UI/components/Header/Component/WalletSelector.ui.js
+++ b/src/modules/UI/components/Header/Component/WalletSelector.ui.js
@@ -1,0 +1,30 @@
+// @flow
+
+import React, {Component} from 'react'
+import type {State} from '../../../../ReduxTypes'
+import {TextAndIconButton} from '../../Buttons/TextAndIconButton.ui'
+import * as Constants from '../../../../../constants/indexConstants'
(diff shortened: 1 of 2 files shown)


src/modules/UI/components/Header/Component/WalletSelector.ui.js
```diff
@@ -0,0 +1,30 @@
...
+
+export default class WalletSelector extends Component<Props, State> {
+  render () {
+    return <TextAndIconButton
+      style={CryptoExchangeSceneStyle.flipWrapper.walletSelector}
+      icon={Constants.KEYBOARD_ARROW_DOWN}
+      iconType={Constants.MATERIAL_ICONS}
+      onPress={this.props.onPress}
+      title={this.props.title}
+      blat={'WalletSelectorConnector'}
```
wolverineks: what is this?

thehobbit85: lol, forgot to remote it, used it to help me find the the module in the react stack when debugging :)
Will remove it

PR #369, opened 2018-01-24
Contact List on Payee Input sorted and remove the blank named contacts

amanalili:


amanalili:
This is in Transaction Details UI

PR #379, opened 2018-01-26
refactor login action initialization.

ahujsak:
• removed over 13 dispatches into one single dispatch to initialize the account

### DIFF
--- a/src/modules/Login/action.js
+++ b/src/modules/Login/action.js
@@ -1,24 +1,19 @@
 // @flow
 
-import {AbcAccount} from 'airbitz-core-types'
+import {AbcAccount, type AbcCurrencyWallet} from 'airbitz-core-types'
 import {Actions} from 'react-native-router-flux'
 
 import type {GetState, Dispatch} from '../ReduxTypes'
-import {displayErrorAlert} from '../UI/components/ErrorAlert/actions'
 
 // Login/action.js
 import * as CONTEXT_API from '../Core/Context/api'
 import * as CORE_SELECTORS from '../Core/selectors'
 import * as ACCOUNT_API from '../Core/Account/api'
-import * as ACCOUNT_ACTIONS from '../Core/Account/action.js'
-import * as SETTINGS_ACTIONS from '../UI/Settings/action.js'
 import * as SETTINGS_API from '../Core/Account/settings.js'
-import * as WALLET_ACTIONS from '../UI/Wallets/action'
 import * as actions from '../../actions/indexActions'
 import * as Constants from '../../constants/indexConstants'
-import * as ADD_TOKEN_ACTIONS from '../UI/scenes/AddToken/action.js'
 import s from '../../locales/strings.js'
-import {updateWalletsRequest} from '../Core/Wallets/action.js'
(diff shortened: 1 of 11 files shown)


src/modules/Login/action.js
```diff
@@ -1,24 +1,19 @@
 // @flow
 
-import {AbcAccount} from 'airbitz-core-types'
+import {AbcAccount, type AbcCurrencyWallet} from 'airbitz-core-types'
```
swansontec: They are both types, so `import type {}` is better.

src/modules/Login/action.js
```diff
@@ -1,24 +1,19 @@
...
-import * as ACCOUNT_ACTIONS from '../Core/Account/action.js'
-import * as SETTINGS_ACTIONS from '../UI/Settings/action.js'
 import * as SETTINGS_API from '../Core/Account/settings.js'
-import * as WALLET_ACTIONS from '../UI/Wallets/action'
 import * as actions from '../../actions/indexActions'
 import * as Constants from '../../constants/indexConstants'
-import * as ADD_TOKEN_ACTIONS from '../UI/scenes/AddToken/action.js'
 import s from '../../locales/strings.js'
-import {updateWalletsRequest} from '../Core/Wallets/action.js'
+// import {updateWalletsRequest} from '../Core/Wallets/action.js'
```
swansontec: Oops?

src/modules/Login/action.js
```diff
@@ -32,85 +27,101 @@ export const initializeAccount = (account: AbcAccount, touchIdInfo: Object) => a
...
+    accountInitObject.bluetoothMode = localFinal.bluetoothMode
+
+    const coreSettings = await SETTINGS_API.getCoreSettings(account)
+    const coreDefaults = SETTINGS_API.CORE_DEFAULTS
+    const coreFinal = {...coreDefaults, ...coreSettings}
+    accountInitObject.pinMode = coreFinal.pinMode
+    accountInitObject.otpMode = coreFinal.otpMode
+    dispatch(actions.dispatchActionObject(Constants.ACCOUNT_INIT_COMPLETE, accountInitObject))
+  } catch (e) {
+    console.log(e)
```
swansontec: Should we tell the user if this happens?

src/modules/Login/action.js
```diff
@@ -32,85 +27,101 @@ export const initializeAccount = (account: AbcAccount, touchIdInfo: Object) => a
...
+    accountInitObject.denominationKeys.push({currencyCode: 'ETH', denominationKey: syncFinal.ETH.denomination})
+    if (customTokens) {
+      customTokens.forEach((token) => {
+        // dispatch(ADD_TOKEN_ACTIONS.setTokenSettings(token))
+        accountInitObject.customTokensSettings.push(token)
+        // this second dispatch will be redundant if we set 'denomination' property upon customToken creation
+        accountInitObject.denominationKeys.push({currencyCode: token.currencyCode, denominationKey: token.multiplier})
+      })
+    }
+    const localSettings = await SETTINGS_API.getLocalSettings(account)
```
swansontec: Does this function return defaults if it can't load the settings? If so, do we really need to merge the settings on the following lines?

PR #384, opened 2018-01-26
Request Scene. Balance will revert back to 0 when currency wallet is …

amanalili:
…changed

NOTICE BEFORE APPROVAL

I needed to **add a setTimeout** here because there seems to be a a bug when changing to ethereum wallet. Did check how the code works and nothing I see that should affect computation of conversion. 

Error comes from IndexEthereum.js: 1840: 16

Attached error file:
screen shot 2018-01-26 at 10 57 09 pm

### DIFF
--- a/src/modules/UI/components/FlipInput/FlipInput.ui.js
+++ b/src/modules/UI/components/FlipInput/FlipInput.ui.js
@@ -120,6 +120,9 @@ export default class FlipInput extends Component<Props, State> {
         secondaryDisplayAmount: UTILS.truncateDecimals(nextProps.secondaryDisplayAmount, 2)
       })
     }
+    if (nextProps.primaryInfo.displayCurrencyCode !== this.props.primaryInfo.displayCurrencyCode) {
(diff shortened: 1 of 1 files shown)


src/modules/UI/components/FlipInput/FlipInput.ui.js
```diff
@@ -120,6 +120,9 @@ export default class FlipInput extends Component<Props, State> {
         secondaryDisplayAmount: UTILS.truncateDecimals(nextProps.secondaryDisplayAmount, 2)
       })
     }
+    if (nextProps.primaryInfo.displayCurrencyCode !== this.props.primaryInfo.displayCurrencyCode) {
```
paullinator: Hmm. this will need to be added to the FlipInput2 component as well since that's the new one we're using. But it's still in PR review at this time. In a different commit, can you also add it to paul/flipinput2 in FlipInput2.ui.js and test.

paullinator: FlipInput2 has been merged to develop. Art please look at adding this change to FlipInput2 as well

PR #388, opened 2018-01-27
Change mining fee display automatically.

amanalili:
Added a process to call processParsedUri() when changing the mining fee.


thehobbit85:
No need for this since it's already solved in a better less "magical" way in:
https://github.com/Airbitz/edge-react-gui/pull/389

PR #399, opened 2018-01-30
Fix Exchange scene to use new FlipInput2

paullinator:
Refactor CryptoExchangeFlipInput to use ExchangedFlipInput2

Fixes feedback loops which cause input corruption and infinite loops. Also speeds up UI interaction
-------------------------------------------------------------------------------------------------
* Add Flow to CryptoExchangeFlipInput. Necessary to do refactor and detect wrong or missing parameters
* Changed GuiWallet from class to type so we could declare emptyGuiWallet for use when objects are not initialized
* Made CryptoExchangeFlipInput a non-connected component to eliminate feedback loops
* Have the CryptoExchangeScene handle where keyboard input came from and only update opposite CryptoExchangeFlipInput
* Removed dead props and code
* Zero out amounts if user hits center flip wallets button
* Adjust precision of exchangeAmount when converting From amounts to To amounts and vice versa. This prevents excessive decimal places
* Remove redundant mining fee title string

Fix Change Mining Fee and Exchange Max on CryptoExchange
-------------------------------------------------------------------------------------------------
* Fix bug where Exchange Max was calling the SendConfirmation action, not the CryptoExchange actions
* Fix CryptoExchangeActions:changeFee() to not just set the redux state, but also initiate a shape shift transaction to trigger fee estimates
* Collapse Constants.SET_CRYPTO_FROM_NATIVE_AMOUNT and Constants.SET_CRYPTO_TO_NATIVE_AMOUNT into one action (Constants.SET_CRYPTO_EXCHANGE_AMOUNTS). This can set all values in one action saving multiple dispatches.
* Allow Constants.SET_CRYPTO_EXCHANGE_AMOUNTS to take an forceUpdateGui boolean param which if true increments the forceUpdateGuiCounter forcing FlipInput to update GUI amounts
* Propagate a forceUpdateGuiCounter all the way down from CryptoExchageFlipInput to FlipInput2 which allows forced updates from events such as Change Mining Fee or Max Spend/Exchange

Note. Changing Mining Fee still doesn't work as there is a bug in Core where makeSpend for ShapeShift transactions doesn't look at the networkFeeOption. Verified fixed using quick hacked Core with one-line fix

Implement Eliran's optimization of only returning primitive types or global object pointers from mapStateToProps


src/locales/en_US.js
```diff
@@ -5,7 +5,6 @@ const strings = {
   bitcoin_remaining: '%1$s Remaining…',
```
thehobbit85: This change is approved

src/modules/UI/scenes/SendConfirmation/SendConfirmation.ui.js
```diff
@@ -179,6 +179,7 @@ export class SendConfirmation extends Component<Props, State> {
                 secondaryCurrencyInfo={{ ...secondaryInfo }}
                 exchangeSecondaryToPrimaryRatio={this.props.fiatPerCrypto}
                 overridePrimaryExchangeAmount={this.state.overridePrimaryExchangeAmount}
+                forceUpdateGuiCounter={this.state.forceUpdateGuiCounter}
```
thehobbit85: This change is approved

src/modules/UI/scenes/SendConfirmation/SendConfirmationOptions.js
```diff
@@ -7,7 +7,7 @@ import s from '../../../../locales/strings.js'
 import {border} from '../../../utils'
 import styles from './styles'
 
-const CHANGE_MINING_FEE_TEXT = s.strings.change_mining_fee_title
+const CHANGE_MINING_FEE_TEXT = s.strings.title_change_mining_fee
```
thehobbit85: This change is approved

PR #406, opened 2018-01-31
Fix flow errors from core typing AbcCurrencyWallet

swansontec:
The core has a branch, `currency-wallet-flow`, that adds proper types for `AbcCurrencyWallet`. Merging that branch causes lots of errors in the GUI. This pull request fixes those.

### DIFF
--- a/src/modules/UI/Wallets/reducer.js
+++ b/src/modules/UI/Wallets/reducer.js
@@ -248,11 +248,12 @@ function schema (wallet: AbcCurrencyWallet): GuiWallet {
   const currencyCode: string = wallet.currencyInfo.currencyCode
   const fiatCurrencyCode: string = wallet.fiatCurrencyCode.replace('iso:', '')
   const isoFiatCurrencyCode: string = wallet.fiatCurrencyCode
-  const symbolImage: string = wallet.currencyInfo.symbolImage
-  const symbolImageDarkMono: string = wallet.currencyInfo.symbolImageDarkMono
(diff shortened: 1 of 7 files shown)


src/modules/UI/Wallets/reducer.js
```diff
@@ -248,11 +248,12 @@ function schema (wallet: AbcCurrencyWallet): GuiWallet {
...
   const isoFiatCurrencyCode: string = wallet.fiatCurrencyCode
-  const symbolImage: string = wallet.currencyInfo.symbolImage
-  const symbolImageDarkMono: string = wallet.currencyInfo.symbolImageDarkMono
+  const symbolImage = wallet.currencyInfo.symbolImage
+  const symbolImageDarkMono = wallet.currencyInfo.symbolImageDarkMono
   const metaTokens: Array<AbcMetaToken> = wallet.currencyInfo.metaTokens
   const denominations: Array<AbcDenomination> = wallet.currencyInfo.denominations
-  const enabledTokens: Array<string> = wallet.enabledTokens || []
+  // TODO: Fetch the token list asynchonously before dispatching `schema`:
+  const enabledTokens: Array<string> = []
```
paullinator: How does this not break the app? enabledTokens will always be blank upon wallet initialization

swansontec: It was always broken. There was never any `wallet.enabledTokens` property, so this was always undefined, so the value was always an empty array. My change doesn't really affect anything; it just brings the existing problem to the forefront.

PR #407, opened 2018-01-31
Fix functions warning

wolverineks:


paullinator:
Tested back and help buttons on iPhone 7

PR #418, opened 2018-02-01
Change PLATFORM to export and import as object rather than default

kylanhurt:
The purpose of this is to switch from `export default` within `theme/variables/platform.js` to `export {PLATFORM}` to defend against a react-native bug.

There is no Asana task and it was mostly search + replace


src/theme/variables/material.js
```diff
@@ -2,8 +2,8 @@ import color from 'color';
 
 import { Platform, PixelRatio } from 'react-native';
 
-const deviceHeight = platform.deviceHeight;
-const deviceWidth = platform.deviceWidth;
+const deviceHeight = PLATFORM.deviceHeight;
```
thehobbit85: This would crash the app since "PLATFORM" is undefined.
But this file is not imported anywhere so why not remove it completely?

src/theme/variables/material.js
```diff
@@ -2,8 +2,8 @@ import color from 'color';
...
+const deviceWidth = PLATFORM.deviceWidth;
```
thehobbit85: Same as above

src/theme/variables/commonColor.js
```diff
@@ -1,11 +1,11 @@
 import color from 'color';
```
thehobbit85: This file is not imported anywhere so why not remove it completely?

src/theme/variables/platform_old.js
```diff
@@ -1,10 +1,10 @@
 import color from 'color';
```
thehobbit85: This file is not imported anywhere so why not remove it completely?

PR #421, opened 2018-02-02
Fix: show token code in exchange scene drop down header

vsashyn:
Exchange scene drop down header doesn’t show token codes (https://app.asana.com/0/361770107085503/498144750192718/f)


src/modules/UI/components/FlipInput/CryptoExchangeFlipInputWrapperComponent.js
```diff
@@ -78,9 +78,10 @@ export class CryptoExchangeFlipInputWrapperComponent extends Component<Props, St
             </View>
         </View>
     }
-    const {guiWallet: {name: guiWalletName, currencyCode}} = this.props
+    const {guiWallet: {name: guiWalletName},
```
paullinator: This syntax is rather confusing and hard to read. Please break this up into several lines that more clearly dereference `this.props`

PR #422, opened 2018-02-02
Reset state of modal when submitting password for confirmation

ahujsak:
Password was never stored in Redux, however the design of our modals allowed it to stay in local state even after the modal was closed. This fixes that.

### DIFF
--- a/src/modules/UI/scenes/Settings/components/ConfirmPasswordModal.ui.js
+++ b/src/modules/UI/scenes/Settings/components/ConfirmPasswordModal.ui.js
@@ -23,11 +23,18 @@ type ConfirmPasswordModalProps = {
 type State = {
   confimPassword: string
 }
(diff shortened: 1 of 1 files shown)


src/modules/UI/scenes/Settings/components/ConfirmPasswordModal.ui.js
```diff
@@ -23,12 +23,19 @@ type ConfirmPasswordModalProps = {
...
 }
-export default class CryptoExchangeConfirmTransactionModal extends Component<ConfirmPasswordModalProps, State> {
+export default class ConfirmPasswordModal extends Component<ConfirmPasswordModalProps, State> {
   componentWillMount () {
     this.setState({
       confimPassword: ''
     })
   }
+  componentWillReceiveProps (nextProps: ConfirmPasswordModalProps) {
+    if (nextProps.showModal) {
```
wolverineks: Please delete the password as soon as the modal is dismissed

PR #423, opened 2018-02-02
Added error handling on Slider Component. Added the error handling on…

amanalili:
… the the send confirmation and exchange scene catch. Put spring animation on the slider.

Please check first the animation of the slider, and comment if wanted to retain or revert back to normal.


amanalili:
Also made the sliding to be on the farthest left before it activates. Meaning you need to slide it really on the left before it activates. Check if it needs to have some reserved area to activate it.

amanalili:
Well there are no error handling states on both the send confirmation and exchange scene. And I don't want to touch the reducer of the send confirmation cause I do not see any extensibility. 

Unless you got a state on both the send confirmation and exchange scene that will emit that the async catches errors. By far this is my most logical and extensible solution I have.

paullinator:
Closing this as it was implemented by Eliran in another PR

PR #424, opened 2018-02-02
Add spinner and gate next button to not work when doing a make spend.

ahujsak:


src/connectors/scene/CryptoExchangeSceneConnector.js
```diff
@@ -102,7 +101,11 @@ export const mapStateToProps = (state: State): CryptoExchangeSceneComponentState
     forceUpdateGuiCounter: state.cryptoExchange.forceUpdateGuiCounter,
     showWalletSelectModal: state.cryptoExchange.walletListModalVisible,
     showConfirmShiftModal: state.cryptoExchange.confirmTransactionModalVisible,
-    showNextButton
+    showNextButton,
+    // Not Sure why flow does not like the gettingTransaction type.
+    // There is no reason to fail it.
+    // $FlowFixMe
+    gettingTransaction: state.cryptoExchange.gettingTransaction
```
paullinator: You need to add `gettingTransaction` to ReduxTypes.js. That will fix the Flow error

src/modules/UI/scenes/CryptoExchange/CryptoExchangeSceneComponent.js
```diff
@@ -191,7 +192,11 @@ export class CryptoExchangeSceneComponent extends Component<Props, State> {
             />
             <View style={style.shim} />
             <View style={style.actionButtonContainer} >
-              {this.renderButton()}
+            <PrimaryButton
```
paullinator: What's the purpose of bypassing `renderButton()`? Seems like you now will show the user the button even if `showNextButton` is `false`. Also, `showNextButton` is no longer used but it is still in the Flow type `CryptoExchangeSceneComponentStateProps`

paullinator: The `processingFlag` doesn't appear to prevent the `PrimaryButton` from showing or being tapped.

src/connectors/scene/CryptoExchangeSceneConnector.js
ahujsak: Fixed and pushed . Thanks

src/modules/UI/scenes/CryptoExchange/CryptoExchangeSceneComponent.js
ahujsak: @paullinator per my note in Asana, it seemed like we should keep the button up all the time, It is a little confusing to go from , not there, to spinning each time we type something. etc.. I have modified the button to not generate a click action if the processingFlag is true

paullinator: It's confusing as well for a button to be there but with no response. Can you do an Alert that pops up the appropriate reason why the button doesn't work?

PR #425, opened 2018-02-02
Android: handleBackButton to exit app

vsashyn:
Back button doesn't work on the login screen (to get back to the previous screen or possibly leave the app)  (https://app.asana.com/0/361770107085503/482586483170717/f)


src/modules/UI/scenes/WalletList/components/HwBackButtonHandler/HwBackButtonHandler.android.js
```diff
@@ -0,0 +1,24 @@
+import { ToastAndroid } from 'react-native'
```
vsashyn: `Eslint` forces us to separate Android and iOS specific cases to corresponding files. Created same file iOS handler  for consistency, but it will not be used.

PR #427, opened 2018-02-02
Fix android permission request

wolverineks:


src/modules/UI/scenes/Scan/Scan.ui.js
```diff
@@ -145,16 +108,7 @@ export default class Scan extends Component<Props, State> {
...
   }
 
-  setCameraPermission = (cameraPermission: boolean) => {
-    this.setState({
-      cameraPermission
-    })
-  }
-
-  _onToggleTorch = () => {
-    this.props.toggleEnableTorch()
```
paullinator: This should be added back for torch support. Please test to make sure it works.

PR #428, opened 2018-02-02
 Exchange modal change icon more closely to designer's design. Close i…

amanalili:
…con on the right adjusted

Added a cancel button/text bottom of the exchange confirm modal


amanalili:
Icon picked is as closes as per the design.

PR #433, opened 2018-02-02
Slight adjustments to layout on CreateWallet scenes to utilize space

kylanhurt:
I made slight adjustments to the spacing for search results in the CreateWallet process in order to utilize space more efficiently.

Asana Task: none

### DIFF
--- a/src/modules/UI/scenes/CreateWallet/CreateWalletSelectFiat.ui.js
+++ b/src/modules/UI/scenes/CreateWallet/CreateWalletSelectFiat.ui.js
@@ -102,7 +102,7 @@ export class CreateWalletSelectFiat extends Component<CreateWalletSelectFiatProp
       return (entry.label.toLowerCase().indexOf(this.state.searchTerm.toLowerCase()) >= 0)
     })
     const keyboardHeight = this.props.dimensions.keyboardHeight || 0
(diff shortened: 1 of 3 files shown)


src/modules/UI/scenes/CreateWallet/CreateWalletSelectFiat.ui.js
```diff
@@ -102,7 +102,7 @@ export class CreateWalletSelectFiat extends Component<CreateWalletSelectFiatProp
       return (entry.label.toLowerCase().indexOf(this.state.searchTerm.toLowerCase()) >= 0)
     })
     const keyboardHeight = this.props.dimensions.keyboardHeight || 0
-    const searchResultsHeight = stylesRaw.usableHeight - keyboardHeight - 50 // substract button area height and FormField height
+    const searchResultsHeight = stylesRaw.usableHeight - keyboardHeight - 36 // substract button area height and FormField height
```
ahujsak: I am not a fan of magic numbers. You have the comment here to tell me what you are subtracting. Would like to see 2 lines between 103 and 104 that is 
const buttonAreaHeight = stylesRaw.usableHeight
const FormFieldHeight = 36 
then change the line to be 
const searchResultsHeight = buttonAreaHeight - keyboardHeight - FormFieldHeight 
this way you don't need the comment to explain, each variable is clean and clear and there are no more magic numbers

PR #435, opened 2018-02-03
Add contacts loader, new permission module

wolverineks:


paullinator:
Can you give a high level view of what this aims to accomplish? Was there an Asana task for this?

PR #437, opened 2018-02-05
Notification when an invalid password is submitted. Account Setting screens

amanalili:


src/modules/UI/scenes/Settings/action.js
```diff
@@ -94,6 +96,9 @@ export const checkCurrentPassword = (arg: string) => async (dispatch: Dispatch,
   const account = CORE_SELECTORS.getAccount(state)
   const isPassword = await account.checkPassword(arg)
   dispatch(SETTINGS_ACTIONS.setSettingsLock(!isPassword))
+  if (!isPassword) {
+    setTimeout(() => Alert.alert(s.strings.fragmet_invalid_password), 200)
```
ahujsak: @amanalili why are we doing a set timeout here?. couldn't we just fire the alert?.

PR #438, opened 2018-02-05
Removed timout when component unmounts for static modal.

ahujsak:


src/components/modals/StaticModalComponent.js
```diff
@@ -16,13 +16,17 @@ type Props = {
   cancel(): void
 }
 class StaticModalComponent extends Component<Props> {
+  reset: number
   componentDidMount () {
     if (this.props.modalDismissTimerSeconds) {
-      setTimeout(() => {
+      this.reset = setTimeout(() => {
```
wolverineks: Have you confirmed what the return type `setTimeout` is?
I think it's an object, but `this.reset` is declared `number`

src/components/modals/StaticModalComponent.js
```diff
@@ -16,13 +16,17 @@ type Props = {
...
         this.props.cancel()
       }, this.props.modalDismissTimerSeconds * 1000)
     }
   }
+  componentWillUnmount () {
+    clearInterval(this.reset)
```
wolverineks: Isn't `this.reset` a timeout?
Have you confirmed that this works?

src/components/modals/StaticModalComponent.js
wolverineks: why is this in `componentDidMount` and not `constructor`?

src/components/modals/StaticModalComponent.js
```diff
@@ -16,13 +16,17 @@ type Props = {
   cancel(): void
 }
 class StaticModalComponent extends Component<Props> {
+  reset: number
```
wolverineks: if this property is set in `componentDidMount`, then shouldn't this be `reset?: number`

PR #441, opened 2018-02-05
Create both ethereum and bitcoin wallet at account creation using phone locale currency

paullinator:


src/modules/Login/action.js
```diff
@@ -62,10 +65,20 @@ export const initializeAccount = (account: AbcAccount, touchIdInfo: Object) => a
...
-      const walletType = Constants.ETHEREUM_WALLET
-      const fiatCurrencyCode = Constants.USD_FIAT
-      const abcWallet = await ACCOUNT_API.createCurrencyWalletRequest(account, walletType, {name: walletName, fiatCurrencyCode})
+      const ethWalletName = s.strings.string_first_ethereum_wallet_name
+      const btcWalletName = s.strings.string_first_bitcoin_wallet_name
+      const ethWalletType = Constants.ETHEREUM_WALLET
+      const btcWalletType = Constants.BITCOIN_WALLET
+      let fiatCurrencyCode = Constants.USD_FIAT
+      if (
+        localeInfo.currencyCode &&
```
thehobbit85: So technically testing for not null/false/undefined is already covered when doing the "typeof" check but it doesn't matter that much

PR #450, opened 2018-02-06
Implement different version of passing sourceWallet to changeMiningFee

kylanhurt:
The purpose of this task is to base the custom mining fee type based on which scene it is coming from, since the exchange scene and sendConfirmation scenes both have source wallets. It will now look at selectedWallet from SendConfirmation scene and fromWallet for the cryptoExchange scene.

Asana Task: https://app.asana.com/0/361770107085503/498144750192722/f


ahujsak:
Verified Build on IOS, and that it does what it should . Android builds successfully with travis, so merging it in

PR #452, opened 2018-02-07
Fix platform imports

wolverineks:


src/theme/variables/material.js
```diff
@@ -2,8 +2,8 @@ import color from 'color';
 
 import { Platform, PixelRatio } from 'react-native';
 
-const deviceHeight = platform.deviceHeight;
-const deviceWidth = platform.deviceWidth;
+const deviceHeight = PLATFORM.deviceHeight;
```
thehobbit85: This would crash the app since "PLATFORM" is undefined here but this file in never used so I think It can be deleted

src/theme/variables/material.js
```diff
@@ -2,8 +2,8 @@ import color from 'color';
...
+const deviceWidth = PLATFORM.deviceWidth;
```
thehobbit85: same as above

src/theme/variables/commonColor.js
```diff
@@ -1,11 +1,11 @@
 import color from 'color';
```
thehobbit85: This file in never used so I think It can be deleted

PR #464, opened 2018-02-08
Make the plus(+) sign appear only to the newer categories that user c…

amanalili:
…reates


amanalili:
Just an FYI, the + sign never really does anything. It has no function to begin with. Its just a text as of now.

PR #466, opened 2018-02-08
forced the input for the currency code to be uppercase

ahujsak:
Some android devices are not responding to autocapitalize

this eliminates the problem

Asana task : https://app.asana.com/0/361770107085503/545202805762044

### DIFF
--- a/src/modules/UI/scenes/EditToken/EditToken.ui.js
+++ b/src/modules/UI/scenes/EditToken/EditToken.ui.js
@@ -184,9 +184,12 @@ export default class EditToken extends Component<EditTokenComponentProps, State>
   }
 
   onChangeCurrencyCode = (input: string) => {
-    input.toUpperCase()
+    const forcedUpperCase = input.toUpperCase()
+    /* forcedUpperCase needed to defend against React Native bug
+      https://github.com/facebook/react-native/issues/11776
+    */
     this.setState({
-      currencyCode: input.substring(0, 5)
+      currencyCode: forcedUpperCase.substring(0, 5)
     })
   }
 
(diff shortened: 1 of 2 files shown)


src/modules/UI/scenes/EditToken/EditToken.ui.js
```diff
@@ -184,6 +184,7 @@ export default class EditToken extends Component<EditTokenComponentProps, State>
   }
 
   onChangeCurrencyCode = (input: string) => {
+    input.toUpperCase()
```
wolverineks: Have you confirmed that this works?

[MDN String.prototype.toUpperCase()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase)

wolverineks: If you put code it due to a bug in 3rd party code, please include a link to documentation
For instance: [TextInput autoCapitalize ignored on Android](https://github.com/facebook/react-native/issues/11776)

src/modules/UI/scenes/EditToken/EditToken.ui.js
```diff
@@ -184,8 +184,12 @@ export default class EditToken extends Component<EditTokenComponentProps, State>
   }
 
   onChangeCurrencyCode = (input: string) => {
+    const forcedUpperCase = input.toUpperCase()
+    /* forcedUpperCase needed to defend against React Native bug
+      https://github.com/facebook/react-native/issues/11776
+    */
     this.setState({
-      currencyCode: input.substring(0, 5)
+      currencyCode: forcedUpperCase.substring(0, 5)
```
wolverineks: Please address the problem as close to the root as possible, so if it ever gets fixed, we only have to remove a single shim.
`src/components/materialWrappers/Input.js`
Comment out the default prop for `keyboardType`, and include a comment/link to the github pull request that identifies the bug

PR #467, opened 2018-02-09
Hook up the PIN login toggle switch

swansontec:


wolverineks:
Merging even though this doesnt pass travis.
Travis iOS break because of yarn

PR #473, opened 2018-02-09
Fix intl formatNumber to use separators from device info

vsashyn:
Fix showstoppers https://app.asana.com/0/361770107085503/493048970757179/f , https://app.asana.com/0/361770107085503/553467619718332/f


wolverineks:
Q/A - passes iOS

PR #482, opened 2018-02-13
Re-adjust WalletListOptions dropdown vertical position to prevent cutoff

kylanhurt:
The purpose of this task is to prevent the WalletListOptions dropdown menu from going off the screen when the wallet item is toward the bottom of the screen.

Asana Task: https://app.asana.com/0/361770107085503/553039148202323/f


src/modules/UI/components/MenuDropDown/MenuDropDown.ui.js
```diff
@@ -36,11 +69,22 @@ export default class MenuDropDown extends Component<Props> {
   }
 
   render () {
+    const deviceHeight = PLATFORM.deviceHeight
+    const verticalBuffer = ((this.state.pageY + this.state.height - 8) > PLATFORM.deviceHeight) ? 12 : 0
+    const lowerLimitOfMenu = this.state.pageY + this.state.height
+    const offset = lowerLimitOfMenu - deviceHeight
+    const newPageY = this.state.pageY - offset - verticalBuffer
     let optionsStyle = {}
     if (this.props.rightSide) {
```
thehobbit85: "rightSide" is not defined in the props type and will never be true.

kylanhurt: @thehobbit85 That property was already in there so I didn't touch it. I can have it removed, if you'd like.

thehobbit85: That's not my concern, the problem I see is that the "if" condition will never be true and therefore this would never happen:
optionsStyle = {
   left: '1%',
   top: this.state.pageY  
 }
If that's ok then go ahead and delete it

kylanhurt:
@thehobbit85 Thank you for clarification. Changes have been made.

PR #491, opened 2018-02-15
Add Contacts Loader

wolverineks:
Enable requesting Contacts permission only when needed on select scenes

* Contacts Permission requested on select scenes
* Permissions stored in redux
* ContactsLoader loads contacts if granted permission
* Contacts stored in redux


thehobbit85:
I get why you want to keep the "Permissions" in the store but why keep the Contacts there as well?
I don't see the point in having all the Contacts in redux when they are easily accessible through a simple API call when needed.

PR #496, opened 2018-02-15
Checking available shapeshift tokens is dynamic

ahujsak:
• Calling correct ShapeShift API
• removed static array and function
• update calls frequently when wallet is changed.


wolverineks:
I don't think hardcoding the supported tokens to shapeshift is a long term solution.
Let's coordinate with William on getting this functionality in the core.

PR #497, opened 2018-02-16
Change string for custom gasPrice to be "Gas Price (GWei)"

paullinator:
Requires update to 0.8.0 of ethereum library which actually uses GWei as the gasPrice parameter

PR https://github.com/Airbitz/edge-currency-ethereum/pull/7 should first be reviewed, merged, and npm published before this PR is accepted. Otherwise we will get a build error due to missing npm version 0.8.0 of the ethereum library

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "edge-react-gui",
-  "version": "1.0.5",
+  "version": "1.0.6",
   "private": true,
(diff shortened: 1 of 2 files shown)


paullinator:
Merging this as the real change was in the ethereum 0.8.0 dependency which was reviewed and merged.

PR #498, opened 2018-02-16
Change View Component on the MenuDropDownUi Component to fix WalletLi…

amanalili:
…stOptions error

Commented the whole render code and add the fix. There are comments on the code same as written here.

MenuOptions component should be followed by MenuOption component because MenuOptions pass props into its children, this case props is inherited by View and not MenuOption. Workaround is to create a new React class with the View and custom style and pass the props to MenuOption component.

https://app.asana.com/0/361770107085503/549547415517480


amanalili:
Seems refs props and onLayout props work on custom components. Just put those on the MenuOptions.

PR #503, opened 2018-02-20
Remove unnecessary text in Exchange scene dropdown headers

kylanhurt:
This is a slight reversion of a previous commit, because the designs I was looking at were old and some parts are redundant (ie the crypto type in the header, since there is already a denomination, crypto logo, and crypto code in the box).

Asana Task: https://app.asana.com/0/361770107085503/475629743533054/f

### DIFF
--- a/src/modules/UI/components/Header/Component/WalletNameHeader.ui.js
+++ b/src/modules/UI/components/Header/Component/WalletNameHeader.ui.js
@@ -8,8 +8,7 @@ type Props = {
   styles: {
     textStyles: Array<{}>
   },
-  name: string,
-  denomination: string
+  name: string
 }
 
 class WalletNameHeader extends React.Component<Props> {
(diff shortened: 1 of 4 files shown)


src/modules/UI/components/Header/Component/WalletNameHeader.ui.js
```diff
@@ -8,24 +8,19 @@ type Props = {
...
     const name = this.props.name
-    const denomination = this.props.denomination
-    const directionPrefix = this.props.walletDirectionString ? this.props.walletDirectionString : ''
 
     return (
       <View style={style.headerNameContainer}>
         <Text style={textStyles} ellipsizeMode={'middle'} numberOfLines={1}>
-          {directionPrefix + ' ' + name}
-          <Text style={[style.cCode, ...textStyles]}> ({denomination})</Text>
+          {name}
```
paullinator: You removed the denomination from the header. I don't believe this should be removed

PR #512, opened 2018-02-28
Allen/small fixes

ahujsak:
Fixes :  https://app.asana.com/0/361770107085503/548710133686877
Fixes : https://app.asana.com/0/361770107085503/572954516359943

### DIFF
--- a/src/modules/UI/scenes/Otp/OtpSettingsSceneComponent.js
+++ b/src/modules/UI/scenes/Otp/OtpSettingsSceneComponent.js
@@ -56,7 +56,7 @@ export default class OtpSettingsScene extends Component<OtpSettingsSceneProps, S
     }
     this.setState({
       showMessageModal: true,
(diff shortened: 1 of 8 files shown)


src/modules/UI/scenes/Otp/OtpSettingsSceneComponent.js
```diff
@@ -54,7 +56,8 @@ export default class OtpSettingsScene extends Component<OtpSettingsSceneProps, S
     }
     this.setState({
       showMessageModal: true,
-      messageModalMessage: s.strings.otp_enabled_modal_part_one + '' + s.strings.otp_enabled_modal_part_two
+      messageModalMessage: s.strings.otp_enabled_modal_part_one + '' + s.strings.otp_enabled_modal_part_two,
```
paullinator: The `messageModalMessage` state seems unnecessary here as only the `messageModalComponent` would get rendered. If so, set to NULL

PR #515, opened 2018-03-02
Allen/flow and excahnge fix

ahujsak:
This upgrades flow, 
Changes some files so flow passed 
fixed bug where exchange reducer would reset the available tokens to []

### DIFF
--- a/src/modules/UI/scenes/ChangeMiningFee/components/CustomFees/CustomFeesModalConnector.js
+++ b/src/modules/UI/scenes/ChangeMiningFee/components/CustomFees/CustomFeesModalConnector.js
@@ -10,11 +10,16 @@ import { updateMiningFees } from '../../../SendConfirmation/action'
 import { getCustomNetworkFee } from '../../../SendConfirmation/selectors.js'
 import { CLOSE_MODAL_VALUE } from '../../../WalletList/components/WalletOptions/action'
 import CustomFeesModal from './CustomFeesModal.ui'
-import type { CustomFees } from './CustomFeesModal.ui'
+import type {
+  CustomFees,
+  CustomFeesModalStateProps,
+  CustomFeesModalOwnProps,
+  CustomFeesModalDispatchProps
+} from './CustomFeesModal.ui'
 
-const mapStateToProps = (state: State, ownProps: any) => {
+const mapStateToProps = (state: State, ownProps: CustomFeesModalOwnProps): CustomFeesModalStateProps => {
   const wallet = ownProps.sourceWallet
-  let customFeeSettings = []
+  let customFeeSettings: Array<string> = []
   if (_.has(wallet, 'currencyInfo.defaultSettings.customFeeSettings')) {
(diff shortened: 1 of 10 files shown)


src/modules/UI/scenes/ChangeMiningFee/components/CustomFees/CustomFeesModalConnector.js
```diff
@@ -10,9 +10,9 @@ import { updateMiningFees } from '../../../SendConfirmation/action'
 import { getCustomNetworkFee } from '../../../SendConfirmation/selectors.js'
 import { CLOSE_MODAL_VALUE } from '../../../WalletList/components/WalletOptions/action'
 import CustomFeesModal from './CustomFeesModal.ui'
-import type { CustomFees, CustomFeesModalOwnProps } from './CustomFeesModal.ui'
+import type { CustomFees } from './CustomFeesModal.ui'
 
-const mapStateToProps = (state: State, ownProps: CustomFeesModalOwnProps) => {
+const mapStateToProps = (state: State, ownProps: any) => {
```
paullinator: This is going backwards. Without an explanation, I wouldn't want to approve this.

src/reducers/permissions/actions.js
```diff
@@ -16,8 +16,8 @@ export const requestPermission = (permission: Permission) => (dispatch: Dispatch
       .catch(console.log)
   })
 }
-
-export const updatePermissions = (permissions: { [Permission]: PermissionStatus }) => ({
+// $flowFixMe
+export const updatePermissions = (permissions: { [Permission]: any }) => ({
```
paullinator: Also don't like this change moving to `any`

src/modules/UI/scenes/ChangeMiningFee/components/CustomFees/CustomFeesModalConnector.js
paullinator: I submitted a commit that fixes this without using `any`. Please fix action.js below.

src/reducers/permissions/actions.js
```diff
@@ -12,12 +12,12 @@ export const UPDATE_PERMISSIONS = PREFIX + 'UPDATE_PERMISSIONS'
...
   RNPermissions.check(permission).then(permissionStatus => {
     request(permission)
-      .then(status => dispatch(updatePermissions({ [permission]: status })))
+      .then(status => dispatch(updatePermissions({ [permission]: status }))) // eslint-disable-line no-undef
       .catch(console.log)
   })
 }
 // $flowFixMe
-export const updatePermissions = (permissions: { [Permission]: any }) => ({
+export const updatePermissions = (permissions: Object) => ({
```
paullinator: This is kind of lame as you used to have a more specific PermissionStatus object type. I realize it's not as bad as `any` but using `Object` is a bit of a cop out. I'll let it slide but I'll be asking later why you couldn't resolve this.

PR #517, opened 2018-03-04
Offset adjustment even on big screens like iPad

amanalili:
https://app.asana.com/0/361770107085503/569927330298854

### DIFF
--- a/src/styles/components/StaticModalStyle.js
+++ b/src/styles/components/StaticModalStyle.js
@@ -7,15 +7,15 @@ const screenDimensions = {
   height: Dimensions.get('window').height,
   width: Dimensions.get('window').width
 }
-const OFFSET_HACK = -19
+const OFFSET_HACK = -((screenDimensions.width * 0.1) - 18.5)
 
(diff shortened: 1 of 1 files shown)


src/styles/components/StaticModalStyle.js
```diff
@@ -7,15 +7,15 @@ const screenDimensions = {
...
+const OFFSET_HACK = -((screenDimensions.width * 0.1) - 18.5)
 
 const StaticModalStyle = {
   container: {
     position: 'absolute',
     top: OFFSET_HACK,
     left: OFFSET_HACK,
-    width: '100%',
-    height: '100%',
+    width: screenDimensions.width * 2,
```
paullinator: This change looks concerning. Please explain the reasoning behind the `width * 2` math.

amanalili: Agreed about being a concern. Width and Height hack so that any top/left offset will still render the grey background without any gaps. I'll close this PR for now and try to implement the modal correctly so that no hacks is needed.

PR #522, opened 2018-03-06
Add ability to chose legacy address

ahujsak:
• BCH and LTC
Asana task https://app.asana.com/0/361770107085503/574671563444725


src/modules/UI/scenes/Request/action.js
```diff
@@ -1,10 +1,10 @@
 import * as CORE_SELECTORS from '../../../Core/selectors.js'
 import * as WALLET_API from '../../../Core/Wallets/api.js'
 import * as UI_SELECTORS from '../../../UI/selectors.js'
+import * as Constants from '../../../../constants/indexConstants.js'
```
wolverineks: only use named imports

src/modules/UI/scenes/Request/reducer.js
```diff
@@ -1,5 +1,5 @@
 import * as ACTION from './action'
-
+import * as Constants from '../../../../constants/indexConstants.js'
```
wolverineks: only use named imports

src/modules/UI/scenes/scenesReducer.test.js
```diff
@@ -42,6 +42,11 @@ test('initialState', () => {
         publicAddress: ''
       }
     },
+    requestType: {
+      useLegacyAddress: false,
+      receiveAddress: {},
```
wolverineks: do not store duplicated data in redux
this data is stored elsewhere

src/reducers/CryptoExchangeReducer.js
```diff
@@ -106,6 +106,7 @@ function cryptoExchangerReducer (state = initialState, action) {
       return result
     }
     case Constants.UPDATE_SHIFT_TRANSACTION:
+      console.log('What')
```
wolverineks: come on dude, read your code before submitting a PR

src/reducers/RequestTypeReducer.js
```diff
@@ -0,0 +1,30 @@
+import * as Constants from '../constants/indexConstants.js'
```
wolverineks: only use named imports

src/reducers/RequestTypeReducer.js
```diff
@@ -0,0 +1,30 @@
...
+
+const initialState = {
```
wolverineks: please add tests when adding new reducers
follow the existing reducer tests as an example

src/reducers/RequestTypeReducer.js
```diff
@@ -0,0 +1,30 @@
...
+  useLegacyAddress: false,
+  receiveAddress: {},
+  uniqueLegacyAddress: false
+}
+export const requestType = (state = initialState, action) => {
+  switch (action.type) {
+    case Constants.UPDATE_RECEIVE_ADDRESS_SUCCESS:
+      let uniqueLegacy = true
+      if (action.data.receiveAddress.legacyAddress) {
+        uniqueLegacy = action.data.receiveAddress.publicAddress === action.data.receiveAddress.legacyAddress
```
wolverineks: what does this do?
what is this used for?

src/modules/UI/scenes/Request/action.js
wolverineks: please flow all files before submitting a PR

src/modules/UI/scenes/Request/reducer.js
wolverineks: please flow files before submitting a PR

src/reducers/RequestTypeReducer.js
wolverineks: please flow files before submitting a PR

wolverineks:
Additionally, there are multiple ways to select a wallet
This PR does not update the address data when selecting a wallet from the wallet list screen
This will crash the app under certain cirumstances

src/connectors/components/HeaderMenuRequestConnector.js
```diff
@@ -0,0 +1,59 @@
+// @flow
+import { connect } from 'react-redux'
+
+import * as actions from '../../actions/indexActions'
+import * as Constants from '../../constants/indexConstants'
+import s from '../../locales/strings.js'
```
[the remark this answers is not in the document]
swansontec: We are not changing the way localization works today.

src/connectors/components/HeaderMenuRequestConnector.js
```diff
@@ -0,0 +1,59 @@
+// @flow
+import { connect } from 'react-redux'
+
+import * as actions from '../../actions/indexActions'
```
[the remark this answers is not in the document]
swansontec: I would like to see indexActions go away and we just import actions from their respective homes, since these are more logic & code.

src/connectors/components/HeaderMenuRequestConnector.js
```diff
@@ -0,0 +1,59 @@
...
+import * as Constants from '../../constants/indexConstants'
+import s from '../../locales/strings.js'
+import type { Dispatch, State } from '../../modules/ReduxTypes.js'
+import { openHelpModal } from '../../modules/UI/components/HelpModal/actions'
+import LinkedComponent from '../../modules/UI/components/MenuDropDown/MenuDropDown.ui'
```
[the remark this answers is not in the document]
swansontec: MenuDropDown.ui is using a default export? This is definitely something we are trying to avoid.

src/connectors/components/HeaderMenuRequestConnector.js
```diff
@@ -0,0 +1,59 @@
...
+        break
+      case Constants.HELP_VALUE:
+        dispatch(openHelpModal())
+        break
+    }
+  }
+  // nextScreen: () => dispatch(actions.nextScreen())
+})
+
+export default connect(mapStateToProps, mapDispatchToProps)(LinkedComponent)
```
swansontec: `export default` is something we agreed to avoid.

src/reducers/RequestTypeReducer.js
```diff
@@ -0,0 +1,30 @@
...
+    case Constants.USE_REGULAR_REQUEST_ADDRESS:
+      return { ...state, useLegacyAddress: false }
+    case Constants.USE_LEGACY_REQUEST_ADDRESS:
+      return { ...state, useLegacyAddress: true }
+    default:
+      return state
+  }
+}
+
+export default requestType
```
swansontec: Let's avoid default exports.

src/connectors/components/HeaderMenuRequestConnector.js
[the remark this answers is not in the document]
ahujsak: William and I talked, and are in agreement on named imports for actions. Fixed

src/connectors/components/HeaderMenuRequestConnector.js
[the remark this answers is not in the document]
ahujsak: This is how we use localization for entire app 
import s from '../../locales/strings.js'. in fact 75 times. This is consistent , not going to change unless a new direction is something we decide on in an engineering meeting and come to a consensus.

src/connectors/components/HeaderMenuRequestConnector.js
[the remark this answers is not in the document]
ahujsak: I am assuming this is the use of LinkedComponent, this has been corrected to reflect the right name and to also remove the default from MenuDropDown.ui

src/modules/UI/scenes/scenesReducer.test.js
ahujsak: This is in the scene reducer test and is consistent with where the redux data is stored and how the test works. Please explain what is wrong . This was necessary to pass tests.

src/reducers/CryptoExchangeReducer.js
ahujsak: oversight, fixed.

src/reducers/RequestTypeReducer.js
ahujsak: fixed

ahujsak:
**Regarding Named imports int he following** 
```
Import * as Constants
Import s from 
Import * as Styles
Import THEME from 
```
are all conventions that we have been using though out the code base for quite some time, and are not going to be changed unless there is something agreed upon in an engineering meeting. 

I spoke with William about this. We did agree that the `import * from actions` should be named imports. 

I do not want to spend a ton of time defending the code practice but here are the highlights. 

**Import * as Constants**
There are constants littered all over the code base and I have been slowly removing them and putting them into the constants folder. The main reason is that we need to be able to access these from anywhere, and they need to be organized. 

The convention is easy for both intellisense in the IDE .. While coding I know the constants “package’ is located at the root of the src project. Anytime I need a constant I can type in Constants+ period, and get the intellisense to show me all my options. This saves having to go all over and remember what constant is in which file I need. 

**Import s from ‘’strings**
This is how we have been doing localization throughout the whole app. If this is going to change it needs to be a full on refactor.  used 75 times in the code base.

Import THEME from 
This pertains to mostly styles.  99% of the time, The way it was architected is as a nested object so items like Colors etc are accessed through THEME.COLORS.PRIMARY.  It is primarily used in Style objects, so when creating a style file, to deconstruct the THEME outside of the style to reference the items in the big object is a bunch of unnecessary coding. 

Import * as Styles
The styles have been set up so that nested styles make it easier to set up style sheets and re-use styles for components. 

An example of this is 
```
import * as Styles from '../../../styles/indexStyles'
const style = {
 container: Styles.Screen,
 modal: Styles.TwoButtonModal,
}
```
Importing the style makes it faster so we don’t have to hunt and peck to fins the specific style. 

An argument can be made that is by using named imports it is more specific, however I find it easier to read code when I know where something came from . 

If I am looking at a style for a color , and I see THEME.COLOR.Primary color, I know in the code review we are getting the color from the right place. 

Same thing goes for Constants.I_AM_CONSTANT.. this lets me know the constant is set in a centralized place, and is also a visual queue that I am using a constant. 

Re-iterating what I said above about Styles. The word Styles lets me know I am referencing a standard style.. visual queues make it easier to read code. 

Per my conversation with @swansontec we are going to leave Constants, s for strings for localization, Styles and Themes will not be changed for the time being

src/connectors/components/HeaderMenuRequestConnector.js
```diff
@@ -0,0 +1,59 @@
+// @flow
+import { connect } from 'react-redux'
+
+import * as actions from '../../actions/indexActions'
+import * as Constants from '../../constants/indexConstants'
```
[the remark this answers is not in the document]
ahujsak: See comments about named below

src/modules/UI/scenes/scenesReducer.test.js
ahujsak: ok, I see what you mean, receive address is not needed actually.. it is removed now

src/connectors/components/HeaderMenuRequestConnector.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+  const help = {
+    label: s.strings.string_help,
+    key: s.strings.string_help,
+    value: {
+      title: Constants.HELP_VALUE
+    }
+  }
+  const data = uniqueLegacyAddress ? [addressToggle, help] : [help]
```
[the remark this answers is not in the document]
ahujsak: Give what a better name?. address toggle is what the button does. If it is available , the button toggles between address options.

src/connectors/components/HeaderMenuRequestConnector.js
```diff
@@ -0,0 +1,59 @@
...
+      ...Styles.MenuDropDownStyleHeader,
+      icon: { ...Styles.MenuDropDownStyle.icon, color: THEME.COLORS.WHITE }
+    },
+    data,
+    rightSide: true
+  }
+}
+
+export const mapDispatchToProps = (dispatch: Dispatch) => ({
+  onSelect: (value: Object) => {
```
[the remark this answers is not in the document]
ahujsak: No. when the MenuDropDown selects something it passed an object. We need to evaluate the object to see what happens. Any time there are more than 2 options for a value a switch statement is preferable to if, else if , etc...

src/modules/ReduxTypes.js
```diff
@@ -144,6 +144,11 @@ export type State = {
           message: string,
           buttons: Array<{ title: string, message: string }>
         }
+      },
+      requestType: {
+        useLegacyAddress: boolean,
+        receiveAddress: {},
```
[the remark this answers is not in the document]
ahujsak: removed receiveAdderss, to one of your previous comments, it is redundant state and it was actually unnecessary

src/modules/UI/scenes/Request/Request.ui.js
```diff
@@ -90,6 +94,7 @@ export class Request extends Component<Props, State> {
 
     WALLET_API.getReceiveAddress(abcWallet, currencyCode)
       .then(receiveAddress => {
+        console.log(receiveAddress)
```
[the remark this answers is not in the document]
ahujsak: done, oversight

src/reducers/RequestTypeReducer.js
ahujsak: done

src/reducers/RequestTypeReducer.js
ahujsak: When on the request scene we have a smart menu at the top left. 
If the requesting wallet has a legacy address property , and that legacy address is different than the public address then the address is unique and we need to display options to choose it or the public address in the drop down . 
So, for displaying and using a Legacy address in the request screen we need to know 1. Is there a unique legacy address then we need to give the option of choosing it. If it is chosen the request screen will show the correct QR code and address

src/modules/UI/scenes/Request/reducer.js
ahujsak: flow'd

src/modules/UI/scenes/Request/action.js
ahujsak: flow'd

src/reducers/RequestTypeReducer.js
ahujsak: flow'd

src/connectors/components/HeaderMenuRequestConnector.js
```diff
@@ -0,0 +1,59 @@
...
+    switch (value.title) {
+      case s.strings.title_use_regular_address:
```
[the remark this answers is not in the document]
ahujsak: I have not tested with a different language. This is the pattern we have been following for all localization throughout the app. I am not sure when new languages get added , or when new strings get translated. This is something we should bring up in the next engineering meeting. We should not be pushing new versions to store without all localization being current for all languages.

ahujsak:
@wolverineks  said "Additionally, there are multiple ways to select a wallet
This PR does not update the address data when selecting a wallet from the wallet list screen
This will crash the app under certain cirumstances"

please  elaborate. The functionality with the types reducer is limited to the drop down menu that appears on the request screen, and the display of the legacy address is limited to the request.ui component.  Changing the wallet in other places will not affect that functionality. and it will only update the request screen based on the wallet selection created. I do not see where this will have repercussions. I have tested changing the wallet in the send and exchange scene as well as wallet list and not run into a crash.

src/connectors/components/HeaderMenuRequestConnector.js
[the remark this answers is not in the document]
wolverineks: data

src/connectors/components/HeaderMenuRequestConnector.js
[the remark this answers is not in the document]
wolverineks: can you use something that is not translated as the `case`?

src/connectors/components/HeaderMenuRequestConnector.js
```diff
@@ -0,0 +1,59 @@
...
+  const help = {
+    label: s.strings.string_help,
+    key: s.strings.string_help,
+    value: {
+      title: Constants.HELP_VALUE
+    }
+  }
+  const data = uniqueLegacyAddress ? [addressToggle, help] : [help]
+  return {
+    style: {
```
wolverineks: please do not create new objects in mapStateToProps, this prevents React from skipping renders if the props don't change

src/connectors/components/HeaderMenuRequestConnector.js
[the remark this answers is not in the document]
wolverineks: can the switch statement be moved into the component, and the mapDispatchToProps just injects multiple functions?

wolverineks:
Alan,

Steps to reproduce:
* select BCH/LTC from header,
* navigate to wallet list
* select non BCH/LTC from wallet list row
* navigate to request
* activate dropdown menu

src/connectors/components/HeaderMenuRequestConnector.js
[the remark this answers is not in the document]
ahujsak: fair enough, changed to dropDownButtons

src/connectors/components/HeaderMenuRequestConnector.js
ahujsak: Oh, good catch , I am actually going to fix in HeaderExchangeConnector.js, this is where some of this code was copied from when I started this connector.

src/connectors/components/HeaderMenuRequestConnector.js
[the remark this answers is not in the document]
ahujsak: No, not without a significant refactor. The component is dumb, it gets an array of objects, this creates the different clickable items. This is how we have handled it elsewhere throughout the app.

src/connectors/components/HeaderMenuRequestConnector.js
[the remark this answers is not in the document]
ahujsak: yup, switched to use a constant.

src/connectors/components/HeaderMenuRequestConnector.js
```diff
@@ -37,7 +37,7 @@ export const mapStateToProps = (state: State) => {
   const dropDownButtons = uniqueLegacyAddress ? [addressToggle, help] : [help]
   return {
     style: dropDownStyle,
-    dropDownButtons,
+    data: dropDownButtons,
```
paullinator: This is passing a statically allocated Object as a prop. This will ALWAYS cause invalidation of props and a re-render even if nothing changed. This is especially costly in performance because it's in the header which is always mounted.

ahujsak: So, here is the issue, and I am open to suggestions on a better way to do this, 
The requirement is the drop down menu needs to have two buttons sometimes, other times it is one button. When the second button is visible, its content needs to change. on whether or not this button displays the content to select regular address or legacy address, so unfortunately it is set up to behave the way it needs to.  If there is a better way to do this, I am open, however this is designed to check itself and re-render.

PR #523, opened 2018-03-07
Fix upload logs crash. Set log file size limit

vsashyn:
Asana task: https://app.asana.com/0/361770107085503/559464521227450/f
Fix logs rotation flow. Add size limit for log file to 1 MB.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "zlib": "browserify-zlib"
   },
   "dependencies": {
+    "async-lock": "^1.1.2",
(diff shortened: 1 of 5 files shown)


paullinator:
There is no time to implement Winston as we would need a custom transport since Winston doesn’t support React Native file systems. 

Can you be more specific with what is wrong with this PR? We need it in today.

paullinator:
Based on verbal convo with @thehobbit85, he approved this PR as long as we replace the logging functionality with Winston in the next release.

PR #524, opened 2018-03-07
Modals used on the app is centered regardless of device width and height. Fix static modal component backdrop and how it is visible.

amanalili:
Remove hacks on "top" style on both stylized modal and static modal.

https://app.asana.com/0/361770107085503/569927330298854


amanalili:
Fixed as requested

src/modules/UI/scenes/Otp/OtpSettingsSceneComponent.js
```diff
@@ -138,6 +130,13 @@ export default class OtpSettingsScene extends Component<OtpSettingsSceneProps, S
           </View>
           {this.renderModals(styles)}
         </View>
+        <StaticModalComponent
+          cancel={this.cancelStatic}
+          body={this.state.messageModalMessage}
+          bodyComponent={this.state.messageModalComponent}
+          isVisible={this.state.showMessageModal}
+          modalDismissTimerSeconds={8}
+        />
```
ahujsak: @amanalili I am not sure about this change. The reason there were two if statements is that different modals were rendered at different times. By putting them in the same if statement and return, you are always going to render the TwoButton Modal, and then the Static Modal on top of it. You may not be showing it because of the isVisible property change. The reason for separate return statements was so we would only render the modal when it is needed and not when it is not. ... Please explain the reason for this change.

src/components/modals/StaticModalComponent.js
```diff
@@ -37,20 +37,17 @@ class StaticModalComponent extends Component<Props> {
...
-              end={{ x: 1, y: 0 }}
-              colors={[THEME.COLORS.GRADIENT.DARK, THEME.COLORS.GRADIENT.LIGHT]}
-            >
-              <Icon style={styles.icon} name={Constants.CHECK_CIRCLE} size={styles.iconSize} type={Constants.SIMPLE_ICONS} />
-            </LinearGradient>
-            <View style={styles.bottom}>
-              <View style={styles.bodyRow}>
-                {this.renderMiddle(styles)}
+            <View style={styles.innerBox}>
+              <Gradient style={styles.header}>
```
ahujsak: Please go back to using Linear Gradient instead of Gradient component. The reason the component was not used is that for the static modals the gradient goes in reverse of the header so from left to right it should be dark to light instead of light to dark like the header. The other option is to refactor the Gradient component to take an argument for the direction of the gradient.

src/modules/UI/scenes/Otp/OtpSettingsSceneComponent.js
ahujsak: I tested the app, with this scene and it works, so my concern may not be valid.

amanalili: Using an always visible modal is un natural (or not standard) way of using modals. Also the styling of the modals is unnatural (like no background color) because it is always visible and it is controlled adding/removing the component itself. If you are having concern of modals appearing at the same time then I suggest a clearer states that falsify the state of one modal when another one is truthy.

amanalili:
Re fixed again as per the notes/reviews

paullinator:
@ahujsak can you finish the review on this PR. thx

PR #546, opened 2018-03-14
Change txList balance box styling to prevent icon cutoff

kylanhurt:
The purpose of this task is to migrate the Transaction List balance box (top of scene) away from Flexbox to ensure that there is sufficient space to fit the cryptocurrency logo.

https://app.asana.com/0/361770107085503/532233498493258/f


kylanhurt:
@paullinator I believe that in order for me to use relative / percentages for the layout of the "Balance Box" on the txList then that most likely means I would have to set the font sizes of some of the text to a percentage of the height of the box. I will give it a try to re-submit the PR.

PR #549, opened 2018-03-17
Upgrade android JavascriptCore

paullinator:
Update JSC on android using jsc-android npm module
App launch gets about a 15% performance boost as mentioned in the jsc-android README. 
Launching app on Pixel 2, timings go from 3.75 to 3.2 seconds.

This will also enable upgrading to 64 bit once react-native-fast-crypto is upgraded to 64 bit since this version of JSC already includes 64 bit support.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -152,6 +152,12 @@ if (hasProperty('storeFile')) {
     android.signingConfigs.release.keyPassword   = "password"
 }
 
+configurations.all {
+    resolutionStrategy {
+        force 'org.webkit:android-jsc:r216113'
+    }
+}
+
 dependencies {
     compile project(':react-native-mail')
     compile project(':react-native-smart-splash-screen')
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -22,6 +22,10 @@ allprojects {
             // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
             url "$rootDir/../node_modules/react-native/android"
         }
+        maven {
+            // Local Maven repo containing AARs with JSC library built for Android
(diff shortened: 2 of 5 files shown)


swansontec:
The Travis error looks worrisome, though:

```
:react-native-vector-icons:transformNativeLibsWithStripDebugSymbolForDebug
:react-native-vector-icons:transformNativeLibsWithIntermediateJniLibsForDebug
:app:transformNativeLibsWithMergeJniLibsForDebug FAILED
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':app:transformNativeLibsWithMergeJniLibsForDebug'.
> More than one file was found with OS independent path 'lib/x86/libc++_shared.so'
* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
* Get more help at https://help.gradle.org
BUILD FAILED in 3m 50s
506 actionable tasks: 506 executed
The command "cd android && ./gradlew assembleDebug && ./gradlew assembleRelease" exited with 1.
before_cache.1
```

paullinator:
Interesting. Let me look into that. I was testing the build with assembleRelease but there might be another fix needed for assembleDebug.

paullinator:
@swansontec Ok I figured out the problem. `sed -e` behaves differently on linux than on macOS. This affects Travis but not any of our developer builds.

paullinator:
@swansontec Ok I figured out the problem. `sed -e` behaves differently on linux than on macOS. This affects Travis but not any of our developer builds.

Pushed a commit that changes the `sed` command to work on both MacOS and Linux. Travis verified that the Android build works now. Merging..

PR #553, opened 2018-03-20
Kevin/new feature/load wallet address rebase 00

wolverineks:
Refresh address after receiving funds

* Load receive address on login
* Refresh receive address on new transactions
* Refresh Request scene

https://app.asana.com/0/239353283679855/565600412505655


src/modules/Core/Wallets/action.js
```diff
@@ -29,15 +30,35 @@ export const updateWalletsRequest = () => (dispatch: Dispatch, getState: GetStat
   const account = CORE_SELECTORS.getAccount(state)
   const { activeWalletIds, archivedWalletIds, currencyWallets } = account
 
-  for (const walletId: string of Object.keys(currencyWallets)) {
-    const edgeWallet: EdgeCurrencyWallet = currencyWallets[walletId]
-    if (edgeWallet.type === 'wallet:ethereum') {
-      if (state.ui.wallets && state.ui.wallets.byId && state.ui.wallets.byId[walletId]) {
-        const enabledTokens = state.ui.wallets.byId[walletId].enabledTokens
-        edgeWallet.enableTokens(enabledTokens)
+  getReceiveAddresses(currencyWallets).then((receiveAddresses) => {
```
thehobbit85: Why do you need to update the `receiveAddresses` for all wallets and not just the affected wallet?

src/modules/Login/action.js
```diff
@@ -181,3 +193,21 @@ export const logout = (username?: string) => ({
   type: Constants.LOGOUT,
   data: { username }
 })
+
+const getReceiveAddresses = (currencyWallets: { [id: string]: EdgeCurrencyWallet }): Promise<{ [id: string]: EdgeReceiveAddress }> => {
```
thehobbit85: This is the same code as in `src/modules/Core/Wallets/action.js` so you can just export/import the function from there.
But I don't understand why even get addresses on login? Why not just get them when we request for one instead of loading them in advance?

src/modules/UI/Wallets/action.js
```diff
@@ -105,10 +105,15 @@ export const upsertWallet = (wallet: EdgeCurrencyWallet) => (dispatch: Dispatch,
     return
   }
 
-  dispatch({
-    type: UPSERT_WALLET,
-    data: { wallet }
-  })
+  wallet.getReceiveAddress().then((receiveAddress) =>
```
thehobbit85: Why is it here?
Why do you need to get a new address when on every upsert?

src/modules/Core/Wallets/action.js
wolverineks: "Updating" a wallet includes getting an `EdgeReceiveAddress`
When updating all the wallets, all the wallets need `EdgeReceiveAddress`es
It's more of a requirement due to how I'm "updating" wallets than anything else

PR #555, opened 2018-03-21
Implement AddToken via Scan QR Code Functionality

kylanhurt:
The purpose of this task is to implement the ability to add a custom ERC20 token to a wallet via scanning a QR code.

Asana Task: https://app.asana.com/0/361770107085503/590850184862899/f


src/modules/UI/scenes/AddToken/AddToken.ui.js
```diff
@@ -20,7 +20,12 @@ export type AddTokenOwnProps = {
   addNewToken: Function,
   currentCustomTokens: Array<CustomTokenInfo>,
   wallet: GuiWallet,
-  onAddToken: string => void
+  onAddToken: Function,
+  // adding properties in case coming from Scan scene (scan QR code to add token)
+  currencyName?: string,
+  currencyCode?: string,
+  contractAddress?: string,
+  decimalPlaces?: string
```
ahujsak: If these are being set in the default state as = to a prop or an empty string
`currencyName: this.props.currencyName || ''`
then we do not need to make these optional.

kylanhurt: Good point.

PR #557, opened 2018-03-23
Default BTC and BCH to use 'bits'

paullinator:
This mirrors the behavior of Airbitz


paullinator:
@thehobbit85 Bits has been requested by more users that mB. Actually mB has never been requested. There are also more wallets defaulting to bits than mB. This really should be consistent with our products. Is there anything wrong with the coding of this change?

PR #561, opened 2018-03-27
Prevent / Remove Deleted "Lingering" Tokens from WalletList

kylanhurt:
The purpose of this task is to prevent tokens from still showing up in the WalletList once they have been deleted. The fix is three-fold:

- Specifically tell the core to only "enable" tokens that are not explicitly hidden
- Specifically tell the GUI to not show any tokens in the WalletList which are explicitly hidden
- Make certain that the local state of the ManageTokens scene removes a deleted token from its list of tokens to enable / disable. This way if the user presses "Save" on the ManageTokens scene, the deleted token is already deleted from the state and therefore will not be enabled.

Asana Task: https://app.asana.com/0/361770107085503/516982647499351/f


src/modules/Login/action.js
```diff
@@ -103,7 +103,15 @@ export const initializeAccount = (account: EdgeAccount, touchIdInfo: Object) =>
       if (edgeWallet.type === 'wallet:ethereum') {
         if (state.ui.wallets && state.ui.wallets.byId && state.ui.wallets.byId[walletId]) {
           const enabledTokens = state.ui.wallets.byId[walletId].enabledTokens
-          edgeWallet.enableTokens(enabledTokens)
+          const customTokens = state.ui.settings.customTokens
```
ahujsak: So, I am fine with this, however to be consistent should you create a getter similar to everything we do for getting ui settings out of redux - .. a `SETTINGS_API.getCustomTokens(state)` instead of the direct path to the piece of state `state.ui.settings.customTokens`.  I do see in this function that other pieces of state are accessed directly. so it is consistent within this function... So this is not a formal change request, just wondering if we should change it. Thoughts?

kylanhurt: I'll bring this up at next engineering meeting. I've found getters to make debugging more of a headache

paullinator: I believe we actually mentioned to avoid abstracting the Redux store and to access it directly to avoid unnecessary code and boilerplate work. We also get better Flow coverage that way.

PR #571, opened 2018-03-31
Wallet List Scene Loading Progress Indicator / Bar

kylanhurt:
Replacement for: https://github.com/Airbitz/edge-react-gui/pull/569

The purpose of this task is to create a "dropdown" on the Wallet List that shows up initially when the scene loads, and gives the user an update as to what % of their wallets have loaded. The component can also be hidden simply by pressing on it. Otherwise it disappears shortly after reaching 100%

Asana Task: https://app.asana.com/0/361770107085503/556823215362117/f


src/modules/Login/action.js
```diff
@@ -99,6 +100,7 @@ export const initializeAccount = (account: EdgeAccount, touchIdInfo: Object) =>
       accountInitObject.currencyCode = currencyCode
     }
     const activeWalletIds = account.activeWalletIds
+    dispatch(insertWalletIdsForProgress(activeWalletIds))
```
wolverineks: can you get this information from the `accountInitAction`?

src/modules/UI/scenes/WalletList/WalletListConnector.js
```diff
@@ -25,6 +25,8 @@ const mapStateToProps = (state: State) => {
   const dimensions = state.ui.scenes.dimensions
   const customTokens = state.ui.settings.customTokens
   const otpResetPending = SETTINGS_SELECTORS.getOtpResetPending(state)
+  const progressPercentage = UI_SELECTORS.getWalletLoadingPercent(state)
```
wolverineks: you call it percent in some places, percentage in others, ratio in others
Pick the most accurate term and use it consistently

src/modules/UI/components/SimpleDropdown/style.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+import { StyleSheet } from 'react-native'
+
+import THEME from '../../../../theme/variables/airbitz.js'
+
+export default StyleSheet.create({
```
wolverineks: Please avoid default exports,
or at the very least, export a named and default
```
export const styles = ...

export default styles
```

src/modules/UI/components/SimpleDropdown/SimpleDropdown.ui.js
```diff
@@ -0,0 +1,29 @@
+// @flow
+
+import React, { Component } from 'react'
+import { TouchableWithoutFeedback, StylesSheet } from 'react-native'
+import styles from './style.js'
+import * as Animatable from 'react-native-animatable'
+
+export type Props = {
+  height?: number,
+  children: any,
```
wolverineks: `children: Node`

src/modules/UI/Wallets/reducer.js
```diff
@@ -182,6 +182,29 @@ export const walletEnabledTokens = (state: any = {}, action: Action) => {
   return state
 }
 
+export const walletLoadingProgress = (state: Object = {}, action: Action) => {
```
wolverineks: Please dont use `type: Object`
figure out what the type is exactly
`type { [walletId: string]: number }` or something like this

src/modules/UI/Wallets/action.js
```diff
@@ -436,6 +438,30 @@ export function addNewTokenThenDeleteOldSuccess (data: any) {
...
+
+  if (newWalletProgress !== 1 && marginalProgress < 0.1) return
+
+  dispatch({
+    type: UPDATE_WALLET_LOADING_PROGRESS,
+    data
+  })
+}
+
+export function insertWalletIdsForProgress (activeWalletIds: Array<string>) {
```
wolverineks: Please make the value of `action.data` and object
```return {
  type: '...',
  data: { activeWalletIds }
}
```

src/modules/UI/selectors.js
```diff
@@ -45,6 +45,26 @@ export const getArchivedWalletIds = (state: State): Array<string> => {
   return archivedWalletIds
 }
 
+export const getWalletLoadingPercent = (state: State) => {
+  const walletsForProgress = state.ui.wallets.walletLoadingProgress
+  const walletKeys = Object.keys(walletsForProgress)
```
wolverineks: these are not wallet keys, these are wallet ids

src/modules/UI/selectors.js
```diff
@@ -45,6 +45,26 @@ export const getArchivedWalletIds = (state: State): Array<string> => {
   return archivedWalletIds
 }
 
+export const getWalletLoadingPercent = (state: State) => {
+  const walletsForProgress = state.ui.wallets.walletLoadingProgress
```
wolverineks: what are `walletsForProgress`?

src/modules/UI/selectors.js
```diff
@@ -45,6 +45,26 @@ export const getArchivedWalletIds = (state: State): Array<string> => {
   return archivedWalletIds
 }
 
+export const getWalletLoadingPercent = (state: State) => {
```
wolverineks: let me know if you want to try writing a test for this

src/modules/Core/Account/callbacks.js
```diff
@@ -21,7 +21,10 @@ const makeAccountCallbacks = (dispatch: Dispatch): EdgeAccountCallbacks => {
     },
 
     onAddressesChecked (walletId: string, transactionCount: number) {
-      console.log(`${walletId} - onAddressesChecked with ratio: ${transactionCount}`)
+      console.log(`${walletId} - onAddressesChecked with progress ratio: ${transactionCount}`)
+      if (transactionCount > 0) {
```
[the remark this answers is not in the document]
kylanhurt: I suppose the variable name `transactionCount` is a slight misnomer. It is a number between 0 and 1 which conveys the portion of addresses that have been checked.

src/modules/Login/action.js
kylanhurt: When I tried this earlier it caused the `onAddressesChecked` callback to start firing off (eg reporting loading progress for a wallet) before we have the list of wallet ID's. For this reason I was unable to calculate the cumulative progress, which would be a major issue if any of the individual wallets immediately report 100% (only one address used in wallet) and would cause the *cumulative* progress % to never reach 100%.

src/modules/UI/selectors.js
kylanhurt: I believe I used that term in the case that we purposefully do not try to calculate the percentage for all of the wallets (eg archived wallets).

src/modules/UI/selectors.js
kylanhurt: Done

src/modules/UI/Wallets/action.js
kylanhurt: Done

src/modules/UI/scenes/WalletList/WalletListConnector.js
kylanhurt: I will keep this in mind going forward

PR #589, opened 2018-04-19
Remove Cryptocurrency Codes from Fiat List

kylanhurt:
The purpose of this task is to remove the cryptocurrency codes (eg "LTC", "BTC", "ETH") from the list of ISO fiat currency codes and therefore remove those items from the list of fiats that you can set for a wallet during the CreateWallet process.

I accomplished this by removing the package from which we were acquiring our list of cryptocurrencies and replaced it with a list of fiat currencies that have ISO codes from Wikipedia.

Asana Task: https://app.asana.com/0/361770107085503/613641312549211/f


kylanhurt:
I would wait until after this point release before merging this into `develop` (in my opinion).

paullinator:
This is worth getting into the point release. Looks safe enough

PR #592, opened 2018-04-19
Change txlist to only incrementally call getTransactions with larger indices

paullinator:
Before this change getTransactions() would be called with the full range from index 0 to the full range necessary for scrolling. Now it only asks for the range of transactions it doesn't already have.

This required moving a lot of logic that was in the TransactionList component into the action.js file. This includes state of the tx list and validation on what has changed.


paullinator:
Merge after v1.1.3 is merged to master

PR #594, opened 2018-04-20
Enforce node and npm versions via "engines" in package.json

Empact:
These min versions are listed in the readme.
https://docs.npmjs.com/files/package.json#engines
https://nodejs.org/en/download/releases/


Empact:
Huh. Well, I'm going to pass on looking further into it. :P

PR #597, opened 2018-04-20
New header on the Send Confirmation screen, Balance display on Send Confirmation Screen

amanalili:


src/modules/UI/scenes/SendConfirmation/SendConfirmationConnector.js
```diff
@@ -5,8 +5,9 @@ import { connect } from 'react-redux'
 
 import { getCurrencyConverter, getExchangeRate } from '../../../Core/selectors.js'
 import type { Dispatch, State } from '../../../ReduxTypes'
+import { convertNativeToExchange } from '../../../utils'
 import { getExchangeDenomination, getSelectedCurrencyCode, getSelectedWallet } from '../../selectors.js'
-import { getDisplayDenomination } from '../../Settings/selectors.js'
+import * as SETTINGS_SELECTORS from '../../Settings/selectors.js'
```
paullinator: Use explicit imports. We are moving away from `import *` as it is harder to debug

PR #603, opened 2018-04-23
Add global redux error alert

wolverineks:
Add global redux error alert

Simple action / Thunks that throw errors/ reject, and are not caught, 
will automatically get logged, and display global error alert


src/modules/UI/Wallets/action.js
```diff
@@ -150,15 +150,13 @@ export const setEnabledTokens = (walletId: string, enabledTokens: Array<string>,
...
-    .then(() => {
-      // let Redux know it was completed successfully
-      dispatch(setTokensSuccess())
-      dispatch(updateWalletEnabledTokens(walletId, enabledTokens))
-      // refresh the wallet in Redux
-      dispatch(refreshWallet(walletId))
-    })
-    .catch(e => dispatch(displayErrorAlert(e.message)))
+  return WALLET_API.setEnabledTokens(wallet, enabledTokens, disabledTokens).then(() => {
+    // let Redux know it was completed successfully
```
kylanhurt: Hypothetically we could consolidate this into one atomic Redux action. Then again I know you guys were debating that pattern the other day so you're probably aware of it.

PR #616, opened 2018-05-01
Batch upsert for all wallets

paullinator:
* Change throttling logic for refreshWallet to batch across all wallets, not per wallet since each dispatch causes a full revalidation of the redux state and rendering pipeline
* Change upsertWallet to allow multiple wallets to be upsert'ed in one call. Reduces the number of dispatches and re-renders.

### DIFF
--- a/src/modules/UI/Wallets/action.js
+++ b/src/modules/UI/Wallets/action.js
@@ -20,7 +20,7 @@ import * as SETTINGS_SELECTORS from '../Settings/selectors'
 
 export const PREFIX = 'UI/Wallets/'
 
-export const UPSERT_WALLET = PREFIX + 'UPSERT_WALLET'
+export const UPSERT_WALLETS = PREFIX + 'UPSERT_WALLETS'
 
 export const ACTIVATE_WALLET_ID = PREFIX + 'ACTIVATE_WALLET_ID'
 export const ARCHIVE_WALLET_ID = PREFIX + 'ARCHIVE_WALLET_ID'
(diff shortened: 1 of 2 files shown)


paullinator:
Note: this PR plus https://github.com/Airbitz/edge-react-gui/pull/615

Will reduce the time to sync edgy40 account (2nd login) on an iPhone 6 from 45s to 35s

src/modules/UI/Wallets/action.js
```diff
@@ -84,53 +84,63 @@ export const selectWallet = (walletId: string, currencyCode: string) => (dispatc
   }
 }
 
-function dispatchUpsertWallet (dispatch, wallet, walletId) {
-  dispatch(upsertWallet(wallet))
-  refreshDetails[walletId].delayUpsert = false
-  refreshDetails[walletId].lastUpsert = Date.now()
+function dispatchUpsertWallets (dispatch, wallets: Array<EdgeCurrencyWallet>) {
+  global.pcount('dispatchUpsertWallets')
```
kylanhurt: @paullinator What is `global.pcount` supposed to do? I am unfamiliar with it.

paullinator: Ack. this snuck in there. It's a debug line that counts the number of calls to that line of code. It's defined in `app.js`. It's safe to leave in, but should be removed.

paullinator: I'll remove it with a future PR since I'm still in this code.

PR #618, opened 2018-05-08
Password reminder modal

wolverineks:


src/modules/UI/components/Icon/Icon.ui.js
```diff
@@ -14,7 +14,7 @@ import * as Constants from '../../../../constants/indexConstants'
 type Props = {
   style: StyleSheet.Styles,
   name: string,
-  size: number,
+  size?: number,
```
ahujsak: This is not going to break this for now, but how are you dealing with size - is there now a default? are handling it with the style?

src/modules/UI/components/PasswordReminderModal/components/PasswordInput.ui.js
```diff
@@ -0,0 +1,135 @@
...
+        affixTextStyle={affixTextStyle}
+        renderAccessory={renderAccessory}
+        onChangeText={onChangeText}
+        onFocus={onFocus}
+        onBlur={onBlur}
+        value={value}
+      />
+    )
+  }
+}
```
ahujsak: Again: not broken. but why not use existing FormFieldinput component. If extra parameters were needed for the api, modifying and expanding on an existing single component.

src/modules/UI/components/PasswordReminderModal/passwordReminderModalReducer.js
```diff
@@ -0,0 +1,46 @@
...
+import type { Action } from '../../../ReduxTypes'
+import { CHECK_PASSWORD_START, CHECK_PASSWORD_SUCCESS, CHECK_PASSWORD_FAIL, PASSWORD_REMINDER_POSTPONED } from './actions.js'
+
+export const initialState = {
+  status: null
+}
+
+export const IS_CHECKING = 'IS_CHECKING'
+export const VERIFIED = 'VERIFIED'
+export const INVALID = 'INVALID'
```
ahujsak: These should be in the located in the constants directory and imported.

PR #631, opened 2018-05-24
Allen/export2

ahujsak:
export QBO and CSV from wallet drop down menu using the share menu


ahujsak:
QB file was not importing correctly

PR #632, opened 2018-05-27
Custom Wallet List Progress Bar

amanalili:
https://app.asana.com/0/361770107085503/498144750192796


src/modules/UI/components/ProgressBar/ProgressBar.ui.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export default class ProgressBar extends Component<Props> {
+  animation: Animated.Value
+
+  componentWillMount () {
+    this.animation = new Animated.Value(this.props.progress)
+  }
+
+  componentDidUpdate (prevProps: Props) {
```
paullinator: If you make this a "PureComponent" then it will always check if the props have changed before calling componentDidUpdate or render()

Change to:

`export default class ProgressBar extends PureComponent<Props> {`

PR #636, opened 2018-06-02
Paul/add ripple monero qtum ufo

paullinator:
Add Ripple, Monero, Qtum, and Ufo blockchain support


wolverineks:
What are postinstall.sh and postinstall.js?

wolverineks:
Why did you update edge-core-js?

wolverineks:
What do these do?
`"https-proxy-agent": false,`
`"tls": "react-native-tcp/tls`

paullinator:
Updated to use npm published modules vs git modules. Ripple lib was using 3 layers deep of git dependencies and I figured it might be causing the problem. 

Travis now passes Android build and iOS gets much further and stalls in the Xcode portion of the build which is an issue we’ve sporadically seen. Given Allen’s approval, I’m merging this now.

PR #649, opened 2018-06-09
Add Unique Identifier

wolverineks:
Add Unique Identifier feature
https://app.asana.com/0/239353283679855/498144750192806


src/modules/UI/scenes/SendConfirmation/SendConfirmation.ui.js
```diff
@@ -61,19 +65,19 @@ type routerParam = {
 
 type Props = SendConfirmationStateProps & SendConfirmationDispatchProps & routerParam
 
-type State = {
+type State = {|
```
kylanhurt: @wolverineks What is point of the pipe syntax?

wolverineks: its a tighter / stricter flow annotation
i think it provides better flow feedback, and could prevent some bugs

PR #655, opened 2018-06-12
Fix drawer button spacing, routing

wolverineks:
Fix drawer button spacing, routing

https://app.asana.com/0/239353283679855/704232837220427
https://app.asana.com/0/239353283679855/704232837220428


src/modules/UI/components/ControlPanel/style.js
```diff
@@ -183,5 +182,9 @@ export default {
   },
   underlay: {
     color: `${THEME.COLORS.PRIMARY}${THEME.ALPHA.LOW}`
+  },
+  debug: {
```
kylanhurt: I'll approve this pull request but it appears that this `debug` style is not being used.

PR #670, opened 2018-06-18
Upgrade simplex plugin to 0.0.2

teneighty:
This upgrade switches the code to use the production servers rather than
the sandbox.


paullinator:
@teneighty Does this fix the error we were seeing?

teneighty:
Yup. This fixes it.

PR #681, opened 2018-06-21
Update settings page button to be consistent for set up

ahujsak:
This updates the button on the settings screen but leaves the title bar alone on the recovery screen. 
Asana task https://app.asana.com/0/361770107085503/498144750192813

### DIFF
--- a/src/locales/en_US.js
+++ b/src/locales/en_US.js
@@ -130,7 +130,7 @@ const strings = {
   settings_button_change_password: 'Change Password',
   settings_button_debug: 'Debug',
   settings_button_lock_settings: 'Tap to Lock Account Settings',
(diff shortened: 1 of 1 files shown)


src/locales/en_US.js
```diff
@@ -130,7 +130,7 @@ const strings = {
   settings_button_change_password: 'Change Password',
   settings_button_debug: 'Debug',
   settings_button_lock_settings: 'Tap to Lock Account Settings',
-  settings_button_password_recovery: 'Password Recovery',
+  settings_button_password_recovery: 'Setup Password Recovery',
```
kylanhurt: You're gonna hate me for this but technically shouldn't it be "Set Up Password Recovery"? "Setup" is a noun, whereas "Set Up" is a verb / action.

PR #688, opened 2018-06-28
Implement 'Show Balance' Persistence for *Account*

kylanhurt:
The purpose of this task is to give persistence to the 'Show Balance' setting toggle for the Wallet List (account level)

Task: https://app.asana.com/0/361770107085503/498144750192820

### DIFF
--- a/src/modules/Core/Account/settings.js
+++ b/src/modules/Core/Account/settings.js
@@ -45,7 +45,7 @@ export const LOCAL_ACCOUNT_DEFAULTS = {
     nonPasswordDaysLimit: 2,
     nonPasswordLoginsLimit: 2
   },
-  isAccountBalanceVisible: false
+  isAccountBalanceVisible: true
 }
 
 const SYNCHED_SETTINGS_FILENAME = 'Settings.json'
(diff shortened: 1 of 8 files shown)


src/modules/Core/Account/settings.js
```diff
@@ -44,7 +44,8 @@ export const LOCAL_ACCOUNT_DEFAULTS = {
     nonPasswordLoginsCount: 0,
     nonPasswordDaysLimit: 2,
     nonPasswordLoginsLimit: 2
-  }
+  },
+  isAccountBalanceVisible: false
```
paullinator: This should default to true.

src/modules/UI/Settings/reducer.js
```diff
@@ -31,7 +31,8 @@ export const initialState = {
   otpResetDate: null,
   otpResetPending: false,
   confirmPasswordError: '',
-  sendLogsStatus: Constants.REQUEST_STATUS.PENDING
+  sendLogsStatus: Constants.REQUEST_STATUS.PENDING,
+  isAccountBalanceVisible: false
```
paullinator: should be true be default

PR #694, opened 2018-07-03
Enable sending to BIP70

wolverineks:
https://app.asana.com/0/239353283679855/590850184862903


src/modules/Core/Wallets/api.js
```diff
@@ -45,6 +56,26 @@ export const getReceiveAddress = (wallet: EdgeCurrencyWallet, currencyCode: stri
   return wallet.getReceiveAddress ? wallet.getReceiveAddress({ currencyCode }) : Promise.resolve(dummyEdgeReceiveAddress)
 }
 
+export const makeSpendInfo = (paymentProtocolInfo: EdgePaymentProtocolInfo): Promise<EdgeSpendInfo> => {
```
thehobbit85: This shouldn't return a Promise, it should just return the object:
{
   networkFeeOption: 'high',
   metadata: {
     name,
      notes
   },
   spendTargets
}

src/modules/UI/scenes/SendConfirmation/action.js
```diff
@@ -28,6 +39,55 @@ export const updateAmount = (nativeAmount: string, exchangeAmount: string, fiatP
...
+  const walletId = getSelectedWalletId(state)
+  const edgeWallet = getWallet(state, walletId)
+
+  Promise.resolve(paymentProtocolURL)
+    .then(paymentProtocolURL => {
+      return getPaymentProtocolInfo(edgeWallet, paymentProtocolURL)
+    })
+    .then(paymentProtocolInfo => {
+      return makeSpendInfo(paymentProtocolInfo)
+    })
```
thehobbit85: And since `makeSpendInfo` is not an async function, then you can remove one of the `then` and just write it as:
.then(paymentProtocolInfo => {
  const spendInfo = makeSpendInfo(paymentProtocolInfo)
  dispatch(newSpendInfo(spendInfo))
  ...etc
 })

src/modules/UI/scenes/SendConfirmation/action.js
```diff
@@ -28,6 +39,55 @@ export const updateAmount = (nativeAmount: string, exchangeAmount: string, fiatP
...
 
+type EdgePaymentProtocolUri = EdgeParsedUri & { paymentProtocolURL: string }
+
+export const paymentProtocolUriReceived = ({ paymentProtocolURL }: EdgePaymentProtocolUri) => (dispatch: Dispatch, getState: GetState) => {
+  const state = getState()
+  const walletId = getSelectedWalletId(state)
+  const edgeWallet = getWallet(state, walletId)
+
+  Promise.resolve(paymentProtocolURL)
+    .then(paymentProtocolURL => {
```
thehobbit85: A tiny styling comment, i think it's prettier like this:
then(paymentProtocolURL => getPaymentProtocolInfo(edgeWallet, paymentProtocolURL))

src/modules/UI/scenes/SendConfirmation/action.js
```diff
@@ -50,8 +110,7 @@ export const updateMaxSpend = () => (dispatch: Dispatch, getState: GetState) =>
 
   getMaxSpendable(edgeWallet, spendInfo)
     .then(nativeAmount => {
```
thehobbit85: I think that this is prettier but you don't have to change it:
then(nativeAmount => dispatch(createTX({ nativeAmount }, true)))

src/modules/UI/scenes/SendConfirmation/reducer.js
```diff
@@ -17,42 +18,90 @@ export const sendConfirmation = (state: SendConfirmationState = initialState, ac
         forceUpdateGuiCounter++
       }
       if (!parsedUri) return { ...state, forceUpdateGuiCounter, error, transaction }
+
```
thehobbit85: are you sure that you that the object `state.parsedUri` is not undefined here?
if not, you might want to init it just in case so that lines like:
!isEqual(state.parsedUri.metadata, metadata) 
won't crash the app

src/modules/Core/Wallets/api.js
```diff
@@ -45,6 +56,26 @@ export const getReceiveAddress = (wallet: EdgeCurrencyWallet, currencyCode: stri
...
+  const { domain, memo, merchant, spendTargets } = paymentProtocolInfo
```
thehobbit85: You also have the param `nativeAmount` inside `paymentProtocolInfo` which is already the total amount to spend so you can use that instead of reducing it inside `src/modules/UI/scenes/SendConfirmation/reducer.js`

src/modules/Core/Wallets/api.js
```diff
@@ -45,6 +56,26 @@ export const getReceiveAddress = (wallet: EdgeCurrencyWallet, currencyCode: stri
...
+
+  const name = domain === BITPAY.domain ? BITPAY.merchantName(memo) : merchant || domain
+  const notes = memo
+
+  return Promise.resolve({
+    networkFeeOption: 'high',
```
thehobbit85: why is the `networkFeeOption` set to `high` here?

wolverineks: I copied the example
https://developer.airbitz.co/javascript/#getpaymentprotocolinfo

src/modules/UI/scenes/SendConfirmation/action.js
wolverineks: I like to line up all the `.then`'s

src/modules/UI/scenes/SendConfirmation/action.js
wolverineks: once i enter async, i keep it async all the way, making each `.then` as small as possible, and using composition as much as possible

src/modules/UI/scenes/SendConfirmation/action.js
wolverineks: - update https://github.com/Airbitz/edge-conventions with your preference
- find a corresponding eslint rule
- update `.eslintrc`

src/modules/Core/Wallets/api.js
wolverineks: This is a function that I created in an attempt to make core related stuff async

src/modules/UI/scenes/SendConfirmation/action.js
wolverineks: i prefer smaller `.then`'s and composition

src/modules/UI/scenes/SendConfirmation/action.js
wolverineks: i prefer to line up all the `.then`'s

src/modules/UI/scenes/SendConfirmation/action.js
wolverineks: agreed, but there are some types that expected `void` as the return value, i dont know if thats the case here though.

src/modules/UI/scenes/SendConfirmation/reducer.js
wolverineks: im not sure. but the reducer and actions make it difficult to tell
i dont think i changed anything except in the code that would affect this

wolverineks:
`isEditable` is required for `BIP70` and possibly for plugins

src/modules/Core/Wallets/api.js
wolverineks: ok

src/modules/UI/scenes/SendConfirmation/action.js
wolverineks: https://eslint.org/docs/rules/arrow-body-style
'arrow-body-style': ['error', 'as-needed'],

PR #696, opened 2018-07-05
Add redux performance logging

wolverineks:
If enabled, this will create a csv file
The filepath to this file is logged to the console

This will record the type, start time, end time, and duration of each 
action that updates redux (simple actions only, no thunks)

This can be used to create charts and graphs to inform performance 
optimization decisions


wolverineks:
example charts with actions sorted by duration

screen shot 2018-07-05 at 7 58 54 am

PR #703, opened 2018-07-10
Implement Buy Crypto UI button in transaction list

amanalili:
- Only appears on BTC, BCH, ETH.
- Added loader when transactions are being loaded.


kylanhurt:
@paullinator My concern here is that according to @rjrs2k the TransactionList still has issues with some transactions and if the user has one or two transactions but the app thinks it does not then it will show these buttons and make the user even more confused that their transactions aren't showing up.

paullinator:
@kylanhurt I don't think the user will be anymore confused. They will want to see their missing transactions either way.

src/modules/UI/scenes/TransactionList/components/BuyCrypto.ui.js
```diff
@@ -0,0 +1,43 @@
...
+    const { wallet } = this.props
+    return wallet.currencyNames[wallet.currencyCode]
+  }
+  render () {
+    return (
+      <TouchableWithoutFeedback onPress={Actions.buysell}>
+        <View style={style.buyCryptoContainer}>
+          <View style={style.buyCryptoBox}>
+            <Image style={style.buyCryptoBoxImage} source={{ uri: this.getLogo() }} resizeMode={'cover'} />
+            <T style={style.buyCryptoBoxText}>Buy { this.getCurrencyName() } with Credit Card</T>
```
kylanhurt: This string will need to be internationalized. Our Edge team will have to make this change.

src/modules/UI/scenes/TransactionList/components/BuyCrypto.ui.js
```diff
@@ -0,0 +1,43 @@
...
+          </View>
+          <View style={style.buyCryptoNoTransactionBox}>
+            <T style={style.buyCryptoNoTransactionText}>No transaction yet!</T>
```
kylanhurt: Same here.

kylanhurt:
@amanalili I am not seeing the "Buy Crypto" stuff showing on the txList for empty Transaction Lists. The `emptyListComponent` isn't even being rendered / triggered.

paullinator:
Kylan. Are you checking a BTC BCH Zoe ETH Wallet? It only shows in those

kylanhurt:
@paullinator @swansontec This PR should be ready for merging

PR #707, opened 2018-07-12
Add firebase for conversion tracking

paullinator:
* Remove all files in ios/Pods/ and force use of `pod install` in postinstall script instead
* Clean up Android build to work with latest NDK and Android Studio
* Add react-native-firebase dependency
* Add events for new account creation
* Add dummy Firebase config files


paullinator:
Closing. This change breaks the release build

PR #735, opened 2018-07-30
Allen/onboarding

ahujsak:
When a user is new and before their first wallets are created, the gui flags that and prepares to launch the on boarding instruction screens. It's a 5 slide display that can be skipped.


src/connectors/components/OnboardingPagingDotsConnector.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+import { connect } from 'react-redux'
+
+// import * as Constants from '../../constants/indexConstants.js'
+import type { Dispatch, State } from '../../modules/ReduxTypes'
+import PagingDots from '../../modules/UI/components/PagingDots/PagingDots.js'
```
kylanhurt: Please change the `PagingDots.js` actual filename to `PagingDots.ui.js`

src/connectors/components/OnboardingPagingDotsConnector.js
```diff
@@ -0,0 +1,17 @@
...
+
+export const mapStateToProps = (state: State) => ({
+  totalItems: state.onBoarding.totalSlides,
```
kylanhurt: Are we not putting this under the `ui` area of the Reducer?

src/connectors/scene/OnBoardingSceneConnector.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import { connect } from 'react-redux'
+
+// import * as Constants from '../../constants/indexConstants.js'
+import * as actions from '../../actions/indexActions'
+import { COMPLETE_ONBOARDING, CHANGE_ONBOARDING_SLIDE } from '../../constants/indexConstants.js'
+import type { Dispatch, State } from '../../modules/ReduxTypes'
+import OnBoardingScene from '../../modules/UI/scenes/OnBoarding/OnBoardingScene.js'
```
kylanhurt: In the future please use `OnBoardingComponent` as the name of the component

src/locales/strings/enUS.json
```diff
@@ -338,5 +338,13 @@
...
 
   "save": "Save",
-  "spending_limits_warning": "** Warning **\\n\\nSetting a password spending limit may prevent you from accessing funds if your password is forgotten. Please make sure you have set up password recovery"  
+  "spending_limits_warning": "** Warning **\\n\\nSetting a password spending limit may prevent you from accessing funds if your password is forgotten. Please make sure you have set up password recovery",
+  "onboarding_button": "Got It",
+  "onboarding_skip_button": "Skip",
+  "onboarding_slide_1": "\n\nTap [+] to create new walllets for other cryptocurrencies.\nTap on a wallet name to send or receive funds for that wallet",
+  "onboarding_slide_2": "Tap the 3 dots to the right of an Ethereum wallet to add token support",
+  "onboarding_slide_3": "To receive cryptocurrency from others [copy] or [share] your wallet address, or have them scan your QR code here ",
+  "onboarding_slide_4": "To send cryptocurrency scan a friend's QR code here, or paste the code via the [Address] button.",
```
kylanhurt: Are we showing the `[button text]` example as capitalized or not? `[Address]` is capitalized but `[copy]` and `[share]` are not... which is not consistent.

src/modules/Login/action.js
```diff
@@ -51,6 +51,7 @@ export const initializeAccount = (account: EdgeAccount, touchIdInfo: Object) =>
   }
   const accountInitObject = {
     account: account,
+    showOnBoarding: false,
```
kylanhurt: Our convention would be `isOnBoardingVisible`

src/modules/ReduxTypes.js
```diff
@@ -210,7 +210,12 @@ export type State = {
   },
   exchangeRates: number,
   permissions: PermissionsState,
-  contacts: ContactsState
+  contacts: ContactsState,
+  onBoarding: {
+    currentIndex: number,
+    totalSlides: number,
+    slides: Array<Object>
```
kylanhurt: Can we be more specific than `<Object>`?

src/modules/UI/scenes/OnBoarding/OnBoardingScene.js
```diff
@@ -0,0 +1,111 @@
+// @flow
```
kylanhurt: Filename should be `OnBoardingScene.ui.js`

src/modules/UI/scenes/OnBoarding/OnBoardingScene.js
```diff
@@ -0,0 +1,111 @@
...
+import PagingDots from '../../../../connectors/components/OnboardingPagingDotsConnector.js'
+import s from '../../../../locales/strings.js'
+import { OnBoardingSceneStyles } from '../../../../styles/indexStyles.js'
+import { PLATFORM } from '../../../../theme/variables/platform'
+import { PrimaryButton } from '../../components/Buttons/'
+import OnBoardingSlide from './OnBoardingSlide'
+
+// import Swiper from 'react-native-swiper'
+
+type Props= {
```
kylanhurt: Should be `OnBoardingOwnProps` and then `OnBoardingState`

src/modules/UI/scenes/OnBoarding/OnBoardingScene.js
```diff
@@ -0,0 +1,111 @@
+// @flow
+import React, { Component } from 'react'
+import { Animated, Modal, View } from 'react-native'
+
+import PagingDots from '../../../../connectors/components/OnboardingPagingDotsConnector.js'
+import s from '../../../../locales/strings.js'
+import { OnBoardingSceneStyles } from '../../../../styles/indexStyles.js'
+import { PLATFORM } from '../../../../theme/variables/platform'
+import { PrimaryButton } from '../../components/Buttons/'
```
kylanhurt: You should be using `PrimaryButton` from `/components/Modals/components/PrimaryButton.ui.js`. We are going to be eventually deprecating the button component that you used.

src/modules/UI/scenes/OnBoarding/OnBoardingSlide.js
```diff
@@ -0,0 +1,69 @@
...
+}
+
+type State = {
+  orientation: string
+}
+
+export default class OnBoardingScene extends Component<Props, State> {
+  constructor (props: Props) {
+    super(props)
+    const { height, width } = Dimensions.get('window')
```
kylanhurt: Consider importing this from the `platform` file

src/styles/indexStyles.js
```diff
@@ -12,3 +12,6 @@ export * from './scenes/ScreenStyles.js'
 export * from './scenes/CryptoExchangeSceneStyles.js'
 export * from './scenes/OtpSettingsScreenStyles.js'
 export * from './scenes/TransactionsExportSceneStyle.js'
+export * from './scenes/OnBoardingSceneStyles.js'
```
kylanhurt: I like this naming convention and I think we should propose to use it everywhere. Easier to find files.

src/styles/scenes/OnBoardingSceneStyles.js
```diff
@@ -0,0 +1,63 @@
...
+  /* modalContainer: {
+    flex: 1,
+    alignItems: 'center',
+    zIndex: 1,
+    elevation: 1
+  }, */
+  modalContainer: {
+    alignItems: 'center',
+    position: 'absolute',
+    top: PLATFORM.deviceHeight * 1 / 8,
```
kylanhurt: Multiplying by 1 is redundant.

src/connectors/components/OnboardingPagingDotsConnector.js
ahujsak: What do you mean the ui area of the reducer?

src/styles/indexStyles.js
ahujsak: Thanks, thats why I do it. :)

src/styles/scenes/OnBoardingSceneStyles.js
ahujsak: it is multiplying the fraction (1/8)

src/connectors/components/OnboardingPagingDotsConnector.js
ahujsak: Changed it to PagingDotsComponent Since that is the name inside the file. export default class PagingDotsComponent

src/connectors/components/OnboardingPagingDotsConnector.js
ahujsak: Because we do not want to re-render the parent component, we just want to have this component controlled by state.

src/modules/UI/scenes/OnBoarding/OnBoardingScene.js
paullinator: This should have failed eslint since there is no space after `=`. Was a linter skipped on precommit?

src/connectors/components/OnboardingPagingDotsConnector.js
paullinator: Allen, this is a classic example of a component that should use our discussed architecture of sending redux state via a selector function and not directly connecting the component to redux since it is a child component of a scene. This would prevent the need for a separate file altogether.

paullinator: Proposed style for a "semi-connected" child component

Use PureComponent instead of Component

Have parent pass in selector functions into the child component as props

In the child component `getDerivedStateFromProps()`, call the props functions and assign local component state the resulting values

ie.

```
static getDerivedStateFromProps(nextProps, prevState) {
  const newState = {
    totalItems: nextProps.getTotalItems()
    currentIndex: nextProps.getCurrencyIndex()
  }
}
```
As a PureComponent, you will not rerender if newState doesn't differ from previous state and React will shallow compare for you. No need to have child be connected to redux

kylanhurt:
Close since @ahujsak will be submitting a new PR.

PR #737, opened 2018-07-31
update Login ui package

ahujsak:
• Fix type issues with updated login ui
• update login-ui-rn to new package

### DIFF
--- a/src/modules/UI/scenes/Settings/action.js
+++ b/src/modules/UI/scenes/Settings/action.js
@@ -1,5 +1,6 @@
 // @flow
 
+import { makeReactNativeFolder } from 'disklet'
 import type { EdgeAccount } from 'edge-core-js'
 import { disableTouchId, enableTouchId } from 'edge-login-ui-rn'
 import { Actions } from 'react-native-router-flux'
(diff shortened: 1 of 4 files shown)


src/modules/UI/scenes/Settings/action.js
```diff
@@ -167,13 +168,11 @@ export const setBitcoinOverrideServerRequest = (overrideServer: string) => (disp
 
 // touch id interaction
 export const updateTouchIdEnabled = (arg: boolean, account: EdgeAccount) => async (dispatch: Dispatch, getState: GetState) => {
-  const context = CORE_SELECTORS.getContext(getState())
-  // dispatch the update for the new state for
   dispatch(SETTINGS_ACTIONS.updateTouchIdEnabled(arg))
   if (arg) {
-    enableTouchId(context, account)
+    enableTouchId(makeReactNativeFolder(), account)
```
paullinator: Did this API change? If so this is a breaking change and the major version should bump accordingly

ahujsak: Yes it did. we removed context.io.folder I will republish Login Ui and do the version accordingly.

PR #740, opened 2018-08-01
API change for createPrivateKey

thehobbit85:


swansontec:
I concur. This must be part of the release.

PR #743, opened 2018-08-02
Update spending limits ux

wolverineks:
Update spending limits ux

### DIFF
--- a/src/modules/UI/Settings/spendingLimits/SpendingLimitsReducer.js
+++ b/src/modules/UI/Settings/spendingLimits/SpendingLimitsReducer.js
@@ -23,12 +23,12 @@ export const newSpendingLimits = (spendingLimits: SpendingLimits) => ({
 // REDUCERS
 export const initialState = {
   transaction: {
-    isEnabled: true,
+    isEnabled: false,
     amount: 0
   }
 }
 
-export const isEnabled = (state: boolean = true, action: Action) => {
+export const isEnabled = (state: boolean = initialState.transaction.isEnabled, action: Action) => {
   if (!action.data) return state
   switch (action.type) {
     case ACCOUNT_INIT_COMPLETE:
@@ -40,7 +40,7 @@ export const isEnabled = (state: boolean = true, action: Action) => {
   }
 }
 
-export const amount = (state: number = 0, action: Action) => {
+export const amount = (state: number = initialState.transaction.amount, action: Action) => {
   if (!action.data) return state
   switch (action.type) {
     case ACCOUNT_INIT_COMPLETE:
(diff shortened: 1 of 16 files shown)


src/modules/UI/Settings/spendingLimits/SpendingLimitsReducer.js
```diff
@@ -23,7 +23,7 @@ export const newSpendingLimits = (spendingLimits: SpendingLimits) => ({
 // REDUCERS
 export const initialState = {
   transaction: {
-    isEnabled: false,
+    isEnabled: true,
```
kylanhurt: Is there a reason we are setting the default to true / enabled?

src/modules/UI/scenes/Scan/action.js
```diff
@@ -115,8 +115,8 @@ export const parseUri = (data: string) => (dispatch: Dispatch, getState: GetStat
       }
 
       // PUBLIC ADDRESS URI
+      Actions.sendConfirmation('fromScan')
```
kylanhurt: This should use a scene constant

src/modules/UI/scenes/Scan/action.js
```diff
@@ -132,6 +132,21 @@ export const parseUri = (data: string) => (dispatch: Dispatch, getState: GetStat
...
+  const state = getState()
+  dispatch(legacyAddressModalDeactivated())
+  const parsedUri = state.ui.scenes.scan.parsedUri
+  setImmediate(() => {
+    if (!parsedUri) {
+      dispatch(enableScan())
+      return
+    }
+
+    Actions.sendConfirmation('fromScan')
```
kylanhurt: This as well

src/modules/UI/scenes/SendConfirmation/action.js
```diff
@@ -109,7 +109,16 @@ export const updateMaxSpend = () => (dispatch: Dispatch, getState: GetState) =>
   const spendInfo = getSpendInfo(state)
 
   getMaxSpendable(edgeWallet, spendInfo)
-    .then(nativeAmount => dispatch(createTX({ nativeAmount }, true)))
+    .then(nativeAmount => {
+      const state = getState()
+      const spendInfo = getSpendInfo(state, { nativeAmount })
+      const authRequired = getAuthRequired(state, spendInfo)
+
+      dispatch(reset())
```
kylanhurt: Are you able to combine these three dispatches into one? Or two?

wolverineks: im not sure how to do that
reset enables the scene to detect the spendInfo/transaction as "new" and not "updated", adn this triggers the TextInput focusing, gross, i know
if you can come up with a better way, im all ears

until then, 3 dispatches instead of 1 for an action that rarely ever happens, and would only happen once, i think is okay for now

src/modules/UI/Settings/spendingLimits/SpendingLimitsReducer.js
wolverineks: it should be false

kylanhurt:
Comments appear to have been handled by Kevin. Merging into develop.

PR #746, opened 2018-08-03
Scale Font on TxDetails Scene

kylanhurt:
The purpose of this task is to adjust the font scaling on the txDetails page to use @amanalili's solution.

Asana Task: https://app.asana.com/0/361770107085503/669010486596396/f


paullinator:
This actually doesn’t use th library art is using. Please use React-Native-size-matters

PR #752, opened 2018-08-05
Fix CreateWallet buttons and fix onboarding typos

kylanhurt:
The purpose of this task is to fix the onboarding regression that modified the CreateWallet button styling. Additionally, I have fixed some minor typos within the onboarding "cards".

Edit: apparently this also fixes a second Asana task as well...

Asana Task #1: https://app.asana.com/0/361770107085503/768634831505674/f
Asana Task #2: https://app.asana.com/0/361770107085503/768634831505675/f

### DIFF
--- a/src/modules/UI/components/Buttons/style.js
+++ b/src/modules/UI/components/Buttons/style.js
@@ -14,7 +14,7 @@ export const styles = {
     height: THEME.BUTTONS.HEIGHT,
     alignItems: 'center',
     justifyContent: 'center',
-    width: '100%',
+    flex: 1,
     borderRadius: 3
   },
   stylizedButtonText: {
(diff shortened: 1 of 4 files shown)


src/modules/UI/components/Buttons/style.js
```diff
@@ -14,7 +14,7 @@ export const styles = {
     height: THEME.BUTTONS.HEIGHT,
     alignItems: 'center',
     justifyContent: 'center',
-    width: '100%',
+    flex: 1,
```
paullinator: This scares me. It would affect every button in the app. How did the create wallet buttons break in the first place?

kylanhurt: @paullinator This is fixing a *regression* by Allen's onboarding PR that created the exact type of incident that you were scared of. It shouldn't have been adjusted in his PR, and I should have caught it when I code reviewed it. In this PR I have changed the button he is using to the one our app is migrating to, but also changing the styling of the old button back to what it should be (so now the Create Wallet buttons will be back to how their intended style).

I hope this makes sense.

PR #756, opened 2018-08-06
Bump login-rn-ui to 0.4.0

ahujsak:
• fix code from breaking change in 0.4.0
• enables fix from ASANA https://app.asana.com/0/361770107085503/693269684549043/f

PR #757, opened 2018-08-06
William/no core io

swansontec:
This stops using `EdgeContext.io`. This is part of a larger cleanup branch, but I've sliced it out to support the upgrade to `edge-login-ui` v0.4.0. This would be a better version of #756.


paullinator:
Please don't merge until next week.

swansontec:
@ahujsak Do we need the changes in login-ui v0.4.0 for this release? If so, maybe we should re-open your #756 PR to get the changes in the release, and then merge this later once the release is over?

swansontec:
All this stuff is already in #798, so we no longer need to merge this.

PR #763, opened 2018-08-07
Remove max spend for monero

wolverineks:
https://app.asana.com/0/239353283679855/761815917484852

### DIFF
--- a/src/modules/UI/scenes/SendConfirmation/sendConfirmationOptions.test.js
+++ b/src/modules/UI/scenes/SendConfirmation/sendConfirmationOptions.test.js
@@ -1,0 +1,77 @@
+/* globals jest describe it expect */
+/* eslint-disable flowtype/require-valid-file-annotation */
+
+import React from 'react'
+import ShallowRenderer from 'react-test-renderer/shallow'
+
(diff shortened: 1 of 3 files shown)


src/modules/UI/scenes/SendConfirmation/sendConfirmationOptions.test.js
```diff
@@ -0,0 +1,77 @@
...
+    expect(actual).toMatchSnapshot()
+  })
+
+  it('XMR', () => {
+    const renderer = new ShallowRenderer()
+
+    const props = {
+      changeMiningFee: jest.fn,
+      openHelpModal: jest.fn,
+      sendMaxSpend: jest.fn,
```
kylanhurt: Does `sendMaxSpend` need to be removed from this test for XMR? @wolverineks

kylanhurt:
@kylan

src/modules/UI/scenes/SendConfirmation/sendConfirmationOptions.test.js
wolverineks: i pass it in for each currency, but if the currencyCode === 'XMR' then the button shouldnt show

PR #786, opened 2018-08-15
Require only one press on button for settings modals

kylanhurt:
The purpose of this task is to reduce number of required presses for Settings modals from one press to two presses. Solution is to place the modal outside of the ScrollView.

Asana Task: https://app.asana.com/0/361770107085503/755725093339248/f


paullinator:
Can’t merge. Some jest tests failed

PR #790, opened 2018-08-16
Fix styling for Sweep Private Key modal

kylanhurt:
The purpose of this task is to give proper styling to the Sweep Private Keys modal buttons.

Asana Task: https://app.asana.com/0/361770107085503/781564217904941/f


src/modules/UI/components/Buttons/style.js
```diff
@@ -14,7 +14,8 @@ export const rawStyles = {
   },
   buttonText: {
     fontFamily: THEME.FONTS.DEFAULT,
-    fontSize: 18
+    fontSize: 18,
+    lineHeight: 18
```
wolverineks: which buttons were affected by this?

kylanhurt: The main buttons affected were the Export Transaction scene buttons, but this should ensure that a.) no more buttons have text getting cut off (top or bottom) and b.) that the button text is properly aligned vertically.

PR #791, opened 2018-08-16
Make Wallet List Progress Bar Own Connected Component

kylanhurt:
The purpose of this task is to have the Wallet List Progress Bar component be connected directly to Redux rather than receive the progress % property from the parent scene (Wallet List).

Asana Task: https://app.asana.com/0/361770107085503/749204124304339/f


paullinator:
This does not utilize the coding convention as we discussed in our Redux meeting last week. Closing this PR for a re-write

PR #803, opened 2018-08-19
Upgrade to React Native 0.55

paullinator:


.flowconfig
```diff
@@ -12,52 +12,56 @@
...
 
-suppress_comment=\\(.\\|\n\\)*\\$FlowFixMe\\($\\|[^(]\\|(\\(>=0\\.\\(4[0-0]\\|[1-3][0-9]\\|[0-9]\\).[0-9]\\)? *\\(site=[a-z,_]*react_native[a-z,_]*\\)?)\\)
-suppress_comment=\\(.\\|\n\\)*\\$FlowIssue\\((\\(>=0\\.\\(4[0-0]\\|[1-3][0-9]\\|[0-9]\\).[0-9]\\)? *\\(site=[a-z,_]*react_native[a-z,_]*\\)?)\\)?:? #[0-9]+
+suppress_comment=\\(.\\|\n\\)*\\$FlowFixMe\\($\\|[^(]\\|(\\(<VERSION>\\)? *\\(site=[a-z,_]*react_native[a-z,_]*\\)?)\\)
+suppress_comment=\\(.\\|\n\\)*\\$FlowIssue\\((\\(<VERSION>\\)? *\\(site=[a-z,_]*react_native[a-z,_]*\\)?)\\)?:? #[0-9]+
 suppress_comment=\\(.\\|\n\\)*\\$FlowFixedInNextDeploy
 suppress_comment=\\(.\\|\n\\)*\\$FlowExpectedError
 
+[version]
+^0.75.0
```
swansontec: Seems like this should blow up, since we are on version 0.66.0.

paullinator: @swansontec

It does and it was removed in the 2nd commit.

swansontec:
I get the following error when trying to run the branch:

```
bundling failed: Error: Unable to resolve module `../utils/mapPropsToStyleNames` from `/Users/william/airbitz/edge-react-gui/node_modules/native-base/dist/src/basic/SwipeRow.js`: The module `../utils/mapPropsToStyleNames` could not be found from `/Users/william/airbitz/edge-react-gui/node_modules/native-base/dist/src/basic/SwipeRow.js`.
```

I'll dig into it a bit.

src/app.js
```diff
@@ -7,7 +7,7 @@ import './util/polyfills'
 import { Client } from 'bugsnag-react-native'
 import React, { Component } from 'react'
 import { AppState, AsyncStorage, Platform, Text, TextInput } from 'react-native'
-import BackgroundTask from 'react-native-background-task'
+import BackgroundFetch from 'react-native-background-fetch'
```
wolverineks: why did this change?

src/app.js
```diff
@@ -46,7 +46,14 @@ console.ignoredYellowBox = IGNORED_WARNINGS
 global.OS = Platform.OS
 global.slowlogOptions = { threshold: 16 }
 // Disable the font scaling
+if (!Text.defaultProps) {
```
wolverineks: why is this needed?

src/modules/UI/scenes/Scan/Scan.ui.js
```diff
@@ -135,9 +136,9 @@ export default class Scan extends Component<Props> {
 
   renderCamera = () => {
     if (this.props.cameraPermission === AUTHORIZED) {
-      const torchMode = this.props.torchEnabled ? Camera.constants.TorchMode.on : Camera.constants.TorchMode.off
+      const torchMode = this.props.torchEnabled ? RNCamera.Constants.FlashMode.on : RNCamera.Constants.FlashMode.off
```
wolverineks: why did this change from torch mode to flash mode?

src/modules/UI/scenes/Scan/Scan.ui.js
```diff
@@ -135,9 +136,9 @@ export default class Scan extends Component<Props> {
...
 
-      return <Camera style={styles.preview} ref="cameraCapture" torchMode={torchMode} onBarCodeRead={this.onBarCodeRead} />
+      return <RNCamera style={styles.preview} type={RNCamera.Constants.Type.back} ref="cameraCapture" flashMode={torchMode} onBarCodeRead={this.onBarCodeRead} />
```
wolverineks: why did this change from torch to flash?

src/modules/UI/scenes/WalletList/components/WalletListRow/FullWalletListRow.ui.js
```diff
@@ -221,7 +221,10 @@ const mapDispatchToProps = dispatch => ({
 })
 
 // $FlowFixMe
-const FullWalletListRowConnected = connect(mapStateToProps, mapDispatchToProps)(FullWalletListRowLoadedComponent)
+const FullWalletListRowConnected = connect(
```
wolverineks: is this just formatting?

src/modules/UI/scenes/WalletList/style.js
```diff
@@ -224,13 +224,13 @@ export const styles = {
     height: 22,
     width: 22,
     marginRight: 5,
-    resizeMode: Image.resizeMode.contain,
+    resizeMode: 'contain',
```
wolverineks: why did this change from an imported string to a hard coded string?

src/modules/UI/scenes/Scan/Scan.ui.js
swansontec: Because it's a different library now, and that's what the new library called it.

src/app.js
swansontec: Because the old library wasn't maintained / compatible with the new RN.

src/app.js
paullinator: New RN does not initialize default props so if you try to access it without initializing it, you’ll throw.

src/modules/UI/scenes/Scan/Scan.ui.js
paullinator: RNCamera replaced RCTCamera and so did the constants.

src/modules/UI/scenes/WalletList/style.js
paullinator: New RN removed the constant. Have to use the hard coded string now. Not my choice.

src/modules/UI/scenes/Scan/Scan.ui.js
wolverineks: are you sure you dont want `RNCamera.Constants.FlashMode.torch`

> RNCamera.Constants.FlashMode.torch turns on torch mode, meaning the flash light will be turned on all the time (even before taking photo) just like a flashlight.

https://github.com/react-native-community/react-native-camera/blob/master/docs/RNCamera.md#flashmode

paullinator: I think you are right. Can you submit a PR to change.

PR #807, opened 2018-08-23
Fix autologout crash

wolverineks:
https://app.asana.com/0/239353283679855/776427948299396

### DIFF
--- a/src/modules/Main.ui.js
+++ b/src/modules/Main.ui.js
@@ -52,6 +52,8 @@ import TransactionsExportSceneConnector from '../connectors/scene/TransactionsEx
 import * as Constants from '../constants/indexConstants'
 import { setIntlLocale } from '../locales/intl'
 import s, { selectLocale } from '../locales/strings.js'
+import { LoadingScene } from '../modules/UI/components/Loading/LoadingScene.ui.js'
+import { ifLoggedIn } from '../modules/UI/components/LoginStatus/LoginStatus.js'
 import { OnBoardingComponent } from '../modules/UI/scenes/OnBoarding/OnBoardingComponent.js'
 import { makeCoreContext } from '../util/makeContext.js'
 import * as CONTEXT_API from './Core/Context/api'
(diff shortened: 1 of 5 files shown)


src/modules/Main.ui.js
```diff
@@ -52,6 +52,8 @@ import TransactionsExportSceneConnector from '../connectors/scene/TransactionsEx
 import * as Constants from '../constants/indexConstants'
 import { setIntlLocale } from '../locales/intl'
 import s, { selectLocale } from '../locales/strings.js'
+import { Loading } from '../modules/UI/components/Loading/Loading.ui.js'
```
swansontec: `LoadingPage` / `LoadingScene`?

src/modules/Main.ui.js
```diff
@@ -68,8 +70,8 @@ import WalletName from './UI/components/Header/WalletName/WalletNameConnector.js
 import HelpModal from './UI/components/HelpModal'
 import { passwordReminderModalConnector as PasswordReminderModal } from './UI/components/PasswordReminderModal/indexPasswordReminderModal.js'
 import TransactionAlert from './UI/components/TransactionAlert/TransactionAlertConnector'
-import type { Permission } from './UI/permissions.js'
 import { CAMERA, CONTACTS } from './UI/permissions.js'
+import type { Permission } from './UI/permissions.js'
```
swansontec: `import { CAMERA, CONTACTS, type Permission } from './UI/permissions.js'` works since the import-sort upgrade.

src/modules/UI/components/LoginStatus/LoginStatus.js
```diff
@@ -0,0 +1,31 @@
...
+import { connect } from 'react-redux'
+
+import { getLoginStatus } from '../../Settings/selectors.js'
+
+export function ifLoggedIn<Klass> (LoggedIn: Klass, LoggedOut: ComponentType<{}>): Klass {
+  class IfLoggedIn extends React.Component<{ loginStatus: boolean, outerProps: any }> {
+    render () {
+      const { loginStatus, outerProps } = this.props
+      // $FlowFixMe
+      return loginStatus ? <LoggedIn {...outerProps} /> : <LoggedOut {...outerProps} />
```
swansontec: `LoggedOut` doesn't take `outerProps` in the types, so maybe we don't need it?

PR #810, opened 2018-08-25
Add Fiat to WalletListModal (Dropdown)

kylanhurt:
The purpose of this task is to add *fiat* to the WalletListModal dropdown. Additionally, I inserted logic to remove wallets that should not be visible on the crypto exchange scene: specifically disallowing users to select a wallet if that wallet is already selected for the opposite side of the trade. Also, I removed same currencies because exchanging one currency to the same currency is illogical.

Asana Task: https://app.asana.com/0/361770107085503/778189856161370/f


src/modules/UI/components/WalletListRow/WalletListRow.ui.js
```diff
@@ -0,0 +1,205 @@
...
+              </Text>
+            </View>
+          </View>
+        </View>
+      </TouchableHighlight>
+    )
+  }
+}
+
+const mapStateToProps = (state: State, ownProps: WalletListRowOwnProps): WalletListRowStateProps => {
```
ahujsak: I thought we were keeping connectors in separate files?

src/modules/utils.js
```diff
@@ -30,6 +30,29 @@ export const findDenominationSymbol = (denoms: Array<EdgeDenomination>, value: s
   }
 }
 
+export const getSetCurrencyMultiplier = (currencyCode: string, settings: Object, denominations: Object) => {
```
ahujsak: Can we get a better name getSet is confusing.

src/modules/utils.js
```diff
@@ -30,6 +30,29 @@ export const findDenominationSymbol = (denoms: Array<EdgeDenomination>, value: s
...
+  const setCurrencyDenomination = settings[currencyCode].denomination
+  const denominationsInfo = denominations[setCurrencyDenomination]
+  const multiplier = denominationsInfo.multiplier
+  return multiplier
+}
+
+// tokens can only have one denomination / multiplier from what I understand
+export const getSetTokenMultiplier = (currencyCode: string, settings: Object, denomination: Object): string => {
```
ahujsak: Same as above

PR #814, opened 2018-08-28
Change 'Pin' instances to 'PIN'

kylanhurt:
Since 'PIN' stands for 'Personal Identification Number', then it should always be in all-uppercase.

Asana Task: https://app.asana.com/0/361770107085503/798004909016442


kylanhurt:
@ahujsak It's possible that I turned off `precommit`, but it's also possible that precommit generated the file but I didn't add the newly generated JSON file to the next commit or didn't push it.

PR #823, opened 2018-08-31
Fix incorrect dates on transactions

wolverineks:
https://app.asana.com/0/239353283679855/775404594737488

Incoming transaction may have dates
that are incorrect by a factor of 1000x

This commit enables detecting incorrect
dates and attempts to fix them


paullinator:
Tests fail with miss-matched snapshots

PR #828, opened 2018-09-01
Implement WiredBalanceBox

paullinator:
This slightly optimizes WalletList.ui.js by separating the top Balance Box into a wired component.

* Get all state for WiredBalanceBox from selectors
* Get total fiat amount from thick selector that does the old `tallyUpTotalCrypto` routine. This needs to be optimized to not always re-calc all wallet fiat values if only one has changed
* Remove unnecessary props from WalletListConnector

### DIFF
--- a/src/modules/UI/Settings/selectors.js
+++ b/src/modules/UI/Settings/selectors.js
@@ -212,3 +212,8 @@ export const getOtpResetPending = (state: State) => {
   const otpResetPending = settings.otpResetPending
   return otpResetPending
 }
+
+export const getIsAccountBalanceVisible = (state: State) => {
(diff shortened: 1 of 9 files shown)


kylanhurt:
@paullinator I've made an adjustment to the code to use a utility function for tallying up the balance. The WalletList should have already been using it, rather than it's own component function, but it must have been an oversight.

Everything else looks good to me.

paullinator:
@kylanhurt `tallyUpTotalCrypto` is now a selector. Something I call a "thick" selector as it has a lot of logic in it. It's no longer in the component so in theory any component can use it. It does live in the walletlist selector but can be easily moved now that it's refactored

PR #844, opened 2018-09-12
Set parameter for number of Redux actions to show in RN Debugger

kylanhurt:
The purpose of this task is to explicitly define the number of actions show in the React Native Debugger so that the `maxAge` parameter can be modified according to a developer's needs.

There is no associated Asana task for this.

### DIFF
--- a/src/lib/configureStore.js
+++ b/src/lib/configureStore.js
@@ -16,7 +16,7 @@ const middleware = [errorAlert, loginStatusChecker, thunk, soundsMiddleware]
 if (ENV.ENABLE_REDUX_PERF_LOGGING) middleware.push(perfLogger)
 
 const composeEnhancers =
-  typeof window === 'object' && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ ? window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({ name: 'ui' }) : compose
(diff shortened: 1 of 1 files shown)


src/lib/configureStore.js
```diff
@@ -16,7 +16,7 @@ const middleware = [errorAlert, loginStatusChecker, thunk, soundsMiddleware]
 if (ENV.ENABLE_REDUX_PERF_LOGGING) middleware.push(perfLogger)
 
 const composeEnhancers =
-  typeof window === 'object' && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ ? window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({ name: 'ui' }) : compose
+  typeof window === 'object' && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ ? window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({ name: 'ui', maxAge: 50 }) : compose
```
wolverineks: does this increase the number of actions displayed in the redux logger?

kylanhurt: No, because the default is 50 so it's just making it more obvious which parameter can be modified to change the number of actions displayed. If you increase the number it will show more actions. William is also thinking of removing the core from Redux at some point.

PR #855, opened 2018-09-14
Remove OTP check from login

paullinator:
This is done in login-ui now. Speeds up logins

### DIFF
--- a/src/modules/Login/action.js
+++ b/src/modules/Login/action.js
@@ -34,17 +34,6 @@ export const initializeAccount = (account: EdgeAccount, touchIdInfo: Object) => 
 
   const state = getState()
   const context = CORE_SELECTORS.getContext(state)
-  let otpResetPending = false
-  try {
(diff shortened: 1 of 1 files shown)


src/modules/Login/action.js
```diff
@@ -59,7 +48,7 @@ export const initializeAccount = (account: EdgeAccount, touchIdInfo: Object) =>
     walletId: '',
     currencyCode: '',
     currencyPlugins: [],
-    otpInfo: { enabled: account.otpKey != null, otpKey: account.otpKey, otpResetPending },
+    otpInfo: { enabled: account.otpKey != null, otpKey: account.otpKey, otpResetPending: false },
```
swansontec: We should probably get rid of `otpInfo.otpResetPending`, but that's scary without Flow types.

paullinator: Hence why i left it there. `false` is always safe for now

PR #856, opened 2018-09-14
Allen/remove async get accounts

ahujsak:
This removes the async fetchAccounts from initialization
moves the modal to the main ui 
and throws the modal if there is a OTP reset when the new function returns and does


src/constants/ActionConstants.js
```diff
@@ -51,6 +51,6 @@ export const ACCOUNT_INIT_COMPLETE = 'accountInitComplete'
 
 // OTP
 export const DISABLE_OTP_RESET = 'disableOTPReset'
-
+export const ENABLE_OTP_RESET = 'enableOTPReset'
```
swansontec: This will probably conflict with @wolverineks new action types as string literals.

PR #857, opened 2018-09-15
Enable flow typing reducer cases / actions

wolverineks:
Flow assumes the "flow type" of a variable to be the type of the data to which the variable resolves.
This means  with `const a = 'some string'`, flow considers the type of `a` to be `string`, not `'some string'`.

The unfortunate effect is that flow cannot distinguish between any 2 actions that use variables in the `type` property. For instance, flow cannot distinguish the difference between the following 2 actions:

```javascript
const SOME_ACTION = 'SOME_ACTION'
const SOME_OTHER_ACTION = 'SOME_OTHER_ACTION'
const someAction = () => ({ type: SOME_ACTION, data: 123 })
const someOtherAction = () => ({ type: SOME_OTHER_ACTION })
```

Consequently, flow cannot provide much assistance inside case statements of reducers.
```javascript
switch (action.type) {
  case SOME_ACTION: {
    ...
   /* flow cannot tell which action has been dispatched */
  }

  case SOME_OTHER_ACTION: {
  ...
   /* flow cannot tell which action has been dispatched */
  }
}
```

To enable flow to provide additional assistance, this PR converts all the action types and corresponding reducer cases to string literals.


src/modules/UI/scenes/Request/reducer.js
```diff
@@ -57,19 +55,7 @@ export const request = (state: RequestSceneState = initialState, action: Action)
...
-        ...state,
-        receiveAddress: {
-          ...receiveAddress,
-          metadata
-        }
-      }
-    }
-
-    case ACTION.UPDATE_AMOUNT_REQUESTED_IN_FIAT: {
+    case 'UPDATE_AMOUNT_REQUESTED_IN_FIAT': {
```
swansontec: We could probably PR this separately so we can verify that this change is correct.

src/reducers/contacts/actions.js
```diff
@@ -2,15 +2,13 @@
 
 import type { GuiContact } from '../../types'
 
-export const PREFIX = 'CONTACTS/'
-
-export const LOAD_CONTACTS_START = PREFIX + 'LOAD_CONTACTS_START'
+// export const LOAD_CONTACTS_START = 'CONTACTS/LOAD_CONTACTS_START'
```
swansontec: Delete this comment?

src/reducers/contacts/actions.js
```diff
@@ -2,15 +2,13 @@
...
 export const loadContactsStart = () => ({
-  type: LOAD_CONTACTS_START
+  type: 'CONTACTS/LOAD_CONTACTS_START'
 })
 
-export const LOAD_CONTACTS_SUCCESS = PREFIX + 'LOAD_CONTACTS_SUCCESS'
+// export const LOAD_CONTACTS_SUCCESS = 'CONTACTS/LOAD_CONTACTS_SUCCESS'
```
swansontec: Delete this comment?

PR #858, opened 2018-09-15
Scaling of Transaction Details Page

amanalili:


package.json
```diff
@@ -127,9 +126,10 @@
     "react-native-popup-menu": "0.13.3",
     "react-native-push-notification": "^3.0.2",
     "react-native-qrcode": "^0.2.3",
-    "react-native-randombytes": "^3.5.0",
-    "react-native-router-flux": "4.0.0-beta.31",
-    "react-native-share": "^1.1.1",
+    "react-native-randombytes": "git://github.com/Airbitz/react-native-randombytes.git",
```
paullinator: why did anything in package.json change other than `react-native-size-matters`?

src/modules/UI/components/Buttons/PrimaryButtonScaled.ui.js
```diff
@@ -0,0 +1,39 @@
+// @flow
```
paullinator: Put scaling into PrimaryButton and remove use of Scaled versions of components. We're looking to get the full app scaled now

src/modules/UI/components/FormattedTextScaled/FormattedTextScaled.ui.js
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable flowtype/require-valid-file-annotation */
```
paullinator: Remove scaled versions of components

package.json
amanalili: I am not really sure why

paullinator: Please rebase this PR without the incorrect changes to package.json

PR #864, opened 2018-09-22
Add type checking to remaining actions / reducers

wolverineks:


src/modules/UI/Settings/reducer.js
```diff
@@ -286,9 +290,10 @@ export const settingsLegacy = (state: SettingsState = initialState, action: Acti
 
     case 'OVERWRITE_THEN_DELETE_TOKEN_SUCCESS': {
       // where oldCurrencyCode is the sender, and tokenObj.currencyCode is the receiver (new code)
-      const receiverCode = data.tokenObj.currencyCode
-      const senderCode = data.oldCurrencyCode
-      const { tokenObj } = data
+      if (!action.data) throw new Error('Invalid action')
+      const receiverCode = action.data.tokenObj.currencyCode
+      const senderCode = action.data.oldCurrencyCode
```
kylanhurt: Should this be checking to make sure that `tokenObj` exists?

src/modules/UI/Settings/reducer.js
```diff
@@ -334,15 +340,17 @@ export const settingsLegacy = (state: SettingsState = initialState, action: Acti
...
-        [currencyCode]: data
+        [currencyCode]: action.data
       }
     }
 
     case 'ADD_NEW_CUSTOM_TOKEN_SUCCESS': {
-      const { tokenObj, newCurrencyCode, settings } = data
+      if (!action.data) throw new Error('Invalid action')
+      const { tokenObj, newCurrencyCode, settings } = action.data
       const customTokens = settings.customTokens
```
kylanhurt: Do we need to check the existence of `settings` before we try to access the `settings.customTokens` property?

src/modules/UI/Settings/reducer.js
wolverineks: the point of this PR is not to fix existing bugs, but to fix bugs caused by the previous action typing PR

wolverineks: the point of this PR is not to fix existing bugs, but to fix bugs caused by the previous action typing PR

PR #866, opened 2018-09-24
William/upgrade core

swansontec:


swansontec:
Updated to include "fake user 1". There seems to be a problem with the wallets for this user, since we didn't add them to the fake git server.

PR #869, opened 2018-09-25
Enable button tap passthrough on change  password

wolverineks:
https://app.asana.com/0/239353283679855/804580670230013


src/modules/UI/scenes/ChangePinPassword/ChangePasswordComponent.ui.js
```diff
@@ -36,7 +36,7 @@ export default class ChangePassword extends Component<ChangePasswordComponent> {
     return (
       <SafeAreaView>
         <Gradient style={styles.gradient} />
-        <ScrollView style={styles.container}>
+        <ScrollView style={styles.container} keyboardShouldPersistTaps={true}>
```
paullinator: `true` is deprecated. Use `always` instead

https://facebook.github.io/react-native/docs/scrollview#keyboardshouldpersisttaps

PR #871, opened 2018-09-26
Update LICENSE

xagau:


paullinator:
We are not updating LICENSE without further discussion.

PR #880, opened 2018-09-28
Update simplex plugin to allow receive addresss wrapping

kylanhurt:
The purpose of this task is to *actually* fix the need for text-wrapping for receive addresses within the Simplex plugin. This is to update the simplex plugin dependency with that fix.

Asana Task: https://app.asana.com/0/361770107085503/716460273877812/f

### DIFF
--- a/package.json
+++ b/package.json
@@ -159,7 +159,7 @@
     "babel-preset-react-native": "^1.9.0",
     "body-parser": "^1.18.2",
     "detox": "^7.4.3",
-    "edge-plugin-simplex": "https://github.com/Airbitz/edge-plugin-simplex.git#c1b81e1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -159,7 +159,7 @@
     "babel-preset-react-native": "^1.9.0",
     "body-parser": "^1.18.2",
     "detox": "^7.4.3",
-    "edge-plugin-simplex": "https://github.com/Airbitz/edge-plugin-simplex.git#c1b81e1",
+    "edge-plugin-simplex": "https://github.com/Airbitz/edge-plugin-simplex.git",
```
swansontec: I would have left the commit hash in there, like `#a0df994290`, but this is fine. We can fix this next time we update the plugin.

PR #920, opened 2018-10-11
Scale most screens on react gui

amanalili:
Suggest to QA almost all screens and component on the react-gui module before approving.


paullinator:
Cleaned up and implemented in PR https://github.com/EdgeApp/edge-react-gui/pull/926

Closing

PR #933, opened 2018-10-17
Fix cut-off 'Skip' button text for onboarding

kylanhurt:
The purpose fo this task is to prevent the 'Skip' button text from getting cut off in the onboarding slides (top-right).

Asana Task: https://app.asana.com/0/361770107085503/867580585829917/f


kylanhurt:
@paullinator Should be fixed / scaled now.

PR #945, opened 2018-10-19
Fine tune various scaling parameters and styles

paullinator:
* Change global scaling from .4 to .3
* Increase height of wallet list and transaction row
* Use same walletlist row component and style for iOS and Android
* Make transaction name and wallet name 2 rows if needed
* Add auto text sizing for iOS for Wallet name
* Decrease padding/margin to right of 3 dot options icon on wallet list
* Shrink height of date in tx list
* Make the wallet list token row formatted with the same alignment as regular wallets so that names and amounts line up

### DIFF
--- a/src/lib/scaling.js
+++ b/src/lib/scaling.js
@@ -17,12 +17,12 @@ const scaleHorizontal = size => {
   return out
 }
 
-export const scaleH = (size: number, factor: number = 0.4) => {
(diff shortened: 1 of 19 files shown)


kylanhurt:
@paullinator I'm not sure I can approve it while there are the following bugs on Android (Galaxy S8):
image

paullinator:
Got it. I'll look into that.

PR #967, opened 2018-10-24
James/open camera settings from scan

jamesrisberg:
#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a

Added a new AppStateManager component that responds to app state changes (backgrounding, forgrounding, etc.). It's only job as of now is to check system permissions and update the global state with them.

If Camera permissions are denied, "Open Settings" button should show on scan screen that takes you to settings. Upon return after giving permission, permissions should be updated and the camera should show (iOS restarts the app on permission changes so this we get for free.)


jamesrisberg:
Big update to this PR. On top of making some changes recommended by Eli, I renamed some things for clarity and consolidated Permissions related code to simplify how permissions are being checked and requested around the app

PR #968, opened 2018-10-24
William/fix action types

swansontec:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a


src/modules/Action.js
```diff
@@ -24,7 +25,9 @@ type ActionType =
   // | 'OPEN_VIEWXPUB_WALLET_MODAL'
   // | 'CLOSE_VIEWXPUB_WALLET_MODAL'
   // | 'UNLOCK_WALLET_SEED'
-  | 'SWAP_FROM_TO_CRYPTO_WALLETS'
+  // | 'SELECT_FROM_WALLET_CRYPTO_EXCHANGE'
+  // | 'SELECT_TO_WALLET_CRYPTO_EXCHANGE'
+  // | 'SWAP_FROM_TO_CRYPTO_WALLETS'
```
jamesrisberg: Should we leave these commented lines here? Or can they be removed?

jamesrisberg: Should we leave these commented lines here? Or can they be removed?

swansontec: When Kevin initially did this work, he had comments for all the properly-typed actions. I don't think this is necessary, but I didn't want to rock the boat, so I left them in.

kylanhurt:
@swansontec You able to fix the merge conflicts for this?

swansontec:
Fixed merge conflicts.

src/modules/Action.js
swansontec: I went ahead and removed them, per your comment.

PR #972, opened 2018-10-25
Update AppStoreDescription.txt

peterobrien:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/localization/AppStoreDescription.txt
+++ b/localization/AppStoreDescription.txt
@@ -1,3 +1,8 @@
+Title: Edge - Bitcoin, Ethereum, Monero, Ripple Wallet
+-----
+Short description: Private, secure & easy to use. Buy, sell, send & receive Bitcoin & other crypto.
+-----
+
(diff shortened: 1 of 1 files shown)


localization/AppStoreDescription.txt
```diff
@@ -1,3 +1,8 @@
+Title: Edge - Bitcoin, Ethereum, Monero, Ripple Wallet
+-----
+Short description: Private, secure & easy to use. Buy, sell, send & receive Bitcoin & other crypto.
```
jamesrisberg: should we say cryptocurrencies here instead of crypto? Maybe take a look at ASO keywords and what would garner the best search results

paullinator: @jamesrisberg This is actually the title as chosen by our ASO company. They are actively experimenting with the wording.

PR #997, opened 2018-11-06
Remove deprecated callback APIs

jamesrisberg:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a

Laying the groundwork for the bridging the core and leveling up our thread support!

https://app.asana.com/0/361770107085503/841900095247890/f


src/modules/Core/Account/EdgeAccountCallbackManager.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  componentDidUpdate () {
+    if (this.props.account.id) this.subscribeToAccount()
+  }
+
+  subscribeToAccount = () => {
+    const { account } = this.props
+
+    account.watch('allKeys', () => {
```
swansontec: We might want to watch `currencyWallets` for this, rather than `allKeys`. Otherwise, we might get missing wallets depending on the timing.

src/modules/Core/Account/EdgeAccountCallbackManager.js
```diff
@@ -0,0 +1,83 @@
...
+      // $FlowFixMe
+      setTimeout(() => this.props.updateWalletsRequest(), 0)
+    })
+
+    account.watch('loggedIn', () => {
+      console.log('onLoggedOut')
```
swansontec: I realize this is just a `console.log`, but maybe we should check that `loggedIn === false`.

src/modules/Core/Account/EdgeAccountCallbackManager.js
```diff
@@ -0,0 +1,83 @@
...
+  componentDidUpdate () {
+    if (this.props.account.id) this.subscribeToAccount()
+  }
+
+  subscribeToAccount = () => {
+    const { account } = this.props
+
+    account.watch('allKeys', () => {
+      // $FlowFixMe
+      setTimeout(() => this.props.updateWalletsRequest(), 0)
```
swansontec: I don't think we need `setTimeout` here anymore. That was to hack around a race condition that the new `watch` system simply eliminates.

src/modules/Core/Account/EdgeAccountCallbackManager.js
```diff
@@ -0,0 +1,83 @@
...
+}
+
+type EdgeAccountCallbackManagerDispatchProps = {
+  updateWalletsRequest: () => any,
+  updateExchangeRates: () => any
+}
+
+type Props = EdgeAccountCallbackManagerStateProps & EdgeAccountCallbackManagerDispatchProps
+
+class EdgeAccountCallbackManager extends React.Component<Props> {
```
swansontec: Should this extend `PureComponent`? If our parent re-renders, we might re-render as well, which could cause double subscribes.

swansontec:
I really like the component-based approach being used in this pull request. Nice and clean.

PR #1026, opened 2018-11-17
Use high performance timers for perf timers

paullinator:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ x] n/a

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -101,7 +101,7 @@ if (ENABLE_PERF_LOGGING) {
   // $FlowFixMe: suppressing this error until we can find a workaround
   global.pnow = function (label: string) {
     const d = makeDate()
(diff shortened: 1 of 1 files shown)


src/app.js
```diff
@@ -113,7 +113,7 @@ if (ENABLE_PERF_LOGGING) {
       perfCounters[label] = 0
     }
     if (typeof perfTimers[label] === 'undefined') {
-      perfTimers[label] = Date.now()
+      perfTimers[label] = global.nativePerformanceNow()
```
thehobbit85: Where do you get this `nativePerformanceNow` from?

paullinator: It’s built into react native.

PR #1042, opened 2018-11-23
Wallet List: Remove paddings on the right side when sorting

amanalili:
Task: Sort wallet: looks to be excessive padding on the right hand side

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android


paullinator:
Looks like more space needs to be given to the wallet names as they get truncated too early.  Please change to allow wallet names to take up 2 rows if needed like the non-sorted view. Also make sure the icon is on a column by itself. On Android it shows above the wallet name.

paullinator:
screenshot_20181123-234058
screen shot 2018-11-23 at 23 40 36

PR #1046, opened 2018-11-26
Allen/settings shape shift on off

ahujsak:
Fixes asana 
https://app.asana.com/0/361770107085503/897596338626223
Add "activate button" in Shapeshift settings

Video of tested video
https://www.screencast.com/t/vzUnmqtoDC

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


paullinator:
Testing this branch, I get this error in the debugger.

```ExceptionsManager.js:71 Warning: Encountered two children with the same key, `shapeshift`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
    in RCTView (at View.js:60)
    in View (at ExchangeSettingsScene.js:115)
    in RCTView (at View.js:60)
    in View (at ExchangeSettingsScene.js:105)
    in RCTView (at View.js:60)
    in View (at ExchangeSettingsScene.js:103)
    in RCTSafeAreaView (at SafeAreaView.ios.js:34)
    in SafeAreaView (at SafeAreaView.ui.js:18)
    in SafeAreaViewComponent (at ExchangeSettingsScene.js:102)
    in ExchangeSettingsComponent (created by Connect(ExchangeSettingsComponent))
    in Connect(ExchangeSettingsComponent) (at navigationStore.js:525)
    in Wrapped (at SceneView.js:17)
    in SceneView (at CardStack.js:466)```

**Comment by paullinator** [comment]
Im' getting bizarre behavior when using this with a test account and logging into/out of Shapeshift in Settings. After login, I get a dark screen with a spinner that never goes away. I reboot and the screen shows "signout" I tap signout and I get a spinner and when it goes away, Sign Out is still there.

File: src/components/scenes/ExchangeSettingsScene.js [hunk 1]
```diff
@@ -48,6 +54,7 @@ export class ExchangeSettingsComponent extends Component<ExchangeSettingsProps,
     super(props)
     this.state = {}
     const { exchanges } = props
+    console.log('exchange: : ', exchanges)
```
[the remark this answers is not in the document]
ahujsak: removed

src/components/common/RowWithButton.js
```diff
@@ -0,0 +1,48 @@
...
+
+import T from '../../modules/UI/components/FormattedText/index'
+import styles, { styles as styleRaw } from '../../styles/scenes/SettingsStyle'
+
+type Props = {
+  logo: string,
+  leftText: string,
+  rightText: string,
+  onToggle(): void
+}
```
kylanhurt: Should this be `onToggle` or `onPress`?

src/styles/scenes/SettingsStyle.js
```diff
@@ -69,6 +69,11 @@ export const styles = {
     backgroundColor: THEME.COLORS.TRANSPARENT
   },
 
+  // /////// Double Row
+  doubleRowContainer: {
+    height: scale(88),
+    backgroundColor: THEME.COLORS.WHITE
```
kylanhurt: Do we really need to set a height for the container? In general I think it's better to not hard-code heights if we don't have to.

ahujsak: ok, removed. I was just looking at what we did for single row. since the rows are defined and we are having two in the container, we can remove

src/components/common/RowWithButton.js
ahujsak: yup, copy paste legacy. fixed.

paullinator:
The `Sign Out` button should fully sign out the user and allow them to login with a different username or create a new account. Otherwise there's not much purpose in the Sign Out

paullinator:
Also need to resolve conflicts

ahujsak:
@paullinator I am open to suggestions on the sign out -> sign in with a different user,, however that is out of our control. We open up a webview, and shapeshift reads a cookie, and if the user has signed in from that device before, then they check a session etc.. and auto sign the person in.

PR #1053, opened 2018-11-29
Wallet List Sort: Fix sorting currency text and symbol, seperate currency logo to text

amanalili:
Task: Sort wallet: looks to be excessive padding on the right hand side

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android

### DIFF
--- a/src/components/common/SortableWalletListRow.js
+++ b/src/components/common/SortableWalletListRow.js
@@ -2,7 +2,7 @@
 
 import { bns } from 'biggystring'
 import React, { Component } from 'react'
-import { ActivityIndicator, Image, Platform, TouchableHighlight, View } from 'react-native'
+import { ActivityIndicator, Image, TouchableHighlight, View } from 'react-native'
 import { connect } from 'react-redux'
 
 import sort from '../../assets/images/walletlist/sort.png'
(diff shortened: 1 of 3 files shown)


src/components/common/SortableWalletListRow.js
```diff
@@ -40,28 +40,24 @@ class SortableWalletListRow extends Component<Props, State> {
...
+                <Image
+                  style={[styles.rowDragCurrencyLogo, { marginLeft: 0, marginRight: 5 }]}
+                  transform={[{ translateY: 3 }]}
+                  source={{ uri: symbolImageDarkMono }}
+                />
               )}
             </View>
+            <View style={[styles.rowNameTextWrapAndroidIos]}>
+              <T style={[styles.rowNameText]} numberOfLines={2} adjustsFontSizeToFit={true} minimumFontScale={0.6}>
+                {Platform.OS === 'ios' ? name : cutOffText(name, 34)}
```
paullinator: For Android please use ellipsesMode instead of just cutting off text. cutOffText knows nothing about the screen size and available space.

https://facebook.github.io/react-native/docs/text#ellipsizemode

paullinator: defaulted to what? If it's already on, then please just remove the cutOffText()

amanalili: It is needed for the ios platform. That is why there the conditional is for.

Also, it is exactly the same code on the unsorted list that we always use.

paullinator: How is it needed for iOS when the conditional above only has it running on Android? If it’s on the unsourced list code, we need to remove it there also. It was always a bad idea to hardcode ellipses to a specific char length.

amanalili: Let me check it out without that code and will evaluate if it cuts off the same as before

amanalili: Updated

PR #1068, opened 2018-12-07
Faa.st exchange plugin support

dylanseago:
Support exchange plugin for the Faa.st service.

Tested on a Google Pixel 3 XL running Android 9.0.

Depends on https://github.com/EdgeApp/edge-core-js/pull/176


src/actions/CryptoExchangeActions.js
```diff
@@ -164,6 +164,10 @@ export const shiftCryptoCurrency = () => async (dispatch: Dispatch, getState: Ge
             name = 'Changelly'
             supportEmail = 'support@changelly.com'
             break
+          case 'faast':
```
swansontec: @ahujsak The plugins already export this information from their `swapInfo`. There should be absolutely no reason for this information to exist in the GUI. We should take a look at fixing this before we add more plugins so it doesn't get worse.

dylanseago: @swansontec One additional thing I noticed is that the Exchange settings page doesn't use pluginName. It looks like it's using a capitalized version of the plugin code name instead.

paullinator:
Closing in favor of PR https://github.com/EdgeApp/edge-react-gui/pull/1212 which fixes missing icons

PR #1134, opened 2019-01-04
Bitrefill

ahujsak:
This adds support for the Bitrefill plugin. Upgrades the PluginView to handle standardized url callbacks, make sepnds, and displaying a success page. 

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


copy-bitrefill.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
kylanhurt: @ahujsak Paul wants us to remove these files, since `copy-plugins.js` should work just fine

ios/edge.xcodeproj/project.pbxproj
```diff
@@ -5,6 +5,7 @@
 	};
```
kylanhurt: Remove this file

src/components/scenes/PluginViewScene.js
```diff
@@ -224,15 +238,49 @@ class PluginView extends React.Component<PluginProps, PluginState> {
...
       return
     }
-    // TODO: improve handling of edge-ret URIs
-    if (navState.url.match(/edge-ret:\/\/plugins/)) {
+    const parsedUrl = parse(navState.url, {}, true)
+    if (parsedUrl.protocol === 'edge:' && parsedUrl.hostname === 'x-callback-url') {
+      switch (parsedUrl.pathname) {
+        case '/paymentUri':
+          this.props.abcWallet.parseUri(parsedUrl.query.uri)
+            .then(result => {
```
kylanhurt: I don't think we should be calling it "abcWallet" anymore, but `coreWallet`. ABC used to mean "Airbitz Core"

src/components/scenes/PluginViewScene.js
```diff
@@ -224,15 +238,49 @@ class PluginView extends React.Component<PluginProps, PluginState> {
...
+      this.bridge.navStackClear()
+      return
+    }
+
     if (!navState.canGoForward) {
       this.bridge.navStackPush(navState.url)
     } else if (!navState.canGoBack) {
       this.bridge.navStackClear()
+    } else {
     }
```
kylanhurt: Unnecessary else clause, but not a deal breaker

src/modules/UI/scenes/Plugins/api.js
```diff
@@ -4,18 +4,21 @@ import type { EdgeReceiveAddress, EdgeTransaction } from 'edge-core-js'
...
-  fiatCurrencyCode: w.fiatCurrencyCode
-})
+const formatWallet = w => {
+  return {
+    id: w.id,
+    name: w.name,
+    type: w.type,
+    currencyCode: w.currencyCode ? w.currencyCode : w.currencyInfo.currencyCode,
+    primaryNativeBalance: w.currencyCode,
+    fiatCurrencyCode: w.fiatCurrencyCode
```
kylanhurt: Just want to point out that `primaryNativeBalance` should not have anything to do with a `currencyCode`

src/modules/UI/scenes/Plugins/api.js
```diff
@@ -66,13 +69,22 @@ export class PluginBridge {
...
+    return Promise.resolve(null)
+  }
+
+  changeWallet (): Promise<any> {
     this.context.toggleWalletList()
     return Promise.resolve(null)
   }
 
+  getAPIKey (): Promise<string> {
+    const key: string = ENV.BIT_REFILL_API_KEY
```
kylanhurt: Please rename this, we may have tons of plugins which means `getApiKey` can't only be relevant for BitRefill, etc.

src/components/scenes/PluginViewScene.js
```diff
@@ -246,10 +294,14 @@ class PluginView extends React.Component<PluginProps, PluginState> {
           allowUniversalAccessFromFileURLs
           onMessage={this._onMessage}
           onNavigationStateChange={this._onNavigationStateChange}
-          originWhitelist={['file://', 'https://']}
+          originWhitelist={['file://', 'https://', 'http://', 'edge://']}
           ref={this._setWebview}
           scalesPageToFit={contentScaling}
           source={this._renderWebView()}
+          setWebContentsDebuggingEnabled={true}
```
paullinator: What is the ramification of leaving this on in production?

src/modules/UI/scenes/Plugins/api.js
paullinator: There are already functions for getting a value from the GUI. For reference, look at how the Simplex plugin was doing this.

copy-bitrefill.sh
ahujsak: copy plugins does something different,... it does not move them from the local repository and this is faster than updotting while developing.

ios/edge.xcodeproj/project.pbxproj
ahujsak: why would I remove this file?...

src/modules/UI/scenes/Plugins/api.js
ahujsak: @kylanhurt changed to a better naming convention that will get the api key from the plugin name in an object of all the plugin api keys

src/components/scenes/PluginViewScene.js
ahujsak: No ramification. It enables you to use safari to debug the webview react native code etc.. used it to see console logs from the plugin that do not surface in react native debugger. removed it anyway.

src/modules/UI/scenes/Plugins/api.js
ahujsak: fixed, was copy past error from previous file

src/components/scenes/PluginViewScene.js
ahujsak: removed

src/components/scenes/PluginViewScene.js
ahujsak: fixed

copy-bitrefill.sh
ahujsak: removed

src/modules/UI/scenes/Plugins/api.js
paullinator: I'm not looking at the code, but in @teneighty 's docs here

https://github.com/EdgeApp/edge-docs/blob/master/source/includes/_plugins.md

He references a `Config` object where you can do a get() call specifically for API_KEYs. Was that not implemented?

paullinator: screen shot 2019-01-05 at 14 18 48

ahujsak: Ok, refactored to use that and it is working. New method has been removed.

ios/edge.xcodeproj/project.pbxproj
ahujsak: non issue after rebasing

PR #1154, opened 2019-01-08
add segwitAddress to RequestScene

ahujsak:
Adding bech32 to the request screen. 
Fixes Asana 
Implement bech32 GUI
https://app.asana.com/0/361770107085503/963606581009777

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/scenes/RequestScene.js
+++ b/src/components/scenes/RequestScene.js
@@ -37,6 +37,7 @@ export type RequestStateProps = {
   loading: false,
   primaryCurrencyInfo: GuiCurrencyInfo,
   publicAddress: string,
+  segwitAddress: string,
   legacyAddress: string,
   secondaryCurrencyInfo: GuiCurrencyInfo,
   showToWalletModal: boolean,
@@ -50,6 +51,7 @@ export type RequestLoadingProps = {
   loading: true,
   primaryCurrencyInfo: null,
   publicAddress: string,
+  segwitAddress: string,
   legacyAddress: string,
   secondaryCurrencyInfo: null,
   showToWalletModal: null,
@@ -69,6 +71,7 @@ export type Props = LoadingProps | LoadedProps
 export type State = {
   publicAddress: string,
   legacyAddress: string,
+  segwitAddress: string,
   encodedURI: string,
   minimumPopupModalState: CurrencyMinimumPopupState
 }
@@ -85,6 +88,7 @@ export class Request extends Component<Props, State> {
     this.state = {
       publicAddress: props.publicAddress,
       legacyAddress: props.legacyAddress,
+      segwitAddress: props.segwitAddress,
       encodedURI: '',
       minimumPopupModalState
     }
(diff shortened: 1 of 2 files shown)


src/connectors/scenes/RequestConnector.js
```diff
@@ -64,6 +65,7 @@ const mapStateToProps = (state: State): RequestStateProps | RequestLoadingProps
     guiWallet,
     publicAddress: guiWallet.receiveAddress.publicAddress || '',
     legacyAddress: guiWallet.receiveAddress.legacyAddress || '',
+    segwitAddress: guiWallet.receiveAddress.segwitAddress || '',
```
swansontec: Could the prop have been `string | void`? I guess this works fine, but I generally dislike blank strings. Maybe a cleanup for another PR.

src/components/scenes/RequestScene.js
```diff
@@ -130,7 +134,8 @@ export class Request extends Component<Props, State> {
     const { edgeWallet, useLegacyAddress } = this.props
     let publicAddress = this.props.publicAddress
     let legacyAddress = this.props.legacyAddress
-    const abcEncodeUri = useLegacyAddress ? { publicAddress, legacyAddress } : { publicAddress }
+    let segwitAddress = this.props.segwitAddress
+    const abcEncodeUri = segwitAddress ? { segwitAddress, publicAddress } : useLegacyAddress ? { publicAddress, legacyAddress } : { publicAddress }
```
swansontec: Does `encodeUri` understand segwitAddress`? Can you verify that if you enter an amount, the QR code still uses the segwit address, and not the public address?

ahujsak: No it does not.

src/connectors/scenes/RequestConnector.js
ahujsak: lets leave for another PR, this was just doing what was there before, really not trying to do a refactor on this while trying to get it in.

src/components/scenes/RequestScene.js
```diff
@@ -272,8 +286,9 @@ export class Request extends Component<Props, State> {
   }
 
   onExchangeAmountChanged = async (amounts: ExchangedFlipInputAmounts) => {
-    const { publicAddress, legacyAddress } = this.state
-    const edgeEncodeUri: EdgeEncodeUri = this.props.useLegacyAddress && legacyAddress ? { publicAddress, legacyAddress } : { publicAddress }
+    const { segwitAddress, publicAddress, legacyAddress } = this.state
+    const edgeEncodeUri: EdgeEncodeUri =
+      this.props.useLegacyAddress && legacyAddress ? { publicAddress, legacyAddress } : segwitAddress ? { segwitAddress, publicAddress } : { publicAddress }
```
swansontec: I looked at the types, and it seems like `EdgeEncodeIUri` does take a `segwitAddress` parameter, which seems wrong to me. I'm not in the mood to refactor this right now, so it can stay as-is.

paullinator:
Do not merge until bech32 support is in.

PR #1161, opened 2019-01-10
Allen/bitrefill tighten up

ahujsak:
This adjusts for new callback structure for purchases. 
It also handles possible race condition with getting selected wallet
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/scenes/PluginViewScene.js
+++ b/src/components/scenes/PluginViewScene.js
@@ -116,6 +116,7 @@ class PluginView extends React.Component<PluginProps, PluginState> {
   updateBridge: Function
   webview: any
   successUrl: string | null
+  openingSendConfirmation: boolean
   constructor (props) {
     super(props)
     this.state = {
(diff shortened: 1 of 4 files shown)


src/components/scenes/PluginViewScene.js
```diff
@@ -240,6 +241,38 @@ class PluginView extends React.Component<PluginProps, PluginState> {
       return
     }
     const parsedUrl = parse(navState.url, {}, true)
+    if (parsedUrl.query['edge-callback']) {
```
paullinator: This `if` should be 

`if (parsedUrl.query) {`

You are assuming the `query` parameter exists on the next line so you should check that it does.

src/components/scenes/PluginViewScene.js
```diff
@@ -240,6 +241,38 @@ class PluginView extends React.Component<PluginProps, PluginState> {
...
+      switch (parsedUrl.query['edge-callback']) {
```
paullinator: This is nearly all duplicate code from below. Can you consolidate this logic into common code. It looks like if someone used `x-callback-url` they would be missing some necessary logic like.

`openingSendConfirmation = true`

src/components/scenes/PluginViewScene.js
ahujsak: no longer relevant

src/components/scenes/PluginViewScene.js
ahujsak: moved the code down before removing this whole workflow

PR #1182, opened 2019-01-18
Fix payment protocol destination detection

swansontec:
The `merchant` field doesn't contain they payee name in the current implementation. Always use the domain name instead to avoid crashes.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a


swansontec:
BitPay is not the only place that supports JSON payment protocol anymore. CoinText is adding support for this. In their case, it would crash and corrupt the app, so fix that.

PR #1183, opened 2019-01-19
Transaction List: Transaction List Loading Bar

amanalili:
Task: Add green loading bar to Tx list

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android


src/connectors/scenes/TransactionListConnector.js
```diff
@@ -77,7 +85,8 @@ const mapStateToProps = (state: State) => {
     showToWalletModal: state.ui.scenes.walletListModal.walletListModalVisible,
     requiredConfirmations,
     numTransactions,
-    isBalanceVisible
+    isBalanceVisible,
+    loadingPercent
```
paullinator: Do not connect `loadingPercent` directly to redux using the Scene's connector. This will cause a re-render of the entire scene on each loadingPercent update.

Please use the WiredProgressBar like it was used in the WalletListScene

amanalili:
Done squashing

PR #1211, opened 2019-02-01
Kylan/repaid synced settings

kylanhurt:
The purpose of this task is to repaid bad data within the synced settings.

Asana Tasks:
https://app.asana.com/0/361770107085503/1103570340937804/f
https://app.asana.com/0/361770107085503/1103777801053757/f

### DIFF
--- a/src/modules/Login/action.js
+++ b/src/modules/Login/action.js
@@ -190,7 +190,11 @@ export const initializeAccount = (account: EdgeAccount, touchIdInfo: Object) => 
   }
 }
 
-export const mergeSettings = (loadedSettings: Object, defaults: Object, types: Object): { finalSettings: Object } => {
+export const mergeSettings = (
+  loadedSettings: Object,
+  defaults: Object,
+  types: Object
+): { finalSettings: Object, isOverwriteNeeded: boolean, isDefaultTypeIncorrect: boolean } => {
   const finalSettings = {}
   // begin process for repairing damaged settings data
   let isOverwriteNeeded = false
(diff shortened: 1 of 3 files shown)


src/modules/Login/action.js
```diff
@@ -176,6 +190,53 @@ export const initializeAccount = (account: EdgeAccount, touchIdInfo: Object) =>
...
+  const finalSettings = {}
+  // begin process for repairing damaged settings data
+  let isOverwriteNeeded = false
+  let isDefaultTypeIncorrect = false
+  for (const key in defaults) {
+    // if the type of the setting default does not meet the enforced type
+    const defaultSettingType = typeof defaults[key]
+    if (defaultSettingType !== types[key]) {
+      isDefaultTypeIncorrect = true
+      console.error('MismatchedDefaultSettingType key: ', key, ' with defaultSettingType: ', defaultSettingType, ' and necessary type: ', types[key])
```
paullinator: Should just throw here. This should be considered a coding error and be caught immediately

src/modules/Login/action.js
```diff
@@ -176,6 +190,53 @@ export const initializeAccount = (account: EdgeAccount, touchIdInfo: Object) =>
...
+    }
+    // if the type of the loaded setting does not meet the enforced type
+    // eslint-disable-next-line valid-typeof
```
paullinator: Why do you have the eslint rule exception?

kylanhurt: The eslint disabling is meant for the `typeof` operator a few lines down, I realize the commented-out console.log lines probably cause confusion (I will remove them). Eslint doesn't like us using the `typoef` operator to compare to anything other than a string (ie 'object' or 'string'), so comparing to a variable that holds the string value of 'object' or 'string' causes eslint to throw an error.

src/modules/Login/action.js
paullinator: Ok fair enough. I didn’t see the try catch.

PR #1218, opened 2019-02-04
Better initializeAccount error-handling and prevent prop of undefined error

kylanhurt:
The purpose of this task is to fix the initializeAccount method to **not** execute the `catch` clause if there is no settings.json file, which was causing a "cannot find property ____ of undefined" error.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

Asana Task: https://app.asana.com/0/361770107085503/1108083253570323/f

### DIFF
--- a/src/modules/Login/action.js
+++ b/src/modules/Login/action.js
@@ -197,47 +197,43 @@ export const mergeSettings = (
   defaults: Object,
   types: Object
 ): { finalSettings: Object, isOverwriteNeeded: boolean, isDefaultTypeIncorrect: boolean } => {
-  try {
-    const finalSettings = {}
-    // begin process for repairing damaged settings data
-    let isOverwriteNeeded = false
-    let isDefaultTypeIncorrect = false
-    for (const key in defaults) {
-      // if the type of the setting default does not meet the enforced type
-      const defaultSettingType = typeof defaults[key]
-      if (defaultSettingType !== types[key]) {
-        isDefaultTypeIncorrect = true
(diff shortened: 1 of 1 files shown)


src/modules/Login/action.js
```diff
@@ -139,7 +140,8 @@ export const initializeAccount = (account: EdgeAccount, touchIdInfo: Object) =>
     accountInitObject.archivedWalletIds = archivedWalletIds
 
     const loadedSyncedSettings = await getSyncedSettings(account)
-    const mergedSyncedSettings = mergeSettings(loadedSyncedSettings, SYNCED_ACCOUNT_DEFAULTS, SYNCED_ACCOUNT_TYPES)
+    const syncedSettings = { ...loadedSyncedSettings } // will prevent mergeSettings trying to find prop of undefined
```
swansontec: So, like `const syncedSettings = loadedSyncedSettings || {}` might be another way to do the same thing.

src/modules/Login/action.js
```diff
@@ -193,43 +197,47 @@ export const mergeSettings = (
...
-  return {
-    finalSettings,
-    isOverwriteNeeded,
-    isDefaultTypeIncorrect
+    return {
+      finalSettings,
+      isOverwriteNeeded,
+      isDefaultTypeIncorrect
+    }
+  } catch (e) {
```
swansontec: The try block is not needed if it's just re-throwing.

PR #1228, opened 2019-02-11
Remove unactivated EOS from WalletListModal

kylanhurt:
The purpose of this task is to omit unactivated wallets (EOS) from dropdowns for requesting and sending.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a

Asana Task: https://app.asana.com/0/361770107085503/1108931850293728/f

### DIFF
--- a/src/components/scenes/RequestScene.js
+++ b/src/components/scenes/RequestScene.js
@@ -40,7 +40,8 @@ export type RequestStateProps = {
   legacyAddress: string,
   secondaryCurrencyInfo: GuiCurrencyInfo,
   showToWalletModal: boolean,
-  useLegacyAddress: boolean
+  useLegacyAddress: boolean,
+  wallets: { [string]: GuiWallet }
 }
 export type RequestLoadingProps = {
   edgeWallet: null,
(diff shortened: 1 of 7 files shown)


src/components/scenes/RequestScene.js
```diff
@@ -225,8 +226,15 @@ export class Request extends Component<Props, State> {
     }
 
     const color = 'white'
-    const { primaryCurrencyInfo, secondaryCurrencyInfo, exchangeSecondaryToPrimaryRatio, onSelectWallet } = this.props
+    const { primaryCurrencyInfo, secondaryCurrencyInfo, exchangeSecondaryToPrimaryRatio, onSelectWallet, wallets } = this.props
     const requestAddress = this.props.useLegacyAddress ? this.state.legacyAddress : this.state.publicAddress
+    const walletsCopy = { ...wallets }
```
paullinator: Logic seems a bit backwards. Instead of copying everything then deleting the inactive wallets. Why not just iterate over all wallets and copy the ones that are activated.

src/components/scenes/ScanScene.js
```diff
@@ -70,8 +72,14 @@ export class Scan extends Component<Props> {
   }
 
   render () {
-    const { onSelectWallet } = this.props
-
+    const { onSelectWallet, wallets } = this.props
```
paullinator: same as comment above.

PR #1233, opened 2019-02-15
wired up developer button

ahujsak:
This turns on a developer mode. Button toggles on and off in Settings. If button is on, then custom plugins are enabled. 
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/locales/en_US.js
+++ b/src/locales/en_US.js
@@ -193,6 +193,8 @@ const strings = {
   settings_account_title_cap: 'Account',
   settings_button_change_password: 'Change Password',
   settings_button_debug: 'Debug',
+  settings_turn_on_developer_mode: 'Turn On Developer',
+  settings_turn_off_developer_mode: 'Turn Off Developer',
   settings_button_lock_settings: 'Tap to Lock Account Settings',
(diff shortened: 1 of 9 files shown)


paullinator:
It doesn't appear that there is any feedback to tell the user that they've turned on or off developer mode. A simple Alert modal would do. The button state also doesn't change so that makes it even more confusing. Can you change the button to "Developer Mode" and throw an Alert modal that says

Title: "Developer Mode"
Message: "Enabled"

or Message: "Disabled"

src/locales/en_US.js
```diff
@@ -193,6 +193,8 @@ const strings = {
   settings_account_title_cap: 'Account',
   settings_button_change_password: 'Change Password',
   settings_button_debug: 'Debug',
+  settings_turn_on_developer_mode: 'Turn On Developer',
```
paullinator: Change text to 'Enable Developer Mode' and 'Disable Developer Mode'.

Shrink text if needed to ensure this fits on an iPhone 5 width screen

PR #1234, opened 2019-02-15
Warn users on login if notifications are disabled

jamesrisberg:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

This is a subtask of https://app.asana.com/0/361770107085503/838231819659034/f

### DIFF
--- a/src/locales/en_US.js
+++ b/src/locales/en_US.js
@@ -490,7 +490,10 @@ const strings = {
   alert_deep_link_no_wallet: 'No wallets exist that support %1$s. You can create a wallet for %2$s by tapping + on the Wallet List screen.',
   load_plugin: 'Load Plugin',
   plugin_url: 'Plugin URL',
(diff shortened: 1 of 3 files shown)


jamesrisberg:
So if a user does not have notifications enabled it will warn them every time they open the app indefinitely?

paullinator:
Let's use this heuristic. It's just a best guess but we can update in the future as we see fit.

1. Show on 1st launch of login scene if the user has had logged in accounts before.
2. Also show on 3rd launch of the edge-login-ui scene if the user has had logged in accounts before. 3. After that, only show once every 2 weeks and then give them an option to "Don't show this warning"

swansontec:
After thinking more about this, we would like to move these changes into https://github.com/EdgeApp/edge-login-ui

PR #1236, opened 2019-02-18
Check for clock skew on login

jamesrisberg:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

Subtask of https://app.asana.com/0/361770107085503/838231819659034/f

### DIFF
--- a/ios/edge/Info.plist
+++ b/ios/edge/Info.plist
@@ -68,6 +68,13 @@
 	<dict>
 		<key>NSExceptionDomains</key>
 		<dict>
+			<key>worldtimeapi.org</key>
+			<dict>
+				<key>NSIncludesSubdomains</key>
+				<true/>
+				<key>NSExceptionAllowsInsecureHTTPLoads</key>
+				<true/>
+			</dict>
 			<key>localhost</key>
 			<dict>
 				<key>NSExceptionAllowsInsecureHTTPLoads</key>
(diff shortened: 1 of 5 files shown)


ios/edge/Info.plist
```diff
@@ -68,6 +68,13 @@
 	<dict>
 		<key>NSExceptionDomains</key>
 		<dict>
+			<key>worldtimeapi.org</key>
+			<dict>
+				<key>NSIncludesSubdomains</key>
+				<true/>
+				<key>NSExceptionAllowsInsecureHTTPLoads</key>
```
paullinator: I don't believe this will work in production apps as Apple is strict about https

src/modules/Login/action.js
```diff
@@ -180,6 +183,30 @@ export const initializeAccount = (account: EdgeAccount, touchIdInfo: Object) =>
     accountInitObject.pinMode = coreFinal.pinMode
     accountInitObject.otpMode = coreFinal.otpMode
 
+    // Check clock skew against atomic time
+    try {
+      const remoteTime = await fetch('http://worldtimeapi.org/api/timezone/Etc/GMT')
+      const body = await remoteTime.json()
```
paullinator: Do not put a blocking network call in initAccount as this can slow down login times for users. This should run in the background and throw a modal once complete.

paullinator: Also, we should probably have this at app startup time instead of initAccount as the user might not be able to login at all. This would entail moving almost the entire PR to the edge-login-ui repo which is a good idea so SDK partners get it as well.

PR #1255, opened 2019-02-25
fix wallet spinner. tie to redux.

ahujsak:
Fixes P): Wallet Selector (exchange screen) - If a create wallet fails, the wallet selector button gets stuck spinning indefinitely
https://app.asana.com/0/361770107085503/1111016622465604

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/scenes/CryptoExchangeScene.js
+++ b/src/components/scenes/CryptoExchangeScene.js
@@ -59,7 +59,8 @@ export type CryptoExchangeSceneComponentStateProps = {
   wallets: { [string]: GuiWallet },
   totalWallets: number,
   supportedWalletTypes: Array<Object>,
-  state: State
+  state: State,
+  creatingWallet: boolean
 }
 
 export type CryptoExchangeSceneComponentDispatchProps = {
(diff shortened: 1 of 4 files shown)


src/components/scenes/CryptoExchangeScene.js
```diff
@@ -323,7 +318,7 @@ export class CryptoExchangeScene extends Component<Props, LocalState> {
     const allowedWallets = []
     for (const id in wallets) {
       const wallet = wallets[id]
-      if (excludedCurrencyCode !== wallet.currencyCode) {
+      if (excludedCurrencyCode !== wallet.currencyCode && wallet.currencyCode !== 'ETH') {
         walletCurrencyCodes.push(wallet.currencyCode)
```
kylanhurt: @ahujsak Could you explain why ETH is being excluded here?

ahujsak: ok, that leaked in from a fix from another bug. I am going to remove it, and it will be in the next PR.

PR #1256, opened 2019-02-26
Fix EOS nativeAmount tx error

kylanhurt:
The purpose of this task is to fix the transaction error that occurs while paying for EOS activation. The fix was to fix a regression by passing the `props` (not `state`) version of `nativeAmount` to the `signBroadcastAndSave` action. We are able to keep `exchangeAmount` as-is because it is only used for updating the displayed fiat amount on SendConfirmation (which the app should route away from on success, anyway).

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

Asana Task: https://app.asana.com/0/361770107085503/1111288373353711/f

### DIFF
--- a/src/components/scenes/SendConfirmationScene.js
+++ b/src/components/scenes/SendConfirmationScene.js
@@ -327,7 +327,9 @@ export class SendConfirmation extends Component<Props, State> {
   }
 
   onSlideToConfirm = () => {
-    this.props.signBroadcastAndSave(this.state.nativeAmount, this.state.exchangeAmount, this.props.fiatPerCrypto.toString())
+    const { nativeAmount, signBroadcastAndSave, fiatPerCrypto } = this.props
+    const { exchangeAmount } = this.state
+    signBroadcastAndSave(nativeAmount, exchangeAmount, fiatPerCrypto.toString())
   }
 
   networkFeeSyntax = () => {
--- a/src/actions/SendConfirmationActions.js
+++ b/src/actions/SendConfirmationActions.js
@@ -119,8 +119,12 @@ export const sendConfirmationUpdateTx = (guiMakeSpendInfo: GuiMakeSpendInfo | Ed
   dispatch(newSpendInfo(spendInfo, authRequired))
 
   await makeSpend(edgeWallet, spendInfo)
-    .then(edgeTransaction => dispatch(updateTransaction(edgeTransaction, guiMakeSpendInfoClone, forceUpdateGui, null)))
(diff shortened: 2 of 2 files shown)


src/components/scenes/SendConfirmationScene.js
```diff
@@ -327,7 +327,9 @@ export class SendConfirmation extends Component<Props, State> {
   }
 
   onSlideToConfirm = () => {
-    this.props.signBroadcastAndSave(this.state.nativeAmount, this.state.exchangeAmount, this.props.fiatPerCrypto.toString())
+    const { nativeAmount, signBroadcastAndSave, fiatPerCrypto } = this.props
+    const { exchangeAmount } = this.state
+    signBroadcastAndSave(nativeAmount, exchangeAmount, fiatPerCrypto.toString())
```
swansontec: Ok, so you are using `this.props.nativeAmount` instead of `this.state.nativeAmount`, but it seems like editing the amount in the send confirmation screen would then be broken. I would like a little more clarity about why this works.

paullinator: @kylanhurt should probably clarify definitively, but from my inspection, it looks like each edit to the amount gets thrown up to redux via updateAmount() which comes back into this scene via props.nativeAmount from the connector, hence why the amount does change. I am however, mostly curious as to how this broke in the first place.

paullinator: Looking at this further, it's likely this PR unfixes the monero race-condition bug. Since the values are pulled from redux, redux might have values that are out of date since the updateAmount() function has to call makeSpend and wait for the results which could be slow on monero. We need to properly fix the problem which I think I just found.

When we update the amount, we call updateAmount() which is async so multiple edits to the amount could have multiple updateAmount() calls in flight at the same time. However the input lock and spinner we have will automatically turn off once the first one updateAmount() returns even though it's not the most recent. Once if the user waits awhile will the last, most up-to-date updateAmount come in. I'd suggest putting in a counter with each updateAmount and make sure that the local `this.state.updateAmountCounter` matches the incoming `this.props.updateAmountCounter` before removing the spinner and input lock.

kylanhurt: I'll have to double-check which lifecycle methods get called when EOS routes to the SendConfirmation scene. It's possible that we can get the component `state.nativeAmount` to get its value from `props.nativeAmount` and change the `signBroadcastAndSave` to use the `state` version. When you route to SendConfirmation from EOS it should lock the component state anyway so there shouldn't be a risk of `state.nativeAmount` and `props.nativeAmount` getting out of sync.

PR #1260, opened 2019-02-27
Fix SendConf Make Spend Issue

kylanhurt:
The purpose of this task is to fix the Send Confirmation bug whereby the order of makeSpend's requested differ from the order of the transactions returned (and therefore result in the user sending a tx other than the latest). Solution was to move some redux action functionality to the scene component.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a

### DIFF
--- a/src/components/scenes/SendConfirmationScene.js
+++ b/src/components/scenes/SendConfirmationScene.js
@@ -2,7 +2,7 @@
 
 import { bns } from 'biggystring'
 import { Scene } from 'edge-components'
-import type { EdgeDenomination, EdgeMetadata } from 'edge-core-js'
+import type { EdgeCurrencyWallet, EdgeDenomination, EdgeMetadata, EdgeSpendInfo, EdgeTransaction } from 'edge-core-js'
 import React, { Component } from 'react'
 import { TouchableOpacity, View } from 'react-native'
 import slowlog from 'react-native-slowlog'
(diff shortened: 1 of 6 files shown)


src/components/scenes/SendConfirmationScene.js
```diff
@@ -12,6 +12,7 @@ import { UniqueIdentifierModalConnect as UniqueIdentifierModal } from '../../con
 import { getSpecialCurrencyInfo } from '../../constants/indexConstants.js'
 import { intl } from '../../locales/intl'
 import s from '../../locales/strings.js'
+import { makeSpend } from '../../modules/Core/Wallets/api.js'
```
swansontec: I would like to stop using these pointless wrappers and just call core methods directly. `wallet.makeSpend(...)` is more obvious than `makeSpend(wallet, ...)`, and it's less code to maintain.

src/components/scenes/SendConfirmationScene.js
```diff
@@ -316,14 +326,35 @@ export class SendConfirmation extends Component<Props, State> {
...
+    const spendInfo = getSpendInfoWithoutState(guiMakeSpendInfoClone, sceneState, currencyCode)
+
+    try {
+      newSpendInfo(spendInfo, isLimitExceeded)
+      const edgeTransaction = await makeSpend(coreWallet, spendInfo)
+      if (count === this.count) {
+        this.lastSeenCount = count
+        updateTransaction(edgeTransaction, guiMakeSpendInfoClone, false, null)
+        this.setState({ showSpinner: false })
+      } else if (count > this.lastSeenCount) {
```
swansontec: This is the case where the core is lagging the GUI, but still making valid progress. We might consider updating the fees still in this case, but I agree that just ignoring the outdated transaction is a valid option too.

paullinator:
Haven’t reviewed yet but conflicts need fixing.

src/components/scenes/SendConfirmationScene.js
paullinator: So so agreed!

paullinator:
While not a huge PR, the changes are quite complicated. I'm having trouble wrapping my head around this. I'm now a bit concerned about the scope of risk at this 11th hour. As I'm short on time to more fully review this, I have to defer to @swansontec as our Chief Architect on the final decision on including this, and if so, properly communicating to QA the test vectors necessary to make sure we don't get any regressions.

src/components/scenes/SendConfirmationScene.js
ahujsak: Amen...Can we do it as an actual task rather than piece by piece. Basically get rid of the Wallets/api.js file and see where flow hates us.

PR #1265, opened 2019-03-02
Fix SendConf spending limits PIN issue

kylanhurt:
The purpose of this task is to fix a regression whereby the Spending Limits PIN display was being decided by the **previous** fiat amount value. The solution was to bring that logic out of the connector (accessing older state) and into the scene method where it can access the "current" fiat value of the FlipInput.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a

Asana Task: https://app.asana.com/0/361770107085503/1112060050249480/f


src/components/scenes/SendConfirmationScene.js
```diff
@@ -337,7 +337,12 @@ export class SendConfirmation extends Component<Props, State> {
 
     const guiMakeSpendInfoClone = { ...guiMakeSpendInfo }
     const spendInfo = getSpendInfoWithoutState(guiMakeSpendInfoClone, sceneState, currencyCode)
-
+    let authType
+    if (spendingLimits.transaction.isEnabled) {
+      authType = amountFiat > spendingLimits.transaction.amount ? 'pin' : 'none'
```
paullinator: How did a missing authType pass flow?

PR #1270, opened 2019-03-04
Protect use of textInput ref

paullinator:
* Check if ref to TextInput components is null before using
* Fix inline function props to be static for performance

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/modules/UI/components/FlipInput/FlipInput2.ui.js
+++ b/src/modules/UI/components/FlipInput/FlipInput2.ui.js
@@ -114,8 +114,8 @@ export class FlipInput extends Component<FlipInputOwnProps, State> {
   backInterpolate: Animated.Value
   androidFrontOpacityInterpolate: Animated.Value
   androidBackOpacityInterpolate: Animated.Value
-  textInputFront: TextInput
-  textInputBack: TextInput
+  textInputFront: TextInput | null
+  textInputBack: TextInput | null
 
   constructor (props: Props) {
     super(props)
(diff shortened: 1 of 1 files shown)


src/modules/UI/components/FlipInput/FlipInput2.ui.js
```diff
@@ -280,18 +292,28 @@ export class FlipInput extends Component<FlipInputOwnProps, State> {
...
           returnKeyType="done"
           underlineColorAndroid={'transparent'}
-          ref={ref => {
-            this.textInputFront = ref
-          }}
-          onFocus={() => this.setState({ textInputFrontFocus: true })}
-          onBlur={() => this.setState({ textInputFrontFocus: false })}
+          ref={this.getTextInputFrontRef}
+          onFocus={this.textInputFrontFocusTrue()}
+          onBlur={this.textInputFrontFocusFalse()}
```
kylanhurt: Should these really have the parentheses at the end? Won't they be called everytime the component is renders, and the return value will be what's executed for `onFocus` or `onBlur`? And neither of those methods have a return value at all (they just re-set state).

src/modules/UI/components/FlipInput/FlipInput2.ui.js
```diff
@@ -307,11 +329,9 @@ export class FlipInput extends Component<FlipInputOwnProps, State> {
...
           returnKeyType="done"
           underlineColorAndroid={'transparent'}
-          ref={ref => {
-            this.textInputBack = ref
-          }}
-          onFocus={() => this.setState({ textInputBackFocus: true })}
-          onBlur={() => this.setState({ textInputBackFocus: false })}
+          ref={this.getTextInputBackRef}
+          onFocus={this.textInputBackFocusTrue()}
+          onBlur={this.textInputBackFocusFalse()}
```
kylanhurt: Same here

PR #1278, opened 2019-03-06
Fix createWalletType crash to display error instead

kylanhurt:
The purpose of this task is to prevent the crashes occurring when a user **quickly** taps the "Create Wallet (+)" button immediately after login (before plugins have loaded). Instead of crashing with an error, I have modified the scene to just display a dropdown error alert.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

Asana Task: https://app.asana.com/0/361770107085503/1112350860288128/f

### DIFF
--- a/src/components/scenes/CreateWalletSelectCryptoScene.js
+++ b/src/components/scenes/CreateWalletSelectCryptoScene.js
@@ -151,7 +151,7 @@ export class CreateWalletSelectCrypto extends Component<Props, State> {
     }
     if (unloadedWalletCount) {
       if (!errorShown) {
-        displayErrorAlert('Error missing wallet type. Unloaded plugins.')
+        displayErrorAlert(s.strings.create_wallet_selcet_crypto_unloaded_plugins_error)
         errorShown = true
       }
     }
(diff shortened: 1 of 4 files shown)


src/components/scenes/CreateWalletSelectCryptoScene.js
```diff
@@ -127,21 +134,29 @@ export class CreateWalletSelectCrypto extends Component<Props, State> {
...
         sortedWalletTypes.push(walletTypesCopy[idx])
         walletTypesCopy.splice(idx, 1)
       } else {
-        throw new Error('Error missing wallet type. Ensure all plugins are loaded')
+        unloadedWalletCount++
+      }
+    }
+    if (unloadedWalletCount) {
+      if (!errorShown) {
+        displayErrorAlert('Error missing wallet type. Unloaded plugins.')
```
paullinator: Error message should be localized in en_US.js

Also recommend a more user friendly message.
"Some wallet types unavailable. Please wait for account to fully load to access all wallet types."

kylanhurt: I'll try it and see how much of that message can fit in the small dropdown alert.

kylanhurt:
screen shot 2019-03-06 at 11 03 52 am

PR #1285, opened 2019-03-11
When add token is clicked and asked if wanted to add a new eth wallet…

amanalili:
…, auto select eth wallet and go to select fiat screen

Task: Add Token Button - If no ETH Wallet exists, the Create Wallet option for ETH should have ETH automatically selected

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android

### DIFF
--- a/src/components/scenes/WalletListScene.js
+++ b/src/components/scenes/WalletListScene.js
@@ -2,7 +2,7 @@
 
 import { createYesNoModal, showModal } from 'edge-components'
 import React, { Component } from 'react'
-import { ActivityIndicator, Animated, FlatList, Image, TouchableOpacity, TouchableWithoutFeedback, View } from 'react-native'
+import { ActivityIndicator, Alert, Animated, FlatList, Image, TouchableOpacity, TouchableWithoutFeedback, View } from 'react-native'
 import { Actions } from 'react-native-router-flux'
 import slowlog from 'react-native-slowlog'
 import SortableListView from 'react-native-sortable-listview'
@@ -26,6 +26,7 @@ import { buyMultipleCryptoStyle } from '../../styles/components/BuyCryptoStyle.j
 import { TwoButtonModalStyle } from '../../styles/indexStyles.js'
 import styles from '../../styles/scenes/WalletListStyle'
 import THEME from '../../theme/variables/airbitz'
+import type { GuiWalletType } from '../../types'
 import { getFiatSymbol, getObjectDiff, getTotalFiatAmountFromExchangeRates } from '../../util/utils'
 import FullWalletListRow from '../common/FullWalletListRow.js'
 import SortableWalletListRow from '../common/SortableWalletListRow.js'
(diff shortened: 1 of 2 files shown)


src/connectors/scenes/WalletListConnector.js
```diff
@@ -28,6 +28,10 @@ const mapStateToProps = (state: State) => {
   const isWalletFiatBalanceVisible = state.ui.settings.isWalletFiatBalanceVisible
   const defaultFiat = SETTINGS_SELECTORS.getDefaultFiat(state)
 
+  const supportedWalletTypes = SETTINGS_SELECTORS.getSupportedWalletTypes(state)
+  const supportedWalletTypesArray: Array<any> = Object.values(supportedWalletTypes)
```
paullinator: This seems redundant. `supportedWalletTypes` is already an array of walletType objects. What's the `Object.values()` call for?

src/components/scenes/WalletListScene.js
```diff
@@ -465,7 +467,12 @@ export default class WalletList extends Component<Props, State> {
         noButtonText: s.strings.string_cancel_cap,
         yesButtonText: s.strings.title_create_wallet
       })
-      return (await showModal(modal)) ? Actions[Constants.CREATE_WALLET_SELECT_CRYPTO]() : null
+
+      if (this.props.ethereumWalletType) {
+        return (await showModal(modal)) ? Actions[Constants.CREATE_WALLET_SELECT_FIAT]({ selectedWalletType: this.props.ethereumWalletType }) : null
+      } else {
+        return Alert.alert(s.strings.create_wallet_invalid_input, s.strings.create_wallet_select_valid_crypto)
```
paullinator: Why would you ever get into this `else` case?

src/connectors/scenes/WalletListConnector.js
amanalili: I "typeof" it before and it shows it is an object with keyed numbers somehow. I was confused also at first

src/components/scenes/WalletListScene.js
amanalili: I checked the code on the create wallet screen and somehow there is a scenario that it can be not supported?

I am really not sure so just to be safe I just copied the code. If the ethereum wallet is always supported then no problem ill remove even the null check.

src/connectors/scenes/WalletListConnector.js
paullinator: typeof for an Array is an object with keyed numbers. Please remove Object.values()

src/components/scenes/WalletListScene.js
paullinator: Ok. let's leave it for now.

src/connectors/scenes/WalletListConnector.js
amanalili: fixed

PR #1288, opened 2019-03-12
Fix invalid synced setting denomination during account initialization

kylanhurt:
The purpose of this task is to repair any invalid currency denominations in the user's settings area.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a

Asana Task: https://app.asana.com/0/361770107085503/1113481040564768/f

### DIFF
--- a/src/constants/WalletAndCurrencyConstants.js
+++ b/src/constants/WalletAndCurrencyConstants.js
@@ -53,6 +53,7 @@ export const CURRENCY_PLUGIN_NAMES = {
   SMART: 'smartcash',
   UFO: 'ufo',
   VTC: 'vertcoin',
+  XLM: 'stellar',
   XZC: 'zcoin'
 }
 
--- a/src/constants/WalletAndCurrencyConstants.js
+++ b/src/constants/WalletAndCurrencyConstants.js
@@ -34,6 +34,7 @@ export const DEFAULT_STARTER_WALLET_NAMES = {
   XLM: s.strings.string_first_stellar_wallet_name
 }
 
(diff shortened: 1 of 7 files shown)


src/constants/WalletAndCurrencyConstants.js
```diff
@@ -53,6 +53,7 @@ export const CURRENCY_PLUGIN_NAMES = {
   SMART: 'smartcash',
   UFO: 'ufo',
   VTC: 'vertcoin',
+  XLM: 'stellar',
   XZC: 'zcoin'
```
kylanhurt: Since the currencyName (based on `currencyCode`) is a property whose subproperties are accessed in the `mergeSettings` function, that means missing currencies in this object will cause an error (which is why I'm filling in Stellar).

src/modules/Login/action.js
```diff
@@ -193,8 +193,9 @@ export const initializeAccount = (account: EdgeAccount, touchIdInfo: Object) =>
 export const mergeSettings = (
   loadedSettings: Object,
   defaults: Object,
-  types: Object
-): { finalSettings: Object, isOverwriteNeeded: boolean, isDefaultTypeIncorrect: boolean } => {
+  types: Object,
+  account?: Object
+): { finalSettings: Object, isOverwriteNeeded: boolean, isDefaultTypeIncorrect: boolean, account?: Object } => {
```
paullinator: It doesn't look like you ever return account but it's in the flow type

kylanhurt: Ah thank you for the heads up, I was getting the param types mixed up with the return types.

PR #1298, opened 2019-03-19
Added a new SafeAreaView dependency to accomodate new iPhone devices

amanalili:
Task: Fix iPhone XS Max safe area

#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,6 +134,7 @@
     "react-native-qrcode": "^0.2.3",
     "react-native-randombytes": "^3.5.0",
     "react-native-router-flux": "4.0.0-beta.31",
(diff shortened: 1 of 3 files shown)


kylanhurt:
@amanalili Please try using the `<SafeAreaView>` component in `import SafeAreaView from 'src/modules/UI/components/SafeAreaView' rather than using a new / additional package

paullinator:
@kylanhurt art’s changes are to replace the RN SafeAreaView with one that works w/iphoneX max. He’s not replacing our internal version

PR #1302, opened 2019-03-21
Don't show Wallet Empty modal when user taps Sweep Private Key from side menu

amanalili:
Task: Title

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android

### DIFF
--- a/src/components/Main.ui.js
+++ b/src/components/Main.ui.js
@@ -167,7 +167,7 @@ type Props = {
   openDrawer: () => void,
   dispatchAddressDeepLinkReceived: (addressDeepLinkData: Object) => any,
   deepLinkPending: boolean,
-  checkAndShowGetCryptoModal: () => void
+  checkAndShowGetCryptoModal: (?string) => void
 }
 
 async function queryUtilServer (context: EdgeContext, folder: DiskletFolder, usernames: Array<string>) {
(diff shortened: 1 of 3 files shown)


src/components/Main.ui.js
```diff
@@ -903,4 +903,10 @@ export default class Main extends Component<Props> {
     Actions.pop()
     return true
   }
+  checkAndShowGetCryptoModal = (props: any) => {
```
paullinator: This function seems unnecessary. Why not just add the sweepPrivateKey check to the original this.props.checkAndShow.. function? Then pass props.data to it

src/modules/UI/components/ControlPanel/Component/Main.js
```diff
@@ -164,7 +164,7 @@ const ScanButton = () => {
 
 const SweepPrivateKeyButton = () => {
   return (
-    <Button onPress={Actions.scan}>
+    <Button onPress={() => Actions.scan('sweepPrivateKey')}>
```
paullinator: Do not pass inline functions as a prop. This will force a rerender. Instead create another function in the class and pass that in

paullinator:
conflicts and xcode project should not be changed.

PR #1310, opened 2019-03-28
new API for deep linking

ahujsak:
This is the new API for deep links with 'plugin' as the path name. It is currently simplex only. Simplex plugin with this deep link has not been integrated in yet.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/modules/DeepLinkingManager.js
+++ b/src/modules/DeepLinkingManager.js
@@ -5,10 +5,12 @@ import { Alert } from 'react-native'
 import { Actions } from 'react-native-router-flux'
 import { connect } from 'react-redux'
 import { sprintf } from 'sprintf-js'
+import parse from 'url-parse'
 
 import { selectWallet } from '../actions/WalletActions.js'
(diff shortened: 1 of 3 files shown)


src/modules/DeepLinkingManager.js
```diff
@@ -33,11 +35,31 @@ class DeepLinkingManager extends React.Component<Props> {
   componentDidUpdate () {
     if (Object.keys(this.props.wallets).length > 0 && this.props.deepLinkPending) this.checkForWallet()
   }
+  processPluginDeepLink = (parsedUrl: Object) => {
+    if (parsedUrl.pathname.includes('simplex')) {
```
swansontec: I would like this to be more plugin-agnostic, but this works for now.

swansontec: Where do we take the `pathname` and inject it into the plugin? How do they know what the URI parameters are?

PR #1319, opened 2019-04-03
CocoaPods

swansontec:
This switches our iOS build infrastructure over to use CocoaPods. This greatly simplifies the upgrade to React Native 0.59, and also un-breaks the `react-native link` command going forward.


swansontec:
Updated podfile to satisfy Travis. The Android failures seem unrelated to these changes.

PR #1324, opened 2019-04-06
Tx with blockHeight greater than current wallet block height to show as pending (not negative)

kylanhurt:
The purpose of this task is to handle the case where the wallet block height is lagging behind the transaction block height and get the TransactionList to show as 'Pending' rather than a negative confirmation number.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

Asana Task: https://app.asana.com/0/361770107085503/1116927447707180/f

### DIFF
--- a/src/components/common/TransactionRow.js
+++ b/src/components/common/TransactionRow.js
@@ -138,8 +138,8 @@ export class TransactionRowComponent extends Component<Props, State> {
     } else if (tx.blockHeight < 0) {
       pendingTimeSyntax = s.strings.fragment_transaction_list_tx_dropped
       pendingTimeStyle = styles.transactionPartialConfirmation
-    } else if (currentConfirmations === 0) {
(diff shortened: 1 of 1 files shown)


src/components/common/TransactionRow.js
```diff
@@ -138,8 +138,11 @@ export class TransactionRowComponent extends Component<Props, State> {
     } else if (tx.blockHeight < 0) {
       pendingTimeSyntax = s.strings.fragment_transaction_list_tx_dropped
       pendingTimeStyle = styles.transactionPartialConfirmation
-    } else if (currentConfirmations === 0) {
-      // if completely unconfirmed or wallet uninitialized
+    } else if (currentConfirmations <= 0) {
+      if (currentConfirmations < 0) {
```
swansontec: Debugging statement?

PR #1338, opened 2019-04-24
Do not produce invalid versions

swansontec:
Most Jenkins builds were failing, due to slashes getting into the version numbers.


swansontec:
Over-the-shoulder with Paul.

PR #1349, opened 2019-05-02
CoinZark Swap Plugin

gertjaap:
This pull request includes the changes for the integration of [CoinZark](https://coinzark.com/) as an Exchange provider in the Edge wallet.  See also [The related PR in `edge-exchange-plugins`](https://github.com/EdgeApp/edge-exchange-plugins/pull/15)

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [X] n/a


env.example.json
```diff
@@ -24,6 +24,10 @@
     "affiliateId": "xxxx",
     "affiliateMargin": 0.5
   },
+  "x_COINZARK_INIT": {
+    "apiKey": "xxxx",
+    "affiliateMargin": 0.5
```
swansontec: The plugin itself seems to be using `affiliateId` and `affiliateFee` here. Is this right?

gertjaap: Good catch. This was indeed incorrect in the sample. Changed that.

swansontec:
We are closing this out, since it has falling behind.

PR #1387, opened 2019-05-23
Revert back the old style settings. P0 task

amanalili:
Task: Reduce row height

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android


kylanhurt:
@amanalili Is this supposed to revert the previous change exactly? Because I'm seeing `justifyContent` and `fontWeight` being properties / styles that were changed (and should be changed back to be consistent?).

amanalili:
I just decide to fix the errors on Exchange Settings Screen

swansontec:
@amanalili If this branch isn't needed, should we delete it?

PR #1392, opened 2019-05-27
two different user-agents

ahujsak:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


swansontec:
I made a demo page to show the user agent, then tried it on a few phones. The returned values pretty much match this PR aside from a few version numbers, so this makes sense.

PR #1435, opened 2019-06-23
William/countries

swansontec:
These are not the best solutions to the bugs, but they are the least-invasive way to get the job done. The keyboard corruption issue works for iOS, but not for Android. Fixing this will require a deeper refactor.

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


swansontec:
Replaced by #1443

PR #1465, opened 2019-07-13
Enforce Totle same wallet as destination of swap

kylanhurt:
Related to: https://app.asana.com/0/361770107085503/1130331036367390/f

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/actions/CryptoExchangeActions.js
+++ b/src/actions/CryptoExchangeActions.js
@@ -281,6 +281,13 @@ const getShiftTransaction = (fromWallet: GuiWallet, toWallet: GuiWallet, info: S
     throw quoteError
   }
 
+  // totle and other DEX's may not allow swapping to a different wallet (received coins must come to same wallet)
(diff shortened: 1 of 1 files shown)


src/actions/CryptoExchangeActions.js
```diff
@@ -281,6 +281,13 @@ const getShiftTransaction = (fromWallet: GuiWallet, toWallet: GuiWallet, info: S
     throw quoteError
   }
 
+  // totle and other DEX's may not allow swapping to a different wallet (received coins must come to same wallet)
+  if (edgeCoinExchangeQuote.pluginName === 'totle') {
```
swansontec: Plugins that have this requirement should declare that fact in their `PluginInfo` in the core, and then the GUI should limit the wallet picker accordingly. However, I realize we don't have time for that this late in the release cycle, so this will do the trick for now.

PR #1466, opened 2019-07-13
fix backbutton to callback for Edge Provider

ahujsak:
This pr solves the issue where a user clicks to make a spend from a partner plugin, and then they hit the back button. Edge provider did not resolve. Now it resolves a transaction to null if there is no transaction so the partner can handle null
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/scenes/SendConfirmationScene.js
+++ b/src/components/scenes/SendConfirmationScene.js
@@ -164,10 +164,7 @@ export class SendConfirmation extends Component<Props, State> {
   componentWillUnmount () {
     this.props.reset()
     if (this.props.guiMakeSpendInfo && this.props.guiMakeSpendInfo.onBack) {
-      const cb = this.props.guiMakeSpendInfo.onBack ? this.props.guiMakeSpendInfo.onBack : ''
-      if (cb) {
-        cb()
-      }
+      this.props.guiMakeSpendInfo.onBack()
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/components/scenes/SendConfirmationScene.js
```diff
@@ -164,7 +164,10 @@ export class SendConfirmation extends Component<Props, State> {
   componentWillUnmount () {
     this.props.reset()
     if (this.props.guiMakeSpendInfo && this.props.guiMakeSpendInfo.onBack) {
-      this.props.guiMakeSpendInfo.onBack()
+      const cb = this.props.guiMakeSpendInfo.onBack ? this.props.guiMakeSpendInfo.onBack : ''
```
swansontec: Over-complicated code. Just do: `const cb = this.props.guiMakeSpendInfo.onBack`, and the false case will be `undefined` automatically.

Also, should we add `onBack: Function` to the `GuiMakeSpendInfo` type?

swansontec: Actually, this whole addition isn't needed - the line above already checks for this case. Also, just checked, and it looks like `onBack` is already in the types, so we are good.

ahujsak: Flow did not like it because onBack is optional.

PR #1481, opened 2019-07-18
Hide zero amount transactions (proxy allowance) from Transaction List 

kylanhurt:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/scenes/TransactionListScene.js
+++ b/src/components/scenes/TransactionListScene.js
@@ -145,7 +145,10 @@ export class TransactionList extends Component<Props, State> {
   }
 
   render () {
-    const txs = this.state.reset ? emptyArray : this.props.transactions
+    const { transactions } = this.props
+    const { reset } = this.state
+    const txs = reset ? emptyArray : transactions
+
     return (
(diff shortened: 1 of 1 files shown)


paullinator:
This should be implemented in the account based repo. That way a new transaction notification does not get sent it to the GUI causing a notification for the blank transaction

kylanhurt:
@paullinator Okay, I will have to check to make sure that hiding the transactions does not mess with the pagination that the Transaction List uses to request transactions from the core / plugins, though.

src/components/scenes/TransactionListScene.js
```diff
@@ -145,7 +145,10 @@ export class TransactionList extends Component<Props, State> {
   }
 
   render () {
-    const txs = this.state.reset ? emptyArray : this.props.transactions
+    const { transactions } = this.props
+    const { reset } = this.state
```
swansontec: Could combine lines here, `    const { reset, transactions } = this.props`

PR #1492, opened 2019-07-23
Adjust the SceneWrapper on Android

swansontec:
The keyboard animation isn't perfect on Android, since the `Dimensions` module is buggy. If we don't need the keyboard animation, skip it entirely.

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


swansontec:
Verbal approval from @paullinator.

PR #1503, opened 2019-07-26
William/last minute fixes

swansontec:
#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


swansontec:
Interesting... Travis doesn't run Flow, so it doesn't catch the problem that exists on develop.

PR #1504, opened 2019-07-26
Fix mining fee scene layout

swansontec:
#### PR Requirements

This will fail Travis because develop has Flow errors.

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


swansontec:
Random build failure on Android only: "Could not resolve com.googlecode.json-simple:json-simple:1.1." Restarting the build might solve it?

PR #1530, opened 2019-08-08
Fix pluginList text overflow

kylanhurt:
#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/styles/scenes/PluginsStyle.js
+++ b/src/styles/scenes/PluginsStyle.js
@@ -36,42 +36,33 @@ export const styles = {
   pluginRow: {
     borderBottomWidth: 1,
     borderBottomColor: THEME.COLORS.GRAY_3,
-    padding: scale(14),
-    paddingRight: scale(30),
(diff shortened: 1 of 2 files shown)


src/styles/scenes/PluginsStyle.js
```diff
@@ -60,6 +61,9 @@ export const styles = {
   },
   textBoxWrap: {
     /* justifyContent: 'center' */
+    height: scale(18 + 4 + 4 * 18),
+    // width minus left logo box minus padding
+    width: PLATFORM.deviceWidth - scale(50 + 16) - scale(40)
```
swansontec: Hard-coded dimensions based on device dimensions are a huge red flag! I will find another way to fix this!

swansontec:
I had @peachbits work with me to find the best visuals for this feature. I think we have something that looks good, but is also immune to device size changes and text overflow issues (in case of loooong translations.)

PR #1559, opened 2019-08-21
exit from plugin view screens.

ahujsak:
This pr adds an exit button to the right of the plugin header for exiting out of a plugin at any time. It also expands on the edgeProvider API to exit programmatically.
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/modules/UI/components/Header/Component/ExitButton.js
+++ b/src/modules/UI/components/Header/Component/ExitButton.js
@@ -8,8 +8,6 @@ import s from '../../../../../locales/strings.js'
 import T from '../../../components/FormattedText'
 import styles from '../style'
 
-const EXIT_TEXT = s.strings.string_exit
-
 type Props = {}
 
 class ExitButton extends Component<Props> {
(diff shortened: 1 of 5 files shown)


src/modules/UI/components/Header/Component/ExitButton.js
```diff
@@ -0,0 +1,27 @@
...
+
+import React, { Component } from 'react'
+import { TouchableOpacity } from 'react-native'
+import { Actions } from 'react-native-router-flux'
+
+import s from '../../../../../locales/strings.js'
+import T from '../../../components/FormattedText'
+import styles from '../style'
+
+const EXIT_TEXT = s.strings.string_exit
```
swansontec: I see this all over. Why do we do this?

src/modules/UI/components/Header/Component/ExitButton.js
```diff
@@ -0,0 +1,27 @@
...
+type Props = {}
+
+class ExitButton extends Component<Props> {
+  popThis = () => {
+    Actions.pop()
+  }
+  render () {
+    return (
+      <TouchableOpacity style={styles.sideTextWrap} onPress={this.popThis}>
+        <T style={[styles.sideText]}>{EXIT_TEXT}</T>
```
swansontec: We don't need the square brackets here.

PR #1580, opened 2019-08-29
Audit every `catch` statement in the codebase

swansontec:
#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


src/actions/OtpActions.js
```diff
@@ -7,30 +7,20 @@ import type { Dispatch, GetState } from '../types/reduxTypes.js'
 export const enableOtp = () => async (dispatch: Dispatch, getState: GetState) => {
   const state = getState()
   const account = CORE_SELECTORS.getAccount(state)
-  try {
```
swansontec: We don't need try/catch here because the `dispatch` includes that by default for all actions

PR #1592, opened 2019-09-05
Add Request Review popup

sfncook:
#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android


src/actions/RequestReviewActions.js
```diff
@@ -0,0 +1,66 @@
...
+    const title = sprintf(s.strings.request_review_question_title, s.strings.app_name_short)
+    await Airship.show(bridge => (
+      <TwoButtonSimpleConfirmationModal
+        bridge={bridge}
+        title={title}
+        subTitle={s.strings.request_review_question_subtitle}
+        cancelText={s.strings.request_review_answer_no}
+        doneText={s.strings.request_review_answer_yes}
+        onCancel={() => {}}
+        onDone={() => Linking.openURL(s.strings.request_review_android_page_link)}
```
swansontec: This modal should not have `onCancel` or `onDone` callbacks. Instead, it should use the promise return value to indicate which button the user pressed.

src/actions/RequestReviewActions.js
```diff
@@ -0,0 +1,66 @@
...
+    swapCountData = JSON.parse(swapCountDataStr)
+  } catch (e) {
+    // File needs init
+    swapCountData = {
+      swap_count: 0,
+      has_review_been_requested: false
+    }
+  }
+
+  const hasReviewBeenRequested = swapCountData.has_review_been_requested
```
swansontec: We tend to use camelCase within our files.

src/actions/RequestReviewActions.js
```diff
@@ -0,0 +1,66 @@
...
+    swapCount++
+    swapCountData.swap_count = swapCount
+    if (swapCount >= MANY_SWAPS_TO_TRIGGER_REQUEST) {
+      await requestReview()
+      swapCountData.has_review_been_requested = true
+    }
+    const swapCountDataStr = JSON.stringify(swapCountData)
+    await account.disklet
+      .setText(SWAP_COUNT_DATA_FILE, swapCountDataStr)
+      .catch(e => console.error(`RequestReviewActions.updateSwapCount: Error writing file ${SWAP_COUNT_DATA_FILE}:`, e))
```
swansontec: The only correct way to handle errors is `showError`, from `AirshipInstance.js`. Otherwise, if you don't want to bother the user with this, please put a comment saying that failure is OK, in which case it's `console.log`, not `console.error` for the logging.

src/actions/RequestReviewActions.js
```diff
@@ -0,0 +1,66 @@
+// @flow
+
+import React from 'react'
+import { Linking, Platform } from 'react-native'
+import * as StoreReview from 'react-native-store-review'
+import { sprintf } from 'sprintf-js'
+
+import { TwoButtonSimpleConfirmationModal } from '../components/modals/TwoButtonSimpleConfirmationModal'
```
swansontec: Missing `.js` extensions on these files.

src/styles/common/textStyles.js
```diff
@@ -34,6 +35,14 @@ const rawStyles = {
     fontFamily: THEME.FONTS.DEFAULT,
     fontSize: THEME.rem(1),
     textAlign: 'center'
+  },
+
+  linkText: {
```
swansontec: These styles are meant to be more primitive, like colors and fonts vs. layout like  width & alignment. Please remove width & alignment from this style.

src/components/modals/TwoButtonSimpleConfirmationModal.js
```diff
@@ -0,0 +1,53 @@
...
+export class TwoButtonSimpleConfirmationModal extends Component<Props> {
+  onCancel () {
+    const { bridge, onCancel } = this.props
+    bridge.resolve()
+    onCancel()
+  }
+
+  onDone () {
+    const { bridge, onDone } = this.props
+    bridge.resolve()
```
swansontec: This should resolve with a usable value, like `true`,

src/components/modals/TwoButtonSimpleConfirmationModal.js
```diff
@@ -0,0 +1,53 @@
...
+    onDone()
+  }
+
+  render () {
+    const { bridge, cancelText, doneText, subTitle, title } = this.props
+    return (
+      <AirshipModal bridge={bridge} onCancel={() => bridge.resolve()}>
```
swansontec: This should resolve with `false`.

src/components/modals/TwoButtonSimpleConfirmationModal.js
```diff
@@ -0,0 +1,53 @@
...
+
+import { PrimaryButton } from 'edge-components'
+import React, { Component } from 'react'
+import { Text, TouchableOpacity } from 'react-native'
+import EntypoIcon from 'react-native-vector-icons/Entypo'
+
+import { type AirshipBridge, AirshipModal, ContentArea, IconCircle, THEME, textStyles } from './modalParts.js'
+
+type Props = {
+  bridge: AirshipBridge<void>,
```
swansontec: The bridge should return `boolean` instead of `void`, making it useful.

src/components/modals/TwoButtonSimpleConfirmationModal.js
```diff
@@ -0,0 +1,53 @@
...
+  title: string,
+  subTitle: string,
+  cancelText: string,
+  doneText: string,
+  onCancel(): void,
```
swansontec: Remove `onCancel` and `onDone`.

src/styles/common/textStyles.js
```diff
@@ -34,6 +35,14 @@ const rawStyles = {
...
+    width: '100%',
+    fontSize: scale(18),
```
swansontec: We are trying to get rid of `scale`. The new replacement is `THEME.rem(x)`. The idea is that all layout should be relative to the height of a line of text, which is 1 rem. So, the `fontSize` here should be 1 rem, or `THEME.rem(1)`.

src/styles/common/textStyles.js
swansontec: More to the point, they are supposed to be composable, so someone could do `style={[textStyles.bodyCenter, textStyles.linkText]}` to get the effect you are going for. So really, `linkText` should just be the color override.

src/actions/RequestReviewActions.js
sfncook: @swansontec Can you please provide an example in the edge-react-ui code of what you're talking about?  I copied this pattern from the other modals in that same directory.  I don't see any other modals that return boolean-resolved-promises.  I'd like to be imitating your preexisting style.  All the other modals appear to be using callbacks.

PR #1598, opened 2019-09-07
William/filter tokens

swansontec:
#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/modules/Login/action.js
+++ b/src/modules/Login/action.js
@@ -12,6 +12,7 @@ import { showError } from '../../components/services/AirshipInstance.js'
 import * as Constants from '../../constants/indexConstants'
 import s from '../../locales/strings.js'
 import type { Dispatch, GetState } from '../../types/reduxTypes.js'
+import { type CustomTokenInfo } from '../../types/types.js'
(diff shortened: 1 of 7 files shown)


swansontec:
Here is the new error message, plus the fact that existing broken tokens will no longer load:

Simulator Screen Shot - iPhone X - 2019-09-06 at 17 26 58

src/modules/Login/action.js
```diff
@@ -289,6 +290,18 @@ export const mergeSettings = (
     }
   }
 
+  // Filter conflicting tokens out of synced settings:
+  if (finalSettings.customTokens && account != null) {
```
kylanhurt: Even if we change the denomination info for a currency here, back in the calling `initializeAccount` function one line 183 it can undo the work we've done in this `mergeSettings` routine

swansontec: If you follow the data-flow, `initializeAccount` line 183 is using the data that comes out of `mergeSettings`. Indeed, `mergeSettings` is the very first thing that happens to the settings after they come off of disk, and is therefore the most definitive place to do an edit like this.

PR #1599, opened 2019-09-07
Prevent customTokens being saved to redux settings, and overwrite wit…

kylanhurt:
…h correct multiplier data

Remove possibilities of adding or enabling disallowed token codes

Refactor error message for duplicate token

Fix cocopod file

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


src/actions/AddTokenActions.js
```diff
@@ -48,6 +49,12 @@ export const addTokenAsync = async (
 ) => {
   // create modified object structure to match metaTokens
   const newTokenObj: CustomTokenInfo = WALLET_ACTIONS.assembleCustomToken(currencyName, currencyCode, contractAddress, denomination)
+  const plugins: Object = state.ui.settings.plugins
+  const allCurrencyInfos = plugins.allCurrencyInfos
+  const allSupportedParentCurrencies = allCurrencyInfos.map(info => info.currencyCode)
+  if (allSupportedParentCurrencies.find(item => item === currencyCode)) {
+    throw new Error(s.strings.error_token_exists)
+  }
```
swansontec: Right, this is one place custom tokens can get into the system - from the user at runtime.

src/modules/Login/action.js
```diff
@@ -179,9 +181,13 @@ export const initializeAccount = (account: EdgeAccount, touchIdInfo: Object) =>
 
     if (accountInitObject.customTokens) {
       accountInitObject.customTokens.forEach(token => {
-        accountInitObject.customTokensSettings.push(token)
-        // this second dispatch will be redundant if we set 'denomination' property upon customToken creation
-        accountInitObject.denominationKeys.push({ currencyCode: token.currencyCode, denominationKey: token.multiplier })
+        if (allSupportedParentCurrencies.find(item => item === token.currencyCode)) {
```
swansontec: This is the other place custom tokens can get into the system - from the settings. If both of these entry points are shut down, then there will not be any conflicting tokens, and the other locations do not need filters.

src/modules/Core/Wallets/action.js
```diff
@@ -29,6 +32,9 @@ export const updateWalletsRequest = () => (dispatch: Dispatch, getState: GetStat
           const customTokens = state.ui.settings.customTokens
           const enabledNotHiddenTokens = enabledTokens.filter(token => {
             let isVisible = true // assume we will enable token
+            if (allSupportedParentCurrencies.includes(token)) {
+              return false
+            }
             const tokenIndex = _.findIndex(customTokens, item => item.currencyCode === token)
```
swansontec: There is no point in doing this filter here, since the next line, which searches for a custom token matching the currency code, will already eliminate these. We already prevent conflicting custom tokens from entering the system in the first place.

src/actions/WalletActions.js
```diff
@@ -245,10 +245,18 @@ export const getEnabledTokens = (walletId: string) => async (dispatch: Dispatch,
...
-    if (tokensToEnable.length) {
+    const plugins: Object = state.ui.settings.plugins
+    const allCurrencyInfos = plugins.allCurrencyInfos
+    // make sure that the token doesn't have same currencyCode
+    // as a parent plugin
+    const allSupportedParentCurrencies = allCurrencyInfos.map(info => info.currencyCode)
+    const filteredTokensToEnable = tokensToEnable.filter(token => {
+      return !allSupportedParentCurrencies.includes(token)
+    })
+    if (filteredTokensToEnable.length) {
```
swansontec: This block of code should have been moved into the loop above, assuming it was needed in the first place. It isn't needed, though, since `tokensToEnable` only includes currency codes that exist, and we have prevented tokens with conflicting currency codes from even existing to begin with. Therefore, the loop above already handles this case without doing anything extra.

src/components/scenes/ManageTokensScene.js
```diff
@@ -74,11 +75,15 @@ export default class ManageTokens extends Component<ManageTokensProps, State> {
...
     const accountMetaTokenInfo = [...this.props.settingsCustomTokens]
     const combinedTokenInfo = UTILS.mergeTokensRemoveInvisible(metaTokens, accountMetaTokenInfo)
-
-    const sortedTokenInfo = combinedTokenInfo.sort((a, b) => {
+    const allCurrencyInfos = plugins.allCurrencyInfos
+    const allSupportedParentCurrencies = allCurrencyInfos.map(info => info.currencyCode)
+    const combinedFilteredTokenInfo = combinedTokenInfo.filter(token => {
+      return !allSupportedParentCurrencies.includes(token.currencyCode)
+    })
+    const sortedTokenInfo = combinedFilteredTokenInfo.sort((a, b) => {
```
swansontec: This isn't needed, since we prevent tokens with conflicting currency codes from even entering the system, so we don't need to filter them out.

src/modules/Login/action.js
```diff
@@ -279,6 +285,10 @@ export const mergeSettings = (
         // for each currency info (each native currency)
         const pluginDenominations = account.currencyConfig[currencyName].currencyInfo.denominations // get denominations for that plugin
         const settingDenominationIndex = pluginDenominations.findIndex(pluginDenom => pluginDenom.multiplier === loadedSettings[key].denomination) // find settings denom in plugin denoms
+        if (loadedSettings[key].multiplier !== pluginDenominations[0].multiplier) finalSettings[key].multiplier = pluginDenominations[0].multiplier
```
swansontec: This logic looks like it would reset all user settings, so if I set my BTC wallet to "bits", it would always go back to BTC. Actually, the surrounding code already does the right thing - it checks to be sure the user setting is one of the valid options, and only resets it if not.

swansontec:
Merged my version instead.

PR #1613, opened 2019-09-13
Use the Android in-app review feature for iOS

swansontec:
#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -27,7 +27,6 @@ target 'edge' do
   pod 'Firebase/Core', '~> 5.3.0'
 
   # React Native linked dependencies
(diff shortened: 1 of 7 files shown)


paullinator:
What’s the reason for this? There iOS  review library is far preferred

swansontec:
The link in the iOS library is not clickable in any of our testing. Maybe the app needs to be published before it becomes clickable, or maybe it never becomes clickable. RJ wanted this backup solution in case it is truly non-functional.

paullinator:
I would presume that it is not clickable since the version of the app running is not available on the App Store to review. Let’s put back the actual iOS library and see how it goes once it goes live in the App Store

PR #1533, opened 2019-08-09
RIF Token integration

alepc253:
-RIF integration 
-Add token action considers both Eth and RSK 
-Add custom token action considers wich wallet type created

To be considered after PR #1483 

#### PR Requirements

If you have made **any** visual changes to the GUI. Make sure you have:
- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

PR #1622, opened 2019-09-20
Official RIF integration

kylanhurt:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


src/locales/strings/de.json
```diff
@@ -311,6 +311,8 @@
   "title_change_pin": "PIN ändern",
```
kylanhurt: You can ignore these translated files, they should just get overwritten by the Crowdin translations

kylanhurt:
Supersedes: https://github.com/EdgeApp/edge-react-gui/pull/1533

PR #1624, opened 2019-09-20
Allen/bity

ahujsak:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/copy-bity.sh
+++ b/copy-bity.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+# Usage: copy-core [path-to-core-ui]
+set -e
+src=${1:-../edge-plugin-bity}
+
+dest=node_modules/edge-plugin-bity
+mkdir -p $dest
+
+cp    $src/package.json $dest/package.json
+cp    $src/manifest.json $dest/manifest.json
+cp -r $src/src/ $dest/src/
+cp -r $src/target/ $dest/target/
+cp -r $src/public/ $dest/public/
--- a/copy-bity.sh
+++ b/copy-bity.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-# Usage: copy-core [path-to-core-ui]
+# Usage: copy-edge-plugin-bity [path-to-bity plugin]
 set -e
 src=${1:-../edge-plugin-bity}
 
--- a/src/constants/plugins/buySellPlugins.js
+++ b/src/constants/plugins/buySellPlugins.js
@@ -66,7 +66,7 @@ export const allPlugins: Array<BuySellPlugin> = [
     permissions: ['camera']
   },
   {
-    pluginId: 'co.edgesecure.bity',
+    pluginId: 'com.bity',
     uri: hostedUri + 'co.edgesecure.bity/index.html',
     name: 'Bity',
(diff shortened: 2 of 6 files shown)


src/constants/plugins/buySellPlugins.js
```diff
@@ -65,6 +65,14 @@ export const allPlugins: Array<BuySellPlugin> = [
     supportEmail: 'support@sendwyre.com',
     permissions: ['camera']
   },
+  {
+    pluginId: 'co.edgesecure.bity',
+    uri: hostedUri + 'co.edgesecure.bity/index.html',
+    name: 'Bity',
+    subtitle: 'ACH bank transfer\nBTC\nFee: 1% / Settlement: 1-5 days\nBuy Limit: $2500 / Unlimited Sell',
+    imageUrl: 'https://edge.app/wp-content/uploads/2019/09/logoBity.png',
+    supportEmail: 'support@sendwyre.com'
```
swansontec: Wrong email address.

src/constants/plugins/buySellPlugins.js
```diff
@@ -65,6 +65,14 @@ export const allPlugins: Array<BuySellPlugin> = [
     supportEmail: 'support@sendwyre.com',
     permissions: ['camera']
   },
+  {
+    pluginId: 'co.edgesecure.bity',
```
swansontec: This pluginId sucks.

src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -339,19 +339,17 @@ export class EdgeProvider extends Bridgeable {
...
-    // this is about bit id signatures.
-    const obj = {
-      publicKey: 'string',
-      // Hex encoded signature
-      signedMessage: 'string'
+    try {
+      const signedMessage = await coreWallet.otherMethods.signMessageBase64(message, guiWallet.receiveAddress.publicAddress)
+      return signedMessage
+    } catch (e) {
+      console.log('EP: E', e)
```
swansontec: This should be:

```js
catch (e) {
  showError(e)
  throw e // <- This way Bity knows it failed? Unless they know some other way?
}
```

copy-bity.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+# Usage: copy-core [path-to-core-ui]
```
swansontec: Wrong comment?

src/modules/UI/scenes/Plugins/EdgeProvider.js
ahujsak: thanks, however Bity handles calling the `showError`. I figure it is better to let the individual plugins decide what to do with an error, and then if they want to use ours they can all `EdgeProvider` back and we can then show it.

src/constants/plugins/buySellPlugins.js
ahujsak: co.edgesecure.bity mirrors what we did with wyre and simplex. I am open to suggestions.

copy-bity.sh
ahujsak: fixed

PR #1625, opened 2019-09-21
Bundle the core plugins for a smaller total size

swansontec:
#### PR Requirements

Minification is enabled on the core, but not the plugins for now. Core = 633 KiB, Plugins = 7.88 MiB.

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


swansontec:
Also pushed this to the test branch.

webpack.config.js
```diff
@@ -0,0 +1,39 @@
...
+const path = require('path')
+
+const webpack = require('webpack')
+
+// Set this to false for easier debugging:
+// TODO: Production mode is currently crashing
+const production = false
+
+const babelOptions = {
+  presets: production ? ['@babel/preset-env'] : [],
```
swansontec: Actually, wait! This will completely break on old Android phones. I need to solve this before we can merge anything.

swansontec: Aha! Found the problem, and we now run with minified plugins. The plugin bundle is now 4.36 MiB.

postinstall.sh
```diff
@@ -52,43 +39,15 @@ cat >ios/edge-core/index.html <<HTML
   </head>
   <body>
     <script src="edge-core.js"></script>
-    <script>
-      var loading = 0
```
paullinator: Why was the code to do delayed loading of the plug-ins removed? Even though there are fewer plug-ins, it would still be helpful to delay loading the full bundle and let the GUI load with the core.

postinstall.sh
```diff
@@ -52,43 +39,15 @@ cat >ios/edge-core/index.html <<HTML
...
+    <script
+      async
+      src="plugin-bundle.js"
+      onerror="lockEdgeCorePlugins()"
+      onload="lockEdgeCorePlugins()"
+    ></script>
   </body>
 </html>
 HTML
-cp ./node_modules/edge-core-js/lib/react-native/edge-core.js ./ios/edge-core
```
paullinator: What replaces this section? Don’t each of the plug-ins still need to serve up their react native side?

postinstall.sh
swansontec: Since there is only one now, the simple `async` prop on the script tag should now handle the delay. However, we should bench it to verify that this is the case.

postinstall.sh
swansontec: Both iOS and Android now pull assets from the same folder now.

paullinator:
Just to fix conflicts and this is approved.

postinstall.sh
swansontec: I went ahead and rolled it back for safety.

PR #1628, opened 2019-09-22
Refactor buy and sell list to highlight payment method versus partner

sfncook:
#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


sfncook:
Screen Shot 2019-09-22 at 12 03 22 PM
Screen Shot 2019-09-22 at 12 03 36 PM
Screen Shot 2019-09-22 at 12 03 49 PM
Screen Shot 2019-09-22 at 12 23 27 PM

src/types/types.js
```diff
@@ -210,28 +210,20 @@ export type SpendingLimits = {
...
   name: string,
-  subtitle: string,
-  imageUrl: string,
   supportEmail?: string,
   permissions?: Array<string>,
   originWhitelist?: Array<string>,
-  isLegacy?: true
-}
-
-export type BuySellFilter = {
```
paullinator: Bring back this flow type and edit it to match the JSON objects for buy and sell plug-ins

src/components/scenes/PluginViewListScene.js
```diff
@@ -189,22 +210,15 @@ class PluginList extends Component<Props> {
...
-    // Remove plugins that don't match our country:
-    if (countryData != null) {
-      plugins = plugins.filter((plugin: BuySellPlugin) => plugin.pluginId === 'custom' || filter[plugin.name.toLowerCase()].countryCodes[countryCode])
-    }
+    let plugins =
+      direction === 'buy'
+        ? getBuyPlugins(Platform.OS, countryCode)
+        : direction === 'sell'
+          ? getSellPlugins(Platform.OS, countryCode)
+          : getAllPlugins(Platform.OS, countryCode)
```
paullinator: From here on out we will only support showing either buy or sell plug-ins but not all. Please remove the all case. There is only one part of the app that calls and with all and that can be changed to request the buy list

PR #1642, opened 2019-09-27
ChangeHero plugin

magmel48:
#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a

Associated request with plugin can be found here https://github.com/EdgeApp/edge-exchange-plugins/pull/47


swansontec:
This PR contains a number of merges. If we ever do go to implement this, we would like a clean PR with just a single commit.

PR #1644, opened 2019-10-01
Implement Hedera support

abonander:
### Blocked on Companion PR: https://github.com/EdgeApp/edge-currency-accountbased/pull/80

This PR is based on `develop` which I notice has diverged from `master` by ~150 commits but both have recent activity. [The workflow conventions](https://github.com/EdgeApp/edge-conventions/blob/master/workflow.md) mentioned in the Contributing section of the README don't specify which one to prefer but `develop` sounded more appropriate.

Refer to the `edge-currency-accountbased` PR for details. This PR probably needs to wait for a new version of that package to be published.

I was able to run `npm run format` in this one but it didn't produce a delta.


abonander:
The build failure appears unrelated to my changes.

abonander:
@swansontec if you're involved on the frontend, this PR is ready for review pending the merge of https://github.com/EdgeApp/edge-currency-accountbased/pull/80

swansontec:
To test this, I went ahead and rebased this PR as [william/rebase-hedera](https://github.com/EdgeApp/edge-react-gui/tree/william/rebase-hedera). Then, when I tried creating a wallet, I got this:

Screenshot_20200513-105828_Edge

I will try getting Hedera in another app and then importing the keys.

abonander:
@swansontec oh sorry, I forgot to update the endpoints in the code; they're still pointing to the local host machine address instead of an actual server. Let me fix that right now.

abonander:
@swansontec fixed, please update `edge-currency-accountbased` and try again.

src/components/scenes/CreateWalletSelectFiatScene.js
```diff
@@ -63,6 +63,8 @@ class CreateWalletSelectFiatComponent extends Component<Props, State> {
       Actions[nextSceneKey]({
         selectedWalletType: selectedWalletType,
         selectedFiat: this.getFiatType(this.state.selectedFiat),
+        // Hedera-like: wallet name is local-only but activation is still necessary
+        needsAccountCreation: specialCurrencyInfo.needsAccountCreation,
```
abonander, line 67: This doesn't need to be passed through props.

swansontec:
After the update, I get this now:

Screenshot_20200514-103725_Edge

swansontec:
Also, for every time I have gone through this process, I have ended up with a new wallet! There is no way to activate these wallets after-the-fact, so now I have a good pile of useless Hedera wallets.

I think we should change the flow so the create-wallet process presents a confirmation screen *before* it creates the wallet, and then replace the transaction list screen with an "activate wallet" screen as long as the wallet remains un-activated. Creating the wallet immediately goes to this new screen. This would most likely apply to EOS, Hedera, XRP, and XLM.

abonander:
@swansontec I noticed that when I try to create an EOS wallet but back out before performing activation, the UI actually knows that the wallet isn't activated and prompts me to continue the process when I try to use it, but I couldn't figure out how to duplicate that for the Hedera wallet.

Also, sorry about that new error, the creator API we have deployed is missing the "get activation cost" route and needs to be updated.

src/actions/WalletActions.js
```diff
@@ -67,26 +67,39 @@ export const selectWallet = (walletId: string, currencyCode: string, from?: stri
...
+
     if (publicAddress) {
       // already activated
       dispatch({
         type: 'UI/WALLETS/SELECT_WALLET',
         data: { walletId, currencyCode }
       })
+    } else if (activationPending) {
+      // Hedera: wallet is pending activation, we have no course of action but to wait
+      Actions.jump(Constants.WALLET_LIST_SCENE)
```
abonander, line 95: @paullinator 

This branch is for Hedera wallets where the activation payment was submitted and now we're just waiting on the backend to confirm the payment and create the account.

I figured if we don't want the user to really access the wallet here and so should kick them back to the wallet list scene while displaying the modal, but I couldn't figure out how to make this work. 

When this branch is triggered the modal seems to pop up over the wallet details scene for an arbitrarily different wallet instead. If I use `Actions.popTo(Constants.WalletListScene)` like I see elsewhere in the code, I get a warning that the route does not exist.

paullinator: You are seeing an arbitrary wallet details scene because you never call dispatch `UI/WALLETS/SELECT_WALLET` so the most recent selected wallet would show. I dont think you need to do this however if youre just going top hop back to the wallet list scene. Not sure if you just have a typo but the command should be 

`Actions.popTo(Constants.WALLET_LIST_SCENE)`

Not `Actions.jump(Constants.WALLET_LIST_SCENE)` and not `Actions.popTo(Constants.WalletListScene)`

abonander: Yeah it was a typo, still having trouble with `Actions.popTo(Constants.WALLET_LIST_SCENE)`. I only tried `Actions.jump()` to see if it would make a difference.

PR #1646, opened 2019-10-02
Implement 'Exchange Rate Loading' on sidebar

kylanhurt:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/modules/UI/components/ControlPanel/ControlPanel.ui.js
+++ b/src/modules/UI/components/ControlPanel/ControlPanel.ui.js
@@ -28,7 +28,7 @@ export type Props = {
   openSelectUser: () => void,
   closeSelectUser: () => void,
   usersView: boolean,
-  exchangeRate: Object | number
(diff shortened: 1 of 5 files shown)


src/modules/UI/components/ControlPanel/ControlPanel.ui.js
```diff
@@ -25,7 +27,8 @@ export type Props = {
   username: string,
   openSelectUser: () => void,
   closeSelectUser: () => void,
-  usersView: boolean
+  usersView: boolean,
+  exchangeRate: Object | number
```
swansontec: It seems to always be a number? Maybe it should be `number | null`?

kylanhurt: The `getExchangeRate` returns `0` if there is no exchange rate, so we'll just use `number`

PR #1655, opened 2019-10-07
Allen/notifications

ahujsak:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -15,6 +15,7 @@ import ENV from '../env.json'
 import * as Constants from './constants/indexConstants.js'
 import s from './locales/strings.js'
 import { log, logToServer } from './util/logger'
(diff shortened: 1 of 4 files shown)


src/app.js
```diff
@@ -183,6 +185,24 @@ async function backgroundWorker () {
         }
       }
     }
+    // After OTP resets, the next thing to check for is are there any other notifications.
+    const notification = await checkForNotifications()
+    if (notification) {
+      const dateOne = new Date(Date.now() + 1000)
```
swansontec: One second in the future? Seems good, but maybe a comment would be good here - I saw the `+ 1000`, and at first I thought it was the `* 1000` we are constantly doing to go from seconds to ms.

PR #1674, opened 2019-10-24
On Send Confirmation Screen, clicking max spend will always make the top flip input crypto

amanalili:
Task Send Confirmation - flip crypto/fiat amount when user selects max send (https://app.asana.com/0/361770107085503/1134961556780288)

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android


src/modules/UI/components/FlipInput/FlipInput2.ui.js
```diff
@@ -204,6 +205,15 @@ export class FlipInput extends Component<Props, State> {
     }
   }
 
+  componentDidUpdate (prevProps: Props) {
+    if (this.props.toggleFiatOnTop != null) {
+      console.log(this.props.toggleFiatOnTop)
```
swansontec: Please remove `console.log`

src/reducers/scenes/SendConfirmationReducer.js
```diff
@@ -245,6 +247,17 @@ export const transaction = (state: EdgeTransaction | null = null, action: Action
...
 }
 
+export const toggleFiatOnTop = (state: boolean | null = null, action: Action) => {
+  switch (action.type) {
+    case 'UI/SEND_CONFIMATION/TOGGLE_FIAT_ON_TOP': {
+      return action.data
+    }
+
+    default:
+      return state
```
swansontec: Once this gets set, it never gets unset, which means the flip input would be stuck flipped forever, I think.

src/modules/UI/components/FlipInput/FlipInput2.ui.js
amanalili: Fixed

src/reducers/scenes/SendConfirmationReducer.js
amanalili: Yeah I guess so, fixed to default state null

src/actions/SendConfirmationActions.js
```diff
@@ -143,6 +148,7 @@ export const updateMaxSpend = () => (dispatch: Dispatch, getState: GetState) =>
   const walletId = getSelectedWalletId(state)
   const edgeWallet = getWallet(state, walletId)
   const spendInfo = getSpendInfo(state)
+  const toggleCryptoOnTopState = state.ui.scenes.sendConfirmation.toggleCryptoOnTop
```
paullinator: Do the incrementing in the reducer versus here. The reducer already has access to the current counter and can simply increment it.

src/reducers/scenes/SendConfirmationReducer.js
```diff
@@ -245,6 +247,17 @@ export const transaction = (state: EdgeTransaction | null = null, action: Action
   }
 }
 
+export const toggleCryptoOnTop = (state: number = 0, action: Action) => {
+  switch (action.type) {
+    case 'UI/SEND_CONFIMATION/TOGGLE_CRYPTO_ON_TOP': {
```
paullinator: This is where you would want to increment the counter instead of in the action

paullinator:
This looks pretty good but I just left a couple small comments which should be easy changes

PR #1678, opened 2019-10-29
When exchange rate is present, the flip input fiat is on the top

amanalili:
Task: Send screen - fiat and crypto values flipped from rest of app (https://app.asana.com/0/361770107085503/1127823520648213)

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/modules/UI/components/FlipInput/ExchangedFlipInput2.js
+++ b/src/modules/UI/components/FlipInput/ExchangedFlipInput2.js
@@ -147,6 +147,13 @@ export class ExchangedFlipInput extends Component<Props, State> {
     this.props.onExchangeAmountChanged({ exchangeAmount, nativeAmount })
   }
 
+  isFiatOnTop = () => {
+    if (this.props.isFiatOnTop) {
+      return this.props.isFiatOnTop
+    }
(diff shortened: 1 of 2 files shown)


src/modules/UI/components/FlipInput/ExchangedFlipInput2.js
```diff
@@ -147,6 +147,13 @@ export class ExchangedFlipInput extends Component<Props, State> {
     this.props.onExchangeAmountChanged({ exchangeAmount, nativeAmount })
   }
 
+  isFiatOnTop = () => {
+    if (this.props.isFiatOnTop) {
+      return this.props.isFiatOnTop
+    }
+    return this.state.exchangeSecondaryToPrimaryRatio !== false
```
paullinator: exchangeSecondaryToPrimaryRatio is a string. This comparison checks it as a boolean. You should be checking it to see if it is equal to 0 using a line such as 

`bns.eq(this.state.exchangeSecondaryToPrimaryRatio, '0')`

amanalili: Updated

PR #1681, opened 2019-11-02
Include addOnUrl with plugin URI

sfncook:
#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/scenes/PluginViewScene.js
+++ b/src/components/scenes/PluginViewScene.js
@@ -176,11 +176,12 @@ class PluginView extends React.Component<Props> {
   }
 
   render () {
-    const { uri, originWhitelist = ['file://*', 'https://*', 'http://*', 'edge://*'] } = this.props.plugin
+    const { uri, addOnUrl = '', originWhitelist = ['file://*', 'https://*', 'http://*', 'edge://*'] } = this.props.plugin
     const userAgent =
       Platform.OS === 'android'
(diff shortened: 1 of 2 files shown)


src/components/scenes/PluginViewScene.js
```diff
@@ -176,11 +176,12 @@ class PluginView extends React.Component<Props> {
   }
 
   render () {
-    const { uri, originWhitelist = ['file://*', 'https://*', 'http://*', 'edge://*'] } = this.props.plugin
+    const { uri, addOnUrl, originWhitelist = ['file://*', 'https://*', 'http://*', 'edge://*'] } = this.props.plugin
```
swansontec: I would use a default, like `addOnUrl = ''`, rather than the `|| ''` later.

src/types/types.js
```diff
@@ -231,7 +231,8 @@ export type BuySellPlugin = {
   cryptoCodes?: Array<string>,
   countryCodes?: { [string]: boolean },
   forPlatform?: string,
-  forCountries?: Array<string>
+  forCountries?: Array<string>,
+  addOnUrl?: string
```
swansontec: If you want to document `addOnUrl`, this would be the place.

src/components/scenes/PluginViewScene.js
sfncook: Done.

src/types/types.js
sfncook: Done.

PR #1688, opened 2019-11-08
tweak conditionals for disabling tokens

ahujsak:
#### PR Requirements
This adjusts disabled conditionals for tokens with 0 crypto but some fiat. 
- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


swansontec:
I just gave this a spin. Before, the balance for my USDC wallet showed as 0, but now it's 1971732 (I wish)! The true balance is 1.97 USDC. The way you are calculating crypto display amounts somehow differs from the rest of the app, so we are still seeing bugs.

PR #1713, opened 2019-11-25
Import `Airship` from NPM

swansontec:
#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


swansontec:
The Travis iOS build keeps timing out on `pod install`, but the problem isn't related to this commit.

PR #1721, opened 2019-11-28
Change the text of the biometry row to what biometry type is supported

amanalili:
Task: Rename rows on settings. Use TouchId, Use FaceId, Use Biometric. - https://app.asana.com/0/361770107085503/1126000976007328

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android

### DIFF
--- a/src/components/scenes/SettingsOverviewScene.js
+++ b/src/components/scenes/SettingsOverviewScene.js
@@ -49,25 +49,17 @@ type Props = {
   toggleDeveloperMode(boolean): void
 }
 
-export default class SettingsOverview extends Component<Props> {
+type State = {
+  touchIdText: string
+}
+
+export default class SettingsOverview extends Component<Props, State> {
   optionModals: Array<Object>
   currencies: Array<Object>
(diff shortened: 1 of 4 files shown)


src/components/scenes/SettingsOverviewScene.js
```diff
@@ -86,6 +87,33 @@ export default class SettingsOverview extends Component<Props> {
...
+  async componentDidMount () {
+    if (!this.props.supportsTouchId) {
+      return null
+    }
+    try {
+      const biometryType = await getSupportedBiometryType()
+      switch (biometryType) {
+        case 'FaceID':
+          this.options.useTouchID.text = s.strings.settings_button_use_faceID
+          this.forceUpdate()
```
paullinator: Do not use forceUpdate(). Instead, move this.options.useTouchID.text to component State to it so that you automatically re-render when it changes

amanalili: Fix on the new commit

PR #1751, opened 2019-12-17
Fix selectors function to use the wallets fiat on components using it

amanalili:
Task: https://app.asana.com/0/361770107085503/1153857373665121

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

### DIFF
--- a/src/components/common/CryptoExchangeWalletListRow.js
+++ b/src/components/common/CryptoExchangeWalletListRow.js
@@ -8,7 +8,7 @@ import { Image, TouchableHighlight, View } from 'react-native'
 import { intl } from '../../locales/intl'
 import * as SETTINGS_SELECTORS from '../../modules/Settings/selectors.js'
 import FormattedText from '../../modules/UI/components/FormattedText/index'
-import { calculateSettingsFiatBalance } from '../../modules/UI/selectors.js'
+import { calculateWalletFiatBalance } from '../../modules/UI/selectors.js'
 import { CryptoExchangeWalletListRowStyle as styles } from '../../styles/indexStyles'
 import type { State } from '../../types/reduxTypes.js'
 import type { GuiWallet } from '../../types/types.js'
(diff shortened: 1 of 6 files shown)


src/components/common/SortableWalletListRow.js
```diff
@@ -10,7 +10,7 @@ import { intl } from '../../locales/intl'
 import s from '../../locales/strings.js'
 import * as SETTINGS_SELECTORS from '../../modules/Settings/selectors'
 import T from '../../modules/UI/components/FormattedText/index'
-import { calculateSettingsFiatBalanceWithoutState } from '../../modules/UI/selectors.js'
+import { calculateWalletFiatBalanceWithoutState } from '../../modules/UI/selectors.js'
```
paullinator: Is the old functions still in use? If not, please delete it

src/components/common/CryptoExchangeWalletListRow.js
```diff
@@ -8,7 +8,7 @@ import { Image, TouchableHighlight, View } from 'react-native'
 import { intl } from '../../locales/intl'
 import * as SETTINGS_SELECTORS from '../../modules/Settings/selectors.js'
 import FormattedText from '../../modules/UI/components/FormattedText/index'
-import { calculateSettingsFiatBalance } from '../../modules/UI/selectors.js'
+import { calculateWalletFiatBalance } from '../../modules/UI/selectors.js'
```
paullinator: Remove the old function as good as it does not appear to be in use anymore

src/modules/UI/selectors.js
```diff
@@ -207,3 +207,17 @@ export const calculateWalletFiatBalance = (wallet: GuiWallet, state: State): str
...
+  const nativeBalance = wallet.nativeBalances[currencyCode]
+  if (!nativeBalance || nativeBalance === '0') return '0'
+  const denominations = settings[currencyCode].denominations
+  const exchangeDenomination = denominations.find(denomination => denomination.name === currencyCode)
+  if (!exchangeDenomination) return '0'
+  const nativeToExchangeRatio: string = exchangeDenomination.multiplier
+  const cryptoAmount: number = parseFloat(convertNativeToExchange(nativeToExchangeRatio)(nativeBalance))
+  fiatValue = convertCurrencyWithoutState(exchangeRates, currencyCode, wallet.isoFiatCurrencyCode, cryptoAmount)
+  return intl.formatNumber(fiatValue, { toFixed: 2 }) || '0'
+}
```
paullinator: This does not appear to address the problem for tokens. I suggest you change this function to except a currency code parameter and use that to determine which currency to calculate the fiat balance for.You can then pass in the currency code from the component which knows whether or not it needs the balance for the parent or the token

PR #1761, opened 2019-12-19
Enable hidden wallets

bruscantini:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a

Not show "hidden" wallets on WalletListScene.
Depends on changes from https://github.com/EdgeApp/edge-core-js/pull/267

### DIFF
--- a/src/components/scenes/WalletListScene.js
+++ b/src/components/scenes/WalletListScene.js
@@ -59,6 +59,7 @@ type State = {
 type Props = {
   activeWalletIds: Array<string>,
   creationTweaks: AppTweaks,
+  hiddenWalletIds: Array<string>,
   customTokens: Array<any>,
   dimensions: DeviceDimensions,
   wallets: any,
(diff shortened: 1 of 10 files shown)


src/components/scenes/WalletListScene.js
```diff
@@ -132,24 +133,21 @@ export default class WalletList extends Component<Props, State> {
   }
 
   render () {
-    const { wallets, activeWalletIds } = this.props
-    const walletsArray = []
-    const activeWallets = {}
```
bruscantini: It doesn't look like these variables were ever used, so I removed them.

bruscantini:
Hi @swansontec or @paullinator .
The integration test is failing because of 'environment variables'. What should I do here? Thanks,

paullinator:
@bruscantini Can you send a screenshot or copy/paste of the environment variable error you are seeing?

paullinator:
As well, are you seeing this error when building the develop branch without your changes?

bruscantini:
Hi @paullinator !
I'm actually not seeing the error at all during building. But the Travis CI for this PR is throwing the error. I just want the PR to pass all automated checks before your review/merge. :)

Here's the link to the Travis CI error: https://travis-ci.org/EdgeApp/edge-react-gui/jobs/633649267?utm_medium=notification&utm_source=github_status

swansontec:
@paullinator I know what this is. The way we store encrypted API keys in the Travis build system doesn't work when third parties initiate a build. Therefore, any PR's from outside the company are going to fail Travis automatically. We should probably just put some plain-text dummy API keys into Travis, rather than using encryption.

If the changes look good to you, Paul, feel free to merge, even without Travis passing.

paullinator:
@swansontec Thanks for the info. I will review the pr in the next couple days.

bruscantini:
Hey @paullinator can you review and merge this PR when you get a chance? Once react-gui can hide wallets, we can link Edge in the Ember Fund app. Thanks!

paullinator:
These changes look good but you also need to add the logic to exclude hidden wallets from the Wallet picker Modal

paullinator:
@swansontec Can you confirm that the appropriate hidden wallets parameter is now in the Edge account object that comes from the core?

bruscantini:
Thanks @paullinator . I updated based on your comment. One thing, though. When running the edge-react-gui locally, account.hiddenWalletIds is empty. But when I run the Ember app, the EdgeAccount object I get has hiddenWalletIds populated.

Is this because I am setting the wallet state to hidden _from_ the _Ember_ app? The goal is to be able to hide the wallets from Ember app so that they do not appear in the edge-react-gui.

swansontec:
It looks like this is blocked on the fact that the `hidden` flag is stored per-account, so setting it on the Ember account isn't setting it on the Edge Wallet account. We need to fix edge-core-js to sync this flag between accounts, and then we can refresh & merge this PR.

PR #1766, opened 2019-12-24
Functions to secure sorting of wallet list on wallet selectors

amanalili:
Task: Wallet List scene order of wallets is not the same as the wallet selector list. (https://app.asana.com/0/361770107085503/1154597586670346)

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android

### DIFF
--- a/src/components/modals/WalletListModal.js
+++ b/src/components/modals/WalletListModal.js
@@ -17,6 +17,7 @@ import { type AirshipBridge, AirshipModal } from './modalParts.js'
 type Props = {
   bridge: AirshipBridge<GuiWallet | Object | null>,
   wallets: Array<GuiWallet>,
+  activeWalletIds: Array<string>,
   existingWalletToFilterId?: string,
   existingWalletToFilterCurrencyCode?: string,
   headerTitle: string,
@@ -53,17 +54,18 @@ export class WalletListModal extends Component<Props, LocalState> {
     let i = 0
     let totalCurrenciesAndTokens = 0
     let totalWalletsToAdd = 0
-    for (i; i < this.props.wallets.length; i++) {
-      const wallet = this.props.wallets[i]
-      if (wallet.type === 'wallet:fio') continue
-      const record = {
-        walletItem: wallet,
-        supportedWalletType: null
-      }
-      records.push(record)
-      totalCurrenciesAndTokens++
-      if (wallet.enabledTokens.length) {
-        totalCurrenciesAndTokens = totalCurrenciesAndTokens + wallet.enabledTokens.length
+    for (i; i < this.props.activeWalletIds.length; i++) {
+      const wallet = this.props.wallets.find(wallet => wallet.id === this.props.activeWalletIds[i])
+      if (wallet) {
+        const record = {
+          walletItem: wallet,
+          supportedWalletType: null
+        }
+        records.push(record)
+        totalCurrenciesAndTokens++
+        if (wallet.enabledTokens.length) {
+          totalCurrenciesAndTokens = totalCurrenciesAndTokens + wallet.enabledTokens.length
+        }
       }
     }
     for (i = 0; i < this.props.supportedWalletTypes.length; i++) {
--- a/src/components/modals/WalletListModal.js
+++ b/src/components/modals/WalletListModal.js
@@ -14,17 +14,20 @@ import type { GuiWallet } from '../../types/types.js'
 import { scale } from '../../util/scaling.js'
 import { type AirshipBridge, AirshipModal } from './modalParts.js'
 
-type Props = {
+export type StateProps = {
+  state: State,
+  activeWalletIds: Array<string>
+}
+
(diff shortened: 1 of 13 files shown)


src/components/modals/WalletListModal.js
```diff
@@ -53,17 +54,18 @@ export class WalletListModal extends Component<Props, LocalState> {
...
-        walletItem: wallet,
-        supportedWalletType: null
-      }
-      records.push(record)
-      totalCurrenciesAndTokens++
-      if (wallet.enabledTokens.length) {
-        totalCurrenciesAndTokens = totalCurrenciesAndTokens + wallet.enabledTokens.length
+    for (i; i < this.props.activeWalletIds.length; i++) {
+      const wallet = this.props.wallets.find(wallet => wallet.id === this.props.activeWalletIds[i])
+      if (wallet) {
```
paullinator: Need to add back the logic that excludes FIO

amanalili:
> Change the WalletListModal to be connected so that you do not have to pass in the active Wallet IDs from all of the locations that use the modal

Fixed

src/components/modals/WalletListModal.js
amanalili: Fixed

src/components/modals/WalletListModal.js
```diff
@@ -14,7 +14,12 @@ import type { GuiWallet } from '../../types/types.js'
 import { scale } from '../../util/scaling.js'
 import { type AirshipBridge, AirshipModal } from './modalParts.js'
 
-type Props = {
+export type connectorProps = {
```
paullinator: This should be called StateProps

src/components/modals/WalletListModal.js
```diff
@@ -14,7 +14,12 @@ import type { GuiWallet } from '../../types/types.js'
...
+  state: State,
+  activeWalletIds: Array<string>
+}
+
+type componentProps = {
```
paullinator: This should be called OwnProps

src/connectors/components/WalletListConnector.js
```diff
@@ -0,0 +1,21 @@
...
+import type { connectorProps } from '../../components/modals/WalletListModal'
+import { WalletListModal } from '../../components/modals/WalletListModal'
+import { getActiveWalletIds } from '../../modules/UI/selectors.js'
+import type { State } from '../../types/reduxTypes.js'
+
+export const mapStateToProps = (state: State): connectorProps => {
+  const wallets = state.ui.wallets.byId
+  const activeWalletIds = getActiveWalletIds(state).filter(id => !(wallets[id] != null && wallets[id].type === 'wallet:fio'))
+  return {
+    state,
```
paullinator: It looks like you no longer need the state. Please remove

src/connectors/components/WalletListConnector.js
```diff
@@ -0,0 +1,21 @@
+// @flow
```
paullinator: This file should be called WalletListModalConnector

src/connectors/components/WalletListConnector.js
```diff
@@ -0,0 +1,21 @@
...
+  const activeWalletIds = getActiveWalletIds(state).filter(id => !(wallets[id] != null && wallets[id].type === 'wallet:fio'))
+  return {
+    state,
+    activeWalletIds
+  }
+}
+
+const WalletListConnector = connect(mapStateToProps)(WalletListModal)
+
+export { WalletListConnector }
```
paullinator: Change the export to WalletListModalConnected

src/components/modals/WalletListModal.js
amanalili: Fix

src/components/modals/WalletListModal.js
amanalili: Fixed

src/connectors/components/WalletListConnector.js
amanalili: State is needed by the WalletListModal itself

src/connectors/components/WalletListConnector.js
amanalili: Fixed

src/connectors/components/WalletListConnector.js
amanalili: Fixed

src/connectors/components/WalletListConnector.js
paullinator: WalletListModal does not appear to use state. From what i see it passes state into CryptoExchangeWalletListRow which is used to call the selector calculateWalletFiatBalance.  Selectors should not be called from within components. Instead you should make CryptoExchangeWalletListRow a connected component and use its connector to call calculateWalletFiatBalance and pass in the result.

amanalili: Fixed, updated PR

PR #1769, opened 2019-12-27
Write Bity data to logs for QA

kylanhurt:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/modules/UI/scenes/Plugins/EdgeProvider.js
+++ b/src/modules/UI/scenes/Plugins/EdgeProvider.js
@@ -362,11 +362,15 @@ export class EdgeProvider extends Bridgeable {
     return transaction
   }
 
+  // log body and signature and pubic address and final message (returned from signMessage)
+  // log response afterwards line 451
   async signMessage (message: string) /* EdgeSignedMessage */ {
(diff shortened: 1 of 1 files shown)


src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -420,55 +423,64 @@ export class EdgeProvider extends Bridgeable {
...
+          const bankDetailResponse = await window.fetch(detailUrl, bankDetailsRequest)
+          if (bankDetailResponse.status === 200) {
+            const parsedResponse = await bankDetailResponse.json()
+            console.log('Bity parsedResponse: ', parsedResponse)
+            return parsedResponse
+          }
         }
       }
+      return orderData
+    } catch (e) {
```
paullinator: When I mentioned to add a catch, it was to add a catch to each of the fetch calls individually so that you know which one actually threw. It would be easiest to add the .catch() notation after each fetch call.

```
await window.fetch(newURL, request).catch(e => {
  console.log(`throw from fetch ${newURL}`, e)
  throw e
})```

PR #1810, opened 2020-02-06
William/preferred swap

swansontec:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


src/reducers/scenes/SettingsReducer.js
```diff
@@ -79,6 +79,7 @@ export type SettingsState = {
   isTouchSupported: boolean,
   loginStatus: boolean | null,
   merchantMode: boolean,
+  preferredSwapPluginId: string,
```
swansontec: string | void

swansontec: Fixed.

PR #1832, opened 2020-02-25
Adds getTransactions to EdgeProvider

peachbits:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/modules/UI/scenes/Plugins/EdgeProvider.js
+++ b/src/modules/UI/scenes/Plugins/EdgeProvider.js
@@ -8,12 +8,14 @@ import { Linking } from 'react-native'
 import Mailer from 'react-native-mail'
 import { Actions } from 'react-native-router-flux'
 import SafariView from 'react-native-safari-view'
+import { sprintf } from 'sprintf-js'
 import { Bridgeable } from 'yaob'
 
 import { createCurrencyWalletAndSelectForPlugins } from '../../../../actions/CreateWalletActions.js'
 import { trackAccountEvent, trackConversion } from '../../../../actions/TrackingActions.js'
 import { selectWallet } from '../../../../actions/WalletActions'
 import { launchModal } from '../../../../components/common/ModalProvider.js'
+import { TwoButtonSimpleConfirmationModal } from '../../../../components/modals/TwoButtonSimpleConfirmationModal.js'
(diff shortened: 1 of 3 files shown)


src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -288,6 +296,60 @@ export class EdgeProvider extends Bridgeable {
...
+        transactions.push(newTx)
+      }
+      const result: EdgeGetTransactionsResult = {
+        fiatCurrencyCode,
+        balance,
+        transactions
+      }
+      console.log('EdgeGetTransactionsResult', JSON.stringify(result))
+      return result
+    } else {
```
swansontec: I prefer an early bail-out style. If you put `if (!confirmTxShare)  throw new Error('User denied permission')` near the top instead, then the rest of the function can be un-indented one level.

src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -288,6 +296,60 @@ export class EdgeProvider extends Bridgeable {
...
+        bridge={bridge}
+        title={s.strings.fragment_wallets_export_transactions}
+        subTitle={sprintf(s.strings.transaction_history_permission, coreWallet.name)}
+        cancelText={s.strings.no}
+        doneText={s.strings.yes}
+      />
+    ))
+
+    // Grab transactions from current wallet
+
```
swansontec: The comment that describes the following block is a great coding style, but there shouldn't be a blank line between it and the block itself.

src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -288,6 +296,60 @@ export class EdgeProvider extends Bridgeable {
...
+          txid: tx.txid,
+          ourReceiveAddresses: tx.ourReceiveAddresses,
+          metadata: tx.metadata
+        }
+        transactions.push(newTx)
+      }
+      const result: EdgeGetTransactionsResult = {
+        fiatCurrencyCode,
+        balance,
+        transactions
```
swansontec: We do we include this extra stuff in the reply? Don't they already have this via other API's? If not, maybe we should name `getTransactions` tp `getWalletHistory` instead, since it does more than it says on the tin.

PR #1836, opened 2020-02-27
Feature/domain resolution

JohnnyJumper:
## Integration with Unstoppabledomains
Using @unstoppabledomains/resolution to do the resolution for blockchain domains.
Screenshots with the results bellow

photo5028715811338037344
photo5028715811338037343
photo5028715811338037342

## Changes
- package.json
- src/components/modals/AddressModal.js
- src/locales/en_US.js
- src/locales/strings/enUS.json
- src/util/utils.js

## Enviroment tested at
 - Android phone ( OnePlus T6 )
 - Android version: 9
 - OxygenOS version 9.0.17

### DIFF
--- a/src/components/modals/AddressModal.js
+++ b/src/components/modals/AddressModal.js
@@ -11,6 +11,7 @@ import * as Constants from '../../constants/indexConstants'
 import s from '../../locales/strings.js'
 import styles from '../../styles/scenes/ScaneStyle'
 import { colors as COLORS } from '../../theme/variables/airbitz.js'
+import resolution from '@unstoppabledomains/resolution'
 
 // INTERACTIVE_MODAL /////////////////////////////////////////////////////////////////////////////
 type AddressModalProps = {
--- a/src/components/modals/AddressModal.js
+++ b/src/components/modals/AddressModal.js
@@ -12,7 +12,6 @@ import * as Constants from '../../constants/indexConstants'
 import s from '../../locales/strings.js'
 import styles from '../../styles/scenes/ScaneStyle'
 import { colors as COLORS } from '../../theme/variables/airbitz.js'
-import { debounce } from '../../util/utils'
 
 // INTERACTIVE_MODAL /////////////////////////////////////////////////////////////////////////////
 type AddressModalProps = {
@@ -85,7 +84,9 @@ export class AddressModal extends Component<AddressModalProps, AddressModalState
   onChangeTextDelayed = (domain: string) => {
     const { currencyCode } = this.props
     console.log(`before bouncing on ${domain} ${currencyCode}`)
-    debounce(this.resolveAddress(domain, currencyCode), 500, false)
+    if (domain.endsWith('.zil') || domain.endsWith('.crypto')) {
+      this.resolveAddress(domain, currencyCode)
+    }
     this.updateUri(domain)
   }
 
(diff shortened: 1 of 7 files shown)


src/components/modals/AddressModal.js
```diff
@@ -60,6 +65,65 @@ export class AddressModal extends Component<AddressModalProps, AddressModalState
...
 
+  setStatusLabel = (status: string) => {
+    this.setState({ statusLabel: status })
+  }
+
+  setCryptoAddress = (address: string) => {
+    this.setState({ cryptoAddress: address })
+  }
+
+  getServiceName = (resolutionObject: Resolution, domain: string) => {
```
paullinator: unused?

src/components/modals/AddressModal.js
```diff
@@ -11,16 +11,20 @@ import * as Constants from '../../constants/indexConstants'
...
 import styles from '../../styles/scenes/ScaneStyle'
 import { colors as COLORS } from '../../theme/variables/airbitz.js'
+import ResolutionError, { ResolutionErrorCode} from '../common/ResolutionError.js'
 
 // INTERACTIVE_MODAL /////////////////////////////////////////////////////////////////////////////
 type AddressModalProps = {
   onDone: any => void,
-  coreWallet: EdgeCurrencyWallet
+  coreWallet: EdgeCurrencyWallet,
+  currencyCode: string
```
paullinator: unused?

src/components/modals/AddressModal.js
```diff
@@ -11,16 +11,20 @@ import * as Constants from '../../constants/indexConstants'
...
 }
 
 type AddressModalState = {
   clipboard: string,
-  uri: string
+  uri: string,
+  statusLabel: string,
+  cryptoAddress?: string
```
paullinator: unused

src/components/modals/AddressModal.js
```diff
@@ -60,6 +65,65 @@ export class AddressModal extends Component<AddressModalProps, AddressModalState
     }
   }
 
+  setStatusLabel = (status: string) => {
+    this.setState({ statusLabel: status })
+  }
+
+  setCryptoAddress = (address: string) => {
```
paullinator: unused

src/components/modals/AddressModal.js
JohnnyJumper: This is needed for the error code in cases when some domain doesn't have a certain ticker configured. 

Example of the error message would be: 
" Domain brad.crypto has no BTC address attached to it "

src/components/modals/AddressModal.js
```diff
@@ -89,11 +148,11 @@ export class AddressModal extends Component<AddressModalProps, AddressModalState
...
                 style={MaterialInputStyle}
                 value={uri}
-                onChangeText={this.updateUri}
+                onChangeText={this.onChangeTextDelayed}
                 error={''}
                 placeholder={s.strings.fragment_send_send_to_hint}
-                label={s.strings.fragment_send_send_to_hint}
-                onSubmit={() => this.props.onDone(uri)}
+                label={statusLabel}
+                onSubmit={() => this.props.onDone(cryptoAddress)}
```
paullinator: It looks like we will never send the URI to the onDone() Call back which is necessary if domain resolution does not succeed

PR #1843, opened 2020-03-06
Fix pin not showing on send confirmation when hitting spend limits

amanalili:
Task - PIN shows invalid even if PIN entry is not shown for BRZ - https://app.asana.com/0/361770107085503/1163402411159817

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/scenes/SendConfirmationScene.js
+++ b/src/components/scenes/SendConfirmationScene.js
@@ -96,8 +96,6 @@ type State = {|
 export class SendConfirmation extends Component<Props, State> {
   pinInput: any
   flipInput: any
-  count: number
-  lastSeenCount: number
 
   constructor (props: Props) {
     super(props)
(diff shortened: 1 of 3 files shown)


src/components/scenes/SendConfirmationScene.js
```diff
@@ -341,34 +337,21 @@ export class SendConfirmation extends Component<Props, State> {
...
 
     const guiMakeSpendInfoClone = { ...guiMakeSpendInfo }
     const spendInfo = getSpendInfoWithoutState(guiMakeSpendInfoClone, sceneState, currencyCode)
-    let authType
-    if (spendingLimits.transaction.isEnabled) {
-      authType = amountFiat > spendingLimits.transaction.amount ? 'pin' : 'none'
-    } else {
-      authType = 'none'
-    }
+    const authType: any = getAuthRequiredDispatch(spendInfo) // Type casting any cause dispatch returns a function
```
paullinator: This should not need to be type casted. Dispatch does not return to function but actually returns a value

amanalili: Lot of errors on flow as it is use on newSpendInfo(spendInfo, authType) which the second parameter expects a type

PR #1857, opened 2020-03-19
adds switchain images and config entries

slowdownitsfine:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


src/assets/images/exchange/index.js
```diff
@@ -16,9 +17,10 @@ import faastLogo from './settingsExchangeFaast.png'
 import foxExchangeLogo from './settingsExchangeFoxExchange.png'
 import godexLogo from './settingsExchangeGodex.png'
 import shapeshiftLogo from './settingsExchangeShapeshift.png'
+import switchainLogo from './settingsExchangeSwitchain.png'
 import totleLogo from './settingsExchangeTotle.png'
 
-export function getSwapPluginIcon (pluginId: string) {
+export function getSwapPluginIcon(pluginId: string) {
```
swansontec: Can you run `yarn format`? This looks like a formatting glitch.

src/util/corePlugins.js
```diff
@@ -80,6 +80,7 @@ export const swapPlugins = {
   foxExchange: ENV.FOX_INIT,
   godex: ENV.GODEX_INIT,
   // shapeshift: ENV.SHAPESHIFT_INIT,
+  switchain: ENV.SWITCHAIN_INIT,
```
swansontec: Can you add your block to env.example.json?

src/assets/images/exchange/index.js
dcusan: Done

src/util/corePlugins.js
dcusan: Added

swansontec:
For the other bug @peachbits found, with the missing icon, we need an entry in `src/components/services/ContactsLoader.js`. I think @peachbits can upload the icon if you add the entry to that file.

dcusan:
added the entry at `ContactsLoader`

swansontec:
> added the entry at `ContactsLoader`

Thanks, I see it now. However, it is a little cut-off with our circle treatment - not a deal-breaker, but maybe you'll want to adjust it.

Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-01 at 17 41 35

As far as final testing goes, I was able to do a swap using your service, which was quite fast. Our GUI isn't showing the `quoteId` your plugin sends back (our bug), but when I capture the `quoteId` in the debugger and paste it onto the end of your URI, I still don't get anything useful, even after performing the swap (https://www.switchain.com/order-status/da0950fc-51b0-3b60-b39e-07f60cb61953). So, it doesn't look like the `orderIdSeed` workaround you guys came up with is actually generating a usable `quoteId` in the end. So, I'll bite the bullet and adjust edge-core-js to allow plugins to return a `quoteId` as part of the `approve` method, thereby solving the problem properly.

dcusan:
Thanks @swansontec, I am unsure about the image cropping, as this image is hosted by edge I think.

I ran back to back exchanges in the simulator and the provided status url was correct. My independent local tests where also successful. I'm not sure what could be happening, maybe the capitalization of the query param `orderIdSeed` and `orderId` is at fault? I ran my tests in the iOS simulator.

Thanks for enabling populating `quoteId` from within `approve`, let me know how to proceed.

swansontec:
Switchain is in the app, so we apparently merged this manually.

PR #1863, opened 2020-03-23
Fix on Logout and auto-logout errors on certain components

amanalili:
For Review, see comments

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android


src/__tests__/__snapshots__/RootReducer.test.js.snap
```diff
@@ -81,6 +81,7 @@ Object {
     "plugins": Array [],
   },
   "exchangeRates": Object {},
+  "isLoggedOut": true,
```
swansontec: Please rename this to `loginSceneStarting`, or something like that. This has more to do with component state than core account state, so the name is confusing.

src/__tests__/__snapshots__/RootReducer.test.js.snap
```diff
@@ -130,7 +131,7 @@ Object {
         "exchangeRates": Object {},
       },
       "fioAddress": Object {
-        "expiration": 2019-01-01T08:00:00.000Z,
+        "expiration": 2018-12-31T16:00:00.000Z,
```
swansontec: This shouldn't have changed. Are you working on top of the latest develop?

src/components/scenes/LoginScene.js
```diff
@@ -28,7 +28,8 @@ type DispatchProps = {
   deepLinkHandled(): void,
   initializeAccount(account: EdgeAccount, touchIdInfo: Object): void,
   logout(): void,
-  showSendLogsModal(): void
+  showSendLogsModal(): void,
+  loggedOut(): void
```
swansontec: Rename `logggedOut()` to `sceneStarted()`

src/types/reduxActions.js
```diff
@@ -157,6 +157,7 @@ export type Action =
     }
   | { type: 'IS_CHECKING_HANDLE_AVAILABILITY', data: boolean }
   | { type: 'LOGOUT', data: { username?: string } }
+  | { type: 'LOGGED_OUT' }
```
swansontec: Rename to `LOGIN_SCENE_STARTED`

src/components/scenes/LoginScene.js
```diff
@@ -44,6 +45,10 @@ class LoginSceneComponent extends Component<Props, State> {
     slowlog(this, /.*/, global.slowlogOptions)
   }
 
+  componentDidMount () {
+    this.props.loggedOut()
```
swansontec: I could be wrong, but it seems like this should only fire one time, when the app launches. What happens if you log in & log out a bunch of times? Does the login scene get re-mounted each time?

src/components/Main.ui.js
```diff
@@ -199,498 +190,511 @@ export class MainComponent extends Component<Props> {
...
-                <Tabs
-                  key={Constants.EDGE}
-                  swipeEnabled={false}
-                  navTransparent={true}
-                  tabBarPosition={'bottom'}
-                  showLabel={true}
-                  tabBarStyle={styles.footerTabStyles}
-                >
-                  <Stack key={Constants.WALLET_LIST} icon={this.icon(Constants.WALLET_LIST)} tabBarLabel={WALLETS}>
+            {this.props.isLoggedOut && (
```
swansontec: There is too much indentation happening here, and has been for a while. Now is a good time to fix it, instead of making it worse!

Please move this block of code into its own method, `renderMainApp` or something, and then this line can be `{this.props.isLoggedOut && this.renderMainApp() }`.

src/__tests__/__snapshots__/RootReducer.test.js.snap
swansontec: This has turned into a merge conflict, but a quick rebase should solve it.

src/types/reduxActions.js
```diff
@@ -157,6 +157,7 @@ export type Action =
     }
   | { type: 'IS_CHECKING_HANDLE_AVAILABILITY', data: boolean }
   | { type: 'LOGOUT', data: { username?: string } }
+  | { type: 'SHOW_MAIN_APP' }
```
swansontec: We have another section for actions with no payload, up above. Please move this up there, sorted alphabetically.

src/components/scenes/LoginScene.js
swansontec: Verified that it does, in fact, re-fire.

PR #1865, opened 2020-03-23
Fix on flip input amount and placeholder on android

amanalili:
Task: Send Screen - Place fiat currency symbol next to input on android - https://app.asana.com/0/361770107085503/1164052871025755

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android

### DIFF
--- a/src/modules/UI/components/FlipInput/styles.js
+++ b/src/modules/UI/components/FlipInput/styles.js
@@ -99,7 +99,8 @@ export const top = StyleSheet.create({
     fontSize: scale(15),
     color: THEME.COLORS.WHITE,
     fontFamily: THEME.FONTS.SYMBOLS,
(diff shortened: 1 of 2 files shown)


src/modules/UI/components/FlipInput/styles.js
```diff
@@ -99,7 +99,8 @@ export const top = StyleSheet.create({
     fontSize: scale(15),
     color: THEME.COLORS.WHITE,
     fontFamily: THEME.FONTS.SYMBOLS,
-    marginRight: scale(5),
+    marginRight: scale(3),
+    marginBottom: Platform.OS === 'ios' ? 0 : 3,
```
paullinator, line 103: Margins should be scaled

PR #1883, opened 2020-04-03
Refactor of otpResetModal to use the new Airship modals

amanalili:
DO NOT PR: For review and discussion


swansontec:
Merged via #1886.

PR #1912, opened 2020-04-17
Refactored Wallet List Modal

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android


src/components/modals/WalletListModal2.js
```diff
@@ -0,0 +1,250 @@
...
+
+export type StateProps = {
+  activeWalletIds: Array<string>,
+  mostRecentWallets: Array<MostRecentWallet>
+}
+
+type OwnProps = {
+  bridge: AirshipBridge<GuiWallet | Object | null>,
+  headerTitle: string,
+  wallets: Array<GuiWallet>,
```
swansontec, line 21: Shouldn't this one just come from Redux?

src/components/modals/WalletListModal2.js
```diff
@@ -0,0 +1,250 @@
...
+import { scale } from '../../util/scaling.js'
+import { type AirshipBridge, AirshipModal } from './modalParts.js'
+
+export type StateProps = {
+  activeWalletIds: Array<string>,
+  mostRecentWallets: Array<MostRecentWallet>
+}
+
+type OwnProps = {
+  bridge: AirshipBridge<GuiWallet | Object | null>,
```
swansontec, line 19: When will this modal return `Object`? Shouldn't we have a proper Flow type for this case, whatever it is?

src/components/modals/WalletListModal2.js
```diff
@@ -0,0 +1,250 @@
...
+  headerTitle: string,
+  wallets: Array<GuiWallet>,
+  createWalletCurrencies: Array<Object>,
+  showCreateWallet: boolean,
```
swansontec, line 23: Can we make `createWalletCurrencies` optional? Actually, can't this just be `allowedCurrencyCodes`, minus any wallet types we already have?

src/components/modals/WalletListModal2.js
```diff
@@ -0,0 +1,250 @@
...
+type Record = {
+  walletItem: GuiWallet | null,
+  createWalletCurrency: Object | null,
+  mostRecentUsed?: boolean,
+  currencyCode?: string | null,
+  headerLabel?: string
+}
+
+type FlatListItem = {
+  item: Record,
```
swansontec, line 38: We have a new `FlatListItem` type in `src/types/types.js` that takes a parameter, so you can use that one as `FlatListItem<Record>`.

src/connectors/components/WalletListModalConnector2.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+
+import { connect } from 'react-redux'
+
+import type { StateProps } from '../../components/modals/WalletListModal2.js'
+import { WalletListModal } from '../../components/modals/WalletListModal2.js'
+import { getActiveWalletIds } from '../../modules/UI/selectors.js'
+import type { State } from '../../types/reduxTypes.js'
+
+export const mapStateToProps = (state: State): StateProps => {
```
swansontec, line 10: Please move the whole connector into the `modals/WalletListModal2.js` file. I am trying to get rid of the separate connector files, since they make the program really hard to read with all the flipping back & forth.

src/modules/Settings/selectors.js
```diff
@@ -125,9 +125,15 @@ export const getPluginInfo = (state: State, type: string): EdgeCurrencyInfo => {
   return currencyInfo
 }
 
-export const getSupportedWalletTypes = (state: State) => {
-  const allCurrencyInfos = getPlugins(state).allCurrencyInfos
-
+export type SupportedWalletTypes = {
```
swansontec, line 128: Please look at src/util/CurrencyCodeHelpers.js, which already has this exact logic.

src/components/modals/WalletListModal2.js
```diff
@@ -1,25 +1,28 @@
...
+  wallets: { [string]: GuiWallet },
   activeWalletIds: Array<string>,
-  mostRecentWallets: Array<MostRecentWallet>
+  mostRecentWallets: Array<MostRecentWallet>,
+  allCurrencyInfos: Array<EdgeCurrencyInfo>
 }
 
 type OwnProps = {
-  bridge: AirshipBridge<GuiWallet | Object | null>,
+  bridge: AirshipBridge<GuiWallet | TokenSelectObject | SupportedWalletTypes | null>,
```
swansontec, line 27: Thanks, this looks much, much better. However, we have a `GuiWalletType` in src/util/CurrencyInfoHelpers.js which is the exact same as your new `SupportedWalletTypes` type, so please use the existing one instead.

PR #1923, opened 2020-04-21
Add Transfer Swap Plugin

passabilities:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [X] Tested on small Android
- [ ] n/a


swansontec:
In order to solve the merge conflicts, can you rebase & squash your commits, so the whole branch is just one clean commit on top of develop.

env.example.json
```diff
@@ -60,6 +60,9 @@
   "X_COINSWITCH_INIT": {
     "apiKey": ""
   },
+  "x_TRANSFER_INIT":{
```
swansontec, line 63: Actually, since the plugin takes no options, we don't actually need this.

src/util/corePlugins.js
```diff
@@ -81,7 +81,8 @@ export const swapPlugins = {
   godex: ENV.GODEX_INIT,
   // shapeshift: ENV.SHAPESHIFT_INIT,
   switchain: ENV.SWITCHAIN_INIT,
-  totle: ENV.TOTLE_INIT
+  totle: ENV.TOTLE_INIT,
+  transfer: ENV.TRANSFER_INIT
```
swansontec, line 85: Here, we can just set this to `true`.

yarn.lock
```diff
@@ -1579,6 +1579,11 @@
   resolved "https://registry.yarnpkg.com/@types/node/-/node-11.11.6.tgz#df929d1bb2eee5afdda598a41930fe50b43eaa6a"
   integrity sha512-Exw4yUWMBXM3X+8oqzJNRqZSwUAaS4+7NdvHqQuFi/d+synz++xmX3QIf+BFqneW8N31R8Ky+sikfZUXq07ggQ==
 
+"@types/node@^8.0.24":
```
swansontec, line 1582: Since nothing changed in package.json, there shouldn't be lockfile changes. Can you revert this whole file?

src/components/modals/WalletListModal.js
```diff
@@ -29,7 +29,8 @@ type OwnProps = {
   showWalletCreators: boolean,
   excludedTokens: Array<string>,
   noWalletCodes: Array<string>,
-  disableZeroBalance: boolean
+  disableZeroBalance: boolean,
+  filter?: string
```
swansontec, line 33: @amanalili has a branch which completely refactors the wallet list modal to take much simpler props, like `supportedCurrencyCodes` / `blockedCurrencyCodes`. I'm pretty sure his PR will conflict with this change, so I'm not sure what the correct order of merging & rebasing is here.

paullinator: @swansontec Arts changes actually remove the filter preventing two identical currency codes from being used in the exchange scene. This basically makes @passabilities changes to wallet list modal unnecessary. Lets wait for Art's changes to get merged and this branch can be rebased on top of it.

PR #1926, opened 2020-04-23
Add a promotion settings screen

swansontec:
I am tired of hearing that promotions / affiliations "aren't working", so I made a screen to show us what is going on.

Screenshot_20200422-174940_Edge

#### PR Requirements

- [ ] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


src/reducers/AccountReducer.js
```diff
@@ -37,7 +37,7 @@ const accountInner: Reducer<AccountState, Action> = combineReducers({
         return { ...referral, promotions }
       }
       case 'ACCOUNT_SWAP_IGNORED': {
-        return { ...state, ignoreAccountSwap: true }
+        return { ...state, ignoreAccountSwap: action.data }
```
paullinator, line 40: This does not look relevant to promotion settings. Was this a bug fix?

swansontec: No, it's a new feature. I added the ability to cancel the user preferred swap setting and go back to the affiliated one, but then I realized that was stupid & unnecessary & removed the UX but left the action.

PR #1929, opened 2020-04-23
Fix set fio addresses when no fio wallets

andreyvEze:


src/modules/FioAddress/action.js
```diff
@@ -25,17 +25,24 @@ export const refreshAllFioAddresses = () => async (dispatch: Dispatch, getState:
...
-  dispatch({
-    type: 'FIO/SET_FIO_ADDRESSES',
-    data: { fioAddresses }
-  })
+    dispatch({
+      type: 'FIO/SET_FIO_ADDRESSES',
+      data: { fioAddresses }
+    })
+  } else {
+    window.requestAnimationFrame(() => {
```
paullinator, line 39: Why not just use `window.requestAnimationFrame` around the dispatch on line 34 therefore always using it ?

andreyvEze: yeah, we can use it always, just weird thing and not a beautiful solution as for me
will update it now

PR #1934, opened 2020-04-25
FIO Address connect wallets fixes

andreyvEze:
* Remove console logs. Labels changed.
* Add 5 per once limit. 
* Refresh pub addresses when fioAddress added.
* FIO Request list check for denomination
* Make FIO address lower case for request/send fields


src/components/scenes/FioConnectWalletScene.js
```diff
@@ -129,7 +133,7 @@ export class FioConnectWalletScene extends Component<Props, State> {
   render () {
     const { notConnectedWallets } = this.props
     const { selectedWalletsMap, fioWalletLoading } = this.state
-    const continueDisabled = !Object.keys(selectedWalletsMap).length
+    const continueDisabled = !Object.keys(selectedWalletsMap).length || Object.keys(selectedWalletsMap).length > 5
```
paullinator, line 136: Not sure I like the limitation of only allowing 5 wallets to be selected at a time. Why did you add this?

src/components/scenes/FioDisconnectWalletScene.js
```diff
@@ -79,6 +80,10 @@ export class FioDisconnectWalletScene extends Component<Props, State> {
     if (selectedToRemove[wallet.key]) {
       delete selectedToRemove[wallet.key]
     } else {
+      if (Object.keys(selectedToRemove).length > 4) {
```
paullinator, line 83: Same comment as above. Dislike the select wallet limitation.

src/locales/en_US.js
```diff
@@ -722,7 +722,9 @@ const strings = {
   fio_wallet_connect_return: 'Do not remove',
   fio_show_connected_wallets: 'Show connected',
   fio_connect_no_wallets: 'There are no wallets',
-  fio_connect_checkbox_text: 'I acknowledge that once connected, addresses from these wallets will become public',
+  fio_connect_only_5_per_once: 'You can connect up to 5 wallets per once',
```
paullinator, line 725: Same comment as above. Dislike the select wallet limitation. As well, this string is grammatically incorrect.

src/modules/FioRequest/components/FioRequestRow.js
```diff
@@ -167,7 +168,13 @@ const mapStateToProps = (state: State, ownProps: OwnProps) => {
       fiatAmount: ''
     }
   }
-  const displayDenomination = getDisplayDenomination(state, fioRequest.content.token_code)
+  let displayDenomination = null
+  try {
+    displayDenomination = getDisplayDenomination(state, fioRequest.content.token_code)
+  } catch (e) {
+    displayDenomination = { name: fioRequest.content.token_code, multiplier: '0' }
```
paullinator, line 175: Do not create inline objects in `mapStateToProps`

src/components/scenes/FioConnectWalletScene.js
ericbutz: The FIO call to add addresses limits a user to adding 5 at time. This was done to prevent abuse of "feeless" call because it is bundle eligible. So, you can add more than 5, but only 5 at a time right now.

PR #1938, opened 2020-04-28
Fio request modal ux change

andreyvEze:


src/components/scenes/RequestScene.js
```diff
@@ -79,11 +80,14 @@ export type State = {
...
   encodedURI: string,
-  minimumPopupModalState: CurrencyMinimumPopupState
+  minimumPopupModalState: CurrencyMinimumPopupState,
+  isFioMode: boolean
 }
 
 export class Request extends Component<Props, State> {
   amounts: ExchangedFlipInputAmounts
+  flipInput: RefObject | null = null
+  inputAccessoryViewID: string = 'cancelHeaderId'
```
peachbits, line 90: This is a constant and never edited so it should be outside the class

src/components/scenes/RequestScene.js
```diff
@@ -386,11 +417,11 @@ export class Request extends Component<Props, State> {
     if (this.amounts) {
       const native = parseFloat(this.amounts.nativeAmount)
```
peachbits, line 418: remove const declration and se bns.lte(native, '0') in if

src/components/scenes/RequestScene.js
```diff
@@ -228,6 +233,17 @@ export class Request extends Component<Props, State> {
     this.onCloseXRPMinimumModal()
   }
 
+  onNext = () => {
+    if (this.state.isFioMode) {
+      this.setState({ isFioMode: false })
+      this.fioAddressModal()
+    }
```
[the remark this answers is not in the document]
andreyvEze: I think we don't need else here. There was no onNext before, so if isFioMode === false the behavior stays the same as was before

src/components/scenes/RequestScene.js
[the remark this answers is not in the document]
andreyvEze: done

src/components/scenes/RequestScene.js
andreyvEze: done

PR #1944, opened 2020-04-30
Promotional links (uri) with currency to wallet address

amanalili:
Task: Add addresses to promo card URLs - https://app.asana.com/0/361770107085503/1173119912688929

Task: 
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android


src/components/common/ReferalLinkActivityIndicator.js
```diff
@@ -0,0 +1,127 @@
...
+type Props = {
+  bridge: AirshipBridge<void>,
+  // The message to show in the toast:
+  message?: string,
+  // If set, the toast will stay up for the lifetime of the promise,
+  // and will include a spinner.
+  activity?: Promise<mixed>
+}
+
+class AirshipToast extends Component<Props> {
```
swansontec, line 29: We need to come up with a better name for this. `FullScreenActivity`?

amanalili: It is not the official, the indicator as of now, maybe I can just rename it as the filename?

swansontec: Yes, the file name & component should both be called `FullScreenSpinner` or something, and then move the `showFullScreenSpinner` function into services/AirshipInstance.js.

PR #1946, opened 2020-04-30
Implementation of new Wallet List Modal on all the component that use it

amanalili:
For review only: **DO NOT MERGE YET**, did not test yet all components implemented by it. Review commit by commit

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


src/components/common/WalletListModalCreateWallet.js
```diff
@@ -2,20 +2,18 @@
...
+  supportedWallet: GuiWalletType,
+  onPress(GuiWalletType): void
 }
-type State = {}
 
-class CryptoExchangeCreateWalletRow extends Component<Props, State> {
-  onPress = () => {
-    this.props.onPress(this.props.supportedWallet)
-  }
+class WalletListModalCreateWallet extends Component<Props> {
```
swansontec, line 15: I think this should be called `WalletListModalCreateRow`. This would be more consistent with the other names we have like `SettingsSwitchRow` / `SettingsHeaderRow` / `SettingsRadioRow`.

swansontec: I tried rebasing this myself, because it seemed easy, but I got a Flow error:

```js
if (showCreateWallet) {
  return <CryptoExchangeCreateWalletRow supportedWallet={createWalletCurrency || {}} ... />
```

Turns out, `|| {}` isn't a valid GuiWalletType, so I fixed the if statement instead:

```js
if (showCreateWallet && createWalletCurrency) {
  return <WalletListModalCreateRow supportedWallet={createWalletCurrency} ... />
```

swansontec: I went ahead and rebased your branch, since it seemed simple enough to do.

PR #1962, opened 2020-05-06
Renew Fio Address

andreyvEze:


src/components/scenes/FioAccountSettingsScene.js
```diff
@@ -0,0 +1,95 @@
...
+type Props = NavigationProps & StateProps
+
+export class FioAccountSettingsScene extends Component<Props, State> {
+  state: State = {
+    fioWallet: null,
+    feeLoading: false,
+    renewalFee: null
+  }
+  componentDidMount (): void {
+    this.setFees()
```
peachbits, line 39: This should be done when the user taps renew rather than when the scene loads. That way the settings screen is visible while it loads and if the user doesn't want to review it doesn't go to the network.

src/components/scenes/FioAddressDetailsScene.js
```diff
@@ -42,6 +42,11 @@ export class FioAddressDetailsScene extends Component<Props, State> {
     Actions[Constants.FIO_CONNECT_TO_WALLETS]({ fioAddressName })
   }
```
peachbits, line 43: This scene is going away. We are submitting designs to put the renew button on the wallet connect screen.

PR #1971, opened 2020-05-08
Check if update is available

passabilities:
#### PR Requirements

- [x] Tested on iOS
- [x] Tested on small Android

image


swansontec:
The modal doesn't look right. Please don't use the "centered" mode of the Airship modal, but instead use the slide-up version. Also, the modal should be "tight" to the contents, instead of expanding to fill the screen like this.

src/components/modals/UpdateModal.js
```diff
@@ -0,0 +1,75 @@
...
+              backgroundColor: THEME.COLORS.PRIMARY
+            }}
+          >
+            <Image source={edgeLogo} />
+          </View>
+        </IconCircle>
+
+        <View
+          style={{
+            padding: THEME.rem(1),
```
swansontec, line 44: Since the ladder layout already adds 1rem all around, this means we end up with 2rem on the outside. Is that intentional?

src/components/modals/UpdateModal.js
```diff
@@ -0,0 +1,75 @@
...
+            alignItems: 'center'
+          }}
+        >
+          <LadderLayout padding={THEME.rem(1)}>
+            <Text style={dayText('title')}>{s.strings.update_header}</Text>
+
+            <Text style={[dayText('center'), { lineHeight: textSize.large }]}>{s.strings.update_fresh}</Text>
+
+            <PrimaryButton style={{ width: '70%' }} onPress={() => this._onPressUpdate()}>
+              <Text style={{ color: THEME.COLORS.WHITE }}>{s.strings.update_now}</Text>
```
swansontec, line 56: This text doesn't have any size or font properties, which is a problem. It needs to use `dayText` or `nightText` as its base style.

src/components/modals/UpdateModal.js
```diff
@@ -0,0 +1,75 @@
...
+            </PrimaryButton>
+
+            <TouchableHighlight onPress={() => this._onPressLater()}>
+              <Text style={{ color: THEME.COLORS.PRIMARY }}>{s.strings.update_later}</Text>
```
swansontec, line 60: Same for this text.

src/components/modals/UpdateModal.js
```diff
@@ -0,0 +1,75 @@
...
+            </TouchableHighlight>
+          </LadderLayout>
+        </View>
+      </AirshipModal>
+    )
+  }
+
+  _onPressUpdate () {
```
swansontec, line 68: There is no need for the extra level of indirection here. Just pass the prop right into the button.

src/components/scenes/LoginScene.js
```diff
@@ -16,13 +19,16 @@ import THEME from '../../theme/variables/airbitz.js'
 import { type DeepLink } from '../../types/DeepLink.js'
 import { type Dispatch, type State as ReduxState } from '../../types/reduxTypes.js'
 import { showHelpModal } from '../modals/HelpModal.js'
+import { UpdateModal } from '../modals/UpdateModal'
```
swansontec, line 22: Local files should have the .js extension.

src/components/scenes/LoginScene.js
```diff
@@ -16,13 +19,16 @@ import THEME from '../../theme/variables/airbitz.js'
...
+import { Airship } from '../services/AirshipInstance'
 import { LoadingScene } from './LoadingScene.js'
 
 type StateProps = {
   account: EdgeAccount,
   context: EdgeContext,
   pendingDeepLink: DeepLink | null,
-  username: string
+  username: string,
```
swansontec, line 30: Can you keep these props alphabetical?

src/components/scenes/LoginScene.js
```diff
@@ -110,7 +154,8 @@ export const LoginScene = connect(
     context: state.core.context,
     account: state.core.account,
     pendingDeepLink: state.pendingDeepLink,
-    username: state.nextUsername == null ? '' : state.nextUsername
+    username: state.nextUsername == null ? '' : state.nextUsername,
+    disklet: state.core.disklet
```
swansontec, line 158: Alphabetize.

src/components/scenes/LoginScene.js
```diff
@@ -17,12 +20,15 @@ import THEME from '../../theme/variables/airbitz.js'
 import { type DeepLink } from '../../types/DeepLink.js'
 import { type Dispatch, type State as ReduxState } from '../../types/reduxTypes.js'
 import { showHelpModal } from '../modals/HelpModal.js'
-import { showError } from '../services/AirshipInstance.js'
+import {  } from '../services/AirshipInstance.js'
```
swansontec, line 23: This looks like a typo, and ESLint won't be happy.

PR #1973, opened 2020-05-08
William/remove core selectors

swansontec:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


swansontec:
I have tested this by doing some basic tapping around, visiting various scenes in the app. It seems to work, but I'm not sure how to test more deeply than that. The changes are pretty "obvious", so it should work if Flow is happy.

PR #1976, opened 2020-05-13
William/upgrade lint

swansontec:
Now that our PR list is the shortest it has been in a while, we should think about upgrading our linting toolchain.


package.json
```diff
@@ -50,19 +49,11 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "npm run precommit"
+      "pre-commit": "lint-staged"
```
peachbits, line 52: Will this run flow? The old npm run precommit does.

swansontec: Oh, yeah, we need to put this line back the way it was. Good catch.

PR #2027, opened 2020-06-03
Sign debug builds using the React Native keys

swansontec:
These keys are now part of the React Native project template, which means developers will need to do a one-time app re-install on their phones to use the new keys. After that, all future debug builds will use identical keys no matter who is doing the development. The release keys will not change.

I am submitting this separately from the React Native upgrade so we can get this out of the way now. Otherwise, switching branches is too painful.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


swansontec:
Actually, this change seems to cause more harm than good - the signing is *still* different between debug RN 59 and debug RN 62, and now we have to mess up life for RN 59 folks too. I need to do more testing.

swansontec:
Ok, this pull request actually doesn't do anything on RN 0.59 - debug builds still use the same key they were using before.

PR #2090, opened 2020-07-15
Add RNS support

alepc253:
RNS provides an architecture which enables the identification of blockchain addresses by human-readable names.
This service is part of RIF Services (RSK Infrastructure Framework) and runs on top of RSK blockchain but it's a multi blockchain naming service.

For testing, you can use `alice.rsk` that has RSK, ETH and BTC addresses already associated -in the respective wallets-.

RNS_RSK
RNS_BTC
RNS_ETH

References:
https://developers.rsk.co/rif/rns/
https://github.com/rnsdomains/rns-js


paullinator:
Due to the impact on code size and load times for react native apps, we have to decline this request since it adds web3 which is a significantly sized library. It also seems to add another dependency which has more potential to bloat our code and slow down application load times

alepc253:
I understand, thanks for checking it. We'll try again when we have a library version more react-native-friendly.

PR #2100, opened 2020-07-29
Add sweep private key mining fee popup text

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android


src/locales/en_US.js
```diff
@@ -483,6 +483,8 @@ const strings = {
   buy_crypto_modal_exchange: 'Exchange',
   exchange_crypto_modal_message: 'Your %s wallet is empty. Would you like to exchange another crypto into %s?',
   private_key_modal_sweep_from_private_address: 'Sweep Funds From Private Key',
+  private_key_modal_sweep_from_private_address_message:
```
amanalili, line 486: This process needs an on-chain transaction from the original wallet to Edge. Mining fees are applied and depend on the network status

PR #2110, opened 2020-08-06
Create multicastUtilServers function

peachbits:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/util/tracking.js
+++ b/src/util/tracking.js
@@ -4,7 +4,7 @@ import { getUniqueId } from 'react-native-device-info'
 import firebase from 'react-native-firebase'
 
 import ENV from '../../env.json'
-
+import { asyncWaterfall } from './utils.js'
 export type TrackingEvent =
   | 'ActivateWalletCancel'
   | 'ActivateWalletSelect'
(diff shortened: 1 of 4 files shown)


src/util/tracking.js
```diff
@@ -86,11 +86,23 @@ async function logToFirebase(event: TrackingEvent, values: TrackingValues) {
   }
 }
 
+export async function multicastUtilServers(path: string, options?: Object): Promise<any> {
+  const serverUrls = ['https://util2.edge.app/', 'https://util1.edge.app/']
+  let reply
+  for (const url of serverUrls) {
+    reply = fetch(`${url}${path}`, options)
```
paullinator, line 93: This should use asyncwaterfall as you do not always want to talk to both servers otherwise you will cause duplicate entries. Put in an eight second time out between each server call. That should just be a parameter to asyncwaterfall

src/util/tracking.js
```diff
@@ -86,11 +86,23 @@ async function logToFirebase(event: TrackingEvent, values: TrackingValues) {
...
+    if (reply.ok) {
+      return reply
```
paullinator, line 95: This will never get hit since reply is a promise and you do not have an await Before the call to fetch

PR #2112, opened 2020-08-09
Show explorer link to prove XMR transactions

peachbits:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/modals/TransactionAdvanceDetails.js
+++ b/src/components/modals/TransactionAdvanceDetails.js
@@ -59,6 +59,7 @@ class TransactionAdvanceDetailsComponent extends PureComponent<Props> {
 
   openProveUrl = () => {
     const { recipientAddress, txid, txSecret } = this.props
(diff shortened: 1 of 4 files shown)


src/components/modals/TransactionAdvanceDetails.js
```diff
@@ -56,6 +57,27 @@ class TransactionAdvanceDetailsComponent extends PureComponent<Props> {
     }
   }
 
+  openProveUrl = () => {
+    const { recipientAddress, txid, txSecret } = this.props
+    if (!recipientAddress || !txid || !txSecret) return
```
paullinator, line 62: Should at least show a toast so the user knows why nothing happened.

peachbits: Tile isn't visible to press if any of these three fields are null. This conditional was only to satisfy flow

PR #2116, opened 2020-08-11
William/airship upgrade

swansontec:
This upgrades the GUI to use the new Airship. The old components continue to work fine, as predicted, but this also goes ahead and ports the themed modal and the toast, since more has changed there (including blur for the modal).

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


swansontec:
Simulator Screen Shot - iPhone SE (2nd generation) - 2020-08-11 at 15 20 21

PR #2120, opened 2020-08-13
Theme comments and fixes

amanalili:
- Put other themes on light mode that is not on the dark mode
- Add comments on not used themes
- Fix on themes like keyboardTopViewBackgroundDefault to keyboardTopViewBackgroundLight

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

### DIFF
--- a/src/theme/variables/edgeDark.js
+++ b/src/theme/variables/edgeDark.js
@@ -45,6 +45,8 @@ const palette = {
 }
 
 export const edgeDark: Theme = {
+  themeId: 'edgedark',
+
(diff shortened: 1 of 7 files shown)


src/theme/variables/edgeDark.js
```diff
@@ -54,7 +54,7 @@ export const edgeDark: Theme = {
   // Icons
```
amanalili, line 54: Comment it out instead of commenting it

PR #2127, opened 2020-08-17
William/theming organization

swansontec:
This gets the GUI and the login UI on the same page, at least in terms of folder organization.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


src/theme/ThemeContext.js
```diff
@@ -1,21 +1,19 @@
...
 
-/**
- * Wrap cacheStyles with our specific Theme type.
- */
-export function cacheStyles<Styles: Object>(callback: (theme: Theme) => Styles): (theme: Theme) => Styles {
-  return cacheStylesInner(callback)
-}
+// Provide the theme context methods:
+const themeContext = makeThemeContext(edgeDark)
+export const { ThemeProvider, useTheme, withTheme, changeTheme, getTheme, watchTheme } = themeContext
```
amanalili: Suggested to re-add `cacheStyles` to be re-added to the exports so all Theme related is can be imported all using this file

amanalili (approved):

src/theme/ThemeContext.js
swansontec: Yeah, that makes sense. Done.

PR #2133, opened 2020-08-22
William/fix settings

swansontec:
The way we have been testing scenes is wrong, since the shallow renderings just have the output of `render()`, not what actually goes on screen. Fix that for the settings screen. 

The next commit has no visual changes, as verified by the unit test.

#2127 is now rebased on top of this.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a


src/components/scenes/SettingsScene.js
```diff
@@ -183,11 +184,17 @@ class SettingsOverviewComponent extends React.Component<Props, State> {
...
             icon={<FontAwesomeIcon name="user-o" color={theme.icon} size={iconSize} />}
-            text={`${s.strings.settings_account_title_cap}: ${this.props.username}`}
+            text={`${s.strings.settings_account_title_cap}: ${account.username}`}
           />
           <SettingsRow
-            text={s.strings[this.props.lockButton]}
-            right={<IonIcon name={this.props.lockButtonIcon} color={theme.iconTappable} size={iconSize} />}
+            text={isLocked ? s.strings.settings_button_unlock_settings : s.strings.settings_button_lock_settings}
+            right={
+              isLocked ? (
```
amanalili, line 192: Suggested `<IonIcon name={isLocked ? "ios-lock" : "ios-unlock"  color={theme.iconTappable} size={iconSize} />`

swansontec: Good catch, fixed.

PR #2139, opened 2020-08-25
Add optional deep link to launch the app with

peachbits:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/env.example.json
+++ b/env.example.json
@@ -65,6 +65,7 @@
   "DEBUG_CORE_BRIDGE": false,
   "USE_FAKE_CORE": false,
   "USE_FIREBASE": true,
+  "YOLO_DEEP_LINK": null,
   "YOLO_PASSWORD": null,
   "YOLO_USERNAME": null
 }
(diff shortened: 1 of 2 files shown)


env.example.json
```diff
@@ -65,6 +65,7 @@
   "DEBUG_CORE_BRIDGE": false,
   "USE_FAKE_CORE": false,
   "USE_FIREBASE": true,
+  "YOLO_DEEP_LINK": "",
```
swansontec, line 68: The default value should be null, according to your logic.

src/components/services/DeepLinkingManager.js
```diff
@@ -5,6 +5,7 @@ import * as React from 'react'
 import { Linking } from 'react-native'
 import { connect } from 'react-redux'
 
+import { YOLO_DEEP_LINK } from '../../../env.json'
```
swansontec, line 8: This needs to be `import ENV from '../../../env.json'`. According to the spec, JSON files only have a single default export, not named exports. Yes, the loosey-goosey RN packager accepts your syntax, but it technically shouldn't.

PR #2150, opened 2020-08-30
Fix bugs on flip input focus and also always flip input on top when there is no exchange rates

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

### DIFF
--- a/src/modules/UI/components/FlipInput/FlipInput2.ui.js
+++ b/src/modules/UI/components/FlipInput/FlipInput2.ui.js
@@ -201,7 +201,16 @@ export class FlipInput extends React.Component<Props, State> {
 
     if (this.props.isFocus) {
       setTimeout(() => {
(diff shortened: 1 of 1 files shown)


src/modules/UI/components/FlipInput/FlipInput2.ui.js
```diff
@@ -201,9 +201,22 @@ export class FlipInput extends React.Component<Props, State> {
...
-        this.textInputBack && this.textInputBack.focus()
+        if (this.state.isToggled) {
+          this.textInputBack && this.textInputBack.focus()
+        } else {
+          this.textInputFront && this.textInputFront.focus()
+        }
       }, 650)
     }
+
+    const { exchangeSecondaryToPrimaryRatio } = this.props
```
amanalili, line 212: consolidated only 204 to 219

PR #2158, opened 2020-09-03
William/upgrade airship

swansontec:
This fixes the problem where autologout would leave various modals still visible.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


swansontec (commented):
Verbal review with @paullinator.

PR #2167, opened 2020-09-10
Fix bug on flip input selection on android

amanalili:
#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android


src/modules/UI/components/FlipInput/FlipInput2.ui.js
```diff
@@ -80,6 +80,11 @@ export type FlipInputOwnProps = {
 
 type Props = FlipInputOwnProps
 
+const resetInputCursor = {
```
amanalili, line 83: defaultInputCursor

PR #2190, opened 2020-10-02
Create CloseButtonModal component for themed modal. Apply to WalletListMenuModal

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

### DIFF
--- a/src/components/themed/CloseModalButton.js
+++ b/src/components/themed/CloseModalButton.js
@@ -1,0 +1,27 @@
+// @flow
+
+import React, { PureComponent } from 'react'
+import { TouchableWithoutFeedback, View } from 'react-native'
+import Entypo from 'react-native-vector-icons/Entypo'
+
+import { type AirshipBridge } from '../modals/modalParts'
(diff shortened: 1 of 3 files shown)


src/components/themed/CloseModalButton.js
```diff
@@ -0,0 +1,27 @@
...
+import Entypo from 'react-native-vector-icons/Entypo'
+
+import { type AirshipBridge } from '../modals/modalParts'
+import { type ThemeProps, withTheme } from '../services/ThemeContext.js'
+
+type OwnProps = {
+  bridge: AirshipBridge<null>
+}
+
+class CloseModalButtonComponent extends PureComponent<OwnProps & ThemeProps> {
```
swansontec, line 14: Please see https://github.com/EdgeApp/edge-login-ui/blob/76aa2e21ad9a0045ea8de344b80118004ab89622/packages/edge-login-ui-rn/src/components/themed/ModalParts.js#L24. We should try to keep the GUI and login UI in sync as much as possible, so use that instead.

swansontec (changes requested):

PR #2237, opened 2020-11-09
Don't initiate onKeyPress for events triggered by tapping a number

peachbits:
Tapping a numeric key on some keyboards, like Samsung Keyboard, will create an event in addition to onChangeText. This causes onKeyPress to be called twice and duplicate the entered value. Fix is to prevent events with a 0-9 key from calling onKeyPress


peachbits:
New commit making the change android only. Tested with ios and android sims as well as affected samsung device.

paullinator (approved):

PR #2258, opened 2020-11-25
William/fix header

swansontec:
#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


paullinator (approved):

paullinator:
The changes look good although some quick comments

1. @amanalili Is making significant changes to the header for theming in the UI2 branch. I would ask holding off on merging this to see if this can get re-based into that branch instead

2. In the future, pretty please make any re-factoring and renaming/moving of files in a separate comment from actual edits to the file. Makes it infinitely easier to see what changed 🙏🏻

swansontec:
@paullinator I think we should merge this right away, because it solves a real visual issue in the app. Otherwise I wouldn't have spent any time on this detour.

Also, this was less of a "move" and more of a new-from-scratch component, using the existing code as a reference. So, I think the commit reflects that methodology, but I could certainly set up the commits so it diffs with one of the old files.

paullinator:
OK. I will merge. I'll just have art cherry pick the change into his branch so it's not to deal with the merge conflict later. No need to make the diff commit now. I already saw the diff by doing a manual rollback

PR #2263, opened 2020-11-27
ui2: Added sliding tutorial on the first 2 login of the user

amanalili:
NOTE: still missing image for the modal. Would like to suggest gif for the image showing how it works
Dependent on `art/touchAndSlide-sendReceive` branch

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android


src/util/tutorial.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+import type { EdgeAccount } from 'edge-core-js'
+
+const TUTORIAL = 'tutorial.json'
+
+export const getUserLoginCount = async (account: EdgeAccount): Promise<number> => {
+  const { disklet, id } = account
```
paullinator, line 7: do not use the account disklet.

this setting should be per device and local only

src/util/tutorial.js
```diff
@@ -0,0 +1,21 @@
...
+  try {
+    const userData = JSON.parse(await disklet.getText(TUTORIAL))
+    return userData[id] ? parseInt(userData[id].loginCount) : 0
```
paullinator, line 10: there is no need to index by account id. the disklet on the account is already account specific

src/util/tutorial.js
```diff
@@ -0,0 +1,21 @@
...
+  } catch (error) {
+    console.log(error)
+    return 0
+  }
+}
+
+export const setUserLoginCount = async (account: EdgeAccount, loginCount: number): Promise<void> => {
+  const { disklet, id } = account
+  const data = JSON.stringify({ [id]: { loginCount } })
+  await disklet.setText(TUTORIAL, data)
```
paullinator, line 20: the tutorial json file needs to be made more flexible for use by other tutorial screens. The way you architected it will make it only work for this one wallet sliding tutorial

paullinator (changes requested):

src/util/tutorial.js
amanalili: fixed

src/util/tutorial.js
amanalili: fixed

src/util/tutorial.js
amanalili: fixed

PR #2265, opened 2020-11-27
FIO. Add transaction after domain transfer.

andreyvEze:


paullinator (commented):
will this work if a user logs in on a new device? we need to be able to query this transaction off the network

andreyvEze:
@paullinator I'm not sure. For now history nodes don't have these transactions (when transferring domain), need to be fixed on the history node side, so it is a temporary fix. Should it pull it from your server if I save it like this?

paullinator:
This should be done on the plugin side. The GUI should only add metadata to the transaction. The creation of the actual transaction and saving in the plug-in database should happen inside the plug-in at the time the transaction is broadcast.

PR #2266, opened 2020-11-28
William/rn63 fixes

swansontec:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


src/components/scenes/FioRequestListScene.js
```diff
@@ -492,15 +492,15 @@ export class FioRequestList extends React.Component<Props, LocalState> {
...
+      (index > 0 && formatDate(new Date(this.state.fioRequestsPending[index + 1].time_stamp)) !== formatDate(new Date(fioRequest.time_stamp)))
     return <FioRequestRow fioRequest={fioRequest} isLastOfDate={isLastOfDate} onSelect={this.selectPendingRequest} />
   }
 
   renderSent = (itemObj: { item: FioRequest, index: number }) => {
     const { item: fioRequest, index } = itemObj
     const isLastOfDate =
       index + 1 === this.state.fioRequestsSent.length ||
-      (index > 0 && formatDate(new Date(this.state.fioRequestsSent[index + 1].time_stamp), true) !== formatDate(new Date(fioRequest.time_stamp), true))
+      (index > 0 && formatDate(new Date(this.state.fioRequestsSent[index + 1].time_stamp)) !== formatDate(new Date(fioRequest.time_stamp)))
```
paullinator, line 503: why the change to monthShort false?

paullinator (changes requested):

src/components/scenes/FioRequestListScene.js
swansontec: This code is trying to see if two things fall on the same day. Rather than using math, they are short-cutting by converting to strings (yuck). So, the comparison ends up working the same with the date being short or long, since `'Tue === Tue'` just the same as `'Tuesday' === 'Tuesday'`, but the longer form seems safer.

PR #2271, opened 2020-12-02
Fio Sent Request Details scene themed

andreyvEze:
Pixel 3
Снимок экрана 2020-12-03 в 18 39 26


paullinator:
The header is not centered

andreyvEze:
@paullinator yes, I know. that's because there is no right button. we need to fix it in header component, or is there some solution I did not find? if no, should I fix this in this PR?

paullinator:
@amanalili can you comment on how to properly center the header?

PR #2286, opened 2020-12-09
FIO Address Register Select Wallet scene themed

andreyvEze:
Снимок экрана 2020-12-08 в 01 27 26


andreyvEze:
add fio address icon themed

PR #2287, opened 2020-12-09
Fio Address List scene themed.

andreyvEze:
fioaddresslist


src/components/scenes/FioAddressListScene.js
```diff
@@ -85,127 +84,98 @@ export class FioAddressListScene extends React.Component<Props> {
...
+
+    const out: StateProps = {
+      fioAddresses,
+      fioDomains,
+      fioWallets,
+      fioPlugin,
+      loading,
+      isConnected: state.network.isConnected
+    }
+    return out
```
andreyvEze, line 175: remove out, just return object

PR #2296, opened 2020-12-14
Ui2 animated numbers

j0shwheeler:
## Currently In Progress 
Things to look into for next coding session: 
- Do we deal with any denominations that may not use commas in the currency format? 
- What are all of the types of supported denominations that can be displayed? Look into this...
- Why do we check for fiatSymbol.length on line 52. Does that length determine the type of denomination? 
- Animated component does not play nicely with EdgeText, so make it behave like an edgeText (first idea that comes to mind...)

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


paullinator:
To answer the above questions 

1. Yes, if a user is in a different region like Europe, they may use periods as the thousands separator and a comma as the decimal separator. 
2. Not sure what the second question is asking.
3. This check is just to save space. If the fiat symbol is long then we only print the currency code. Otherwise we print both of the symbol and currency code. 
4. EdgeText is just a wrapper that provides default styling. You're right, we should create an EdgeSpinningNumber that adds the same default styling. Note that we are migrating the name of EdgeText to ThemedText, so you can just go straight to naming it ThemedSpinningNumber.

j0shwheeler:
Just discovered this library breaks when using floating point values. 
image

Instead of this library will be opting for https://github.com/browniefed/react-native-ticker
Which has floating points and various currencies shown to be working in its online demos.

paullinator:
good find. thanks

j0shwheeler:
I worked around this build issue last weekend, though am running into it again which updating my dependencies. Noting this here to investigate later: 
image

Building project with the recommended node version in the readme (v.14.5.0) 
Though it looks like the project is also using a dependency called "scrypt" that is outdated and no longer maintained. 
Looks like it will need to be updated in future updates to this project. 

https://github.com/joepie91/scrypt-for-humans/issues/7

paullinator:
@j0shwheeler  try node 12. we started noticing issues with 14

j0shwheeler:
Thanks @paullinator, 
I was able to build with node 14 using yarn though will keep node 12 in mind for the future. 

Getting closer! 
This new dependency worked out, tested it for USD and Euro numbers. By default it will only spin on the numbers when given arbitrary strings. It also allows for custom spins between arbitrary characters with its "Tick" component. 

I attached a gif of what it looks like so far. 

Just something basic working for now. 

Next steps I see are to:
1. Refactor the ticker component into a component called "ThemedSpinningNumbers"
2. Used the ThemedSpinningNumbers component in all areas where we want spinning text. Not 100% clear yet on all the places we want it to spin on exactly. Will figure that out with you on the dev slack channel @paullinator, hoping to get this ready to merge into the main branch on the next iteration of updates. Ideally before the holidays start. 
3. We may or may not like the easing of the animations. If it's too fast or choppy, we can fork the dependency and update it to tweak any animation parameters. (IMHO the react-native-animated-numbers dependency had a smoother animation and I liked the bounce)
4. Test everything in iOS and Android using some real test transactions to find any possible bugs

j0shwheeler:
Getting closer though ran into a limitation of this ticker dependency as well. 
Getting to a point where I think it might be easier to write a custom implementation of this ticker animation for us using these libraries as a source of ideas for how it could work for us. 

In my last updated I wrapped this into a ticker component. 

Will need to work out why this animation is breaking and determine a solution. 

BrokenNumbersAnimation

ios/Podfile
```diff
@@ -19,4 +19,5 @@ target 'edge' do
 
   # Edge additions
   pod 'RNVectorIcons', :path => '../node_modules/react-native-vector-icons'
+  pod 'RNReanimated', :path => '../node_modules/react-native-reanimated'
```
j0shwheeler, line 22: Added this to podfile in order to overcome compilation errors in iOS

src/components/scenes/TransactionListScene.js
```diff
@@ -217,14 +218,10 @@ export class TransactionList extends React.Component<Props, State> {
                           <T numberOfLines={1} style={[styles.currentBalanceBoxBits, styles.symbol]}>
                             {displayDenomination.symbol + ' '}
                           </T>
-                          <T numberOfLines={1} style={[styles.currentBalanceBoxBits, styles.symbol]}>
-                            {cryptoAmountString}
-                          </T>
+                          <ThemedTicker style={[styles.currentBalanceBoxBits, styles.symbol]}>{cryptoAmountString}</ThemedTicker>
```
j0shwheeler, line 221: ThemedTicker does not honor the "numberOfLines" prop. 
Could that cause any issues down the road? Look into this

src/components/themed/ThemedTicker.js
```diff
@@ -0,0 +1,37 @@
...
+  ellipsizeMode?: string,
+  durations?: number,
+  [x: string]: any
+}
+
+export const ThemedTicker = ({ children, style, duration, ...props }: Props) => {
+  const theme = useTheme()
+
+  // TODO: The nature of the Ticker compoenent does not apply textProps to the string as a whole so these propds dont work.
+  // May not be able to have the text automatically adjust to fit with this ticker. If this is needed, may need more creative solution.
```
j0shwheeler, line 21: Edgetext autosizes to fit its container width and will honor the number of lines prop. 
ThemedTicker does not offer this out of the box. 

Determine if this will end up being an issue or not.

src/components/themed/ThemedTicker.js
```diff
@@ -0,0 +1,37 @@
...
+  const defaultTickerStyle = {
+    color: theme.primaryText,
+    fontFamily: theme.fontFaceDefault,
+    fontSize: theme.rem(1)
+  }
+
+  return (
+    <Ticker {...props} textProps={textProps} textStyle={[defaultTickerStyle, style]} duration={duration || 500}>
+      {children}
+    </Ticker>
```
j0shwheeler, line 35: Two issues with the themed ticker: 

1. The ticker dependency itself had to be modified in node_modules in order to work, 
which means we will need to fork the repo for this module and apply a custom update to work around a bug with it. 

2. There is a noticable lag time for the numbers to render in my simulator. User has to wait about a second for the numbers inside this component to render. This lag time might be much quicker on real devices. Test this further. Might need to update the forked library to improve performance of the initial render.

src/components/scenes/TransactionListScene.js
paullinator: should be fine. I don’t imagine using spinning numbers on multiple lines

src/components/themed/ThemedTicker.js
paullinator: should be ok

src/components/themed/ThemedTicker.js
paullinator: the simulator is pretty slow when using a debug build. try a release build

src/components/themed/ThemedTicker.js
j0shwheeler: You sure we wont have any people on the wallet app the are investing 23423857349573094875390875340985? :p 

Just kidding, that makes sense! Appreciate your quick feedback! will move forward with this information

src/components/scenes/TransactionListScene.js
```diff
@@ -121,8 +122,43 @@ class TransactionListComponent extends React.PureComponent<Props, State> {
 
   emptySection = () => [{ title: s.strings.transaction_list_search_no_result, data: [] }]
 
-  renderEmptyComponent = () =>
-    this.props.numTransactions ? <EmptyLoader /> : <BuyCrypto walletId={this.props.selectedWalletId} currencyCode={this.props.selectedCurrencyCode} />
```
j0shwheeler, line 125: TODO: This file did not get merged completely right after the last merge conflict resolution. This file should be using the BuyCrypto component to stay in sync with the ui2 branch.

PR #2320, opened 2020-12-30
Update Banxa icon and limits

peachbits:


src/modules/Logs/api.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-const API_PREFIX = 'https://info1.edgesecure.co:8444/v1/'
+const API_PREFIX = 'https://logs1.edge.app:8444/v1/'
```
paullinator, line 3: not part of this pull request

as well remove the port for logs1

paullinator (changes requested):

src/modules/Logs/api.js
peachbits: My branches are polluting each other. Removed.

PR #2327, opened 2021-01-04
Remove unsafe method in TransactionList

wolverineks:
no visual changes

verified functionality on ios simulator

- [X] transactions are updated when wallet changes


swansontec (approved):
Approved. We can merge after a quick rebase onto develop.

PR #2328, opened 2021-01-04
Remove unsafe method from PasswordReminderModal

wolverineks:
no visual changes

verified on ios simulator

- [X] password reminder state loaded is updated


swansontec:
This is no longer needed, because Art re-wrote the whole modal.

PR #2331, opened 2021-01-05
Remove unsafe method from SendConfirmation

wolverineks:
no visual changes

not sure what this code does or how to test that it didn't break anything, but this code looks to me like it doesnt change anything

tested on android emulator
- [X] spend max updates
- [X] change fee updates
- [X] flip input still flips
- [X] crypto amount updates
- [X] fiat amount updates

### DIFF
--- a/src/components/scenes/SendConfirmationScene.js
+++ b/src/components/scenes/SendConfirmationScene.js
@@ -89,7 +89,6 @@ type SendConfirmationRouterParams = {
 type Props = SendConfirmationStateProps & SendConfirmationDispatchProps & SendConfirmationRouterParams
 
 type State = {|
-  secondaryDisplayDenomination: GuiDenomination,
   nativeAmount: string,
   overridePrimaryExchangeAmount: string,
   forceUpdateGuiCounter: number,
@@ -107,11 +106,6 @@ export class SendConfirmation extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props)
     this.state = {
-      secondaryDisplayDenomination: {
-        name: '',
-        multiplier: '1',
-        symbol: ''
-      },
       overridePrimaryExchangeAmount: '',
       keyboardVisible: false,
       forceUpdateGuiCounter: 0,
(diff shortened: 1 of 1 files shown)


src/components/scenes/SendConfirmationScene.js
```diff
@@ -223,6 +218,7 @@ export class SendConfirmation extends React.Component<Props, State> {
     const displayAddress = fioAddress ? '' : address
 
     const feeCalculated = !!networkFee || !!parentNetworkFee
+    const showSpinner = !(feeCalculated || errorMsg || nativeAmount === '0' || !this.state.showSpinner || !this.props.pending)
```
swansontec, line 223: If we invert this, we get `const showSpinner = !feeCalculated && !errorMsg && nativeAmount !== '0' && this.state.showSpinner && this.props.pending`, but before it was `... this.state.showSpinner || this.props.pending`, so we have an `&&` turned to a `||`, which is a change in logic.

swansontec (changes requested):
I think I found a bug? Let me know if you agree. In any case, I would like to rebase onto the latest develop before merging.

wolverineks:
so, i took another stab at this and it's changed behavior somehow
to me, it looks like it should work, but it doesnt

typing in '$1' gets rewritten as '$0.99'
is that a bug anyone has seen before?

wolverineks:
ok, so `SecondaryDisplayDenomination` and the spinner state were moved into render, and the forceGuiCounter stuff was moved into `componentDidMount`

the forceGuiCounter stuff looks like it should work directly in render, but apparently it doesnt

this version looks like it works

swansontec (approved):
This version looks much better, thanks.

swansontec:
@wolverineks I merged this, but:

1. The PR was never rebased onto to the latest develop.
2. Hence, the unit tests broke after the merge.

I have fixed the unit tests directly on develop, but I should have insisted on a rebase before merging this. Oops.

wolverineks:
> @wolverineks I merged this, but:
> 
> 1. The PR was never rebased onto to the latest develop.
> 2. Hence, the unit tests broke after the merge.
> 
> I have fixed the unit tests directly on develop, but I should have insisted on a rebase before merging this. Oops.

whoops, sorry 'bout that

PR #2350, opened 2021-01-14
Add keysOnlyMode to disable request scene, exchange to, and wallet creation for WAX

peachbits:


src/components/scenes/RequestScene.js
```diff
@@ -250,54 +250,59 @@ export class Request extends React.Component<Props, State> {
...
+            ) : null}
+
+            <View style={styles.qrContainer}>
+              <QrCode data={this.state.encodedURI} size={qrSize} />
+            </View>
+            <RequestStatus requestAddress={requestAddress} addressExplorer={addressExplorer} />
           </View>
-          <RequestStatus requestAddress={requestAddress} addressExplorer={addressExplorer} />
-        </View>
+        )}
```
paullinator, line 299: This should have an `else` case that tells the user the asset has been deprecated. ie

"Support for WAX has been deprecated. To continue using this wallet, extract the wallet private key and import it into a WAX supporting wallet."

paullinator (changes requested):

PR #2351, opened 2021-01-15
Fix csv export

peachbits:


src/components/scenes/TransactionsExportScene.js
```diff
@@ -210,7 +214,7 @@ class TransactionsExportSceneComponent extends React.PureComponent<Props, State>
     if (isExportCsv) {
       files.push({
         contents: csvFile,
-        mimeType: 'text/csv',
+        mimeType: 'text/comma-separated-values',
```
paullinator, line 217: Please make sure this change works on iOS. If so this PR is approved and can be merged

paullinator (changes requested):

peachbits:
The additional commit to remove the message from iOS has been thoroughly tested on my device. When I tested initially, I didn't realize that iOS uninstalled the mail app (to make room for an OS upgrade that I did to troubleshoot why I couldn't connect to it in xcode) which is why it wasn't appearing in the share dialog with or without `message: ''` included in the share options.

PR #2360, opened 2021-01-19
ui2: Search List Fixes

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

### DIFF
--- a/src/components/themed/EdgeTextField.js
+++ b/src/components/themed/EdgeTextField.js
@@ -1,20 +1,26 @@
 // @flow
 
 import * as React from 'react'
+import { TouchableOpacity, View } from 'react-native'
 // $FlowFixMe = OutlinedTextField is not recognize by flow
 import { type TextFieldProps, OutlinedTextField, TextField } from 'react-native-material-textfield'
+import AntDesignIcon from 'react-native-vector-icons/AntDesign'
 
 import { unpackEdges } from '../../util/edges.js'
-import { type ThemeProps, withTheme } from '../services/ThemeContext.js'
+import { type Theme, type ThemeProps, cacheStyles, withTheme } from '../services/ThemeContext.js'
 
 type EdgeTextFieldProps = {
-  fieldRef?: ?React.ElementRef<typeof OutlinedTextField>,
   marginRem?: number | number[]
 }
 
-type Props = EdgeTextFieldProps & ThemeProps & TextFieldProps
+type EdgeOutlinedTextFieldProps = {
+  fieldRef?: ?React.ElementRef<typeof OutlinedTextField>,
+  marginRem?: number | number[],
+  isClearable: boolean,
+  onClear: () => void
+}
 
-class EdgeTextFieldComponent extends React.PureComponent<Props> {
+class EdgeTextFieldComponent extends React.PureComponent<EdgeTextFieldProps & ThemeProps & TextFieldProps> {
   render() {
     const { theme, marginRem = 0.5, ...rest } = this.props
     const margin = unpackEdges(marginRem)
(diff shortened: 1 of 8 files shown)


paullinator:
please break up the last commit into several commits. One per fix

amanalili:
> please break up the last commit into several commits. One per fix

Break up what can be logically be broken down

src/components/themed/TransactionListTop.js
```diff
@@ -176,7 +176,7 @@ class TransactionListTopComponent extends React.PureComponent<Props, State> {
                 <TouchableOpacity onPress={this.clearText} style={styles.searchClearIcon}>
                   <AntDesignIcon name="close" color={theme.icon} size={theme.rem(1)} />
                 </TouchableOpacity>
-                <TouchableOpacity onPress={this.disabledTextFieldFocus} style={styles.searchDoneButton}>
+                <TouchableOpacity onPress={this.disableTransactionSearching} style={styles.searchDoneButton}>
```
paullinator: Should be named `handleSearchDone`

src/components/themed/WalletListHeader.js
```diff
@@ -101,7 +101,7 @@ class WalletListHeaderComponent extends React.PureComponent<Props, State> {
               <TouchableOpacity onPress={this.clearText} style={styles.searchClearIcon}>
                 <AntDesignIcon name="close" color={theme.icon} size={theme.rem(1)} />
               </TouchableOpacity>
-              <TouchableOpacity onPress={this.disabledTextFieldFocus} style={styles.searchDoneButton}>
+              <TouchableOpacity onPress={this.disableWalletSearching} style={styles.searchDoneButton}>
```
paullinator: Should be named `handleSearchDone`

src/components/themed/EdgeTextField.js
```diff
@@ -1,20 +1,26 @@
...
-  fieldRef?: ?React.ElementRef<typeof OutlinedTextField>,
   marginRem?: number | number[]
 }
 
-type Props = EdgeTextFieldProps & ThemeProps & TextFieldProps
+type EdgeOutlinedTextFieldProps = {
+  fieldRef?: ?React.ElementRef<typeof OutlinedTextField>,
+  marginRem?: number | number[],
+  isClearable: boolean,
+  onClear: () => void
```
paullinator, line 20: This callback should not be needed. The clear button should simply clear the text value which would automatically call the `onChangeText` callback.

paullinator (changes requested):

src/components/themed/EdgeTextField.js
amanalili: This one is a bit tricky because of how the Transaction List search handles the filter. Transaction List only filters when the keyboard search button is pushed or the Input is blurred. It is not the same as the Wallet List where every type you do it filters the list. Why the Transaction List filters like that is because getTransaction filter is expensive as you discussed before. And suggest this might be the most optimal way.

Now, here are the scenarios when pressing the clear button if it only changes the input state and no callback:
- Keyboard is up and you type stuff, you press clear anytime. It just erase the text, the list is not affected until you dismiss the keyboard. No problem
- Keyboard is down, the list is filtered already, and you press clear. This will clear the text BUT IT WON'T RE-FILTER the list. Meaning to clear the filter again, you need to open up keyboard/focus the input and press search button to drop the keyboard, thus trigger the filter after.

No callbacks are viable if the Transaction List is filtered per type on the keyboard. Or if the keyboard down scenario is ok

src/components/themed/TransactionListTop.js
amanalili: Fixed

src/components/themed/WalletListHeader.js
amanalili: Fixed

paullinator (approved):

paullinator:
Resolve conflicts

amanalili:
> Resolve conflicts

Resolved

paullinator (approved):

PR #2368, opened 2021-01-19
Use HiddenMenuButtons in wallet list row.

andreyvEze:


andreyvEze:
@thehobbit85 not sure about last commit, let's discuss on a call next time

PR #2372, opened 2021-01-20
Animated Numbers

j0shwheeler:
### WORK IN PROGRESS
(will remove this note once feature is ready for review) 

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


paullinator:
please re-base on top of develop first to eliminate the extraneous commits

ios/Podfile
```diff
@@ -19,4 +19,5 @@ target 'edge' do
 
   # Edge additions
   pod 'RNVectorIcons', :path => '../node_modules/react-native-vector-icons'
+  pod 'RNReanimated', :path => '../node_modules/react-native-reanimated'
```
j0shwheeler, line 22: Note: Added this as the react-native-ticker needs to use the RNReanimated lib. Without this compilation fails

j0shwheeler:
Thanks for pointing that out, 
I did a git pull as rebasing was putting me on a long journey of resolving merge conflicts which was faster for me to resolve with a pull merge though I was not thinking about how it was stacking up so many commits. 

I squashed together all of the commits so that way this doesn't mess with the commit history much.

paullinator:
Sorry to be a pain, but could you re-factor the PR to not have any merge commits. Our standard workflow is to have feature branches only rebased on develop and any merge commits only go from a feature branch into develop. Without this workflow, it can be very difficult to review a PR because a merge commit from develop into your feature branch can easily have a change that goes unnoticed.

j0shwheeler:
Got it, good to know and np! I acknowledge the need for standards like this. Will remember that for the future and make that update.

package.json
```diff
@@ -173,7 +174,6 @@
     "react-native-swipe-gestures": "^1.0.2",
     "react-native-swipe-list-view": "^3.2.5",
     "react-native-tcp": "git://github.com/EdgeApp/react-native-tcp.git#rn62",
-    "react-native-ticker": "3.2.0",
```
paullinator: Why is this repo removed? Isn't this required?

paullinator (changes requested):

package.json
paullinator: I cloned your branch and tried to build and run but I definitely got the missing library error. I added this back and rebuilt and the app runs but I get the following error

Screen Shot 2021-01-19 at 23 50 11

j0shwheeler: Thanks for taking the time to check this :) 
In retrospect I should have put this up as a Draft PR as this is indeed not ready for prime time. I ran into that error as well. 

To address where the first question: 
Repo was removed because it was referencing the repo in the npm artifactory not our forked repo. 

I pointed the react-native-ticker to the following to fix some things locally: 
git://github.com/j0shwheeler/react-native-ticker#build_fixes

In order to fix that react-native ticker issue. There are two known issues with this repository that popped up while testing that I need to fix: 
1) First issue, that measurement.height property. The value measurement object is being determined dynamically by the ticker library based on its child components and it is buggy. I am going to patch our forked repo in order to resolve this
2) Second issue, Running into build issues when importing react-native-ticker into edge-react-gui. It is a typescript project that compiles properly when it is published to npm. Though compiling it locally when including it in the project is causing an issue. I've tried compiling it in a few different ways though there were different issues with each method. Easiest way to resolve this I think would be to actually publish this repo to npm with our own edge namespace. Though I think that I can still get it to work having to publish it just need to fiddle with it a bit more. 

Once those issues are completed and the react-native-ticker is back in action I will be able to clean up that ThemedTicker component and get it in all the places we want and finally do a manual QA on iOS and Android and request a final review!! 

@paullinator

j0shwheeler: Quick update (1) is fixed. The UI now does not error when measurement is undefined. It waits for the height and width properties to be available before it attempts to render using those values now. 
https://github.com/EdgeApp/react-native-ticker/pull/3/files (still WIP)

Next opportunity I have this week, I will fix the build issues and then just keep pressing forward. 
The new UI looks awesome btw

@paullinator

paullinator: Great to hear. Happen to have any video demos you can share?

j0shwheeler: Yep! Here is a screen recording. 
UI2RotatingNums

paullinator: looks great !

package.json
```diff
@@ -189,6 +191,7 @@
     "@babel/core": "^7.8.4",
     "@babel/preset-env": "^7.6.0",
     "@babel/runtime": "^7.8.4",
+    "@react-native-community/toolbar-android": "^0.1.0-rc.2",
```
j0shwheeler, line 194: Note sure if we need toolbar-android, added this to fix a compilation error related to react-native-vector-icons

package.json
j0shwheeler: Thanks, 
Made more progress today. 

All of the implementation code in the PR should be good to go pending review. 
Testing on my iOS simulator checks out. 

Still need to write unit tests and try this app out on a real android and iOS device. 

I made the numbers spin only for the large numbers on top of the wallet list and transaction list screens. 

The other smaller numbers had issues so did not convert those. 

The screenshot in this comment shows what happens when the EdgeText is converting to a spinning number for the numbers on the individual wallets. ThemedTicker does not support the adjustFontSizeToFit property. 
Screen Shot 2021-02-01 at 7 14 44 PM

That said, its getting closer. 

One question - 
Do you want the .lock files included in the commits or should they be removed? 

Ready for feedback @paullinator

paullinator: Hi Josh. Awesome to hear things are looking good. Definitely commit the lock files.I will take a look at the PR this week

package.json
paullinator: What compilation error are you seeing? Nobody else has brought up issues

paullinator:
The commit history seems to go back-and-forth adding and then removing and fixing code. It looks like the changes can be compressed down to two commits. One that changes the package and another that actually implements the spinning numbers. If that is too difficult, it would be fine to re-base this and squash everything into one commit since it is not that big a change. However the significant changes to the lock files are concerning. It is as if you deleted the lock files and then reran yarn and pod install. Please try to revert the changes to the lock files and re-run yarn, and pod install to see what actually changes and commit those changes.

paullinator (changes requested):
See comments

j0shwheeler:
@paullinator thanks for your feedback, working on wrapping up these tasks tonight. 
I will re-write the git history to be one commit at the end when I believe that all updates are complete and this is ready for a final review

j0shwheeler:
@paullinator 
Cleaned up the code and re-tested on the ios-simulator. Its all looking good now I believe. Reverted the lock files. There is one last PR for react-native-ticker to merge. https://github.com/EdgeApp/react-native-ticker/pull/5

Once that is merged I can re-run yarn install and pod install to update the lock files in this PR. 

After that everything should be good to go, just need to do a final round of testing on some real devices.

paullinator:
I reviewed and commented on your PR. Looks simple enough but it is missing the corresponding changes to the lock file

https://github.com/EdgeApp/react-native-ticker/pull/5

PR #2373, opened 2021-01-20
Fix Android export transactions

peachbits:
Remove extension when sharing as base64 string
Share .qbo as file rather than base64 string
Upgrade react-native-share


src/components/scenes/TransactionsExportScene.js
```diff
@@ -239,7 +239,12 @@ class TransactionsExportSceneComponent extends React.PureComponent<Props, State>
   }
 
   async shareAndroid(title: string, file: File): Promise<void> {
-    const url = `data:${file.mimeType};base64,${base64.stringify(utf8.parse(file.contents))}`
+    const directory = RNFS.ExternalCachesDirectoryPath
+    let url = `data:${file.mimeType};base64,${base64.stringify(utf8.parse(file.contents))}`
+    if (file.mimeType === 'application/vnd.intu.qbo') {
+      url = `file://${directory}/${file.fileName}`
+      await RNFS.writeFile(`${directory}/${file.fileName}`, file.contents, 'utf8')
```
paullinator, line 246: This will likely require additional device permissions. What is the motivation to share a file versus a base 64 string?

paullinator (changes requested):

src/components/scenes/TransactionsExportScene.js
peachbits: The required WRITE_EXTERNAL_STORAGE is already requested by the app and isn't new. 

Android is supposed to use the mimetype in the data url but it's returning null. This could be a bug with react-native-share or Android I wasn't able to narrow it down. It would produce a qbo file with the extension .qbo.null.

paullinator (approved):

PR #2380, opened 2021-01-21
Add verboseLogging toggle

peachbits:


swansontec (approved):
This just needs the core fix to avoid resetting itself.

PR #2381, opened 2021-01-21
Prevent user from entering custom fee values less than 1

peachbits:


paullinator (commented):
It feels like the logic to restrict the custom fee should be in the plug-in instead of the GUI. The GUI is making some heavy assumptions as to the key value pairs that it is restricting.

Instead, the make spend routine can throw an error that the custom fee is too low. This allows us to change the minimum on a per currency basis.

The user might not get immediate feedback but at least they are protected and any GUI built on top of our platform is also protected.

PR #2382, opened 2021-01-22
UI2 minor fixes, see description for the fix list

amanalili:
- Wallet List Scene: wallet rows token crypto amount should also be truncated
- The progress icon will still retain the spacings even when no icons are present. Also remove the loader when no icons are present
- Flip the sidemenu background. Also remove unused background options on SceneWrapper
- Make EdgeText component add \n characters to numberOfLines. This will fix the bug not seeing the limits on plugin descriptions
- Replace some exchange setting icons for dark mode
- Updated some plugin icons to use dark mode version
- Edge Text component will always use adjustsFontSizeToFit regardless of OS

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android


src/components/themed/WalletProgressIcon.js
```diff
@@ -73,7 +72,7 @@ export class WalletProgressIconComponent extends React.PureComponent<Props, Stat
         backgroundColor={theme.walletProgressIconBackground}
         rotation={0}
       >
-        {() => <Image style={iconSize} source={{ uri: icon }} />}
+        {() => (icon ? <Image style={iconSize} source={{ uri: icon }} /> : <View style={iconSize} />)}
```
paullinator, line 75: Why is there an in-line function here? This component is rendered many times on the Wallet list and will always re-render due to the in-line function

src/components/themed/EdgeText.js
```diff
@@ -16,12 +16,16 @@ class EdgeTextComponent extends React.PureComponent<OwnProps & ThemeProps> {
...
+      numberOfLines = numberOfLines + (children.match(/\n/g) || []).length
+    }
     return Platform.OS === 'ios' ? (
-      <Text style={[text, style]} numberOfLines={this.props.numberOfLines || 1} adjustsFontSizeToFit {...props}>
+      <Text style={[text, style]} numberOfLines={numberOfLines} adjustsFontSizeToFit {...props}>
         {children}
       </Text>
     ) : (
-      <Text style={[text, style]} numberOfLines={this.props.numberOfLines || 1} ellipsizeMode={this.props.ellipsizeMode || 'middle'} {...props}>
+      <Text style={[text, style]} numberOfLines={numberOfLines} ellipsizeMode={this.props.ellipsizeMode || 'middle'} {...props}>
```
paullinator, line 28: adjustsFontSizeToFit Now works on android in rn63. Please enable

paullinator (changes requested):

src/components/themed/WalletProgressIcon.js
amanalili: The inline function is how it works for that library for it to have children. I actually did try to remove that before but could not make it work. See examples at https://github.com/bartgryszko/react-native-circular-progress. In the actual source code of it actually is a function `  children: PropTypes.func`.

src/components/themed/EdgeText.js
amanalili: Created a new commit for that fix. Fixed

paullinator (approved):

PR #2383, opened 2021-01-22
Fixed incorrect multiplier being used for payout currency in exchange data

jacobburrell:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x ] Tested on iOS emulator
- [ ] n/a


jacobburrell:
@thehobbit85 I believe it would be best to fix the prop " passed in.

Currently, every coin returns mostly a destinationDenomination object with blank values except for "multiplier" which always has a value of 1.

destinationDenomination.multiplier should be what I'm looking up via "payoutMultiplier" via settings object I pulled from TransactionDetailsScene's connector.

src/components/scenes/TransactionDetailsScene.js
```diff
@@ -251,16 +252,22 @@ export class TransactionDetailsComponent extends React.Component<Props, State> {
     ))
   }
 
-  renderExchangeData = (symbolString: string) => {
-    const { destinationDenomination, destinationWallet, edgeTransaction, guiWallet, walletDefaultDenomProps, theme } = this.props
+
+  renderExchangeData = () => {
```
thehobbit85, line 256: By removing "(symbolString: string)" it breaks the code.

thehobbit85 (changes requested):

src/components/scenes/TransactionDetailsScene.js
jacobburrell: Added back in.

paullinator: What is this change tested at all? You created a syntax error which could not possibly run

paullinator (changes requested):
See comments. As well fix the conflict by rebasing your changes

src/components/scenes/TransactionDetailsScene.js
jacobburrell: Tested and rebased ontop of Develop. 
image

paullinator:
This PR as it stands cannot be merged! There are merge commits that go from develop into this feature branch which is not allowed per our workflow

PR #2386, opened 2021-01-25
Prevent user from entering invalid custom fee

peachbits:


src/components/scenes/ChangeMiningFeeScene.js
```diff
@@ -66,8 +68,15 @@ export class ChangeMiningFee extends React.Component<Props, State> {
...
-    Actions.pop()
+    const { wallet, spendTargets = [] } = this.props
+    const testSpendInfo = { spendTargets, networkFeeOption, customNetworkFee }
+    wallet
+      .makeSpend(testSpendInfo)
+      .then(() => {
+        this.props.onSubmit(networkFeeOption, customNetworkFee)
+        Actions.pop()
+      })
+      .catch(e => showError(`${s.strings.invalid_custom_fee} ${e.message}`))
```
paullinator, line 79: Cannot assume all throws are invalid custom fees. Plugins need to set a specific e.name which you need to check for here.

paullinator (changes requested):

samholmes (approved):

PR #2388, opened 2021-01-25
Password Reminder Modal on focus height fix

amanalili:
Added a Scrollview on Password Reminder Modal messages so when keyboard is opened it will shrink to a minimum size to fix bug when modal overlaps the upper screen of the phone

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android


paullinator:
Incorrect fix for problem. Correct fix is to simply not auto focus the text field.

PR #2399, opened 2021-01-26
prevent auto creating wallets on non-new accounts

wolverineks:
no visual changes

### DIFF
--- a/src/modules/Login/action.js
+++ b/src/modules/Login/action.js
@@ -342,9 +342,17 @@ async function createCustomWallets(account: EdgeAccount, fiatCurrencyCode: strin
  */
 async function createDefaultWallets(account: EdgeAccount, fiatCurrencyCode: string, dispatch: Dispatch) {
   // TODO: Run these in parallel once the Core has safer locking:
-  await safeCreateWallet(account, 'wallet:bitcoin', s.strings.string_first_bitcoin_wallet_name, fiatCurrencyCode, dispatch)
-  await safeCreateWallet(account, 'wallet:bitcoincash', s.strings.string_first_bitcoincash_wallet_name, fiatCurrencyCode, dispatch)
-  await safeCreateWallet(account, 'wallet:ethereum', s.strings.string_first_ethereum_wallet_name, fiatCurrencyCode, dispatch)
+  if (!account.allKeys.find(({ type }) => type === 'wallet:bitcoin')) {
+    await safeCreateWallet(account, 'wallet:bitcoin', s.strings.string_first_bitcoin_wallet_name, fiatCurrencyCode, dispatch)
+  }
+
+  if (!account.allKeys.find(({ type }) => type === 'wallet:bitcoincash')) {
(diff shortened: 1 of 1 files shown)


swansontec (commented):
This seems fine to me, but it does mean that users can delete all their wallets, and the account will simply stay empty. Before, we would create the default wallets whenever the list is empty.

This could potentially cause a problem where we fail to create the wallets on first login, and then we never create the wallets, because the account is no longer "new". Before, we would recover from this case.

Maybe we could change the logic to include a check for `EdgeAccount.allKeys.length <= 2`, and then create the wallets in that case. That way, deleted wallets won't trigger new wallets to be created, but we can still recover from botched account creations.

swansontec:
I would like @paullinator to read my review summary and give a yay or nay on the proposed change, or just leave it as-is.

paullinator:
> I would like @paullinator to read my review summary and give a yay or nay on the proposed change, or just leave it as-is.

I totally agree with your summary. When servers are busy, we will frequently fail to create one or two of the wallets and it would be good to properly clean up the next time a user logs in.

wolverineks:
what about something like?
```ts
if (!account.allKeys.find(({ walletType ) => walletType === 'wallet:Bitcoin'))) 
  createWallet(account, 'BTC')
if (!account.allKeys.find(({ walletType ) => walletType === 'wallet:Ethereum')))
  createWallet(account, 'ETH')
if (!account.allKeys.find(({ walletType ) => walletType === 'wallet:BitcoinCash'))) 
  createWallet(account, 'BCH')
```

paullinator:
> what about something like?
> 
> ```ts
> if (!account.allKeys.find(({ walletType ) => walletType === 'wallet:Bitcoin'))) 
>   createWallet(account, 'BTC')
> if (!account.allKeys.find(({ walletType ) => walletType === 'wallet:Ethereum')))
>   createWallet(account, 'ETH')
> if (!account.allKeys.find(({ walletType ) => walletType === 'wallet:BitcoinCash'))) 
>   createWallet(account, 'BCH')
> ```

I am good with this suggestion, just double check the wallet types. I think they are always lowercase and Bitcoin might have a suffix

swansontec:
> what about something like?
> 
> ```ts
> if (!account.allKeys.find(({ walletType ) => walletType === 'wallet:Bitcoin'))) 
>   createWallet(account, 'BTC')
> if (!account.allKeys.find(({ walletType ) => walletType === 'wallet:Ethereum')))
>   createWallet(account, 'ETH')
> if (!account.allKeys.find(({ walletType ) => walletType === 'wallet:BitcoinCash'))) 
>   createWallet(account, 'BCH')
> ```

This works, specifically, because `allKeys` includes deleted wallets as well. So, we create each one at least once, and then never again. But, the wallet types should be double-checked for spelling, as mentioned.

swansontec (approved):
This looks good. I will merge once its rebased onto develop.

swansontec:
@wolverineks I can merge this once it's rebased onto the fresh `develop`.

PR #2404, opened 2021-01-27
Wallet list swipe row refactor

andreyvEze:


thehobbit85:
Decided not to go that route

PR #2406, opened 2021-01-28
Check for pending vouchers in the native code

swansontec:
#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a

### DIFF
--- a/ios/edge/AppDelegate.m
+++ b/ios/edge/AppDelegate.m
@@ -107,34 +107,34 @@ static void InitializeFlipper(UIApplication *application) {
 - (void)application:(UIApplication *)application
   performFetchWithCompletionHandler:(void (^)(UIBackgroundFetchResult))completionHandler
 {
-  [self fetchMessagesWithCompletion:^(NSArray *resetUsers, NSError *error) {
+  [self fetchMessagesWithCompletion:^(NSArray *problemUsers, NSError *error) {
     if (error) {
       return completionHandler(UIBackgroundFetchResultNoData);
     }
 
-    NSString *message = @"Someone is trying to reset the 2-factor for these accounts: ";
-    for (NSUInteger i = 0; i < resetUsers.count; ++i) {
+    NSString *message = @"Another device is trying to log into: ";
+    for (NSUInteger i = 0; i < problemUsers.count; ++i) {
       if (i == 0) {
-        message = [message stringByAppendingString:resetUsers[i]];
+        message = [message stringByAppendingString:problemUsers[i]];
       } else {
-        message = [message stringByAppendingFormat:@", %@", resetUsers[i]];
+        message = [message stringByAppendingFormat:@", %@", problemUsers[i]];
       }
     }
 
     dispatch_async(dispatch_get_main_queue(), ^(void) {
-      application.applicationIconBadgeNumber = resetUsers.count;
-      if (resetUsers.count == 0) {
+      application.applicationIconBadgeNumber = problemUsers.count;
+      if (problemUsers.count == 0) {
         return completionHandler(UIBackgroundFetchResultNoData);
       }
 
       UNMutableNotificationContent *content = [[UNMutableNotificationContent alloc] init];
-      content.title = @"Urgent";
+      content.title = @"Urgent Security Issue";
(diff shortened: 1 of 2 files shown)


swansontec:
This basically works, but it does have a problem - it shows reminders for users who are trying to log in on the *current* device, which the edge-login-ui repo had special logic to avoid. I need to replicate that logic, and it will be good-to-go.

paullinator:
Can all of the background a querying logic somehow be moved into core JS as a native module that runs with any app that uses core

swansontec:
All problems are fixes, and it's thoroughly tested. A notification arrived last night while I was asleep, which is great because I never used to get *any* voucher notifications before.

swansontec:
> Can all of the background querying logic somehow be moved into core JS as a native module that runs with any app that uses core

Options include moving this native code into edge-login-ui-rn, which already has native logic for the keychain, or maybe adding native code to edge-core-js, which could either be this exact logic, *or* logic to spin up an extremely simple JS environment with limited IO do do the fetch & return the messages array.

swansontec:
Screenshot_20210128-111206

ios/edge/AppDelegate.m
```diff
@@ -107,34 +107,34 @@ - (NSURL *)sourceURLForBridge:(RCTBridge *)bridge
...
-      application.applicationIconBadgeNumber = resetUsers.count;
-      if (resetUsers.count == 0) {
+      application.applicationIconBadgeNumber = problemUsers.count;
+      if (problemUsers.count == 0) {
         return completionHandler(UIBackgroundFetchResultNoData);
       }
 
       UNMutableNotificationContent *content = [[UNMutableNotificationContent alloc] init];
-      content.title = @"Urgent";
+      content.title = @"New Device Login";
```
paullinator, line 131: I would recommend keeping the "Urgent" Text to elevate the urgency

ios/edge/AppDelegate.m
```diff
@@ -247,22 +247,31 @@ - (void)fetchMessagesWithCompletion:(MessagesHandler)completionHandler {
...
+        ((NSArray *)pendingVouchers).count > 0;
+
       id otpResetPending = message[@"otpResetPending"];
-      if (otpResetPending &&
+      BOOL hasReset = otpResetPending &&
         [otpResetPending isKindOfClass:[NSNumber class]] &&
-        [((NSNumber *)otpResetPending) boolValue]) {
-        [resetUsers addObject:username];
+        [((NSNumber *)otpResetPending) boolValue];
+      printf("heya!\n");
```
paullinator, line 267: Stray logging?

paullinator (changes requested):

ios/edge/AppDelegate.m
swansontec: Oops

ios/edge/AppDelegate.m
swansontec: Maybe "Urgent Security Issue"? I want to provide more context here, before we get to the main wall of text. The bare word "Urgent" triggers my "this can be ignored" reflex, since it's so common in spam.

ios/edge/AppDelegate.m
```diff
@@ -128,9 +128,9 @@ - (void)application:(UIApplication *)application
       }
 
       UNMutableNotificationContent *content = [[UNMutableNotificationContent alloc] init];
-      content.title = @"New Device Login";
+      content.title = @"Urgent Security Issue";
       content.body = message;
-      UNTimeIntervalNotificationTrigger* trigger = [UNTimeIntervalNotificationTrigger
+      UNTimeIntervalNotificationTrigger* trigger = [UNTimeIntehervalNotificationTrigger
```
paullinator, line 133: UNTimeIntehervalNotificationTrigger (sp). This should have caused a compile error

paullinator (changes requested):

ios/edge/AppDelegate.m
swansontec: I blame the toddler for extra keystrokes.

paullinator (approved):

PR #2411, opened 2021-01-29
Show progressbar on resync

andreyvEze:


src/components/themed/WalletProgressIcon.js
```diff
@@ -36,6 +36,11 @@ export class WalletProgressIconComponent extends React.PureComponent<Props, Stat
         this.setState({ isDone: true })
       }, 5000)
     }
+    if (this.props.progress === 0) {
+      setTimeout(() => {
+        this.setState({ isDone: false })
+      }, 5000)
```
paullinator, line 42: Five seconds seems like overkill. Please set to 500 ms here as well as on line 37

src/reducers/scenes/WalletsReducer.js
```diff
@@ -185,6 +185,14 @@ const walletLoadingProgress = (state = {}, action: Action): $PropertyType<Wallet
       }
     }
 
+    case 'RESET_WALLET_LOADING_PROGRESS': {
+      if (!action.data) throw new Error('Invalid action')
+      return {
+        ...state,
+        [action.data.walletId]: 0
```
paullinator, line 192: Progress at zero does not show any progress indicator. Set to 5 instead

paullinator (changes requested):

andreyvEze:
@paullinator just as summary of these changes, e top progress-bar also would animate backwards, is it ok?

paullinator:
> @paullinator just as summary of these changes, top progress-bar also would animate backwards, is it ok?

Sure. And that is fine

thehobbit85 (approved):

PR #2423, opened 2021-02-05
Changed reset wallet progress value.

andreyvEze:


paullinator:
I don't understand. Why does setting the default a to 0 fix the problem? We should never have the progress as zero because no progress line shows up. This makes it look like the progress is complete

PR #2426, opened 2021-02-05
Added simplified Utils function to lookup settings currency multiplier

jacobburrell:
Added simplified Utils function to lookup settings currency multiplier and used it to lookup correct multiplier in exchange details.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a
- [x ] Tested on iOS emulator

### DIFF
--- a/src/components/scenes/TransactionDetailsScene.js
+++ b/src/components/scenes/TransactionDetailsScene.js
@@ -47,7 +47,8 @@ type StateProps = {
   destinationWallet?: GuiWallet,
   guiWallet: GuiWallet,
   subcategoriesList: string[],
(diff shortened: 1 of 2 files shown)


jacobburrell:
Found existing functionality in settings selectors that did what my util function did.

The problem was the currency code being passed in was being passed in lowercase. "eth" instead of "ETH".
Also:
`(currencySettings == null || currencySettings.denominations == null)` was always returning true since currencySettings.denominations is an optional parameter and is often null.

The return was to return an empty EdgeDenomination object: `[emptyEdgeDenomination]`

I changed the return to lookup the currency multiplier from settings and return that.

PR #2440, opened 2021-02-15
Throw error if there is no internet connectivity to send logs

peachbits:
Fetch function doesn't throw when there's no connectivity (like airplane mode).


paullinator (changes requested):
While this works for the case that you are in airplane mode, it doesn’t fix the issue where you have a network error. You could probably reproduce the issue by changing the URL to an invalid path.

paullinator (approved):

PR #2447, opened 2021-02-22
Fix bug on wallet list modal where if wallet is created then tries to select it after but created wallet is not yet on the redux state.

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

### DIFF
--- a/src/components/modals/WalletListModal.js
+++ b/src/components/modals/WalletListModal.js
@@ -292,7 +292,14 @@ class WalletListModalConnected extends React.Component<Props, State> {
   createAndSelectWallet = async ({ currencyCode, walletType }: CreateWalletType) => {
     try {
       const wallet = await this.createWallet(currencyCode, walletType)
-      this.props.bridge.resolve({ walletId: wallet.id, currencyCode: wallet.currencyInfo.currencyCode })
+      const checkWalletsThenSelect = () => {
+        if (this.props.wallets[wallet.id]) {
+          this.props.bridge.resolve({ walletId: wallet.id, currencyCode: wallet.currencyInfo.currencyCode })
+        } else {
+          setTimeout(checkWalletsThenSelect, 50)
+        }
+      }
+      checkWalletsThenSelect() // To avoid race condition of selecting a wallet that is not yet on the redux state
     } catch (error) {
       showError(error)
     }


src/components/modals/WalletListModal.js
```diff
@@ -292,7 +292,15 @@ class WalletListModalConnected extends React.Component<Props, State> {
   createAndSelectWallet = async ({ currencyCode, walletType }: CreateWalletType) => {
     try {
       const wallet = await this.createWallet(currencyCode, walletType)
-      this.props.bridge.resolve({ walletId: wallet.id, currencyCode: wallet.currencyInfo.currencyCode })
+      const checkWalletsThenSelect = async () => {
```
paullinator, line 295: No need to make this an async function.

src/components/modals/WalletListModal.js
```diff
@@ -292,7 +292,15 @@ class WalletListModalConnected extends React.Component<Props, State> {
...
+        if (this.props.wallets[wallet.id]) {
+          this.props.bridge.resolve({ walletId: wallet.id, currencyCode: wallet.currencyInfo.currencyCode })
+        } else {
+          await setTimeout(() => {}, 50)
```
paullinator, line 299: No need to await. Just set `checkWalletsThenSelect` as the callback of setTimeout.

paullinator (changes requested):

amanalili:
Fixed

src/components/modals/WalletListModal.js
```diff
@@ -292,7 +292,14 @@ class WalletListModalConnected extends React.Component<Props, State> {
   createAndSelectWallet = async ({ currencyCode, walletType }: CreateWalletType) => {
     try {
       const wallet = await this.createWallet(currencyCode, walletType)
-      this.props.bridge.resolve({ walletId: wallet.id, currencyCode: wallet.currencyInfo.currencyCode })
+      const checkWalletsThenSelect = () => {
+        if (this.props.wallets[wallet.id]) {
+          this.props.bridge.resolve({ walletId: wallet.id, currencyCode: wallet.currencyInfo.currencyCode })
+        } else {
+          setTimeout(() => checkWalletsThenSelect(), 50)
```
paullinator, line 299: Why do you have the anonymous function? just put checkWalletsThenSelect as the callback

paullinator (changes requested):

paullinator (approved):

PR #2481, opened 2021-03-11
Send Scene: Refactor the design Flip Input Modal component

amanalili:
#### PR Requirements

Note: Remove Modal Close Button. as the Done Button or the check button on the android keyboard does the same thing

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

Screen Shot 2021-03-11 at 7 10 55 PM

### DIFF
--- a/src/components/modals/FlipInputModal.js
+++ b/src/components/modals/FlipInputModal.js
@@ -3,7 +3,7 @@
 import { bns } from 'biggystring'
 import { errorNames } from 'edge-core-js'
 import * as React from 'react'
-import { ScrollView, View } from 'react-native'
+import { ScrollView, TouchableOpacity, View } from 'react-native'
 import { type AirshipBridge } from 'react-native-airship'
 import { MenuProvider } from 'react-native-popup-menu'
 import { connect } from 'react-redux'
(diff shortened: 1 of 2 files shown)


src/components/modals/FlipInputModal.js
```diff
@@ -92,6 +91,19 @@ class FlipInputModalComponent extends React.PureComponent<Props, State> {
 
   handleSendMaxAmount = () => this.props.updateMaxSpend(this.props.walletId, this.props.currencyCode)
 
+  renderTitle = () => {
+    const styles = getStyles(this.props.theme)
+    return (
+      <View style={styles.headerContainer}>
```
swansontec, line 96: You should try using `justifyContent: "space-between"` on this container component, so you can get rid of the `styles.spacer` view.

src/components/modals/FlipInputModal.js
```diff
@@ -164,12 +176,13 @@ class FlipInputModalComponent extends React.PureComponent<Props, State> {
...
         <ThemedModal bridge={this.props.bridge} onCancel={this.handleCloseModal}>
           <ScrollView style={{ maxHeight: this.props.theme.rem(23) }} keyboardShouldPersistTaps="handled">
-            <ModalTitle>{s.strings.string_enter_amount}</ModalTitle>
+            {this.renderTitle()}
             {this.renderExchangeRates()}
             {this.renderBalance()}
             {this.renderFlipInput()}
             {this.renderFees()}
             <PrimaryButton label={s.strings.send_confirmation_max_button_title} marginRem={1} onPress={this.handleSendMaxAmount} />
+            <ModalCloseArrow onPress={this.handleCloseModal} />
```
swansontec: Wait, isn't this supposed to go away, according to the description?

src/components/services/Services.js
```diff
@@ -73,6 +73,7 @@ export class Services extends React.PureComponent<Props> {
     return (
       <MenuProvider>
         <Main />
+        <Airship avoidAndroidKeyboard statusBarTranslucent />
```
swansontec, line 76: Oh, yeah, good fix! This makes sense.

swansontec (approved):
It looks like you could just squash the last commit into the first commit, since you add a close arrow & then remove it at the end.

You could also eliminate one view using a different CSS style, but I won't hold up the release over these two little issues. You can fix them if you feel like that's worth doing.

amanalili:
updated

PR #2483, opened 2021-03-11
Send Scene Fixes/Revisions

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android


src/components/modals/AddressModal.js
```diff
@@ -35,7 +35,8 @@ type OwnProps = {
   showPasteButton?: boolean,
   isFioOnly?: boolean,
   useUserFioAddressesOnly?: boolean,
-  checkAddressConnected?: boolean
+  checkAddressConnected?: boolean,
+  autoFocus?: boolean
```
swansontec, line 39: Why is this optional? Shouldn't the modal just do this by default?

If there is a good reason, then carry on - approved.

src/components/scenes/SendScene.js
```diff
@@ -314,18 +315,19 @@ class SendComponent extends React.PureComponent<Props, State> {
...
       return (
-        <Tile
-          type={lockInputs ? 'static' : 'touchable'}
-          title={s.strings.fio_request_amount}
-          onPress={lockInputs ? undefined : this.handleFlipinputModal}
-          body={amountSyntax}
-        />
+        <Tile type={lockInputs ? 'static' : 'touchable'} title={s.strings.fio_request_amount} onPress={lockInputs ? undefined : this.handleFlipinputModal}>
+          <EdgeText style={{ fontSize: theme.rem(2) }}>{cryptoAmountSyntax}</EdgeText>
+          {!!fiatAmountSyntax && <EdgeText>{fiatAmountSyntax}</EdgeText>}
```
swansontec, line 329: I like a more explicit version like:

`fiatAmountSyntax == null ? null : <EdgeText>{fiatAmountSyntax}</EdgeText>`

But the version you have is fine if you want to keep it.

src/components/scenes/SendScene.js
```diff
@@ -314,18 +315,19 @@ class SendComponent extends React.PureComponent<Props, State> {
         const displayAmount = bns.div(nativeAmount, cryptoDisplayDenomination.multiplier, UTILS.DIVIDE_PRECISION)
         const exchangeAmount = bns.div(nativeAmount, cryptoExchangeDenomination.multiplier, UTILS.DIVIDE_PRECISION)
         const fiatAmount = convertCurrencyFromExchangeRates(exchangeRates, selectedCurrencyCode, guiWallet.isoFiatCurrencyCode, parseFloat(exchangeAmount))
-        amountSyntax = `${displayAmount || '0'} ${cryptoDisplayDenomination.name} = (${fiatSymbol} ${fiatAmount.toFixed(2) || '0'})`
+        cryptoAmountSyntax = `${displayAmount || '0'} ${cryptoDisplayDenomination.name}`
```
swansontec, line 318: We should probably get into the habit of using `??` instead of `||` for fallback values like this. It's more explicit about what the intent it, and React Native + Flow support it.

swansontec (approved):
Approved, but with some small questions and comments if you want to address them.

amanalili:
updated

PR #2484, opened 2021-03-12
Send Scene: Added Pin Input module

amanalili:
#### PR Requirements

NOTE: When the pin error occurs upon slide this does NOT reset the slider. This is also happening on the old send confirmation scene and the code does indicate it does not reset the slider.

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

Screen Shot 2021-03-16 at 3 57 28 AM

### DIFF
--- a/src/components/scenes/SendScene.js
+++ b/src/components/scenes/SendScene.js
@@ -35,6 +35,8 @@ import { PinDots } from '../themed/PinDots.js'
 import { SelectFioAddress } from '../themed/SelectFioAddress.js'
 import { Tile } from '../themed/Tile.js'
 
+const PIN_MAX_LENGTH = 4
+
 type StateProps = {
   account: EdgeAccount,
   authRequired: 'pin' | 'none',
@@ -248,7 +250,7 @@ class SendComponent extends React.PureComponent<Props, State> {
 
   handleChangePin = (pin: string) => {
     this.props.onChangePin(pin)
-    if (pin.length >= 4 && this.pinInput) {
+    if (pin.length >= PIN_MAX_LENGTH && this.pinInput) {
       this.pinInput.current.blur()
     }
   }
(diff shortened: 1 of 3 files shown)


src/components/themed/PinDots.js
```diff
@@ -0,0 +1,41 @@
...
+import { View } from 'react-native'
+
+import { type Theme, cacheStyles, useTheme } from '../services/ThemeContext.js'
+
+export function PinDots(props: { pin: string }) {
+  const theme = useTheme()
+  const styles = getStyles(theme)
+  return (
+    <View style={styles.container}>
+      <View style={[styles.circle, props.pin.length > 0 && styles.circleFilled]} />
```
thehobbit85, line 13: Make this part a bit more 'generic'.
Use a map/loop instead of manually adding and checking each circle.

src/components/scenes/SendScene.js
```diff
@@ -235,6 +240,19 @@ class SendComponent extends React.PureComponent<Props, State> {
...
 
+  handleFocusPin = () => {
+    if (this.pinInput && this.pinInput.current) {
+      this.pinInput.current.focus()
+    }
+  }
+
+  handleChangePin = (pin: string) => {
+    this.props.onChangePin(pin)
+    if (pin.length >= 4 && this.pinInput) {
```
thehobbit85, line 251: Don't use unnamed numbers in the code (like this 4). 
Create a constant and and use that. Also pass it in as 'maxLength' for the 'TextInput' component

src/components/scenes/SendScene.js
```diff
@@ -429,6 +447,33 @@ class SendComponent extends React.PureComponent<Props, State> {
...
   }
 
+  renderAuthentication() {
+    const { authRequired, pin, theme } = this.props
+    const styles = getStyles(theme)
+
+    if (authRequired === 'pin') {
+      return (
+        <Tile type="touchable" title={s.strings.four_digit_pin} onPress={this.handleFocusPin}>
+          <PinDots pin={pin} />
```
thehobbit85, line 457: Pass in the constant from the previous commit as a prop for the PinDots component.

src/components/themed/PinDots.js
```diff
@@ -0,0 +1,41 @@
+// @flow
+
+import * as React from 'react'
+import { View } from 'react-native'
+
+import { type Theme, cacheStyles, useTheme } from '../services/ThemeContext.js'
+
+export function PinDots(props: { pin: string }) {
```
thehobbit85, line 8: Add the 'pinCodeLength' as a prop

thehobbit85 (changes requested):
All the comments have to do with 'generalizing' the component.

thehobbit85 (approved):

PR #2498, opened 2021-03-22
Implements default fee

jacobburrell:
#### PR Requirements

- [x ] Tested on iOS Emulator
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/scenes/ChangeMiningFeeScene.js
+++ b/src/components/scenes/ChangeMiningFeeScene.js
@@ -111,7 +111,7 @@ export class ChangeMiningFee extends React.Component<Props, State> {
           {customFormat != null ? this.renderCustomFee(customFormat) : null}
           {this.renderFeeWarning()}
 
-          <SwitchButton label="Make Default Setting" value={this.state.isDefault} onChange={() => this.toggleDefaultFee()} />
+          <SwitchButton label={s.strings.settings_toggle_default_fee} value={this.state.isDefault} onChange={() => this.toggleDefaultFee()} />
 
           <PrimaryButton onPress={this.onSubmit} style={styles.saveButton}>
             <PrimaryButton.Text>{s.strings.save}</PrimaryButton.Text>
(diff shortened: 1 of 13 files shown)


src/components/scenes/ChangeMiningFeeScene.js
```diff
@@ -96,6 +108,17 @@ export class ChangeMiningFee extends React.Component<Props, State> {
           {customFormat != null ? this.renderRadioRow('custom', s.strings.mining_fee_custom_label_choice) : null}
           {customFormat != null ? this.renderCustomFee(customFormat) : null}
           {this.renderFeeWarning()}
+          <View>
```
jacobburrell, line 111: Remove extra view

src/components/scenes/ChangeMiningFeeScene.js
```diff
@@ -96,6 +108,17 @@ export class ChangeMiningFee extends React.Component<Props, State> {
...
+            <View style={styles.row}>
```
jacobburrell, line 112: Turn into component which receives the props:

-  label/children
- init value
- on change (function)

src/components/scenes/ChangeMiningFeeScene.js
```diff
@@ -196,21 +216,37 @@ const rawStyles = {
...
+      currencySettings: state.ui.settings[getSelectedCurrencyCode(state)],
+      spendTargets: getGuiMakeSpendInfo(state).spendTargets
+    }
+  },
   (dispatch: Dispatch): DispatchProps => ({
-    onSubmit(networkFeeOption: string, customNetworkFee: Object, walletId: string, currencyCode?: string) {
-      dispatch(sendConfirmationUpdateTx({ networkFeeOption, customNetworkFee }, true, walletId, currencyCode))
+    // dispatch = setter
+    onSubmit(networkFeeOption: FeeOption, isDefault: boolean, customFee: JsonObject, walletId: string, currencyCode?: string) {
+      if (isDefault && currencyCode) {
```
thehobbit85, line 246: There's a bug here which prevents ever turning off the default fee option.

thehobbit85 (changes requested):
Fix the bug and also make sure that if there is a default fee, you use that when creating a tx and not 'standard'.

src/components/scenes/ChangeMiningFeeScene.js
jacobburrell: Fixed and pushed with the setting of default fee to 'none' and checking for that in the case the user toggles the default fee off.

PR #2499, opened 2021-03-22
Jacob/readme

jacobburrell:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ x] n/a


M4RcB4:
I am not setup to create a review. Here are some comments from my experience on getting the app up and running.

Cocoa Pods Install
I don't see where the user is called to do a pod install. This will need to be run inside the iOS folder.

CMake
The CMake box in the Android SDK settings was not checked by default for me. We may want to instruct them to check that box. 

=====================
**...Not as important stuff..**
=====================

**Order and Structure**
- The different steps for the two platforms are somewhat intermingled. It would be nice if they were strictly organized more by the platform you are trying to get up and running. 
- The use of line breaks(<br>) could help create some space between sections and improve seek-ability 

**Verification Steps**
- It would be helpful to have checkpoint steps in the process where a user could run command line arguments to verify the necessary versions are available in PATH.

**iOS sims and Android emus**
- README assumes the user has simulators and emulators up and running. It might be helpful if we linked or provided instructions on getting an simu/emulator up and running. At a minimum we should make it clear that these should be launch-able prior to attempting deploy or they may receive error messages that require a little bit of investigation. This is the error message that will throw if you set the --debug flag but not without it(https://prnt.sc/10svfok). 
- also had to select the Commandline Tools Version https://prnt.sc/10svyea otherwise the pod install would die here https://prnt.sc/10svzkz

README.md
```diff
@@ -70,11 +70,15 @@ Use Android Studio's Preferences to install NDK (Side by side):
...
+![Android Studio Preferences Screenshot](docs/images/android-studio-ndk-version.png)
+
 1. Open Preferences
 2. Navigate to `Appearance & Behavior -> System Settings -> Android SDK` in the sidebar
 3. Navigate to `SDK Tools` tab
-4. Check `NDK (Side by side)` checkbox
-5. Click OK or Apply and follow the download instructions.
+4. Check `Show Package Details` checkbox at the bottom right-hand corner.
+5. Check Version `21.0.6011959` under `NDK (Side by side)`
+6. Click OK or Apply and follow the download instructions.
```
M4RcB4, line 82: CMake is was not checked by default. We should instruct the user to check that as well.

README.md
```diff
@@ -70,11 +70,15 @@ Use Android Studio's Preferences to install NDK (Side by side):
```
M4RcB4, line 70: ^^^ For installing CocoaPods we should use ```brew install cocoapods```. The user will also need to run ```pod install``` prior to trying to run the iOS app.

M4RcB4 (changes requested):

README.md
paullinator: `pod install` is automatically run when the user runs `yarn`. Don't think we need to have that added to the read me

PR #2507, opened 2021-03-29
Use https://content.edge.app/ for image files

peachbits:


thehobbit85 (changes requested):
Create a constant for 'https://content.edge.app/' and use that constant instead of having the same string all over the place.

thehobbit85 (approved):

PR #2519, opened 2021-03-31
Always export cryptocurrency amount in the base denomination

peachbits:
This only affects major currencies and ignores tokens since ERC20 tokens have a single denomination.


paullinator (changes requested):
Please no. We need exports in other denominations for compatibility with Quick books

peachbits:
This is a request from taxbit since it's not possible to tell from the export what the denomination is.

paullinator:
I would instead add another column to the export which is the denomination

PR #2536, opened 2021-04-12
Change FIO Registration Error Message

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

### DIFF
--- a/src/locales/en_US.js
+++ b/src/locales/en_US.js
@@ -647,7 +647,7 @@ const strings = {
   fio_address_list_domain_register: 'Register Custom FIO Domain',
   fio_address_list_register_domain: 'Register New Domain',
   fio_get_fee_err_msg: 'There was an issue getting the FIO fee. Please try again later.',
-  fio_get_reg_info_err_msg: 'There was an issue getting registration information.',
+  fio_get_reg_info_err_msg: 'Domain does not exist',
   fio_register_address_err_msg: 'There was an issue registering an address. Please try later.',
   fio_error_invalid_address: 'Invalid FIO address',
   fio_renew_err_msg: 'There was an issue renewing your %s. Please try again later.',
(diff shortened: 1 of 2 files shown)


paullinator:
Do not merge

amanalili:
Updated

src/modules/FioAddress/util.js
```diff
@@ -13,6 +13,9 @@ import { truncateDecimals } from '../../util/utils'
 const CONNECTED_WALLETS = 'ConnectedWallets.json'
 const FIO_ADDRESS_CACHE = 'FioAddressCache.json'
 
+const INVALID_DOMAIN_ERROR_1 = 'Domain does not exist'
+const INVALID_DOMAIN_ERROR_2 = 'Domain is not registered'
+
```
thehobbit85, line 18: remove this constants

src/modules/FioAddress/util.js
```diff
@@ -618,7 +621,9 @@ const buyAddressRequest = async (
     })
 
     if (buyAddressResponse.error) {
-      console.log(buyAddressResponse.error)
+      if (buyAddressResponse.error === INVALID_DOMAIN_ERROR_1 || buyAddressResponse.error === INVALID_DOMAIN_ERROR_2) {
```
thehobbit85, line 624: How is this expected to work is the `buyAddressResponse.error` that we get from the Fio plugin is never equal to any of those strings?

Instead, the currencyInfo for Fio already exposed the possible error codes so you should use them to check what is the type of the error.

Here you can find the error codes:
https://github.com/EdgeApp/edge-currency-accountbased/blob/1724bf40a7d431fb2ee068da9138ef62723ae2ec/src/fio/fioInfo.js#L41

thehobbit85 (changes requested):
See comments

PR #2538, opened 2021-04-12
Increase tap area on text buttons

amanalili:
#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android


src/components/navigation/HeaderTextButton.js
```diff
@@ -45,6 +56,14 @@ const getStyles = cacheStyles((theme: Theme) => ({
     paddingLeft: theme.rem(1),
     paddingRight: theme.rem(2.5),
     height: 44 // This is a fixed height of the navigation header no matter what screen size. Default by router-flux
+  },
+  left: {
```
thehobbit85, line 60: Padding left 1 and right 2.5 is already the default, so you don't need another style for 'left'.
Just use left as the default placement.

src/components/navigation/HeaderTextButton.js
```diff
@@ -9,8 +9,19 @@ import { showHelpModal } from '../modals/HelpModal.js'
...
+
+export const PLACEMENT = {
+  LEFT: 'left',
+  RIGHT: 'right'
+}
+
 type Props = {
-  type: 'exit' | 'help'
+  type: typeof TYPE.EXIT | typeof TYPE.HELP,
+  placement: typeof PLACEMENT.LEFT | typeof PLACEMENT.RIGHT
```
thehobbit85, line 24: Don't use the constants, just use string literals directly.

src/components/navigation/HeaderTextButton.js
```diff
@@ -9,8 +9,19 @@ import { showHelpModal } from '../modals/HelpModal.js'
...
+export const TYPE = {
+  EXIT: 'exit',
+  HELP: 'help'
+}
+
+export const PLACEMENT = {
+  LEFT: 'left',
+  RIGHT: 'right'
+}
+
```
thehobbit85: No need to use the constants, this is exactly what typing is for.

src/components/Main.ui.js
```diff
@@ -61,7 +61,7 @@ import { BackButton } from './navigation/BackButton.js'
 import { CurrencySettingsTitle } from './navigation/CurrencySettingsTitle.js'
 import { EdgeLogoHeader } from './navigation/EdgeLogoHeader.js'
 import { handlePluginBack, renderPluginBackButton } from './navigation/GuiPluginBackButton.js'
-import { HeaderTextButton } from './navigation/HeaderTextButton.js'
+import { HeaderTextButton, PLACEMENT as HEADER_BUTTON_PLACEMENT, TYPE as HEADER_BUTTON_TYPE } from './navigation/HeaderTextButton.js'
```
thehobbit85, line 64: remove the constants:
```PLACEMENT as HEADER_BUTTON_PLACEMENT, TYPE as HEADER_BUTTON_TYPE```

File: src/components/Main.ui.js [hunk 5]
```diff
@@ -183,7 +183,7 @@ export class MainComponent extends React.Component<Props> {
                 navTransparent
                 component={ifLoggedIn(WalletListScene)}
                 renderTitle={<EdgeLogoHeader />}
-                renderLeftButton={<HeaderTextButton type="help" />}
+                renderLeftButton={<HeaderTextButton type={HEADER_BUTTON_TYPE.HELP} placement={HEADER_BUTTON_PLACEMENT.LEFT} />}
```
thehobbit85, line 186: No need to pass "placement" as left since left should be the default.

src/components/scenes/RequestScene.js
```diff
@@ -347,8 +347,14 @@ export class RequestComponent extends React.PureComponent<Props, State> {
                 <QrCode data={this.state.encodedURI} size={this.state.qrCodeContainerHeight - theme.rem(1)} />
               )}
             </View>
-            <RightChevronButton paddingRem={[0, 0, 0, 0]} onPress={this.handleAddressBlockExplorer} text={s.strings.request_qr_your_receiving_wallet_address} />
-            <EdgeText style={styles.publicAddressText}>{requestAddress}</EdgeText>
+            <TouchableOpacity onPress={this.handleAddressBlockExplorer}>
```
thehobbit85, line 350: If you are wrapping both of them, then this `RightChevronButton` should just be an icon and text, and not a button.
Otherwise, you can just make the `EdgeText` tappable.

src/components/themed/ThemedButtons.js
```diff
@@ -150,6 +150,15 @@ export function RightChevronButton(props: { text: string, onPress: () => void, p
   )
 }
 
+export function IconButton(props: Props) {
```
thehobbit85, line 153: This component is redundant, you already have the 'ButtonBox' component is the same file which is exactly the same.

thehobbit85 (changes requested):
Address comments

src/components/navigation/HeaderTextButton.js
amanalili: Remove the default on container

thehobbit85 (changes requested):
The code looks ok.
Just resolve the conflicts and add screenshots of the scenes that are affected.

thehobbit85 (approved):

PR #2543, opened 2021-04-17
Eliran/removing string duplication

thehobbit85:


swansontec (approved):
Does Flow type-check this correctly? If we do s.strings.blah, where blah doesn't exist, do we get a nice error? If so, this is a great simplification & definitely approved! Otherwise, I think we might not want to do this.

PR #2551, opened 2021-04-21
Side Menu to close drawer when on the same page

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android


thehobbit85 (approved):
I approve this since it does fix the bug. But this scene needs a complete refactor.
It makes no sense to duplicate the code for all buttons and not a component for that.

PR #2570, opened 2021-04-28
Exchange Scene, change string of the min and max amount limit

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


thehobbit85 (changes requested):
Why is there a change in the pods file?

thehobbit85 (approved):

PR #2582, opened 2021-05-04
Reorganized and Reformatted README

M4RcB4:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ X ] n/a

---

Reformatted and Reorganized the README for better readability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,15 @@ Edge Wallet is:
 # Working with the Source Code
 <br>
 
+#### __Note:__ _This guide __assumes a Mac/OSX environment.__ You can setup a windows/linux environment with small adaptations to the insturctions._
+
+<br>
+
 ## Requirements
 
-#### Note: 
+<br>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,12 +73,31 @@ Edge is known to build with this exact tool set. Updated versions of these tools
...
+<br>
+<br>
+<br>
+<br>
+
+## Getting Started - iOS Setup
+
+### Install CocoaPods
+
+    brew install cocoapods
```
thehobbit85, line 85: I know you can use both brew and gem.
Is there a specific reason you changed it from gem to brew?

thehobbit85 (changes requested):
Don't use '<br>'.
Technically, you can use html tags in markdown files but it's not recommended. 
Instead you can use the more native `&nbsp;`.

README.md
M4RcB4: Consistency. Thoughts?

PR #2605, opened 2021-05-12
Stuck spinning if all wallets are deleted

dbo1dyrev:


thehobbit85:
The code looks good, but can you just add some screenshots?
You should always add screenshots of the App when you make any changes that adds/removes/changes the design in any way.

dbo1dyrev:
Screenshot 2021-05-14 at 12 34 04

thehobbit85 (changes requested):
Actually, now that I see it, can you please split the text after the dot and add a 1 rem spacing between the first and the second line like this?

```
Edge requires at least one wallet in this account. 

If you’d like to delete this wallet, you’ll need to add an additional wallet(s) to this account.
```

I know that's not what the design showed but this looks very crowded.

dbo1dyrev:
Screenshot 2021-05-14 at 13 39 20
Like this?

paullinator:
Please change the text to remove the “(s)” at the end of “wallet”. Having this doesn’t translate well.

dbo1dyrev:
Screenshot 2021-05-17 at 10 25 19

thehobbit85 (approved):

PR #2608, opened 2021-05-13
Add “Address” option Sweep Private Key Side Menu

dbo1dyrev:


src/components/scenes/ScanScene.js
```diff
@@ -45,6 +45,12 @@ export class Scan extends React.Component<Props> {
                 <T style={styles.bottomButtonText}>{s.strings.fragment_send_flash}</T>
               </View>
             </TouchableHighlight>
+            <TouchableHighlight style={styles.bottomButton} onPress={this._onToggleAddressModal} underlayColor={THEME.COLORS.SECONDARY}>
+              <View style={styles.bottomButtonTextWrap}>
+                <IonIcon style={styles.addressBookIcon} name="ios-book" size={scale(24)} />
+                <T style={styles.bottomButtonText}>{s.strings.fragment_send_address}</T>
```
thehobbit85, line 51: The text of the button should not say "Address", instead something like "Enter key" is better.

src/components/scenes/ScanScene.js
```diff
@@ -45,6 +45,12 @@ export class Scan extends React.Component<Props> {
                 <T style={styles.bottomButtonText}>{s.strings.fragment_send_flash}</T>
               </View>
             </TouchableHighlight>
+            <TouchableHighlight style={styles.bottomButton} onPress={this._onToggleAddressModal} underlayColor={THEME.COLORS.SECONDARY}>
+              <View style={styles.bottomButtonTextWrap}>
+                <IonIcon style={styles.addressBookIcon} name="ios-book" size={scale(24)} />
```
thehobbit85, line 50: The icon should not be an address book, but instead look like the "enter" icon in the send scene, the one used in this line - https://github.com/EdgeApp/edge-react-gui/blob/b7ba44615690d09e721a40189fb517d223af376e/src/components/themed/AddressTile.js#L275

src/components/scenes/ScanScene.js
```diff
@@ -45,6 +45,12 @@ export class Scan extends React.Component<Props> {
                 <T style={styles.bottomButtonText}>{s.strings.fragment_send_flash}</T>
               </View>
             </TouchableHighlight>
+            <TouchableHighlight style={styles.bottomButton} onPress={this._onToggleAddressModal} underlayColor={THEME.COLORS.SECONDARY}>
```
thehobbit85, line 48: This button should be on the left and the flash on the right

thehobbit85 (changes requested):
1. Submit screenshots of new scene in the App that has the changes you made. 
You should always do that when you made any changes that adds/removes/changes the design in any way.

2. The modal that is opened when clicking on the button, should have "Enter Private Key" in the header and not "Enter Recipient FIO, ENS, of Public Address".

3. On the same modal, in the text box, it should say "Private Key" and not "Address".

4. The button should only be visible if we got here by selecting "Sweep private key" and not in any other case (like if trying to send money for example).

5. Address the rest of the comments next to the code.

dbo1dyrev:
Screenshot 2021-05-20 at 13 17 01
Screenshot 2021-05-20 at 13 16 53

src/components/modals/SingleInputModal.js
```diff
@@ -0,0 +1,78 @@
+// @flow
```
thehobbit85, line 1: The general idea of using a component like this is good.
I'm just wondering, can't you get the same behavior by using the "AddressModal", maybe with some minor tweaks?

thehobbit85 (changes requested):
See comments

src/modules/UI/components/ControlPanel/Component/Main.js
```diff
@@ -186,7 +189,7 @@ const SweepPrivateKeyButton = ({ onSelectWallet }) => {
     )).then(({ walletId, currencyCode }: WalletListResult) => {
       if (walletId && currencyCode) {
         onSelectWallet(walletId, currencyCode)
-        Actions.jump(Constants.SCAN, { data: 'sweepPrivateKey' })
+        Actions.jump(Constants.SCAN, { isSweepPrivateKey: true })
```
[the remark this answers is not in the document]
dbo1dyrev: I can rewrite it back, but for my opinion, boolean flags are less prone to errors such as typos. I will add string constant for 'sweepPrivateKey'

src/components/scenes/ScanScene.js
```diff
@@ -26,31 +27,40 @@ type Props = {
   walletId: string,
   currencyCode: string,
   qrCodeScanned: (data: string) => void,
-  parseScannedUri: (data: string) => Promise<void>,
+  parseScannedUri: (data: string, customErrorTitle: string, customErrorDescription: string) => Promise<void>,
```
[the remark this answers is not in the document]
dbo1dyrev: this function is a wrapper in src/connectors/scenes/ScanConnector.js,  because in this component it always should use arguments such as  `customErrorTitle ` and `customErrorDescription` I make them required

src/modules/UI/components/ControlPanel/Component/Main.js
```diff
@@ -88,13 +88,16 @@ export default class Main extends React.Component<Props> {
...
-  } else {
-    Actions.jump(scene)
+
+  if (currentScene !== scene) {
+    Actions.jump(scene, sceneProps)
+  } else if (sceneProps) {
+    Actions.refresh(sceneProps)
   }
+
+  drawerClose()
```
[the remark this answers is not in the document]
dbo1dyrev: No, I need it. Because when you try to navigate between two menu options `Sweep Private Key` and `Scan QR (Edge Login) `you get a bug, new button` Enter Key` will be visible on both of them. This is happen because `Sweep Private Key` and `Scan QR` use same `ScanScene` component and when you use `Actions.jump` nothing happens(even component rerender) that's why I use `Actions.refresh(sceneProps)`. Secondary, even if I rerender  `ScanScene` I get the same `data: 'sweepPrivateKey'` in props inside `Scan QR` route(if I previously  was on `Sweep Private Key` route).
In this case I add line `const popToSendScan = () => goToScene(Constants.SCAN, { isSweepPrivateKey: false })`

src/modules/UI/components/ControlPanel/Component/Main.js
```diff
@@ -160,7 +163,7 @@ const WalletsButton = () => {
   )
 }
 
-const popToSendScan = () => goToScene(Constants.SCAN)
+const popToSendScan = () => goToScene(Constants.SCAN, { isSweepPrivateKey: false })
```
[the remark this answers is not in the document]
dbo1dyrev: Reason described above

src/components/modals/SingleInputModal.js
dbo1dyrev: I probably can, but

`AddressModal` is very overloaded with logic and all of it about `Address`. In my case with `Private Key` all of this code not needed, even more all of this empty space on `AddressModal` I don't need too, because this space used only for filtered `fioAddresses` which were found after the user has filled in the input.

I think we should write simple, dumb and self-documenting code which will be easy to read and maintain in future

src/modules/UI/components/ControlPanel/Component/Main.js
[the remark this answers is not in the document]
thehobbit85: 👍🏼

src/components/scenes/ScanScene.js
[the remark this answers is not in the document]
thehobbit85: 👍🏼

src/components/modals/SingleInputModal.js
thehobbit85: "I think we should write simple, dumb and self-documenting code which will be easy to read and maintain in future" 

I couldn't agree more. And you are right about not using the `AddressModal`.  I missed a lot of things regarding that component.
But, part of having "easy to maintain" code, is also not having duplicated, almost identical components.
And if you kept looking in the `modals` folder, you would have found the `SendLogsModal`.

That one is almost identical to the one you are using here.
But at this point, since you already wrote the component, you might as well replace the `SendLogsModal` with the modal you wrote. Or the other way around, but do try and merge them.

src/components/modals/SingleInputModal.js
```diff
@@ -0,0 +1,78 @@
...
+
+// $FlowFixMe
```
thehobbit85, line 3: Same comment as in here: https://github.com/EdgeApp/edge-react-gui/pull/2625#discussion_r636652888

thehobbit85 (changes requested):
Updated my comments.

src/components/modals/SingleInputModal.js
dbo1dyrev: because branch about `$FlowFixMe` already merged, I use `src/utils/hooks` now

PR #2622, opened 2021-05-17
William/selector cleanup

swansontec:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


src/actions/SendConfirmationActions.js
```diff
@@ -23,7 +23,7 @@ import s from '../locales/strings.js'
 import { addToFioAddressCache, recordSend } from '../modules/FioAddress/util'
 import { getExchangeDenomination as settingsGetExchangeDenomination } from '../modules/Settings/selectors.js'
 import { type AuthType, getAuthRequired, getSpendInfo, getSpendInfoWithoutState, getTransaction } from '../modules/UI/scenes/SendConfirmation/selectors'
-import { convertCurrencyFromExchangeRates, getExchangeRate, getSelectedCurrencyCode, getSelectedWallet, getSelectedWalletId } from '../modules/UI/selectors.js'
+import { convertCurrencyFromExchangeRates, getExchangeRate } from '../modules/UI/selectors.js'
```
andreyvEze, line 26: should we move this methods to util file or somethins?

src/components/scenes/FioAddressDetailsScene.js
```diff
@@ -172,7 +171,7 @@ const getStyles = cacheStyles((theme: Theme) => ({
 }))
 
 export const FioAddressDetailsScene = connect((state: RootState) => {
-  const fioWallets = getFioWallets(state)
+  const fioWallets = state.ui.wallets.fioWallets
```
andreyvEze, line 174: could we make in this way?
```
const { fioWallets } = state.ui.wallets
```

or maybe even this: 
```
const out: StateProps = {
    fioWallets: state.ui.wallets.fioWallets
  }
  return out
```

src/components/scenes/FioSentRequestDetailsScene.js
```diff
@@ -98,7 +97,7 @@ const getStyles = cacheStyles((theme: Theme) => ({
 }))
 
 const FioSentRequestDetailsScene = connect((state: RootState) => {
-  const wallet = UI_SELECTORS.getSelectedWallet(state)
+  const wallet = state.ui.wallets.byId[state.ui.wallets.selectedWalletId]
```
andreyvEze, line 100: should we add a type here? `const wallet: GuiWallet`

src/components/scenes/TransactionDetailsScene.js
```diff
@@ -16,7 +16,7 @@ import * as Constants from '../../constants/indexConstants'
 import { formatNumber } from '../../locales/intl.js'
 import s from '../../locales/strings.js'
 import { getDisplayDenomination, getPlugins, getSettings } from '../../modules/Settings/selectors.js'
-import { convertCurrencyFromExchangeRates, convertNativeToExchangeRateDenomination, getSelectedWallet, getWallet } from '../../modules/UI/selectors.js'
+import { convertCurrencyFromExchangeRates, convertNativeToExchangeRateDenomination } from '../../modules/UI/selectors.js'
```
andreyvEze, line 19: should this methods also be in some kind of util file?

src/components/scenes/WalletListScene.js
```diff
@@ -10,7 +10,7 @@ import { updateActiveWalletsOrder } from '../../actions/WalletListActions.js'
 import XPubModal from '../../connectors/XPubModalConnector.js'
 import s from '../../locales/strings.js'
 import { getIsAccountBalanceVisible } from '../../modules/Settings/selectors.js'
-import { getActiveWalletIds, getWalletLoadingPercent } from '../../modules/UI/selectors.js'
+import { getWalletLoadingPercent } from '../../modules/UI/selectors.js'
```
andreyvEze, line 13: could we also remove this? and pass `state.ui.wallets.walletLoadingProgress` instead of state?

src/components/themed/WalletList.js
```diff
@@ -10,7 +10,7 @@ import { selectWallet } from '../../actions/WalletActions.js'
 import { WALLET_LIST_SCENE } from '../../constants/indexConstants.js'
 import s from '../../locales/strings'
 import { SYNCED_ACCOUNT_DEFAULTS } from '../../modules/Core/Account/settings.js'
-import { calculateWalletFiatBalanceUsingDefaultIsoFiat, getActiveWalletIds } from '../../modules/UI/selectors.js'
+import { calculateWalletFiatBalanceUsingDefaultIsoFiat } from '../../modules/UI/selectors.js'
```
andreyvEze, line 13: util file

src/modules/Logs/action.js
```diff
@@ -54,11 +54,7 @@ export const sendLogs = (text: string) => async (dispatch: Dispatch, getState: G
       // Wallet info
       const wallet = currencyWallets[walletId]
       if (wallet && logOutput.loggedInUser) {
-        const currencyCode = wallet.currencyInfo.currencyCode
-          ? wallet.currencyInfo.currencyCode
-          : typeof wallet.currencyCode === 'string'
-          ? wallet.currencyCode
-          : ''
+        const currencyCode = wallet.currencyInfo.currencyCode ?? ''
```
andreyvEze, line 57: so we don't need to set it to `wallet.currencyCode` if no `wallet.currencyInfo.currencyCode`?

src/modules/UI/scenes/SendConfirmation/selectors.js
```diff
@@ -6,10 +6,9 @@ import { type GuiMakeSpendInfo } from '../../../../reducers/scenes/SendConfirmat
 import { type RootState } from '../../../../types/reduxTypes.js'
 import { convertNativeToExchange } from '../../../../util/utils.js'
 import { getExchangeDenomination } from '../../../Settings/selectors.js'
-import { convertCurrency, getSceneState } from '../../selectors.js'
+import { convertCurrency } from '../../selectors.js'
```
andreyvEze, line 9: do we have a task to remove sendComfirmation scene and connector?

andreyvEze (commented):
please check comments

src/components/scenes/FioAddressDetailsScene.js
swansontec: Good catch, I will put a return type of the `mapStateToProps` method, and use destructuring.

src/actions/SendConfirmationActions.js
swansontec: I think that would make sense eventually, but for now the goal was to just get rid of the selectors.

src/components/scenes/FioSentRequestDetailsScene.js
swansontec: Flow is able to infer the type here just fine, since it's just a property lookup off of `RootState`, which has solid type definitions.

src/components/scenes/TransactionDetailsScene.js
swansontec: I agree that these don't seem like normal selectors, but my goal in this PR isn't to organize everything - just get rid of selectors.

src/components/scenes/WalletListScene.js
swansontec: Actually, I have a commit that does this, but I didn't submit it with the PR because I thought it was out-of-scope. If you want, I will throw that commit on the end of this PR so you can look at it too.

src/modules/UI/scenes/SendConfirmation/selectors.js
swansontec: Oh, dang, you are right! If the scene is going away, maybe I should do that first, and then rebase this on top.

src/components/scenes/WalletListScene.js
andreyvEze: ok, let's leave it in next PR

src/components/scenes/TransactionDetailsScene.js
andreyvEze: ok

src/actions/SendConfirmationActions.js
andreyvEze: ok

src/components/scenes/FioSentRequestDetailsScene.js
swansontec: I put in the requested type definitions when I put back `getSelectedWallet`.

src/components/scenes/WalletListScene.js
swansontec: I went ahead and added it to this PR, since the commit was ready to go.

src/modules/UI/scenes/SendConfirmation/selectors.js
swansontec: Created a separate task to get rid of the send confirmation scene.

src/modules/Logs/action.js
swansontec: There is no `EdgeCurrencyWallet.currencyCode`. This was a type error the whole time, but Flow couldn't see the problem because the selectors were messing up its type inference.

src/components/themed/WiredProgressBar.js
```diff
@@ -85,3 +75,20 @@ const getStyles = cacheStyles((theme: Theme) => ({
...
+    const walletsForProgress = state.ui.wallets.walletLoadingProgress
+    const walletIds = Object.keys(walletsForProgress)
+    if (walletIds.length === 0) return { progress: 0 }
+
+    let sum = 0
+    for (const walletId of walletIds) {
+      sum += walletsForProgress[walletId]
+    }
+    let ratio = sum / walletIds.length
+    if (ratio > 0.9999) ratio = 1
```
andreyvEze, line 90: should it be the same value as was before? `0.9999`

src/components/themed/WiredProgressBar.js
```diff
@@ -85,3 +75,20 @@ const getStyles = cacheStyles((theme: Theme) => ({
...
+    return { progress: ratio * 100 }
+  },
+  (dispatch: Dispatch): {} => ({})
```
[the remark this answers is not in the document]
swansontec: Fixed.

src/components/themed/WiredProgressBar.js
swansontec: Oops, fixed.

andreyvEze (approved):

PR #2626, opened 2021-05-18
William/delete send confirmation

swansontec:
This PR replaces #2653. It has the same code, but has been rebased into 3 logical commits.

### Fixed
- Clicking a legacy or payment protocol link will go to the new Send screen.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


swansontec:
Dennis, when you review this PR, I want you to go to the commit tab and look at the commits one-by-one:

Screen Shot 2021-06-11 at 12 25 51 PM

I have taken your changes and reorganized them so they tell a story. There is no more "wip", and there is no more back-and-forth where one commit makes a change and another commit adjusts the change. For example, the first commit adds the `// eslint-disable-next-line react-hooks/exhaustive-deps` comments in their final form, so we don't have to see the in-between `// eslint-disable-next-line` change in our history anymore. Same for the `// BORKED` comments I had added - now, the code just goes directly to a working solution in a single commit.

dbo1dyrev (approved):

PR #2629, opened 2021-05-19
Add Flow type wrappers for the React hooks

swansontec:
I haven't figured out how to test these, but the types are based on the docs.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/util/hooks.js
+++ b/src/util/hooks.js
@@ -1,0 +1,60 @@
+// @flow
+
+import * as React from 'react'
+
+type SetState<S> = (value: S | ((state: S) => S)) => void
+
+type UseCallback = <T: (...args: any[]) => any>(callback: T, deps: any[]) => T
+
+type UseContext = <T>(context: React.Context<T>) => T
+
+type UseDebugValue = <T>(value: T, format?: (value: T) => any) => void
+
+type UseEffect = (effect: () => void | (() => void), deps?: any[]) => void
+
+type UseImperativeHandle = (ref: any, init: () => any, deps?: any[]) => void
+
+type UseMemo = <T>(init: () => T, deps?: any[]) => T
+
+type UseReducer = {
+  // Normal version:
+  <State, Action>(reducer: (state: State | void, action: Action) => State, init: State | void): [State, (action: Action) => void],
+
+  // Initializer version:
+  <State, Action, Init>(
+    reducer: (state: State | void, action: Action) => State,
+    init: Init,
+    initializer: (init: Init) => State
+  ): [State, (action: Action) => void]
+}
+
+type UseRef = {
+  // Component ref:
+  <ElementType: React$ElementType>(): { current: null | React$ElementRef<ElementType> },
+
+  // Value container:
+  <T>(init: T): { current: T }
+}
+
+type UseState = <S>(init: S | (() => S)) => [S, SetState<S>]
+
+// $FlowFixMe
+export const useCallback: UseCallback = React.useCallback
+// $FlowFixMe
+export const useContext: UseContext = React.useContext
+// $FlowFixMe
+export const useDebugValue: UseDebugValue = React.useDebugValue
+// $FlowFixMe
+export const useEffect: UseEffect = React.useEffect
+// $FlowFixMe
+export const useImperativeHandle: UseImperativeHandle = React.useImperativeHandle
+// $FlowFixMe
+export const useLayoutEffect: UseEffect = React.useLayoutEffect
+// $FlowFixMe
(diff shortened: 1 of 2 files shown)


swansontec:
@thehobbit85 Bad news: This is not working at run-time for some reason. I am still investigating the cause.

swansontec:
OK, so the problem was elsewhere (the app was giving a misleading stack trace). However, I have updated this PR, since solving the crash bug forced me to make some other improvements, which are good in any case.

src/util/hooks.js
```diff
@@ -0,0 +1,60 @@
...
+
+type UseEffect = (effect: () => void | (() => void), deps?: any[]) => void
+
+type UseImperativeHandle = (ref: any, init: () => any, deps?: any[]) => void
+
+type UseMemo = <T>(init: () => T, deps?: any[]) => T
+
+type UseReducer = {
+  // Normal version:
+  <State, Action>(reducer: (state: State | void, action: Action) => State, init: State | void): [State, (action: Action) => void],
```
thehobbit85, line 21: Can you do something like this instead of duplicating the type definition?

`
<State, Action, Init = void>(
    reducer: (state: State | void, action: Action) => State,
    init: Init | State | void,
    initializer?: (init: Init) => State
  ): [State, (action: Action) => void]
  `

src/util/hooks.js
```diff
@@ -0,0 +1,60 @@
...
+  <State, Action, Init>(
+    reducer: (state: State | void, action: Action) => State,
+    init: Init,
+    initializer: (init: Init) => State
+  ): [State, (action: Action) => void]
+}
+
+type UseRef = {
+  // Value container:
+  <T>(init: T): { current: T },
```
thehobbit85, line 33: Can you merge this two into one type or having the readonly only on off them really matters?
If not, can't you do something like this? 
`type UseRef = <T>(init?: T): { +current: T | null }`

src/util/hooks.js
```diff
@@ -0,0 +1,60 @@
+// @flow
+
+import * as React from 'react'
+
+type SetState<S> = (value: S | ((state: S) => S)) => void
+
+type UseCallback = <T>(callback: T, deps: any[]) => T
```
thehobbit85, line 7: Isn't callback suppose to be a function? same for the return type?

thehobbit85 (changes requested):
See comments

src/util/hooks.js
swansontec: No, because it's not actually correct. In the case with 2 arguments, the initial value must be `State`, but in the case with 3 values, it just has to match whatever argument 3 expects. Since argument 2 has different types depending on the argument count, we need two separate versions.

src/util/hooks.js
swansontec: Yeah, it should be a function. Whatever function you pass, will get returned out unchanged (but memoized so it's `===` if no deps have changed). Thus, the `T` you pass is the `T` you get (and `T` is a function type). I could try adding a constraint that `<T: Function>`, but these constraints don't really work well in Flow (it seems to randomly ignore them).

src/util/hooks.js
swansontec: This would be incorrect, because in the first case, `current` will not be `null`, and requiring a bunch of `null` tests would be annoying. In the second case, it totally can be `null`, because React links it to the mounted / unmounted lifetime of a component. The other difference is that in case 1, the user is allowed to write to `current`, but in the second case `current` is read-only (because React manages it).

swansontec: After testing this extensively & hitting my head against a wall, I do believe that the overload is still needed, but the React side did need some changes. With these fixes, you can either do:

```
const ref: { current: null | typeof SomeComponent } = useRef()
return <SomeComponent ref={ref} />
```

Or you can do:

```
const ref: { current: string } = useRef('')
```

Getting both cases to work at the same time is tricky, and I think the overload may be the only way to do it. The overload is *certainly* necessary in Typescript, so it stands to reason Flow would need it too.

src/util/hooks.js
swansontec: I have added the type constraint. Now it has to be a function.

src/util/hooks.js
thehobbit85: That's why I was asking 👍🏼

src/util/hooks.js
thehobbit85: 👍🏼

thehobbit85 (approved):

PR #2632, opened 2021-05-19
Fixed allowed currencies array for wallet picker in fio request scene

andreyvEze:


src/components/scenes/FioRequestListScene.js
```diff
@@ -372,7 +372,7 @@ class FioRequestList extends React.Component<Props, LocalState> {
     const { content } = selectedFioPendingRequest
     const chainCode = content.chain_code.toUpperCase()
     const tokenCode = content.token_code.toUpperCase()
-    const allowedFullCurrencyCode = chainCode !== tokenCode && tokenCode && tokenCode !== '' ? [`${chainCode}:${tokenCode}`] : [chainCode]
+    const allowedFullCurrencyCode = chainCode !== tokenCode && tokenCode && tokenCode !== '' ? [tokenCode] : [chainCode]
```
thehobbit85, line 375: It looks like removing the token code is not the right solution.
If you look at this code: https://github.com/EdgeApp/edge-react-gui/blob/fdc0144c9e44f7b9c0ad606eabffe773699bb0c2/src/util/utils.js#L609

It looks the problem is using `:` instead of `-` as a delimiter. 

And I checked and looks like it was `-` for at least a couple of months if not more. So I don't know how or if it even worked until now.

thehobbit85 (changes requested):
See comment

src/components/scenes/FioRequestListScene.js
andreyvEze: @thehobbit85 with current check method it works for not fullCurrencyCode format, just chainCode or tokenCode. I tested current changes and it is working, could you provide me an example when it would not work?

thehobbit85: I honestly can't think of an example.
But, can you at least try and change it into `[`${chainCode}-${tokenCode}`]` and see if that solves the issue.
It so, I prefer keeping both `chainCode` and `tokenCode` if it's possible.

andreyvEze: `[`${chainCode}-${tokenCode}`]` works

thehobbit85 (approved):

PR #2642, opened 2021-05-22
William/upgrade lint

swansontec:
#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


src/reducers/scenes/WalletsReducer.js
```diff
@@ -169,7 +169,7 @@ const walletLoadingProgress = (state = {}, action: Action): $PropertyType<Wallet
       if (!action.data) throw new Error('Invalid action')
       const activeWalletIdList = action.data.activeWalletIds
       const activeWalletIdProgress = {}
-      activeWalletIdList.map(item => {
+      activeWalletIdList.forEach(item => {
```
thehobbit85, line 172: How about just using reduce:
```
return activeWalletIdList.reduce((r, item) => ({...r, [item]: 0 }), {})
```

thehobbit85 (changes requested):
Other than the minor suggestion I wrote in the comment, the code looks correct.

There is one thing that does bother me a bit about the new linting rules, the line length.
I think that the new rules allows for lines that are way too long. Is it something we can change? And do we even want to change it?

src/reducers/scenes/WalletsReducer.js
swansontec: I didn't want to do a big refactor on any of the code, just make ESLint happy. Also, the replacement you suggest is O(n^2), not O(n) like the current version, so it's not clearly superior.

swansontec:
> There is one thing that does bother me a bit about the new linting rules, the line length.

This is a long-standing issue, but we have been waiting for a good point in time to hit the big red button, since it will cause huge merge conflicts in all our outstanding PR's. To do it, we just delete one line in `.prettierrc.json`, and then run `yarn fix` to get the ~30K line diff.

The purpose of this PR is to get the majority of the lint rules in line with our other repos. We still aren't 100% in-line, as you have noticed, but we are vastly closer.

PR #2653, opened 2021-05-26
Delete send confirmation

dbo1dyrev:
Here I delete old SendConfirmation screen and all depended components and actions.


src/actions/ScanActions.js
```diff
@@ -149,7 +148,7 @@ export const parseScannedUri = (data: string, customErrorTitle?: string, customE
 
       if (isLegacyAddressUri(parsedUri)) {
         // LEGACY ADDRESS URI
-        return dispatch(showLegacyAddressModal())
+        // BUSTED!
```
swansontec, line 151: We cannot leave these two in a busted state.

src/actions/ScanActions.js
```diff
@@ -159,8 +158,7 @@ export const parseScannedUri = (data: string, customErrorTitle?: string, customE
 
       if (isPaymentProtocolUri(parsedUri)) {
         // BIP70 URI
-        // $FlowFixMe
-        return dispatch(paymentProtocolUriReceived(parsedUri))
+        // BUSTED!
```
swansontec, line 161: We cannot leave these two in a busted state. We need to somehow route these two URI's to the new Send scene. This is the reason we haven't already removed the SendConfirmation scene - doing so would break deep linking.

swansontec (changes requested):

src/actions/ScanActions.js
```diff
@@ -120,90 +123,105 @@ export const parseScannedUri = (data: string, customErrorTitle?: string, customE
...
-          walletId: selectedWalletId,
-          wallet: guiWallet,
-          onAddToken: noOp
-        }
-        return Actions[ADD_TOKEN](parameters)
+  try {
+    const parsedUri: EdgeParsedUri = await edgeWallet.parseUri(data, currencyCode)
+    dispatch({ type: 'PARSE_URI_SUCCEEDED', data: { parsedUri } })
+
+    console.log('parsedUri parsedUri parsedUri parsedUri parsedUri', parsedUri)
```
swansontec, line 130: Delete console.log

src/components/themed/Fade.js
```diff
@@ -27,6 +27,7 @@ const FadeComponent = ({ visible: propsVisible, hidden, children }: Props) => {
     })
   }
 
+  // eslint-disable-next-line
```
swansontec, line 30: These should be `eslint-disable-next-line rule-name`, so we know what rule is being switched off. Also, these cleanups should be in a separate commit.

swansontec (changes requested):
I think this approach works well, but I don't like the way the commits are organized. I would like to see 3 commits with these names:

- Clean up eslint warnings
- Use the new send screen for deep links
- Delete the send confirmation screen

So, this will take a bit of rebasing to get right.

thehobbit85:
Also, resolve the conflicts

swansontec (approved):
The code looks OK, but the commits are disorganized. It would be preferable to squash these, at this point, rather than merge them as-is.

paullinator (changes requested):
Agree with @swansontec on this one. Please clean up the commits before this is merged.

swansontec (changes requested):
This is still totally wrong, so I'll do the rebase myself, so we are perfectly clear about *what* a clean patch series looks like, then submit a new PR. Dennis can then review my PR, to get an idea for what we are looking for here.

swansontec:
Replaced by #2626, which has the same changes rebased into a series of standalone commits, instead of jumbled together like in this branch.

PR #2661, opened 2021-05-28
Remove split param btc segwit

axenov-it:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/modals/SendLogsModal.js
+++ b/src/components/modals/SendLogsModal.js
@@ -34,7 +34,13 @@ const SendLogsModalComponent = ({ bridge, isConnected, submitLogs }: Props) => {
   }
 
   return (
-    <SingleInputModal bridge={bridge} title={s.strings.settings_button_send_logs} label={s.strings.settings_modal_send_logs_label} onSubmit={handleSubmit} />
+    <SingleInputModal
+      bridge={bridge}
+      title={s.strings.settings_button_send_logs}
+      label={s.strings.settings_modal_send_logs_label}
+      onSubmit={handleSubmit}
+      returnKeyType="send"
+    />
   )
 }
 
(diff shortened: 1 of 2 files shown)


src/components/modals/WalletListMenuModal.js
```diff
@@ -83,16 +83,33 @@ class WalletListMenuModalComponent extends PureComponent<Props> {
...
     }
   }
 
+  checkSplitExclude() {
+    const { walletName } = this.props
+
+    if (!walletName) return false
+
+    return (
+      EXCLUDE_SPLIT_WALLETS.findIndex((item: string) => {
```
thehobbit85, line 106: Using "find" makes more sense here than "findIndex", since all you want is a "true/false" value anyway.

src/components/modals/WalletListMenuModal.js
```diff
@@ -83,16 +83,33 @@ class WalletListMenuModalComponent extends PureComponent<Props> {
...
+        const regex = new RegExp(item, 'i')
```
thehobbit85, line 107: Use "indexOf" or "includes" but not a regex.
Check out the performance differences, its massive:

Screen Shot 2021-06-08 at 20 34 16

thehobbit85 (changes requested):
See comments

src/components/modals/WalletListMenuModal.js
axenov-it: The index of method is not insensitive if I accidentally set segwit to EXCLUDE_SPLIT_WALLETS = ['segwit'] because the lowercase but wallet name is “Btc (Segwit)” . IndexOf method doesn't working for me. We can added individual param (isSegwit) to wallet for it if you diсide that it is actual.

PR #2677, opened 2021-06-07
Remove unused exchange rate data sources

peachbits:
---
  - https://app.asana.com/0/0/1200432242640024


thehobbit85 (approved):

swansontec (approved):
I am approving this, but DO NOT MERGE until we fix the app to show historical rates for old transactions with 0 fiat, since *all* transactions will have 0 fiat after this.

paullinator (approved):

PR #2705, opened 2021-06-22
William/selector cleanup

swansontec:
#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


src/modules/UI/selectors.js
```diff
@@ -47,12 +46,12 @@ export const getDefaultDenomination = (state: RootState, currencyCode: string):
...
   const currencySettings = settings[currencyCode]
   const defaultMultiplier = currencySettings.denomination
-  const denomination = _.find(currencySettings.denominations, denom => denom.multiplier === defaultMultiplier)
+  const denomination = currencySettings.denominations.find(denom => denom.multiplier === defaultMultiplier)
   if (!denomination) throw new Error('Edge: Denomination not found. Possible invalid currencyCode.')
   return denomination
 }
 
-export const getExchangeDenomination = (state: RootState, currencyCode: string, specificWallet?: GuiWallet): GuiDenomination => {
+export const getTokenExchangeDenomination = (state: RootState, currencyCode: string, specificWallet?: GuiWallet): GuiDenomination => {
```
thehobbit85: Why did you change the name?
If I understand correctly, this function returns the denom for a mainnet currency's OR a token's one.

thehobbit85 (changes requested):
See comment + Resolve conflict

src/modules/UI/selectors.js
swansontec: The problem is that there are two functions with the same `getExchangeDenomination` name, one in `UI/selectors.js` and another in `settings/selectors.js`. This one is aware or tokens, and the other doesn't seem to be, so I added "token" to the name.

swansontec: Please advise.

swansontec: I changed the name to `getPrimaryExchangeDenomination` because every single place it is used has the shape `const primaryExchangeDenomination = getPrimaryExchangeDenomination(...)`. Since the left-hand side always has the name `primaryExchangeDenomination`, that's what it should probably be called.

swansontec:
I have just rebased this to avoid merge conficts, and have addressed your comment.

PR #2707, opened 2021-06-22
Sam/Git Blame Ignore Revs

samholmes:
### Added
- .git-blame-ignore-revs file to exclude formatting commits

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,114 @@
+##
+# This file lists commits that perform bulk code-formatting operations.
+# This allows `git blame` to skip over these commits and give better results.
+#
+# To use this file, set the following configuation:
+#    git config blame.ignoreRevsFile .git-blame-ignore-revs
+#
+# Running `yarn fix` should always happen in its own commit,
+# separate from any other changes, so git can skip those cleanly.
+# Tooling changes, such as upgrading prettier or changing options,
+# should happen in the commit right before running `yarn fix`.
+#
+# When adding commits to this file, always mention the date and
+# any tooling changes from the previous commit.
+# 
+# Verify all "yarn fix" commits are present in this file:
+#   git log --all -i --grep='yarn\sfix' --pretty=%H
+##
+
+# 2021-05-24
+8b0bc2bc945073f89b05f9b064b84f11af6299c3
+
+# 2020-12-17 (Upgrade all linting devDependencies)
+17372b17e0b91ee827e339f889a3ac19253a3c73
+
+# 2020-10-09 (Auto-format README.md)
+1739e9a175ee7d3c098316220c9740c9b425477b
+
+# 2020-09-17 (Prefer simpler array syntax)
+13b995842ad952740304372d174a733d65c03640
+
+# 2020-06-17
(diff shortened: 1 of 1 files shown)


.git-blame-ignore-revs
```diff
@@ -0,0 +1,17 @@
+##
+# All "yarn fix" commit revs should be included in this file.
+#
+# Add to your local configuration:
+#   git config --global blame.ignoreRevsFile .git-blame-ignore-revs
```
swansontec, line 5: Doing it globally messes up other repos, since Git expects this file to exist everywhere. Making the config local to the repo is the only safe option.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,17 @@
...
+#   git log --all -i --grep='yarn\sfix' --pretty=%H
+##
+
+8b0bc2bc945073f89b05f9b064b84f11af6299c3
+17372b17e0b91ee827e339f889a3ac19253a3c73
+13b995842ad952740304372d174a733d65c03640
+a7ba8d46e6f2fc3041af03c0f30b4d09ace46f44
+99c56fb565b1886ad756aa296f3c0170602529fc
+da8a3a1eb5cb5aff987867631b47a36055aaae5a
+7bbc8c715947383a6240a5d17512b52dc42cb958
```
swansontec, line 17: Please add:

```
# 2018-02-13 Initial prettier formatting
c990e091589dddfbf56c3bdbe166f25b22d53d00
```

Putting a date comment above each of the other hashes might helpful too, since otherwise this is pretty opaque.

swansontec (changes requested):

.git-blame-ignore-revs
samholmes: It was defined this way in the task requirements, but you are very correct about the breakage here. I'll change the comment and also change the requirement in the task

.git-blame-ignore-revs
```diff
@@ -0,0 +1,111 @@
...
+#
+# Each entry should include a date, and a description of format config change 
+# where possible.
+#
+# Add to git config:
+#   git config blame.ignoreRevsFile .git-blame-ignore-revs
+# 
+# Verify all "yarn fix" commits are present in this file:
+#   git log --all -i --grep='yarn\sfix' --pretty=%H
+##
```
swansontec, line 15: Englilishs with betur grammers!

Also, I did line-wrapping at thought boundaries (commas or new sentences), which is a more oldschool technique:

```
# This file lists commits that perform bulk code-formatting operations.
# This allows `git blame` to skip over these commits and give better results.
#
# To use this file, set the following configuation:
#    git config blame.ignoreRevsFile .git-blame-ignore-revs
#
# Running `yarn fix` should always happen in its own commit,
# separate from any other changes, so git can skip those cleanly.
# Tooling changes, such as upgrading prettier or changing options,
# should happen in the commit right before running `yarn fix`.
#
# When adding commits to this file, always mention the date and
# any tooling changes from the previous commit.
```

swansontec (approved):
Approved. You can take my comment-block edits or leave them - I couldn't resist some English editing.

PR #2708, opened 2021-06-23
Fix negative value fiat fees on Send Scene

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android


thehobbit85:
Closing since this is not the way to solve this.

PR #2716, opened 2021-06-24
v2.0.14

peachbits:


CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
 # edge-react-gui
 
+## 2.0.14 (2021-06-24)
+
+- Show FIO renewal reminders
```
thehobbit85, line 5: This should be removed and changed with:
`Only allow for one FIO expired pop-up per login `

CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
 # edge-react-gui
 
+## 2.0.14 (2021-06-24)
```
thehobbit85, line 3: Missing:
- Logs and CSV Export TX's show's what device was used to send the transaction

thehobbit85 (changes requested):
See comments

PR #2729, opened 2021-07-02
William/constants cleanup

swansontec:
This removes the `indexConstants.js` file, and instead imports things directly, the normal way.

#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


src/components/Main.ui.js
```diff
@@ -463,18 +463,6 @@ export class MainComponent extends React.Component<Props> {
             />
           </Stack>
 
-          <Stack key={Constants.PLUGIN_EARN_INTEREST}>
```
thehobbit85, line 466: I'll admit I don't know anything about this plugin.
Are you sure we don't use/need it anywhere?

thehobbit85 (changes requested):
Resolve the conflicts and see my question.
Other than that, you can consider this Approved.

src/components/Main.ui.js
swansontec: Cred took our customer's money and ran, so Paul made a public statement that we will never integrate a custodial service again. If customers want interest going forward, they can buy some DeFi interest-bearing tokens. We can promote these tokens in the app using our normal promo-card stuff.

PR #2736, opened 2021-07-08
Fix incorrect image icon URL in Add Token row of wallet selector

peachbits:


peachbits:
verbal approval from @thehobbit85

PR #2743, opened 2021-07-12
Fix convertCurrencyWithoutState function multiplication to use Biggystring

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

### DIFF
--- a/src/components/themed/WalletList.js
+++ b/src/components/themed/WalletList.js
@@ -78,10 +78,11 @@ type Props = OwnProps & StateProps & DispatchProps & ThemeProps
 
 class WalletListComponent extends React.PureComponent<Props> {
   sortWalletList(walletList: WalletListItem[]): WalletListItem[] {
-    const getFiatBalance = (wallet: GuiWallet, fullCurrencyCode: string) => {
+    const getFiatBalance = (wallet: GuiWallet, fullCurrencyCode: string): number => {
       const { settings, exchangeRates } = this.props
       const currencyCode = getSortOptionsCurrencyCode(fullCurrencyCode)
-      return calculateWalletFiatBalanceUsingDefaultIsoFiat(wallet, currencyCode, settings, exchangeRates)
+      const fiatBalanceString = calculateWalletFiatBalanceUsingDefaultIsoFiat(wallet, currencyCode, settings, exchangeRates)
+      return parseFloat(fiatBalanceString)
     }
 
     if (this.props.walletsSort === 'name') {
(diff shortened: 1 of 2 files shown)


src/components/themed/WalletList.js
```diff
@@ -79,10 +79,11 @@ type Props = OwnProps & StateProps & DispatchProps & ThemeProps
 
 class WalletListComponent extends React.PureComponent<Props> {
   sortWalletList(walletList: WalletListItem[]): WalletListItem[] {
-    const getFiatBalance = (wallet: GuiWallet, fullCurrencyCode: string) => {
+    const getFiatBalance = (wallet: GuiWallet, fullCurrencyCode: string): number => {
       const { settings, exchangeRates } = this.props
       const currencyCode = getSortOptionsCurrencyCode(fullCurrencyCode)
-      return calculateWalletFiatBalanceUsingDefaultIsoFiat(wallet, currencyCode, settings, exchangeRates)
+      const fiatBalanceString = calculateWalletFiatBalanceUsingDefaultIsoFiat(wallet, currencyCode, settings, exchangeRates)
+      return parseFloat(fiatBalanceString)
```
thehobbit85, line 86: No need to convert back to number.
Instead, when you use this function to compare balances, just use the bns comparison functions: 

https://www.npmjs.com/package/biggystring#comparisons

thehobbit85 (changes requested):
See comment

src/components/themed/WalletList.js
amanalili: It's not use for comparison but for the Array.sort function so it is needed to have a numerical result.

src/selectors/WalletSelectors.js
```diff
@@ -53,10 +53,10 @@ export const convertCurrency = (state: RootState, fromCurrencyCode: string, toCu
   return convertedAmount
 }
 
-const convertCurrencyWithoutState = (exchangeRates: { [string]: number }, fromCurrencyCode: string, toCurrencyCode: string, amount: number = 1) => {
+const convertCurrencyWithoutState = (exchangeRates: { [string]: number }, fromCurrencyCode: string, toCurrencyCode: string, amount: string = '1'): string => {
```
peachbits, line 56: The goal is for these functions to only accept strings as arguments. For every argument that passes numbers, like exchangeRates, create a utility function to convert to strings before a function like convertCurrencyWithoutState is called. Then, update convertCurrencyWithoutState to use biggystring.

PR #2744, opened 2021-07-12
Fix convertCurrency function multiplication to use Biggystring

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


thehobbit85 (changes requested):
The point is to not use "parseFloat" but use only strings.
And like I mentioned on previous PRs, you can use this if you need to compare two strings:

https://www.npmjs.com/package/biggystring#comparisons

thehobbit85 (approved):

thehobbit85 (approved):

PR #2754, opened 2021-07-19
William/outlined text field

swansontec:
This synchronizes the outlined text field with edge-login-ui-rn.

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/themed/EdgeOutlinedField.js
+++ b/src/components/themed/EdgeOutlinedField.js
@@ -1,14 +1,13 @@
 // @flow
 
-// $FlowFixMe = forwardRef is not recognize by flow?
-import React, { forwardRef, useRef } from 'react'
-import { Platform, Text, TextInput, TouchableOpacity, TouchableWithoutFeedback, View } from 'react-native'
-import Animated, { Extrapolate, interpolate, interpolateColor, SharedValue, useAnimatedStyle, useSharedValue, withTiming } from 'react-native-reanimated'
+import * as React from 'react'
+import { type TextStyle, type ViewStyle, Platform, Text, TextInput, TouchableOpacity, TouchableWithoutFeedback, View } from 'react-native'
+import Animated, { Extrapolate, interpolate, interpolateColor, useAnimatedStyle, useSharedValue, withTiming } from 'react-native-reanimated'
 import AntDesignIcon from 'react-native-vector-icons/AntDesign'
 
-import { unpackEdges } from '../../util/edges'
-import { useCallback, useEffect, useImperativeHandle, useMemo, useState } from '../../util/hooks'
-import { type Theme, cacheStyles, useTheme } from '../services/ThemeContext'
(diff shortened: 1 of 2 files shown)


swansontec:
Twin PR: https://github.com/EdgeApp/edge-login-ui/pull/324

swansontec:
Spot-checked a few text fields, and they seem to work fine. Flow is happy, which is where most of the changes are.

src/components/themed/EdgeOutlinedField.js
```diff
@@ -22,25 +22,36 @@ const ANIMATION_STATES = {
...
+  autoFocus?: boolean,
+  inputAccessoryViewID?: string,
+  keyboardType?: 'default' | 'number-pad' | 'decimal-pad' | 'numeric' | 'email-address' | 'phone-pad',
+  maxLength?: number,
+  onSubmitEditing?: () => void,
+  returnKeyType?: 'done' | 'go' | 'next' | 'search' | 'send',
+  secureTextEntry?: boolean,
+  testID?: string,
+  value?: string
+|}
```
samholmes, line 53: Shouldn't we export the `Props` type from `EdgeTextFieldOutlined` for reuse? Seems like there is a lot of code dup here, but there my be a good reason for it that I am not seeing.

samholmes (approved):
Approved but with some suggestions and questions

src/components/themed/EdgeOutlinedField.js
```diff
@@ -138,29 +155,29 @@ const EdgeTextFieldOutlinedComponent = forwardRef((props: InputOutlineProps, ref
     hintLeftMargin
   } = getSizeStyles(theme, size, showSearchIcon)
 
-  const errorState = useCallback(() => error !== null && error !== undefined && error !== '', [error])
+  const errorState = useCallback(() => error != null && error !== '', [error])
```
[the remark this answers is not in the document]
swansontec: Yep, this needs to be fixed.

src/components/themed/EdgeOutlinedField.js
swansontec: Components should never export `Props`. Doing that is a big red flag.

The reason we have to manually put in the `TextInput` properties is because we don't have working Flow types for the `react-native` package, so everything there is just `any`. If we were to switch to typescript, we could just do `Pick<TextInputProps, 'autoCapitalize' | 'autoCorrect' | ...>` to avoid the duplication. I do want to restrict which `TextInput` props are valid, since we don't want anybody passing `style`, for instance.

swansontec: Oh, I saw the comment in the other repo. Yes, the idea is to eventually share the entire component, moving it to another library called `edge-components`. It will be a while before we get there.

thehobbit85 (approved):

PR #2763, opened 2021-07-23
Move exhange powered to bottom right

axenov-it:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/Main.ui.js
+++ b/src/components/Main.ui.js
@@ -456,6 +456,7 @@ export class MainComponent extends React.Component<Props> {
               />
               <Scene
                 key={EXCHANGE_QUOTE_SCENE}
+                hideNavBar
                 navTransparent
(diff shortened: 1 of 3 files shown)


axenov-it:
Screenshot from 2021-07-23 03-13-36
Screenshot from 2021-07-23 03-12-37

thehobbit85:
Not even close. This is how the scene should look like.
Where is the scene header that says "Exchange"? Or the divider under it? 
And all the spacing of the icon are wrong. From ALL sides, top, bottom, right, left, not even one is correct.

THIS IS HOW IT SHOULD LOOK LIKE:

Screen Shot 2021-07-22 at 23 34 38

thehobbit85 (changes requested):
See comment

PR #2764, opened 2021-07-23
Add device name to transaction details

axenov-it:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/modals/TransactionAdvanceDetails.js
+++ b/src/components/modals/TransactionAdvanceDetails.js
@@ -2,6 +2,7 @@
 
 import React, { PureComponent } from 'react'
 import { Linking, Platform, ScrollView, StyleSheet, View } from 'react-native'
+import { getDeviceName } from 'react-native-device-info'
 import SafariView from 'react-native-safari-view'
 
 import s from '../../locales/strings.js'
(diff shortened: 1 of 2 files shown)


axenov-it:
image

src/components/modals/TransactionAdvanceDetails.js
```diff
@@ -26,6 +26,7 @@ const feeString = {
 
 type OwnProps = {
   bridge: AirshipBridge<null>,
+  deviceName: string,
```
thehobbit85, line 29: Why do you need to pass the deviceName as a prop and not just call "getDeviceName" in the modal?

src/components/scenes/TransactionDetailsScene.js
```diff
@@ -63,7 +64,8 @@ type State = {
   direction: string,
   bizId: number,
   category: string,
-  subCategory: string
+  subCategory: string,
+  deviceName: string
```
thehobbit85, line 68: Why is the "deviceName" part of the state for this component. 
This is such a waste of code. Just get it directly on the scene above.

thehobbit85 (changes requested):
Fix

thehobbit85 (approved):

PR #2768, opened 2021-07-23
William/primary button

swansontec:
### Fixed

- Change the primary button to bold, and add a minimum width as per the new design.

This also merges the `PrimaryButton`, `SecondaryButton`, and `PromiseButton` into one component. See the changes in the `CreateWalletReviewScene` for how this works - returning a promise from `onPress` will set the button spinning until the promise resolves, removing the need to constantly re-build this exact same logic on every scene.

Also, by setting `alignSelf="center"`, the button centers itself with the minimum width, so we can remove all types of layout hacks.

#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


swansontec:
Merged #2783 instead.

PR #2771, opened 2021-07-26
Fix fiatAmount function on FioSentRequestDetailsScene.js

amanalili:
Component to use bns operations

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

### DIFF
--- a/src/components/scenes/FioSentRequestDetailsScene.js
+++ b/src/components/scenes/FioSentRequestDetailsScene.js
@@ -1,5 +1,6 @@
 // @flow
 
+import { bns } from 'biggystring'
 import * as React from 'react'
 
 import { FIAT_CODES_SYMBOLS } from '../../constants/WalletAndCurrencyConstants.js'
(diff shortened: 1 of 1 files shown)


src/components/scenes/FioSentRequestDetailsScene.js
```diff
@@ -34,7 +35,7 @@ class FioSentRequestDetailsComponent extends React.PureComponent<Props> {
     const fiatPerCrypto = exchangeRates[rateKey] ? exchangeRates[rateKey] : 0
```
peachbits, line 35: Use exchangeRatesToString here so fiatPerCrypto is a string

src/components/scenes/FioSentRequestDetailsScene.js
```diff
@@ -34,7 +35,7 @@ class FioSentRequestDetailsComponent extends React.PureComponent<Props> {
...
     const amountToMultiply = parseFloat(amount)
```
peachbits, line 36: amount is already a string you can use it directly in bns.mul() without converting to and from a number.

peachbits (changes requested):

PR #2779, opened 2021-07-26
Fix React DevTools for Flipper

samholmes:
This fixes backend compatibility issues with newer version of Flipper.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a


package.json
```diff
@@ -74,6 +74,7 @@
     "long": "4.0.0",
     "process-nextick-args": "2.0.0",
     "randombytes": "2.1.0",
+    "react-devtools-core": "4.13.0",
```
swansontec, line 77: This line isn't needed - the change to `yarn.lock` alone is enough. I do see the benefit of putting this line in, but I'm worried we might forget to remove it the next time we upgrade React Native or whatever. It's hard to say which way is better, but I guess being explicit is better than implicit? Leave it in?

swansontec (approved):
Please rebase onto develop. I did the string changes as an explicit stand-alone commit.

package.json
samholmes: I'd say it's better to be explicit here. If we forget to remove it, at least it is noticeable why we get that version in the `yarn.lock`.

samholmes: I'd say it's better to be explicit here. If we forget to remove it, at least it is noticeable why we get that version in the `yarn.lock`.

swansontec: Great, let's merge it then!

samholmes: I have a concern that this change may be the reason why I cannot connect Hermes Debugger (RN) to my iOS device for debugging. I say put this change on hold until we're are more confident that it doesn't cause more issues that is solves.

PR #2799, opened 2021-07-31
Fix slider reset prop effect

samholmes:
Add `useEffect` for the reset prop to invoke `resetSlider` once when 
reset prop changes from false to true.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a

### DIFF
--- a/src/modules/UI/components/Slider/Slider.js
+++ b/src/modules/UI/components/Slider/Slider.js
@@ -9,7 +9,7 @@ import leftArrowImg from '../../../../assets/images/slider/keyboard-arrow-left.p
 import { type Theme, type ThemeProps, cacheStyles, withTheme } from '../../../../components/services/ThemeContext.js'
 import { EdgeText } from '../../../../components/themed/EdgeText'
 import s from '../../../../locales/strings.js'
-import { useEffect, useState } from '../../../../types/reactHooks.js'
+import { useCallback, useEffect, useState } from '../../../../types/reactHooks.js'
 
 const COMPLETE_POINT: number = 3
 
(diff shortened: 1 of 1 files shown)


thehobbit85 (approved):

src/modules/UI/components/Slider/Slider.js
```diff
@@ -70,7 +70,10 @@ export const SliderComponent = (props: Props) => {
     setCompleted(true)
   }
 
-  if (reset && translateX.value === 0) resetSlider()
```
thehobbit85, line 73: Are you sure that getting rid of the condition `translateX.value === 0` won't cause a problem?
If I'm not mistaken, it was to prevent reseting a slider before it reaches the end.

thehobbit85 (changes requested):
See comment

samholmes:
The intention behind that condition was to prevent the infinite rerender
issue. It wasn’t the correct fix. This is.

src/modules/UI/components/Slider/Slider.js
```diff
@@ -58,20 +58,18 @@ export const SliderComponent = (props: Props) => {
...
 
-  const resetSlider = () => {
+  const resetSlider = useCallback(() => {
     translateX.value = withTiming(upperBound, {
       duration: 500,
       easing: Easing.inOut(Easing.exp)
     })
     setCompleted(false)
-  }
+  }, [])
```
samholmes, line 67: Perhaps `upperBound` should be a dep

thehobbit85 (approved):

PR #2805, opened 2021-08-03
Scan clipboard for address on return to send scene

samholmes:
When the app is coming into foreground, read from the clipboard in case
a new address was copied while the app was in the background.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a

### DIFF
--- a/src/components/themed/AddressTile.js
+++ b/src/components/themed/AddressTile.js
@@ -3,7 +3,7 @@
 import Clipboard from '@react-native-community/clipboard'
 import type { EdgeCurrencyConfig, EdgeCurrencyWallet, EdgeParsedUri } from 'edge-core-js'
 import * as React from 'react'
-import { TouchableOpacity, View } from 'react-native'
+import { AppState, TouchableOpacity, View } from 'react-native'
 import FontAwesome from 'react-native-vector-icons/FontAwesome'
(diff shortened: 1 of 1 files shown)


src/components/themed/AddressTile.js
```diff
@@ -53,16 +53,22 @@ const isPaymentProtocolUri = (parsedUri: EdgeParsedUri): boolean => {
...
+
   constructor(props: Props) {
     super(props)
 
     this.state = {
       clipboard: '',
       loading: false
     }
+
+    this.handleAppStateChange = this.handleAppStateChange.bind(this)
```
samholmes, line 66: Try defining this as an arrow function; if you can't get it to work, then go with this.

src/components/themed/AddressTile.js
```diff
@@ -80,6 +88,12 @@ class AddressTileComponent extends React.PureComponent<Props, State> {
     }
   }
 
+  handleAppStateChange(appState) {
+    if (appState === 'active') {
+      this._setClipboard(this.props)
+    }
```
samholmes, line 94: One-liner

thehobbit85 (approved):
Nice

PR #2825, opened 2021-08-09
Close keyboard when opening the Sidemenu

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android


swansontec (approved):
Yeah, this is a reasonable solution until we switch routers.

PR #2821, opened 2021-08-09
William/unique id modal

swansontec:
#### PR Requirements

If you have made **any** visual changes to the GUI. Make sure you have:
- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a

PR #2827, opened 2021-08-09
Replace the xpub modal with a buttons modal

swansontec:
Both buttons are now secondary buttons, per Julia's quick in-person design review.

#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


swansontec:
Built on #2821

swansontec:
Before:

simulator_screenshot_874626CA-D936-4E2E-AF60-579E64C5C696

After:
simulator_screenshot_619A7582-EE5B-4A20-A197-7E392699E29F

swansontec:
Both buttons are now secondary buttons, per Julia's quick in-person design review.

src/actions/WalletListMenuActions.js
```diff
@@ -89,10 +91,36 @@ export function walletListMenuAction(walletId: string, option: WalletListMenuKey
...
+            bridge={bridge}
+            buttons={
+              xpubExplorer != null
+                ? {
+                    copy: { label: s.strings.fragment_request_copy_title, type: 'secondary' },
+                    link: { label: s.strings.transaction_details_show_advanced_block_explorer, type: 'secondary' }
+                  }
+                : {
+                    copy: { label: s.strings.fragment_request_copy_title, type: 'secondary' }
+                  }
```
thehobbit85, line 108: Create consts and merge the right ones together.

src/actions/WalletListMenuActions.js
```diff
@@ -89,10 +91,36 @@ export function walletListMenuAction(walletId: string, option: WalletListMenuKey
       return (dispatch: Dispatch, getState: GetState) => {
         const state = getState()
         const { currencyWallets } = state.core.account
-        const wallet = currencyWallets[walletId]
-        const xPub = wallet.displayPublicSeed
-        const xPubExplorer = wallet.currencyInfo.xpubExplorer && xPub ? sprintf(wallet.currencyInfo.xpubExplorer, xPub) : ''
-        dispatch({ type: 'OPEN_VIEWXPUB_WALLET_MODAL', data: { xPub, walletId, xPubExplorer } })
+        const { displayPublicSeed, currencyInfo } = currencyWallets[walletId]
```
thehobbit85, line 94: Default displayPublicSeed to ''

thehobbit85 (changes requested):
Small comments

src/actions/WalletListMenuActions.js
swansontec: Derp, defaults don't work with `null`. This is an API design mistake in the core - we should have made it `| void` so defaults would work, instead of `| null` where defaults don't work.

src/actions/WalletListMenuActions.js
swansontec: Done, much nicer.

PR #2857, opened 2021-08-19
Replace exchange max with a new mini button component

swansontec:
#### PR Requirements

- [x] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/scenes/CryptoExchangeScene.js
+++ b/src/components/scenes/CryptoExchangeScene.js
@@ -24,6 +24,7 @@ import { CryptoExchangeMessageBox } from '../themed/CryptoExchangeMessageBoxComp
 import type { ExchangedFlipInputAmounts } from '../themed/ExchangedFlipInput'
 import { LineTextDivider } from '../themed/LineTextDivider'
 import { MainButton } from '../themed/MainButton.js'
+import { MiniButton } from '../themed/MiniButton.js'
 import { SceneHeader } from '../themed/SceneHeader'
 
 type StateProps = {
@@ -56,7 +57,7 @@ type StateProps = {
 type DispatchProps = {
   onSelectWallet: (walletId: string, currencyCode: string, direction: 'from' | 'to') => void,
   getQuoteForTransaction: SetNativeAmountInfo => void,
-  exchangeMax: () => void
+  exchangeMax: () => Promise<void>
 }
 type Props = StateProps & DispatchProps & ThemeProps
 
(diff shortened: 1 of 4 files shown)


swansontec:
simulator_screenshot_3CB09D33-9BB7-4221-A286-DF9960E72EDF

src/components/themed/MiniButton.js
```diff
@@ -0,0 +1,98 @@
...
+
+  // Whether to center the button or stretch to fill the screen.
+  // Defaults to 'auto', letting the parent component be in charge:
+  alignSelf?: 'auto' | 'stretch' | 'center',
+
+  // True to dim the button & prevent interactions:
+  disabled?: boolean,
+
+  // If this is set, the component will insert a text node before the other children:
+  label?: string,
```
swansontec, line 25: This should not be optional, since the component doesn't take other children.

swansontec (commented):
Very PRY - please repeat yourself. However, this is a good thing, since we tweak styles and eliminate props based on the specific, more-limited needs of this component.

src/components/scenes/CryptoExchangeScene.js
```diff
@@ -377,8 +373,8 @@ export const CryptoExchangeScene = connect<StateProps, DispatchProps, {}>(
       dispatch(selectWalletForExchange(walletId, currencyCode, direction))
       dispatch(updateMostRecentWalletsSelected(walletId, currencyCode))
     },
-    exchangeMax() {
-      dispatch(exchangeMax())
+    async exchangeMax() {
+      await dispatch(exchangeMax())
```
samholmes, line 377: Was ESLint yelling about this?

src/components/themed/MiniButton.js
samholmes: Agreed. It should not be optional if there is not a `children` prop. If we later add `children` prop we would need to make this optional again.

src/components/themed/MiniButton.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  // Styles:
+  const theme = useTheme()
+  const styles = getStyles(theme)
+  const dynamicStyles = {
+    alignSelf,
+    opacity: disabled || pending ? 0.7 : 1,
+    ...sidesToMargin(mapSides(fixSides(marginRem, 0), theme.rem))
+  }
```
samholmes, line 58: In an effort to make it a little DRY, could we move some of this shared functionality into hooks in order to use the in both `MainButton` and `MiniButton`? This makes a "has a" inheritance model rather than an "is a" model. If this is feasible, then it's worth exploring.

samholmes (approved):
Approved with optional changes/comments to consider.

src/components/scenes/CryptoExchangeScene.js
swansontec: No, it's a genuine change - before it returned `void`, but now it returns `Promise<void>` so we can show a spinner until it's done working.

src/components/themed/MiniButton.js
swansontec: The styles are unique, and PRY is the right approach. The button spinning logic could maybe be moved into a hook, so I'll give that a shot.

samholmes (approved):

samholmes:
Excellent! Refactoring the functionality into its own hook was the right way to go. Looks great!

PR #2865, opened 2021-08-23
Error message when non allowed characters on Flip Input

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

photo_2021-08-23 19 04 48

photo_2021-08-23 19 04 45

photo_2021-08-23 19 04 50

### DIFF
--- a/src/components/modals/FlipInputModal.js
+++ b/src/components/modals/FlipInputModal.js
@@ -58,7 +58,8 @@ type DispatchProps = {
 
 type State = {
   overridePrimaryExchangeAmount: string,
-  forceUpdateGuiCounter: number
+  forceUpdateGuiCounter: number,
+  errorMessage?: string
 }
 
 type Props = OwnProps & StateProps & DispatchProps & ThemeProps
@@ -78,6 +79,8 @@ class FlipInputModalComponent extends React.PureComponent<Props, State> {
     const { walletId, currencyCode, updateTransactionAmount } = this.props
     updateTransactionAmount(nativeAmount, exchangeAmount, walletId, currencyCode)
   }
+
+  handleAmountChangeError = (errorMessage?: string) => this.setState({ errorMessage })
 
   componentDidUpdate(prevProps: Props) {
     if (this.props.forceUpdateGuiCounter !== this.state.forceUpdateGuiCounter) {
@@ -106,11 +109,11 @@ class FlipInputModalComponent extends React.PureComponent<Props, State> {
   }
 
   renderExchangeRates = () => {
-    const { primaryInfo, secondaryInfo, fiatPerCrypto, errorMessage, theme } = this.props
+    const { primaryInfo, secondaryInfo, errorMessage = this.state.errorMessage, fiatPerCrypto, theme } = this.props
     const styles = getStyles(theme)
     return (
       <View style={styles.exchangeRateContainer}>
-        {errorMessage ? (
+        {errorMessage != null ? (
           <EdgeText style={styles.exchangeRateErrorText}>{errorMessage}</EdgeText>
         ) : (
           <ExchangeRate primaryInfo={primaryInfo} secondaryInfo={secondaryInfo} secondaryDisplayAmount={fiatPerCrypto} />
@@ -147,6 +150,7 @@ class FlipInputModalComponent extends React.PureComponent<Props, State> {
             overridePrimaryExchangeAmount={overridePrimaryExchangeAmount}
             forceUpdateGuiCounter={0}
             onExchangeAmountChanged={this.handleExchangeAmountChange}
+            onError={this.handleAmountChangeError}
             onNext={this.handleCloseModal}
             keyboardVisible={false}
             isFocus
--- a/src/components/themed/CryptoExchangeFlipInputWrapperComponent.js
+++ b/src/components/themed/CryptoExchangeFlipInputWrapperComponent.js
@@ -42,13 +42,13 @@ type StateProps = {
   cryptoAmount?: string
 }
 
+type State = {
+  errorMessage?: string
+}
+
 type Props = OwnProps & StateProps & ThemeProps
 
-class CryptoExchangeFlipInputWrapperComponent extends React.Component<Props> {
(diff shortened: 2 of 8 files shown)


src/components/themed/ErrorText.js
```diff
@@ -0,0 +1,29 @@
...
+
+import * as React from 'react'
+import { StyleSheet } from 'react-native'
+
+import { type Theme, type ThemeProps, cacheStyles, withTheme } from '../services/ThemeContext'
+import { EdgeText } from './EdgeText'
+
+type OwnProps = {
+  error?: string,
+  style?: StyleSheet.Styles
```
swansontec, line 11: This is not OK. New components should *never* accept raw CSS styles.

We actually had a meeting about this a few months ago, and we agreed to do it this way: `<EdgeText><Error>{s.strings.whatever}</Error></EdgeText>`. Basically, the `Error` component would change the text color to red, but let the `EdgeText` manage the other styles (font, size, margin, padding, etc.).

We also want a `<Warning>` for yellow text, and a `<B>` for bold text.

swansontec (changes requested):

src/components/themed/ErrorText.js
amanalili: Removed and fixed

src/components/themed/FlipInput.js
```diff
@@ -373,11 +374,20 @@ class FlipInputComponent extends React.PureComponent<Props, State> {
...
+      this.props.onError(error)
+    }
+  }
+
   onKeyPress(keyPressed: string, decimalAmount: string, maxEntryDecimals: number, setAmounts: (Props, string) => Amounts) {
     keyPressed = keyPressed.replace(',', '.')
     if (keyPressed === 'Backspace') {
       this.setStateAmounts(truncateDecimals(decimalAmount.slice(0, -1), maxEntryDecimals), setAmounts)
-    } else if (checkKeyPress(keyPressed, decimalAmount)) {
+    } else if (!checkKeyPress(keyPressed, decimalAmount)) {
```
thehobbit85, line 387: Instead of flipping the condition, keep it the same and then your else would be the error.

src/components/modals/FlipInputModal.js
```diff
@@ -79,6 +80,8 @@ class FlipInputModalComponent extends React.PureComponent<Props, State> {
     updateTransactionAmount(nativeAmount, exchangeAmount, walletId, currencyCode)
   }
 
+  handleAmountChangeError = (error?: string) => this.setState({ errorMessage: error })
```
thehobbit85, line 83: change the param name to "errorMessage" for 2 reasons.
1. Error usually represent an error object and not the message.
2. By changing the param name, the code would look like this:
`(errorMessage?: string) => this.setState({ errorMessage })`

src/components/modals/FlipInputModal.js
```diff
@@ -106,8 +109,9 @@ class FlipInputModalComponent extends React.PureComponent<Props, State> {
   }
 
   renderExchangeRates = () => {
-    const { primaryInfo, secondaryInfo, fiatPerCrypto, errorMessage, theme } = this.props
+    const { primaryInfo, secondaryInfo, fiatPerCrypto, theme } = this.props
```
thehobbit85, line 112: Instead of removing the "errorMessage" from the destructor, you can use a default:
```
const { 
   primaryInfo,
   secondaryInfo, 
   fiatPerCrypto,
   errorMessage = this.state.errorMessage, 
   theme 
} = this.props
```

src/components/modals/FlipInputModal.js
```diff
@@ -106,8 +109,9 @@ class FlipInputModalComponent extends React.PureComponent<Props, State> {
...
     const styles = getStyles(theme)
+    const errorMessage = this.state.errorMessage ?? this.props.errorMessage
     return (
       <View style={styles.exchangeRateContainer}>
         {errorMessage ? (
```
thehobbit85, line 117: We agreed that if you work on code that has a "fuzy-boolean" like this: `errorMessage ?` you change it to an implicit check like this: `errorMessage != null` or whatever is needed.

src/components/themed/CryptoExchangeFlipInputWrapperComponent.js
```diff
@@ -51,6 +56,13 @@ class CryptoExchangeFlipInputWrapperComponent extends React.Component<Props & Th
     )
   }
 
+  onError = (error?: string) => this.setState({ error })
+
+  focusMe = () => {
+    this.onError()
```
thehobbit85, line 62: Why are you calling onError on focus and not only when you actually have an error?

src/components/themed/CryptoExchangeFlipInputWrapperComponent.js
```diff
@@ -33,8 +33,13 @@ export type Props = {
...
 }
 
-class CryptoExchangeFlipInputWrapperComponent extends React.Component<Props & ThemeProps> {
+type State = {
+  error?: string
+}
+
+class CryptoExchangeFlipInputWrapperComponent extends React.Component<Props & ThemeProps, State> {
   launchSelector = () => {
+    this.onError()
```
thehobbit85, line 42: Why are you calling onError here and not only when you actually have an error?

thehobbit85 (changes requested):

src/components/themed/CryptoExchangeFlipInputWrapperComponent.js
amanalili: this.clearError()

src/components/themed/FlipInput.js
```diff
@@ -373,11 +374,20 @@ class FlipInputComponent extends React.PureComponent<Props, State> {
     }
   }
 
+  handleOnError(error?: string) {
+    if (this.props.onError != null) {
+      this.props.onError(error)
+    }
+  }
+
```
amanalili, line 382: clearError() {
      this.handleOnError()
}

thehobbit85 (approved):

PR #2869, opened 2021-08-25
Dismiss keyboard when searching and going out of the Wallet List scene

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

### DIFF
--- a/src/components/themed/WalletListSwipeRow.js
+++ b/src/components/themed/WalletListSwipeRow.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as React from 'react'
-import { Dimensions, Platform, View } from 'react-native'
+import { Dimensions, Keyboard, Platform, View } from 'react-native'
 import { SwipeRow } from 'react-native-swipe-list-view'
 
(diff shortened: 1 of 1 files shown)


src/components/themed/WalletListSwipeRow.js
```diff
@@ -70,6 +70,7 @@ class WalletListSwipeRowComponent extends React.PureComponent<Props & ThemeProps
 
     this.closeRow()
     this.props.selectWallet(walletId, currencyCode)
+    Keyboard.dismiss()
```
thehobbit85, line 73: Why not just put the "Keyboard.dismiss()" inside the "this.closeRow" function instead of calling it multiple times?

thehobbit85 (changes requested):

src/components/themed/WalletListSwipeRow.js
amanalili: Whoops, good catch. Fixed

thehobbit85 (approved):

PR #2875, opened 2021-08-27
Remove react-native-router-flux from all scenes

swansontec:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/scenes/CreateWalletAccountSelectScene.js
+++ b/src/components/scenes/CreateWalletAccountSelectScene.js
@@ -20,7 +20,7 @@ import SafeAreaView from '../../modules/UI/components/SafeAreaView/SafeAreaView.
 import { getDefaultDenomination } from '../../selectors/DenominationSelectors.js'
 import { THEME } from '../../theme/variables/airbitz.js'
 import { connect } from '../../types/reactRedux.js'
-import { type RouteProp, Actions } from '../../types/routerTypes.js'
+import { type RouteProp } from '../../types/routerTypes.js'
 import type { GuiWallet } from '../../types/types.js'
 import { getCurrencyIcon } from '../../util/CurrencyInfoHelpers.js'
 import { scale } from '../../util/scaling.js'
@@ -106,10 +106,6 @@ class CreateWalletAccountSelect extends React.Component<Props, State> {
 
   componentDidMount() {
     logEvent('ActivateWalletSelect')
-  }
-
-  onBack = () => {
-    Actions.pop()
   }
 
   onPressSelect = () => {
--- a/src/components/scenes/CreateWalletReviewScene.js
+++ b/src/components/scenes/CreateWalletReviewScene.js
@@ -5,10 +5,9 @@ import { Image, Keyboard, View } from 'react-native'
 
 import { createCurrencyWallet } from '../../actions/CreateWalletActions.js'
 import CheckIcon from '../../assets/images/createWallet/check_icon_lg.png'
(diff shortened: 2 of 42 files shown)


src/components/scenes/CreateWalletAccountSelectScene.js
```diff
@@ -108,10 +108,6 @@ class CreateWalletAccountSelect extends React.Component<Props, State> {
     logEvent('ActivateWalletSelect')
   }
 
-  onBack = () => {
```
thehobbit85, line 111: Can I assume this is unused anyway?

src/components/scenes/CreateWalletReviewScene.js
```diff
@@ -43,7 +43,10 @@ class CreateWalletReviewComponent extends React.Component<Props, State> {
     Keyboard.dismiss()
   }
 
-  goToWalletList = () => Actions.popTo(WALLET_LIST_SCENE)
+  goToWalletList = () => {
```
thehobbit85, line 46: why not keep it `goToWalletList = () => this.props.navigation.navigate('walletListScene')`?
Is there a reason you don't want to return the value of navigate?

src/components/scenes/FioAddressSettingsScene.js
```diff
@@ -55,24 +55,28 @@ class FioAddressSettingsComponent extends React.Component<Props, LocalState> {
...
     showToast(s.strings.fio_request_renew_ok_text)
-    Actions.pop()
+    navigation.goBack()
     if (refreshAfterRenew) {
       window.requestAnimationFrame(() => {
-        Actions.refresh({ fioAddressName, expiration: expiration ? new Date(expiration) : '' })
+        navigation.setParams({
+          fioWallet,
+          fioAddressName,
+          expiration
```
thehobbit85, line 72: What about converting expiration to a date like before: `new Date(expiration)`?

src/components/scenes/ScanScene.js
```diff
@@ -48,10 +48,10 @@ export const LOGIN_QR = 'loginQR'
 
 export class Scan extends React.Component<Props> {
   componentDidUpdate(prevProps: Props) {
-    const { route } = this.props
+    const { navigation, route } = this.props
     const { data } = route.params
-    if (data !== prevProps.route.params.data && Actions.currentScene !== 'DrawerOpen') {
-      Actions.drawerClose()
+    if (data !== prevProps.route.params.data) {
```
thehobbit85, line 53: Why is it ok to remove the 'Actions.currentScene !== 'DrawerOpen'' without a replacement?

thehobbit85 (changes requested):
1. Resolve small conflict
2. Address comments
3. Don't merge it yet. Even though it's a pretty "simple" PR, it's still very long and include changes to every scene in the app. I'd rather wait until we get the current staging (2.2.0) into master before we merge this so we can avoid any conflicts if we need to merge/cherry pick anything from develop.

src/components/scenes/CreateWalletAccountSelectScene.js
swansontec: Right, nobody every called this guy.

src/components/scenes/ScanScene.js
swansontec: Because `Actions.currentScene !== 'DrawerOpen'`  is always `true`.

swansontec (commented):
Found one more thing.

PR #2876, opened 2021-08-27
Fix the expired FIO button centering

swansontec:
The button supports this natively now, so we can delete a bunch of code.

Also fixed the Flow type of the modal, which would have triggered the sketchy-null checks if we had them enabled.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [x] Tested on small Android
- [ ] n/a


swansontec:
Screenshot:

signal-2021-08-27-160323_002

thehobbit85 (approved):

PR #2880, opened 2021-09-01
Fix send using fio address

andreyvEze:
Refactor set fio data in other params of spend target

https://user-images.githubusercontent.com/22767467/132348375-4dfeb69f-445a-46d7-8ce9-8b0e9e313611.mp4

https://user-images.githubusercontent.com/22767467/132362145-7a7ea566-3e05-4e12-8dc5-dee838220ce1.mp4


src/actions/ScanActions.js
```diff
@@ -194,6 +194,13 @@ export const parseScannedUri = (data: string, customErrorTitle?: string, customE
       }
     ]
 
+    if (fioAddress) {
```
peachbits, line 197: No more implicit null checks in the GUI. Only booleans should be used this way and other types should be null-checked explicitly. I'm only adding this single comment but this appears in a few places across the PR.

peachbits (changes requested):
I wasn't able to wrap my head around how this addressed the issue in the task so I tested the code and the incoming received transactions are still not being tagged with a sender's fio address when a payment request is paid. If this was working in your testing can you show that with screenshots and steps?

src/actions/ScanActions.js
andreyvEze: yes, sure, my bad, it was just moved from other place and I did not rewrite it from scratch

src/actions/FioActions.js
```diff
@@ -51,22 +55,47 @@ export const refreshConnectedWallets = async (dispatch: Dispatch, getState: GetS
...
         }
+      } catch (e) {
+        //
+      }
+      const obtForTx: FioObtRecord | void = obtDataRecords.find(obtRecord => obtRecord.content.obt_id === transaction.txid)
+      if (obtForTx == null) return
+
+      if (edgeMetadata.notes == null) edgeMetadata.notes = ''
+      let fioNotes = `${s.strings.fragment_transaction_list_sent_prefix}${s.strings.word_to_in_convert_from_to_string} ${obtForTx.payee_fio_address}`
+      if (obtForTx.content.memo != null) fioNotes += `\n${s.strings.fio_sender_memo_label}: ${obtForTx.content.memo}`
```
peachbits, line 89: Also check that the memo isn't an empty string before adding this line to the notes.

src/components/scenes/SendScene.js
```diff
@@ -272,15 +278,16 @@ class SendComponent extends React.PureComponent<Props, State> {
...
   submit = async () => {
-    const { updateSpendPending, signBroadcastAndSave, route } = this.props
+    const { updateSpendPending, isSendUsingFioAddress, signBroadcastAndSave, route } = this.props
     const { guiMakeSpendInfo } = route.params
     const { selectedWalletId, selectedCurrencyCode } = this.state
 
     this.setState({ loading: true })
 
-    if (guiMakeSpendInfo && (guiMakeSpendInfo?.isSendUsingFioAddress || guiMakeSpendInfo?.fioPendingRequest)) {
+    const isFioPendingRequest = guiMakeSpendInfo && guiMakeSpendInfo?.fioPendingRequest
```
peachbits, line 287: Remove "guiMakeSpendInfo &&"

src/actions/FioActions.js
```diff
@@ -51,22 +55,47 @@ export const refreshConnectedWallets = async (dispatch: Dispatch, getState: GetS
 }
 
 export const checkFioObtData = (walletId: string, transactions: EdgeTransaction[]) => async (dispatch: Dispatch, getState: GetState) => {
+  if (checkFioObtDataCounter > MAX_OBT_DATA_CHECKS) return // to prevent loop if account does not have fio wallets
```
peachbits, line 58: This seems pretty arbitrary and could mean that FIO obt data isn't checked after some time. Can you explain the rationale for 15 or is there a better way to determine if FIO names are present and this function can exit early?

peachbits (changes requested):

src/actions/FioActions.js
andreyvEze: there is no important meaning of the value, just 15 * 400 is 6 seconds, and it is just an assumption that wallets should be loaded in this time. as I remember it all was removed before because we wanted to refactor this logic to be in the `core` lib, but for some reason it was forgotten. not sure we should rewrite the logic in current PR, is it ok to create separate task for it?

peachbits:
The autosquash failed because the fixup messages weren't recognized so I created a new squashed PR here: https://github.com/EdgeApp/edge-react-gui/pull/2887

PR #2884, opened 2021-09-01
Fix bug on Transaction Details where fiat amount is blank and closing the modal will make the fiat amount blank

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android

### DIFF
--- a/src/components/scenes/TransactionDetailsScene.js
+++ b/src/components/scenes/TransactionDetailsScene.js
@@ -182,7 +182,12 @@ export class TransactionDetailsComponent extends React.Component<Props, State> {
         amount={this.state.amountFiat}
         onChange={this.onChangeFiat}
       />
-    )).then(_ => {})
+    )).then(_ => {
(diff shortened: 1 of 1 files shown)


src/components/scenes/TransactionDetailsScene.js
```diff
@@ -182,7 +182,12 @@ export class TransactionDetailsComponent extends React.Component<Props, State> {
         amount={this.state.amountFiat}
         onChange={this.onChangeFiat}
       />
-    )).then(_ => {})
+    )).then(_ => {
+      const { amountFiat } = this.state
```
thehobbit85, line 186: Use this util function to check if a string number is zero: 
https://github.com/EdgeApp/edge-react-gui/blob/develop/src/util/utils.js#L149

thehobbit85 (changes requested):

thehobbit85 (approved):

PR #2891, opened 2021-09-11
Update env.example.json

trustwalletvn:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


thehobbit85:
This is actually wrong.
The "etherscanApiKey" is an array containing multiple keys.

PR #2925, opened 2021-10-01
Upgraded TransactionDetails fiat modal to use new dark theme

itayplav:
#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

https://user-images.githubusercontent.com/4023066/135560767-471d81c6-f24e-4cf5-9f63-69c4fab19211.mp4


src/components/scenes/TransactionDetailsScene.js
```diff
@@ -175,19 +175,25 @@ export class TransactionDetailsComponent extends React.Component<Props, State> {
...
+        inputLabel={sprintf(s.strings.transaction_details_fiat_modal_header, fiatCurrencyCode)}
+        returnKeyType="send"
+        keyboardType="decimal-pad"
+        title={s.strings.transaction_details_fiat_label}
+        onSubmit={async amountInput => {
+          const amountCleaned = amountInput.replace(',', '.').replace(/[^\d.,]/g, '')
+          const amountFiat = bns.toFixed(amountCleaned, 2, 2)
+          await this.onChangeFiat(amountFiat)
+          return true
+        }}
```
itayplav, line 194: - This needs to be moved to utils.js 
- Create unit tests for cleaning amountInput
- Format should follow $NUMBER1.$NUMBER2, everything after $NUMBER2 gets removed
- If $NUMBER1 is a number and $NUMBER2 is not a number, then ignore $NUMBER2
- If $NUMBER1 is not a number return 0.00
- If return value evaluates to 0 (null / undefined / invalid), then return 0.00

itayplav:
If returning the wrong number, don't crash. Worst case use try..catch. Example breaking input: 1ne0

PR #2936, opened 2021-10-07
Remove Transaction Details save button

thehobbit85:
Remove the "save" button at the bottom of the Transaction Details scene.
Instead, every time the user edit's a value and hit submit/done in the modal, it would trigger a save.

https://user-images.githubusercontent.com/4012041/136318280-4a85b9b4-0a53-4403-8f95-c0fb3800a8fb.mp4


swansontec:
This was a hallway discussion between me, Juila, and Eliran. The button made sense when we had the text fields inline, but now that we have to confirm modals, it's an extra button press we don't need.

PR #2942, opened 2021-10-13
Add state reset on Crypto Exchange Scene

amanalili:
Note: Suggest that the correct fix is to have a reset function on navigation/action to remount the Scene

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android


swansontec (approved):
Fantastic.

thehobbit85 (approved):

PR #2969, opened 2021-11-05
Add queries section to Android manifest

peachbits:
This is a new requirement for Android version 30. Requests to launch other apps are filtered by this setting and is necessary to allow users to bounce out to the browser to view transaction and address explorer links.


peachbits:
OTS approved by @swansontec

PR #2982, opened 2021-11-10
Fix MainButton component long text being cutoff

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [x] Tested on small Android

photo_2021-11-10 19 16 10


thehobbit85:
Not the right solution

PR #2990, opened 2021-11-12
Fix Wallet Connect initialization 

peachbits:
- Prevent calling wcInit when switching wallet
- Pass public address and wallet ID to plugin (the connectors no longer have wallet scope)
- Call disconnect to close the websocket if the user leaves the scene without approving the connection


peachbits:
to be merged after accountbased is updated

thehobbit85 (approved):

PR #2995, opened 2021-11-15
Move zcash checkpoints from edge-currency-accountbased into android asset folder

peachbits:


peachbits:
Relies on https://github.com/EdgeApp/edge-currency-accountbased/pull/339

scripts/prepare.sh
```diff
@@ -24,7 +24,7 @@ node ./scripts/makeNativeHeaders.js
 
 # Create zcash checkpoints
 mkdir -p android/app/build/intermediates/merged_assets/debug/out/saplingtree/mainnet
-node ./node_modules/edge-currency-accountbased/bin/zecCheckpoints.js || true
+cp -R node_modules/edge-currency-accountbased/lib/zcash/zecCheckpoints android/app/build/intermediates/merged_assets/debug/out/saplingtree/mainnet
```
thehobbit85, line 27: Make it a bit more robust by adding `2>/dev/null || :` to the end of this line so that it would not fail if the zcash checkpoint folder is not there.

thehobbit85 (changes requested):

PR #2999, opened 2021-11-18
Fix contacts permission modal height

amanalili:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [x] Tested on small Android

contacts_permission_modal

---
  - https://app.asana.com/0/0/1201050090794912

### DIFF
--- a/src/components/modals/ContactsPermissionModal.js
+++ b/src/components/modals/ContactsPermissionModal.js
@@ -38,7 +38,7 @@ export function ContactsPermissionModal(props: { bridge: AirshipBridge<any> }) {
       <EdgeText numberOfLines={0} style={styles.message}>
         {s.strings.contacts_permission_modal_text_2}
       </EdgeText>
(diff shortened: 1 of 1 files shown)


src/components/modals/ContactsPermissionModal.js
```diff
@@ -61,15 +60,12 @@ const getStyles = cacheStyles((theme: Theme) => ({
...
   header: {
-    marginVertical: theme.rem(1),
+    marginVertical: theme.rem(0.5),
     fontFamily: theme.fontFaceMedium,
     fontSize: theme.rem(1.25),
     includeFontPadding: false
   },
   message: {
     marginBottom: theme.rem(1)
-  },
```
thehobbit85, line 71: This is not the solution, we do want to keep the margin at the bottom if it fits in the screen.

thehobbit85 (changes requested):

PR #3032, opened 2021-12-07
Whitelist bankid and swish packages

peachbits:


peachbits:
Approved OTS by @thehobbit85

PR #3063, opened 2022-01-05
Gateway Address warnings

thehobbit85:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1201616315654662


src/actions/ScanActions.js
```diff
@@ -8,8 +8,8 @@ import URL from 'url-parse'
 
 import { selectWalletForExchange } from '../actions/CryptoExchangeActions.js'
 import { ButtonsModal } from '../components/modals/ButtonsModal.js'
+import { ConfirmContinueModal } from '../components/modals/ConfirmContinueModal.js'
 import { paymentProtocolUriReceived } from '../components/modals/paymentProtocolUriReceived.js'
-import { shouldContinueLegacy } from '../components/modals/shouldContinueLegacy.js'
```
peachbits, line 12: Technically shouldn't be removed yet as it's still used in this file as of this commit

src/locales/en_US.js
```diff
@@ -884,6 +884,11 @@ const strings = {
   confirm_continue_modal_body: 'Please confirm your understanding below:',
   confirm_continue_modal_button_text: 'I understand and agree to the terms',
 
+  // Gateway Warning Modal
+  gateway_agreement_modal_title: '%s Gatway Address Detected',
+  gateway_agreement_modal_body:
+    'This address is only valid for 24 hours and can only be used once.\n\nDo not send funds after the address has expired and/or send multiple transactions to this address.',
```
peachbits, line 890: Is 24 hours currency-specific? Also, we can just use 'or' instead of 'and/or' I have a suspicion that won't translate well

peachbits (changes requested):

src/components/scenes/SendScene.js
```diff
@@ -171,27 +171,21 @@ class SendComponent extends React.PureComponent<Props, State> {
...
-              }
-            }
-          )
-        }
+        if (walletId == null || currencyCode == null) return
+        selectWallet(walletId, currencyCode)
+        this.setState({
+          ...this.state,
+          ...this.setWallets(this.props, walletId, currencyCode),
+          recipientAddress: undefined
```
peachbits, line 183: Should be an empty string

peachbits (changes requested):

src/actions/ScanActions.js
thehobbit85: Correct, I'll fix it.

src/locales/en_US.js
thehobbit85: "Is 24 hours currency-specific?" I don't think so.
I do recall seeing longer times but never shorter, so it's better be safe.

And I kinda like the 'and/or', very accurate ;)

src/components/scenes/SendScene.js
thehobbit85: Good catch

PR #3074, opened 2022-01-11
Show bundle purchase reminder if attempting to send request with only one bundle remaining

peachbits:
FIO requests can cost "up to two transactions" but the API won't return a new bundle fee unless you're at 0 remaining transactions. This adds a check for the corner case. We can leave the fee check as that would provide the most up to date information in the small chance state is out of date.


peachbits:
Reference https://developers.fioprotocol.io/docs/fio-protocol/fio-fees

thehobbit85 (approved):

PR #3095, opened 2022-01-15
Allow deletion of broken wallets

paullinator:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1201676547072842


thehobbit85 (commented):
Do we want to show the "Confirm delete wallet" modal for the "rawDelete" option or just archive the wallet without prompting the user?

paullinator:
@thehobbit85 Just archive the Wallet immediately. Likely the wallet is broken anyway and the user can always recover it in the future if really needed.

thehobbit85 (approved):

PR #3107, opened 2022-01-21
Add settings support for duplicate currency codes

peachbits:


peachbits:
In addition to fixing tests and custom tokens I'll need to redo these terrible commit messages

PR #3086, opened 2022-01-13
FIO staking scenes.

andreyvEze:
#### PR Requirements

If you have made **any** visual changes to the GUI. Make sure you have:
- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

PR #3114, opened 2022-01-25
FIO Staking

samholmes:
Replaces: https://github.com/EdgeApp/edge-react-gui/pull/3086

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1200689708225859


samholmes:
Known things to be aware for changes (quoted from Andrey):

> here last things I think need to be done I mentioned Eliran about:
> * please check how I used FlipInput and decide how better to refactor it (I started with sendConfirmation redux state and it was a big mess, so I made it in other way). Main issue with the implementation is that the amount state is not persisted between mount/unmount of the FlipInput component's life-cycle.
> * the same for makeSpend/broadcast: the architecture here is questionable about whether there is refactor needed here.
> * proper error handling (error message on stake/unstake scene)
> * there is something wrong with `stakingStatus` updates (it is not changed on GUI side, maybe something wrong with plugin interaction)

src/components/Main.ui.js
```diff
@@ -366,19 +366,6 @@ export class MainComponent extends React.Component<Props> {
...
-                onEnter={props => {
-                  this.props.requestPermission('camera')
-                  this.props.dispatchEnableScan()
-                }}
-                onExit={this.props.dispatchDisableScan}
-                component={ifLoggedIn(ScanScene)}
-                renderLeftButton={<BackButton onPress={this.handleBack} />}
-                renderRightButton={<SideMenuButton />}
-              />
-
```
samholmes: Why is this removed? I want more context to this.

andreyvEze: @samholmes because it caused build error and not used, also we don't have `SCAN` const. and could you explain why it was added here - 418baae83494db8790f71872af7162e22d3cec25 ? I assume it was accidentally

samholmes: I don't have the context as to where it is there if it's not used. I'll have to punt this question to other members of the team to determine if it's the right thing to remove.

samholmes: Resolved: It was an artifact of rebasing the old/original FIO Staking Scenes PR. It should be removed along with a couple of other scenes (fixed here https://github.com/EdgeApp/edge-react-gui/pull/3118).

PR #3122, opened 2022-01-27
FIO staking scenes fixes

andreyvEze:


src/components/scenes/StakingChangeScene.js
```diff
@@ -96,7 +96,7 @@ export const StakingChangeSceneComponent = (props: Props) => {
 
     // If the selectedFioAddress is not defined, then we will not be able to complete the transaction.
     if (selectedFioAddress == null) {
-      setError(new Error('Unable to stake without a FIO addresses'))
+      setError(new Error(s.strings.staking_no_fio_address_error))
```
samholmes, line 101: 👍 

_I should have done this from the beginning._

src/components/scenes/StakingChangeScene.js
```diff
@@ -205,15 +205,18 @@ export const StakingChangeSceneComponent = (props: Props) => {
   }, [amount, currencyPlugin])
 
   useEffect(() => {
-    if (!selectedFioAddress) {
-      const fioAddress = fioAddresses.find(({ walletId: fioAddressWalletId }) => fioAddressWalletId === walletId)
+    if (!selectedFioAddress && fioAddresses != null && fioAddresses.length > 0) {
```
samholmes, line 210: `fioAddresses` should never be nullish according to the prop's type. We can remove the null-check, and if you're concerned that the static types don't give use guarantees about it not being nullish, use `?.length`.

src/components/scenes/StakingChangeScene.js
```diff
@@ -205,15 +205,18 @@ export const StakingChangeSceneComponent = (props: Props) => {
...
+      const fioAddress = fioAddresses
+        .filter(({ walletId: fioAddressWalletId }) => fioAddressWalletId === walletId)
+        .sort(({ bundledTxs }, { bundledTxs: bundledTxs2 }) => (bundledTxs <= bundledTxs2 ? 1 : -1))[0]
 
       // If no address is found, we do not define the selectedFioAddress
       if (fioAddress == null) return
+      if (fioAddress.bundledTxs < 2) return setError(new Error(s.strings.staking_no_bundled_txs_error))
```
samholmes, line 217: Add a comment explaining the purpose of this line.

src/components/scenes/StakingOverviewScene.js
```diff
@@ -80,6 +76,18 @@ export const StakingOverviewSceneComponent = (props: Props) => {
...
+  currencyWallet.watch('stakingStatus', (stakingStatus: EdgeStakingStatus) => {
+    formatLocks(stakingStatus)
+  })
+
+  useEffect(() => {
+    refreshAllFioAddresses()
+  }, [refreshAllFioAddresses])
+
+  useEffect(() => {
+    formatLocks(stakingStatus)
```
samholmes, line 90:
```js
const [stakingStatus, setStakingStatus] = useState(currencyWallet.stakingStatus)

useEffect(() => {
    // Inline formatLocks again (no more abstraction)
 }, [stakingStatus, currencyDenomination])
 
useEffect(() => {
  currencyWallet.watch('stakingStatus', (stakingStatus: EdgeStakingStatus) => {
    setStakingStatus(stakingStatus)
  })
  // return an unwatch cleanup if the core provides that API
}, [currencyWallet])
```

src/components/scenes/StakingChangeScene.js
```diff
@@ -221,7 +223,7 @@ export const StakingChangeSceneComponent = (props: Props) => {
   const sliderDisabled = tx == null || amount == null || amount === '0' || error != null
 
   const renderAdd = () => {
-    const apyValue = sprintf(s.strings.staking_estimated_return, `${apy}%`)
+    const apyValue = sprintf(maxApy != null && apy === maxApy ? s.strings.staking_estimated_return_up_to : s.strings.staking_estimated_return, `${apy}%`)
```
samholmes, line 226: If `apy` is never nullish, then you can shorten it:

```js
sprintf(apy === maxApy ? s.strings.staking_estimated_return_up_to : s.strings.staking_estimated_return, `${apy}%`)
```

But if `apy` is ever nullish, then why do we always allow the string interpolation?

samholmes (changes requested):

src/components/scenes/StakingChangeScene.js
samholmes:
```js
// Restrict staking to addresses with only 2 or more bundled transactions to avoid user needing to pay fee
```

samholmes: Double-check what the minimum bundleTx value is (1 or 2)

samholmes: Move to accountbased instead of in the GUI (if it's not too difficult).

src/locales/strings/enUS.json
```diff
@@ -309,6 +309,8 @@
   "staking_unstake_success": "Successfully unstaked",
   "staking_estimated_rewards": "Estimated Rewards",
   "staking_estimated_return": "Estimated Return: %1s APY",
+  "staking_no_fio_address_error": "Unable to stake without a FIO addresses",
+  "staking_no_bundled_txs_error": "There are not enough bundled transactions to proceed",
```
samholmes, line 314: "Not enough bundled transactions for FIO address"

samholmes: Or use `fio_no_bundled_add_err_msg`

src/components/scenes/StakingChangeScene.js
samholmes: We're not going to worry about moving the `bundledTxs` logic to the plugin for now.

PR #3127, opened 2022-01-29
Jon/fix/bit pay v2 fees

Jon-edge:
#### PR Requirements

- [ x] Tested on iOS Tablet
- [x ] Tested on small Android
- [ ] n/a


src/types/BitPayError.js
```diff
@@ -31,10 +33,12 @@ export const BitPayErrorCode = {
...
   [BitPayErrorCode.EmptyOutputInvoice]: () => s.strings.error_bitpay_empty_output_invoice,
   [BitPayErrorCode.EmptyVerificationHexReq]: () => s.strings.error_bitpay_empty_verification_hex_req,
-  [BitPayErrorCode.FetchFailed]: (params: { header: string, statusCode: string, text: string }) =>
-    sprintf(s.strings.error_bitpay_fetch, params.header, params.statusCode, params.text),
+  [BitPayErrorCode.FetchFailed]: (params: { header: string, statusCode: string, text?: string }) =>
+    sprintf(s.strings.error_bitpay_fetch, params.header, params.statusCode, params.text ?? ''),
+  [BitPayErrorCode.InvalidPaymentOption]: (params: { text: string }) => sprintf(s.strings.error_bitpay_invalid_payment_option, params.text),
   [BitPayErrorCode.MultiOutputInvoice]: () => s.strings.error_bitpay_multi_output_invoice,
   [BitPayErrorCode.MultiInstructionInvoice]: () => s.strings.error_bitpay_multi_tx_invoice,
+  [BitPayErrorCode.InvalidPaymentOption]: (params: { text: string }) => sprintf(s.strings.error_bitpay_invalid_payment_option, params.text),
```
peachbits, line 41: This is a duplicate key. Probably meant NoPaymentOption?

src/actions/BitPayActions.js
```diff
@@ -117,30 +147,38 @@ export async function launchBitPay(dispatch: Dispatch, uri: string, currencyWall
...
+      paymentProtocolInfo: {
+        merchant: {
+          requiredFeeRate
+        }
+      }
+    }
   }
-  const wallet = currencyWallets[walletId]
-  const unsignedTx = await wallet.makeSpend(spendInfo)
+  // $FlowFixMe
```
peachbits, line 167: Flow doesn't like optional chaining when calling methods but checking for null beforehand should be enough. Line 118 should be sufficient to satisfy flow and you can remove the ?s (but we don't like sketchy nulls so it should be if (selectedWallet == null) return

src/actions/BitPayActions.js
```diff
@@ -117,30 +147,38 @@ export async function launchBitPay(dispatch: Dispatch, uri: string, currencyWall
...
+  const unsignedTx = await selectedWallet?.makeSpend(spendInfo)
   const unsignedHex = unsignedTx.otherParams?.txJson?.hex ?? ''
-  const signedTx = await wallet.signTx(unsignedTx)
+  // $FlowFixMe
+  const signedTx = await selectedWallet?.signTx(unsignedTx)
   const signedHex = signedTx.signedTx ?? ''
 
-  errorData = { ...errorData, spendInfo, walletId, unsignedTx, signedTx }
+  errorData = { ...errorData, spendInfo, selectedWallet, unsignedTx, signedTx }
```
peachbits, line 174: An EdgeCurrencyWallet has keys so it's a bad idea to attach it to an object that may be logged. Passing the walletId is a better option.

src/components/themed/AddressTile.js
```diff
@@ -109,11 +109,7 @@ class AddressTileComponent extends React.PureComponent<Props, State> {
 
       // Check is PaymentProtocolUri
       if (!!parsedUri.paymentProtocolURL && !parsedUri.publicAddress) {
-        const guiMakeSpendInfo: ?GuiMakeSpendInfo = await paymentProtocolUriReceived(parsedUri, coreWallet)
-
-        if (guiMakeSpendInfo) {
-          onChangeAddress(guiMakeSpendInfo)
-        }
+        await launchBitPay(parsedUri.paymentProtocolURL, { wallet: coreWallet }).catch(showError)
```
peachbits, line 112: This change appears to break payment protocol v1 support if you're pasting (or god forbid) typing in the URI. We still support v1 in the scan functionality so it should still be supported here as well.

peachbits (changes requested):

PR #3167, opened 2022-02-16
Refactor Splash Screen

thehobbit85:
1. Updated iOS to use storyboard instead of xib for the splash screen and use a higher res image.
2. Updated both iOS and Android to not show the white "loading" screen without using react-native-smart-splash-screen.
3. Remove react-native-smart-splash-screen

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

https://user-images.githubusercontent.com/4012041/154242189-0fc20533-2f22-487f-bafa-376b95f74bd4.mov

---
  - https://app.asana.com/0/0/1201269161541987


src/components/services/EdgeCoreManager.js
```diff
@@ -94,32 +92,17 @@ export class EdgeCoreManager extends React.PureComponent<Props, State> {
...
-    this.setState(
-      state => ({ context, counter: state.counter + 1 }),
-      () => this.hideSplash()
-    )
+    this.setState(state => ({ context, counter: state.counter + 1 }))
   }
 
   onError = (error: Error) => {
     console.log('EdgeContext failed', error)
-    this.hideSplash()
```
swansontec, line 122: Where do we animate the app's opacity, so everything is transparent until the core first loads, and then fades in to the login screen?

android/app/src/main/res/values/styles.xml
```diff
@@ -4,9 +4,7 @@
     <style name="AppTheme" parent="Theme.AppCompat.DayNight.NoActionBar">
         <!-- Customize your theme here. -->
         <item name="android:textColor">#000000</item>
-        <!-- Doing <item name="android:windowIsTranslucent">true</item>
-          as requested by react-native-smart-splash-screen crashes on Oreo
-          because we also request portrait mode. -->
+        <item name="android:windowDisablePreview">true</item>
```
swansontec, line 7: What does this new line do?

swansontec (commented):
I have some questions about this. It mostly looks good, but I think we may be regressing on the first-launch animation.

src/components/services/EdgeCoreManager.js
thehobbit85: On iOS we don't need this anymore. The new storyboard combined with the changes I made to the AppDelegate file makes this unnecessary.

https://reactnative.dev/docs/publishing-to-app-store#pro-tips

android/app/src/main/res/values/styles.xml
thehobbit85: This came from here: https://stackoverflow.com/a/56803562

PR #3212, opened 2022-03-17
JSON Payment Protocol V2 - Transmit Signed Tx - Multiple Outputs

stevenzeiler:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/actions/BitPayActions.js
+++ b/src/actions/BitPayActions.js
@@ -147,23 +147,19 @@ export async function launchBitPay(
   }
   const invoiceInstruction = invoiceResponse.instructions[0]
   errorData = { ...errorData, invoiceInstruction }
-  if (invoiceInstruction.outputs) {
-    if (invoiceInstruction.outputs.length > 1) {
-      throw new BitPayError('MultiOutputInvoice', { errorData })
-    }
-  } else throw new BitPayError('EmptyOutputInvoice', { errorData })
-  const instructionOutput = invoiceInstruction.outputs[0]
-
+  if (!invoiceInstruction.outputs || invoiceInstruction.outputs.length === 0) {
+    throw new BitPayError('EmptyOutputInvoice', { errorData })
+  }
   // Make the spend to generate the tx hexes
   const requiredFeeRate = invoiceInstruction.requiredFeeRate
   const spendInfo: EdgeSpendInfo = {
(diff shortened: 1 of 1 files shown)


src/actions/BitPayActions.js
```diff
@@ -147,23 +147,19 @@ export async function launchBitPay(
...
   const invoiceInstruction = invoiceResponse.instructions[0]
   errorData = { ...errorData, invoiceInstruction }
-  if (invoiceInstruction.outputs) {
-    if (invoiceInstruction.outputs.length > 1) {
-      throw new BitPayError('MultiOutputInvoice', { errorData })
-    }
-  } else throw new BitPayError('EmptyOutputInvoice', { errorData })
-  const instructionOutput = invoiceInstruction.outputs[0]
-
+  if (!invoiceInstruction.outputs) {
```
paullinator, line 150: Need to check `outputs.length > 1`

paullinator (changes requested):

src/actions/BitPayActions.js
stevenzeiler: Do you mean > 0?

paullinator: Ah yes

stevenzeiler: Okay it now checks for an empty array of outputs.

src/actions/BitPayActions.js
```diff
@@ -225,6 +221,20 @@ export async function launchBitPay(
...
+        fetchBitPayJsonResponse(uri, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/payment',
+            'x-paypro-version': '2'
+          },
+          body: JSON.stringify({
+            chain: requestCurrencyCode,
+            currency: requestCurrencyCode,
+            transactions: [{ tx: signedHex, weightedSize: signedHex.length / 2 }]
```
stevenzeiler, line 234: It appears that the `signedHex` which is transmitted here results in a different txid than the one Edge displays in the payment history.

That could possibly be the cause of the tx-mempool-conflict I am receiving when pushing transactions to the blockchair API.

paullinator (approved):

PR #3227, opened 2022-03-30
Patch/stake flip input

samholmes:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a

---
  - https://app.asana.com/0/0/1202037902009665

### DIFF
--- a/src/components/modals/FlipInputModal.js
+++ b/src/components/modals/FlipInputModal.js
@@ -31,7 +31,7 @@ import { MiniButton } from '../themed/MiniButton.js'
 import { ThemedModal } from '../themed/ThemedModal.js'
 
 type OwnProps = {
-  bridge: AirshipBridge<void>,
+  bridge: AirshipBridge<{ nativeAmount: string, exchangeAmount: string }>,
   walletId: string,
   currencyCode: string,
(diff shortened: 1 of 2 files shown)


src/components/modals/FlipInputModal.js
```diff
@@ -85,11 +88,16 @@ class FlipInputModalComponent extends React.PureComponent<Props, State> {
...
+      forceUpdateGuiCounter: 0,
+      nativeAmount: '',
+      exchangeAmount: ''
     }
   }
 
-  handleCloseModal = () => this.props.bridge.resolve()
+  handleCloseModal = () => {
+    const { nativeAmount, exchangeAmount } = this.state
+    this.props.bridge.resolve(Promise.resolve({ nativeAmount, exchangeAmount }))
```
thehobbit85, line 99: Can this modal close when the nativeAmount and/or exchangeAmount are still equal to empty string?
And if so, what would happen with whoever calls this modal? Can they handle the empty string?

thehobbit85 (changes requested):

thehobbit85 (approved):
/rebase-autosquash

PR #3231, opened 2022-04-01
Remove guiwallet from EdgeProvider, Scan, and SendConfirmation

paullinator:
#### PR Requirements

- [] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1202060614291040

### DIFF
--- a/src/modules/UI/scenes/Plugins/EdgeProvider.js
+++ b/src/modules/UI/scenes/Plugins/EdgeProvider.js
@@ -1,7 +1,7 @@
 // @flow
 
 import { abs } from 'biggystring'
-import type { EdgeCurrencyWallet, EdgeMetadata, EdgeNetworkFee, EdgeSpendTarget, EdgeTransaction, JsonObject } from 'edge-core-js'
+import type { EdgeCurrencyWallet, EdgeMetadata, EdgeNetworkFee, EdgeReceiveAddress, EdgeSpendTarget, EdgeTransaction, JsonObject } from 'edge-core-js'
 import * as React from 'react'
 import { Linking } from 'react-native'
 import Mailer from 'react-native-mail'
@@ -19,13 +19,8 @@ import s from '../../../../locales/strings'
 import { type GuiPlugin, type GuiPluginQuery } from '../../../../types/GuiPluginTypes.js'
 import { type Dispatch, type RootState } from '../../../../types/reduxTypes.js'
 import { Actions } from '../../../../types/routerTypes.js'
-import { type GuiMakeSpendInfo, type GuiWallet } from '../../../../types/types.js'
+import { type GuiMakeSpendInfo } from '../../../../types/types.js'
 
-type EdgeReceiveAddress = {
-  publicAddress?: string,
-  segwitAddress?: string,
-  legacyAddress?: string
-}
 type WalletDetails = {
   name: string,
   receiveAddress: {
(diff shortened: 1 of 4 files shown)


src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -133,29 +129,32 @@ export class EdgeProvider extends Bridgeable {
...
       name: walletName,
-      receiveAddress: wallet.receiveAddress,
-      currencyCode: currentCode,
-      fiatCurrencyCode: wallet.fiatCurrencyCode,
-      currencyIcon: wallet.symbolImage,
-      currencyIconDark: wallet.symbolImageDarkMono
+      receiveAddress,
+      currencyCode,
+      fiatCurrencyCode: isoFiatSplatter(coreWallet.fiatCurrencyCode).fiatCurrencyCode,
+      currencyIcon: coreWallet.currencyInfo.symbolImage, // TODO: get images for tokens
```
thehobbit85, line 156: We have a function you can use to coverts a pluginId + contractAddress (if it's a token) to the correct icon links: https://github.com/EdgeApp/edge-react-gui/blob/64c512fe17bc196aa8b1e9aed9215a719967edd3/src/util/CurrencyInfoHelpers.js#L21

src/util/utils.js
```diff
@@ -18,6 +18,14 @@ import { getWalletFiat } from '../util/CurrencyWalletHelpers.js'
 export const DECIMAL_PRECISION = 18
 export const DEFAULT_TRUNCATE_PRECISION = 6
 
+export function isoFiatSplatter(cc: string): { isoFiatCurrencyCode: string, fiatCurrencyCode: string } {
```
thehobbit85, line 21: We already have a function that adds an iso if needed: https://github.com/EdgeApp/edge-react-gui/blob/64c512fe17bc196aa8b1e9aed9215a719967edd3/src/util/utils.js#L273
And removing it is just `code.replace('iso:', '')` so not sure that we even need a function for that.

thehobbit85 (changes requested):
Those are VERY welcomed changes.

But since you removed the `guiWallet`, why not call the `coreWallet` just `wallet` since there is only 1 now?
Or even use the name `edgeWallet` as it's closer to the type name.

src/modules/UI/scenes/Plugins/EdgeProvider.js
paullinator: Single English word variable names such as wallet are very ambiguous. I am open to edgeWallet which I am just now noticing is used in other places.

src/util/utils.js
paullinator: We are very inconsistent in the use of fiatCurrencyCode vs isoFiatCurrencyCode. Especially when used in the core versus the GUI Wallet. This makes it really easy to accidentally use one versus another.

The new routine is the most flexible in that it can take either format and output both formats in one line.

src/modules/UI/scenes/Plugins/EdgeProvider.js
paullinator: I was only trying to remove the use of GUI Wallet and not change additional functionality like how we get icons

paullinator: That deserves a different PR

src/util/utils.js
thehobbit85: This doesn't help and just adds extra code and unnecessary complexity.

You shifted the problem from needing to know which function to call to needing knowing which value you need to destruct from the returned value.

You still have the exact same problem only with an extra step and in a more implicit way which actually makes it actually harder to debug and find errors.

If you really want to solve the issue, then we need to only use the iso code in the GUI and then we won't even an issue.

thehobbit85 (changes requested):

PR #3235, opened 2022-04-02
Remove receive address from redux

paullinator:
Perf optimization to prevent needing to call getReceiveAddress() for all wallets each time a wallet update occurs. 

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1202060614291040

### DIFF
--- a/src/components/scenes/RequestScene.js
+++ b/src/components/scenes/RequestScene.js
@@ -15,10 +15,10 @@ import { refreshReceiveAddressRequest, selectWalletFromModal } from '../../actio
 import { getSpecialCurrencyInfo, SPECIAL_CURRENCY_INFO } from '../../constants/WalletAndCurrencyConstants.js'
 import s from '../../locales/strings.js'
 import { getDisplayDenomination, getExchangeDenomination } from '../../selectors/DenominationSelectors.js'
-import { getExchangeRate, getSelectedWallet } from '../../selectors/WalletSelectors.js'
+import { getExchangeRate } from '../../selectors/WalletSelectors.js'
 import { connect } from '../../types/reactRedux.js'
 import { type NavigationProp } from '../../types/routerTypes.js'
-import type { GuiCurrencyInfo, GuiDenomination, GuiWallet } from '../../types/types.js'
+import type { GuiCurrencyInfo, GuiDenomination } from '../../types/types.js'
 import { getCurrencyIcon } from '../../util/CurrencyInfoHelpers.js'
 import { getAvailableBalance, getWalletName } from '../../util/CurrencyWalletHelpers.js'
 import { convertNativeToDenomination, getCurrencyInfo, getDenomFromIsoCode, getObjectDiff, truncateDecimals } from '../../util/utils.js'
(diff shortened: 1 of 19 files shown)


src/components/themed/WalletListSwipeRow.js
```diff
@@ -70,7 +76,7 @@ function WalletListSwipeRowComponent(props: Props) {
       guiWallet != null &&
       // Some wallets launch an activation screen when selected,
       // so avoid going to the transaction list in that case:
-      (isToken || !getSpecialCurrencyInfo(guiWallet.type).isAccountActivationRequired || guiWallet.receiveAddress.publicAddress)
+      (isToken || !getSpecialCurrencyInfo(edgeWallet.type).isAccountActivationRequired || walletAddress !== '')
```
swansontec, line 79: This is probably not the best solution. The real issue is that sometimes `dispatch(selectWallet)` on the line above will launch the wallet activation scene. In that case, we shouldn't launch the transaction list scene too. If the `selectWallet` action could be modified to return a boolean, we could avoid loading the wallet address altogether in this component.

swansontec (changes requested):
Found a few things, as well as one existing bug that isn't even part of this PR.

src/components/scenes/RequestScene.js
```diff
@@ -127,10 +124,20 @@ export class RequestComponent extends React.Component<Props, State> {
 
   async generateEncodedUri() {
     const { edgeWallet, useLegacyAddress, currencyCode } = this.props
+    let legacyAddress = ''
+    let publicAddress = ''
+    if (edgeWallet != null) {
+      const receiveAddress = await edgeWallet.getReceiveAddress()
```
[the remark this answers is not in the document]
paullinator: Can't do it here due to scope

src/components/themed/WalletListSwipeRow.js
paullinator: This is a good point, but seems out of scope for this change. I don't believe we have a `walletActivated` boolean in the plugin or core. We are for now relying on the wallet having an address to determine activation.

src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -140,20 +141,24 @@ export class EdgeProvider extends Bridgeable {
   async getCurrentWalletInfo(): Promise<WalletDetails> {
     const edgeWallet = this._state.core.account.currencyWallets[this._state.ui.wallets.selectedWalletId]
     const currencyCode = this._state.ui.wallets.selectedCurrencyCode
-    let walletName = edgeWallet.name ? edgeWallet.name : ''
+    let walletName = edgeWallet.name ?? ''
     const receiveAddress = await edgeWallet.getReceiveAddress()
     const enabledTokens = await edgeWallet.getEnabledTokens()
+    const contractAddress = edgeWallet.currencyInfo.metaTokens.find(token => token.currencyCode === currencyCode)?.contractAddress
```
swansontec, line 147: This doesn't account for custom tokens, but then again, those don't have images on our system anyhow. This code is fine, and gets the job done.

The `currencyInfo.metaTokens` field is deprecated, though. The information has moved to the `EdgeCurrencyConfig` object, which in this context would be accessed as:

```js
const currencyConfig = this._state.core.account.currencyConfig[edgeWallet.currencyInfo.pluginId]
currencyConfig.builtinTokens // map of tokenId -> EdgeToken
currencyConfig.customTokens // map of tokenId -> EdgeToken
```

I have a task to make it available as `EdgeCurrencyWallet.currencyConfig`, which would be much shorter to type than this crazy-long thing.

swansontec (approved):

PR #3247, opened 2022-04-05
Remove GUI wallet and reduce unnecessary renders

thehobbit85:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


src/components/themed/WalletListRow.js
```diff
@@ -1,80 +1,73 @@
...
 
-type OwnProps = {
+type Props = {
   currencyCode: string,
+  exchangeRateText?: string | null,
+  exchangeRateType?: null | string, // Should be any of this values: 'neutral' | 'positive' | 'negative'
   children?: React.Node,
   icon?: React.Node,
+  iconUri?: string,
+  iconSize?: number,
```
swansontec, line 19: This should be called `iconSizeRem`. We are consistent with this naming, so it's clear which ones are in rem and which ones are in points. No suffix = points.

src/__tests__/components/WalletListCreateRow.test.js
```diff
@@ -7,7 +7,7 @@ import ShallowRenderer from 'react-test-renderer/shallow'
 import { getTheme } from '../../components/services/ThemeContext.js'
 import { WalletListCreateRowComponent } from '../../components/themed/WalletListCreateRow.js'
 
-describe('WalletListCreateRow', () => {
+describe.skip('WalletListCreateRow', () => {
```
swansontec, line 10: Can we re-enable the test? If not, let's delete it.

src/actions/CreateWalletActions.js
```diff
@@ -16,26 +16,30 @@ import type { Dispatch, GetState } from '../types/reduxTypes.js'
...
+  fiatCurrencyCode?: string,
+  importText?: string // for creating wallet from private seed / key
+}
+
+export const createWallet = async (account: EdgeAccount, { walletType, walletName, fiatCurrencyCode, importText }: CreateWalletOptions) => {
+  // Try and get the new format param from the legacy walletType if it's mentioned
+  const [type, format] = walletType.split('-')
+  const opts = {
+    name: walletName,
+    fiatCurrencyCode: fiatCurrencyCode,
```
swansontec, line 31: This is literally the same thing?

src/components/themed/WalletListCreateRow.js
```diff
@@ -1,185 +1,131 @@
...
+export const WalletListCreateRowComponent = (props: WalletListCreateRowProps) => {
+  const dispatch = useDispatch()
+  const theme = useTheme()
+  const styles = getStyles(theme)
+  const { currencyCode = '', currencyName = '', walletType, symbolImage = '', symbolImageDarkMono = '', parentCurrencyCode, onPress } = props
+
+  const handlePress = useCallback(() => {
+    const handleRes = walletId => (onPress != null ? onPress(walletId, currencyCode) : null)
+    if (walletType != null) dispatch(createAndSelectWallet({ walletType })).then(handleRes)
+    if (parentCurrencyCode != null) dispatch(createAndSelectToken({ currencyCode, parentCurrencyCode })).then(handleRes)
```
swansontec, line 106: The two `if` statements should be mutually exclusive, so we need an `else` here. Otherwise, `onPress` might be called twice.

src/__tests__/components/WalletListCurrencyRow.test.js
```diff
@@ -7,7 +7,7 @@ import ShallowRenderer from 'react-test-renderer/shallow'
 import { getTheme } from '../../components/services/ThemeContext.js'
 import { WalletListRowComponent } from '../../components/themed/WalletListCurrencyRow.js'
 
-describe('WalletListRow', () => {
+describe.skip('WalletListRow', () => {
```
swansontec, line 10: Same thing. Fix or delete the test.

swansontec (changes requested):
I found a number of small issues in the review. We also need to worry about a new problem  - it is possible to tap on a wallet that doesn't have a `GuiWallet` stored in state, whereas before we showed a spinner for this case. I don't know how easy this is to do, but if users do manage it, random other scenes may crash.

src/components/themed/WalletListCurrencyRow.js
```diff
@@ -1,228 +1,203 @@
...
+
+  return (
+    <WalletListRow
+      currencyCode={currencyCode}
+      exchangeRateText={showRate ? exchangeRateText : null}
+      exchangeRateType={showRate ? exchangeRateType : null}
+      icon={icon}
+      onPress={onPress}
+      onLongPress={onLongPress}
+      walletName={walletName ?? name ?? `My ${getCurrencyInfo(allCurrencyInfos, currencyCode)?.displayName ?? ''}`}
```
swansontec, line 188: We can ditch `getCurrencyInfo` and `allCurrencyInfos`, since it's just `edgeWallet.currencyInfo`.

paullinator:
> I found a number of small issues in the review.
> 
> We also need to worry about a new problem - it is possible to tap on a wallet that doesn't have a `GuiWallet` stored in state, whereas before we showed a spinner for this case. I don't know how easy this is to do, but if users do manage it, random other scenes may crash. I think we need to bite the bullet on this one and just deal with the fallout.

Fair enough. Hopefully the delay in tapping a wallet then going to another screen like Send or Request is slow enough to get the guiwallet up and running. Of course longer term we need to just get rid of guiwallet in those scenes too. It's possible those scenes also have a check for `guiwallet != null` ::crossing fingers::

src/components/themed/ManageTokensRow.js
```diff
@@ -33,7 +33,7 @@ export function ManageTokensRow(props: Props) {
   const isEditable = metaTokens.every(token => token.currencyCode !== currencyCode)
 
   const onPress = () => {
-    isEditable ? goToEditTokenScene(currencyCode) : (() => {})()
+    if (isEditable) goToEditTokenScene(currencyCode)
```
swansontec, line 36: This should have been done in `fixup! Remove a few unused or unneeded util functions`, but I'll take it.

swansontec (approved):
Approved. Now to squash & merge.

PR #3249, opened 2022-04-06
William/fix sides

swansontec:
Based on our conversation earlier today, I have gone ahead and eliminated `unpackEdges`. Whether we go the next step to a hook is up to @thehobbit85 - we are already grabbing the theme in most cases, so a hook might be overkill.

#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1202157545945305


src/components/themed/Alert.js
```diff
@@ -52,17 +52,17 @@ export const useAlertTypeIcon = (type: AlerType, color: string, size: number, st
 
 export function Alert({ type, title, message, numberOfLines = 2, marginRem, paddingRem, onPress, children }: Props) {
   const theme = useTheme()
+  const margin = sidesToMargin(mapSides(fixSides(marginRem, 0), theme.rem))
```
thehobbit85, line 55: Shouldn't this be set to 1 rem like on the left?

src/components/themed/Alert.js
```diff
@@ -52,17 +52,17 @@ export const useAlertTypeIcon = (type: AlerType, color: string, size: number, st
...
+  const padding = sidesToPadding(mapSides(fixSides(paddingRem, 0.5), theme.rem))
```
thehobbit85, line 56: Shouldn't this be set to 1 rem like on the left?

thehobbit85 (changes requested):

src/components/themed/Alert.js
swansontec: Oh, darn! I didn't realize the `useSpacing` hook had 1 and 1 as the defaults.

thehobbit85 (approved):

PR #3250, opened 2022-04-08
Eliran/icon fixes

swansontec:
@thehobbit85 I rebased your branch on top of develop, removing the commits we talked about. If you do another `/rebase-autosquash`, it will remove my icon from the commits.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1200962167853001


src/components/scenes/WcConnectScene.js
```diff
@@ -126,7 +125,7 @@ export const WcConnectScene = (props: Props) => {
       return <SelectableRow onPress={showWalletListModal} title={s.strings.wc_confirm_select_wallet} arrowTappable />
     } else {
       const walletNameStr = truncateString(walletName || '', MAX_ADDRESS_CHARACTERS)
-      const walletImage = <FastImage style={styles.currencyLogo} source={{ uri: walletImageUri }} />
+      const walletImage = <CurrencyIcon tokenId={contractAddress} walletId={selectedWallet.walletId} resizeMode="contain" />
```
swansontec, line 128: Please do not pass the contract address as the tokenId. I don't want landmines scattered throughout the program. Only the currency plugin has the ability to pick a tokenId. Everyone else has to look it up from the core.

src/modules/FioAddress/components/ConnectWallets.js
```diff
@@ -141,8 +141,8 @@ class ConnectWallets extends React.Component<Props, LocalState> {
         <View style={[styles.wallet, disabled ? styles.walletDisabled : null]} underlayColor={theme.secondaryButton}>
           <View style={styles.rowContainerTop}>
             <View style={styles.containerLeft}>
-              {wallet.symbolImage ? (
-                <FastImage style={styles.imageContainer} source={{ uri: wallet.symbolImage }} resizeMode="contain" />
+              {wallet != null ? (
+                <CurrencyIcon walletId={wallet.id} currencyCode={wallet.currencyCode} resizeMode="contain" />
```
swansontec, line 145: I don't think we want the spinner here either, so let's do the pluginId.

src/components/themed/CryptoExchangeFlipInputWrapperComponent.js
```diff
@@ -55,11 +54,11 @@ export class CryptoExchangeFlipInputWrapperComponent extends React.Component<Pro
     this.props.onCryptoExchangeAmountChanged(amounts)
   }
 
-  renderLogo = (logo: string) => {
+  renderLogo = () => {
     const styles = getStyles(this.props.theme)
     return (
       <View style={styles.iconContainer}>
-        <FastImage style={styles.currencyIcon} source={{ uri: logo || '' }} />
+        <CurrencyIcon walletId={this.props.walletId} sizeRem={1.75} />
```
swansontec, line 61: I think the spinner would look weird here too. Let's use the pluginId instead.

src/components/modals/WalletListMenuModal.js
```diff
@@ -101,27 +101,20 @@ export function WalletListMenuModal(props: Props) {
...
-  let walletName: string | void
-  if (guiWallet != null) {
-    walletName = guiWallet.name
-
-    const { metaTokens } = guiWallet
-    const contractAddress = metaTokens.find(token => token.currencyCode === currencyCode)?.contractAddress
-    image = getCurrencyIcon(getPluginId(guiWallet.type), contractAddress).symbolImage
-  }
+  const walletName = edgeWallet?.name ?? ''
+  const { contractAddress } = edgeWallet?.currencyInfo.metaTokens.find(token => token.currencyCode === currencyCode) ?? {}
```
swansontec, line 111: This line doesn't pass Flow. I get "Flow does not yet support method or property calls in optional chains." In any case, passing the contract address as a tokenId is not a valid assumption, and I don't want our GUI to randomly break in a few years when we add some weird coin where that's not true.

swansontec (commented):
Here are the things I have found so far. I am working on a fixup commit that fixes the circle animation problems, but the baby woke up before I could get it done.

src/components/scenes/WcConnectScene.js
thehobbit85: It's temporary. 
We need to update everywhere to use pluginId and tokenId instead of currencyCode and contractAddress.
So for now it works but we would change it very soon.

src/components/themed/CurrencyIcon.js
```diff
@@ -0,0 +1,126 @@
...
+  }
+  // If we still don't have a pluginId, try to get a pluginId and tokenId for a token
+  if (pluginId == null) {
+    pluginId = Object.keys(account.currencyConfig).find(id => {
+      tokenId = getTokenId(account, id, currencyCode)
+      return tokenId != null
+    })
+  }
+  // Assume a tokenId can be a contract address for now until we start using the correct ones
+  if (tokenId != null) tokenId = removeHexPrefix(tokenId)
```
thehobbit85, line 50: Remove this hack

src/actions/WalletActions.js
```diff
@@ -48,7 +48,7 @@ export const selectWallet = (walletId: string, currencyCode: string, alwaysActiv
       data: { walletId, currencyCode }
     })
     const wallet: EdgeCurrencyWallet = currencyWallets[walletId]
-    wallet
+    return wallet
```
thehobbit85, line 51: Add an await to line 39

src/components/themed/WalletSyncCircle.js
```diff
@@ -0,0 +1,88 @@
...
+      position: 'absolute',
+      top: -2 * strokeWidth,
+      left: -2 * strokeWidth
+    }),
+    [strokeWidth]
+  )
+
+  return (
+    <Svg height={svgSize} width={svgSize} viewBox={viewBox} style={svgStyle}>
+      <AnimatedCircle cx={center} cy={center} r={radius} strokeDasharray={circumference} strokeWidth={strokeWidth} animatedProps={animatedProps} />
```
thehobbit85, line 85: Rounded corners

thehobbit85 (changes requested):

thehobbit85 (approved):

PR #3273, opened 2022-04-13
Prevent crash when archiving wallet

peachbits:
The bug is that the walletId still exists after archiving but the guiWallet does not so it's the wrong check in the if. But since we hate the guiWallet it's for the best to remove that too.


peachbits:
closing in favor of https://github.com/EdgeApp/edge-react-gui/pull/3250

PR #3294, opened 2022-04-19
Add EdgeTokenIdExtended to Edge Provider's chooseCurrencyWallets

peachbits:
---
  - https://app.asana.com/0/0/1201752545090997

### DIFF
--- a/src/types/types.js
+++ b/src/types/types.js
@@ -375,3 +375,12 @@ export type AppConfig = {
   website: string,
   appStore: string
 }
+
+export type EdgeTokenId = {
+  pluginId: string,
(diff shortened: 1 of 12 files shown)


src/types/types.js
```diff
@@ -360,3 +360,12 @@ export type wcGetConnection = {
   uri: string,
   timeConnected: number
 }
+
+export type EdgeTokenId = {
```
peachbits, line 364: Check if type exists in core

thehobbit85 (approved):
Approved, but needs to resolve conflicts and rebase before merging.

PR #3296, opened 2022-04-19
Enable TombSwap Plugin

Jon-edge:
-Enable TombSwap Plugin
-Move exchange icons to CDN, update naming scheme so that pluginId maps 1:1 to the filename.
-Begin consolidation of CDN references

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


src/constants/CdnConstants.js
```diff
@@ -0,0 +1,25 @@
...
+// flags, currency icons, partners, etc
+
+export const EDGE_CONTENT_SERVER_URL = 'https://content.edge.app'
+
+export const FLAG_LOGO_URL = `${EDGE_CONTENT_SERVER_URL}/country-logos`
+
+const directory = Platform.OS === 'ios' ? RNFS.DocumentDirectoryPath : RNFS.ExternalCachesDirectoryPath
+const BACKGROUND_IMAGE_FILE_NAME = 'login_bg.gif'
+export const BACKGROUND_IMAGE_URI = `${EDGE_CONTENT_SERVER_URL}/${BACKGROUND_IMAGE_FILE_NAME}`
+export const BACKGROUND_IMAGE_LOCAL_URI = `file://${directory}/${BACKGROUND_IMAGE_FILE_NAME}`
```
samholmes, line 17: End all identifiers with `URI` or `URL` to be consistent. I lean towards `URI` for some reason, but I have no conviction behind that. Here's some reasoning thought:

Screen Shot 2022-04-19 at 15 21 05

src/constants/CdnConstants.js
```diff
@@ -0,0 +1,25 @@
+// @flow
```
samholmes, line 1: Move this file out of `constants` too?

samholmes (changes requested):

src/constants/CdnConstants.js
Jon-edge: My understanding is that URL describes a path/folder while the URI describes the actual resource

Jon-edge: Yeah you're right, all URI's are URL's, but not vice-versa

Jon-edge:
Renaming this branch didn't seem to get picked up by this PR, closing and re-opening

samholmes:
@Jon-edge  When closing a PR because it's replaced by a new PR, we should always reference the new branch.

New branch: https://github.com/EdgeApp/edge-react-gui/pull/3301

PR #3303, opened 2022-04-20
William/token fixes

swansontec:
This is prep work for putting the core in charge of custom tokens, instead of the GUI.

#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1202186486803487

### DIFF
--- a/src/components/themed/WalletList.js
+++ b/src/components/themed/WalletList.js
@@ -4,6 +4,7 @@ import * as React from 'react'
 import { FlatList, RefreshControl, SectionList } from 'react-native'
 
 import { selectWallet } from '../../actions/WalletActions.js'
+import { useAllTokens } from '../../hooks/useAllTokens.js'
 import s from '../../locales/strings'
 import { getExchangeDenominationFromState } from '../../selectors/DenominationSelectors.js'
 import { calculateFiatBalance } from '../../selectors/WalletSelectors.js'
@@ -84,7 +85,6 @@ export function WalletList(props: Props) {
     [dispatch, onPress]
   )
   const account = useSelector(state => state.core.account)
-  const customTokens = useSelector(state => state.ui.settings.customTokens)
   const exchangeRates = useSelector(state => state.exchangeRates)
   const mostRecentWallets = useSelector(state => state.ui.settings.mostRecentWallets)
   const walletsSort = useSelector(state => state.ui.settings.walletsSort)
@@ -93,6 +93,9 @@ export function WalletList(props: Props) {
   // Subscribe to the wallet list:
   const [activeWalletIds, setActiveWalletIds] = useState(account.activeWalletIds)
   useEffect(() => account.watch('activeWalletIds', setActiveWalletIds), [account])
+
+  // Subscribe to all the tokens in the account:
+  const allTokens = useAllTokens(account)
 
   function sortWalletList(walletList: WalletListItem[]): WalletListItem[] {
     const getFiatBalance = (wallet: GuiWallet, fullCurrencyCode: string): number => {
@@ -161,10 +164,12 @@ export function WalletList(props: Props) {
           key: walletId
         })
       } else if (wallet != null) {
-        const { enabledTokens, name, currencyCode, currencyNames } = asSafeDefaultGuiWallet(wallet)
+        const { enabledTokens, name, pluginId } = asSafeDefaultGuiWallet(wallet)
+        const { currencyInfo } = account.currencyConfig[pluginId]
+        const { currencyCode, displayName } = currencyInfo
 
         // Initialize wallets
-        if (checkFilterWallet({ name, currencyCode, currencyName: currencyNames[currencyCode] }, searchText, allowedCurrencyCodes, excludeCurrencyCodes)) {
+        if (checkFilterWallet({ name, currencyCode, currencyName: displayName }, searchText, allowedCurrencyCodes, excludeCurrencyCodes)) {
           walletList.push({
             id: walletId,
             fullCurrencyCode: currencyCode,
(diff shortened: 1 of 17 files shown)


swansontec:
Here is proof that `denominations[0].name` always equals `currencyCode`:

```js
  useEffect(() => {
    console.log('!!! start scanning')
      const currencyConfig = account.currencyConfig[pluginId]
      const { currencyInfo } = currencyConfig
      if (currencyInfo.currencyCode !== currencyInfo.denominations[0].name) {
        console.log('!!!!! busted', currencyInfo.denominations[0])
      }
      for (const mt of currencyInfo.metaTokens) {
        if (mt.denominations[0].name !== mt.currencyCode) {
          console.log('!!!!! busted token', currencyInfo.denominations[0])
        }
      }
    }
    console.log('!!! done scanning')
  }, [account])
```

None of the error cases get hit.

src/components/themed/WalletList.js
```diff
@@ -161,39 +160,30 @@ export function WalletList(props: Props) {
           key: walletId
         })
       } else if (wallet != null) {
-        const { enabledTokens, name, currencyCode, currencyNames } = asSafeDefaultGuiWallet(wallet)
+        const currencyConfig = account.currencyConfig[wallet.pluginId]
+        const { builtinTokens = {}, currencyInfo, customTokens = {} } = currencyConfig
```
swansontec, line 164: We actually need to subscribe to the custom tokens, since those are subject to change. We can probably use the `useWalletSubscriptions` hook introduced in #3295 to accomplish this in a simple & dynamic way.

swansontec (commented):
Found one issue while working on the core token management.

src/components/themed/WalletList.js
swansontec: This is fixed now. I did not actually need `useWalletSubscriptions`, since the list of plugins is static. I will need `useWalletSubscriptions` later when we start tracking the enabled tokens as well.

src/components/themed/WalletList.js
```diff
@@ -161,39 +160,30 @@ export function WalletList(props: Props) {
...
-          return isVisible
-        })
-
         // Initialize tokens
-        for (const tokenCode of enabledNotHiddenTokens) {
+        const allTokens = { ...customTokens, ...builtinTokens }
+        for (const tokenCode of enabledTokens) {
+          const tokenId = Object.keys(allTokens).find(id => allTokens[id].currencyCode === tokenCode)
+          if (tokenId == null) continue
+          const token = allTokens[tokenId]
```
thehobbit85: This is exactly what I also had in mind that we need to do here.
This is what I did in my (now irrelevant) walletPickerModal PR:
```js
   // Creating the "add token" rows for token wallets
    const allTokens = { ...builtinTokens, ...customTokens }
    Object.keys(allTokens).forEach(tokenId => {
      // Don't add a "add token" row if we already have an existing wallet of this type
      if (existingPlugins[`${pluginId}.${tokenId}`] === true) return
      const { currencyCode, displayName } = allTokens[tokenId]
      if (filter({ tokenId, pluginId }, pluginId, filterCreate)) return
      addTokenRows.unshift({ currencyCode, displayName, tokenId, pluginId, ...curriedCurrencyIcon(tokenId) })
    })
 ```

**Review by thehobbit85 — CHANGES_REQUESTED** [review]

File: src/selectors/DenominationSelectors.js [hunk 3]
```diff
@@ -4,7 +4,7 @@ import { type EdgeCurrencyInfo, type EdgeDenomination } from 'edge-core-js'
 
 import type { Dispatch, GetState, RootState } from '../types/reduxTypes.js'
 
-export const emptyEdgeDenomination: EdgeDenomination = {
+const emptyEdgeDenomination: EdgeDenomination = {
```
thehobbit85, line 7: Good catch, exporting a const can lead to trouble.
Also, can you not return the constant directly in the 'getExchangeDenomination' function and instead destruct it into a new object `return {...emptyEdgeDenomination}`.
It's better if we don't share an object between calls.

src/hooks/useAllTokens.js
```diff
@@ -0,0 +1,44 @@
...
+      const currencyConfig = account.currencyConfig[pluginId]
+      function update(): void {
+        setOut(out => ({ ...out, [pluginId]: getAllTokens(currencyConfig) }))
+      }
+      cleanups.push(currencyConfig.watch('builtinTokens', update))
+      cleanups.push(currencyConfig.watch('customTokens', update))
+    }
+
+    return () => cleanups.forEach(cleanup => cleanup())
+  }, [account])
```
thehobbit85, line 36: I was trying to create a useCoreWatcher which receives:
1. Watchable Core Object
2. List of keys to watch for

And returns a ref'ed object with the watched values 

But I kept running into issues with types.
Is there a chance you can+want to do it?
Watching over a Core object (especially an account and a wallet) is a super common pattern we have which is already used in a lot of places but it's going to be everywhere now that we are slowly getting rid of the GUI wallet.

thehobbit85 (changes requested):

src/hooks/useAllTokens.js
swansontec: It is pretty easy with TypeScript, and already exists in the login UI. I don't believe it's possible with Flow, unfortunately, at least in a generic way.

The best we can do are individual ones like `const item = useWalletWatcher(wallet, 'item')` or `const another = useAccountWatcher(account, 'another')`.

swansontec:
I have fixed the one thing you asked, @thehobbit85 .

src/hooks/useAllTokens.js
thehobbit85: I'm actually pretty fine with the individual approach for now:
`const item = useWalletWatcher(wallet, 'item')`

But let's keep it for a later PR so we close this one. And I'll take a look at what you did in the login-ui

thehobbit85 (approved):

thehobbit85 (approved):

PR #3322, opened 2022-04-27
Various Fixes for the sync circle component

thehobbit85:
- Fix the outer radius calculation.
- Never compare floating-point values using ===.
- Add a new theme prop instead os using the primaryText for non-text visuals.
- Use a smaller minimum percentage, since this looks better with the rounded end caps.
- Store the done flag in a ref, as it is not used in any worklets.

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/themed/WalletSyncCircle.js
+++ b/src/components/themed/WalletSyncCircle.js
@@ -17,7 +17,7 @@ const DONE_RATIO = 0.999
 type Props = {
   // The diameter of the inner currency icon:
   size: number,
-  edgeWallet: EdgeCurrencyWallet
+  wallet: EdgeCurrencyWallet
 }
 
 /**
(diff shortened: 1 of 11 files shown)


src/components/themed/WalletSyncCircle.js
```diff
@@ -24,36 +27,35 @@ export const WalletSyncCircle = (props: Props) => {
...
-            isDone.value = true
-            stroke.value = theme.primaryText
+        if (!isDone.current) {
+          if (ratio > DONE_RATIO) {
+            isDone.current = true
+            stroke.value = theme.walletProgressIconDone
           }
           if (ratio < BASE_RATIO) syncRatio.value = withTiming(BASE_RATIO, { duration: 1000 })
-          else if (ratio > MAX_RATIO && ratio < 1) syncRatio.value = withTiming(MAX_RATIO, { duration: 1000 })
+          else if (ratio > MAX_RATIO && ratio < DONE_RATIO) syncRatio.value = withTiming(MAX_RATIO, { duration: 1000 })
```
swansontec, line 51: Please just let `syncRatio.value` be the true sync ratio (with animation). The visual minimum & maximum should be implemented using in your `useAnimatedProps` hook using `Math.max(syncRatio.value, MIN_RATIO)` or such. This separates the business logic (tracking the core state) from the display logic. I don't think the max ratio is needed - it was an artifact of the way the old circle component worked.

swansontec (changes requested):
One change would still be needed.

PR #3325, opened 2022-04-28
Remove @react-native-community/art dependency

swansontec:
Since we are using react-native-svg, we don't need a second native graphics library.

#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/patches/edge-login-ui-rn+0.9.32.patch
+++ b/patches/edge-login-ui-rn+0.9.32.patch
@@ -1,0 +1,20 @@
+diff --git a/node_modules/edge-login-ui-rn/lib/components/common/QrCode.js b/node_modules/edge-login-ui-rn/lib/components/common/QrCode.js
+index e10398f..7009159 100644
+--- a/node_modules/edge-login-ui-rn/lib/components/common/QrCode.js
++++ b/node_modules/edge-login-ui-rn/lib/components/common/QrCode.js
+@@ -1,4 +1,4 @@
+-import { Shape, Surface } from '@react-native-community/art';
++import Svg, { Path } from 'react-native-svg';
+ import qrcodeGenerator from 'qrcode-generator';
+ import * as React from 'react';
+ /**
+@@ -14,6 +14,7 @@ export function QrCode(props) {
+     const path = svg.replace(/.*d="([^"]*)".*/, '$1');
+     // Create a drawing transform to scale QR cells to device pixels:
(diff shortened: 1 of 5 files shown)


patches/edge-login-ui-rn+0.9.32.patch
```diff
@@ -0,0 +1,20 @@
...
+@@ -14,6 +14,7 @@ export function QrCode(props) {
+     const path = svg.replace(/.*d="([^"]*)".*/, '$1');
+     // Create a drawing transform to scale QR cells to device pixels:
+     const sizeInCells = code.getModuleCount() + 2 * padding;
+-    return (React.createElement(Surface, { height: size, width: size, style: { backgroundColor } },
+-        React.createElement(Shape, { d: path, fill: foregroundColor, scale: size / sizeInCells })));
++    const viewBox = `0 0 ${sizeInCells} ${sizeInCells}`
++    return (React.createElement(Svg, { height: size, width: size, style: { backgroundColor }, viewBox: viewBox },
++        React.createElement(Path, { d: path, fill: foregroundColor })));
+ }
```
thehobbit85, line 20: Is there a reason why we don't just fix it directly in edge-login-ui-rn?
I don't really like patching our own libraries since we do update them quite often (unlike 3rd party dependencies) which can break the patch or the very least throw an annoying warning about mismatching versions.

src/components/themed/QrCode.js
```diff
@@ -81,8 +81,5 @@ const getStyles = cacheStyles((theme: Theme) => ({
     position: 'absolute',
     right: 0,
     top: 0
-  },
-  surface: {
-    backgroundColor: theme.qrBackgroundColor
```
thehobbit85, line 86: Why was this removed? Wouldn't it change the way it looks?

thehobbit85 (changes requested):

patches/edge-login-ui-rn+0.9.32.patch
swansontec: Breaking change to the native deps, which affects third-party integrations. We would need a major version bump.

thehobbit85: Good enough reason

thehobbit85 (approved):

thehobbit85 (approved):

PR #3326, opened 2022-04-28
Fix filtering of EdgeTokenIdExtended arrays in EdgeProvider chooseCurrencyWallet

peachbits:
---
  - https://app.asana.com/0/0/1202193781379741

### DIFF
--- a/src/modules/UI/scenes/Plugins/EdgeProvider.js
+++ b/src/modules/UI/scenes/Plugins/EdgeProvider.js
@@ -120,17 +120,27 @@ export class EdgeProvider extends Bridgeable {
   // $FlowFixMe // Default empty array is not typed
   async chooseCurrencyWallet(allowedCurrencyCodes: string[] | EdgeTokenIdExtended[] = []): Promise<string> {
     const selectedWallet: WalletListResult = await Airship.show(bridge => (
-      <WalletListModal bridge={bridge} showCreateWallet allowedCurrencyCodes={allowedCurrencyCodes} headerTitle={s.strings.choose_your_wallet} />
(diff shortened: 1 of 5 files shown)


src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -493,6 +503,23 @@ export class EdgeProvider extends Bridgeable {
...
+  // Convert an array of currency codes to an array of EdgeTokenIdExtended objects by creating all possible plugin/token combinations, ignoring any plugin passed as filterPlugins.
+  convertCurrencyCodesToEdgeTokenIds(currencyCodes: string[], filterPlugins?: string[] = []): EdgeTokenIdExtended[] {
+    const tokenIds: EdgeTokenIdExtended[] = []
+    for (const pluginId of Object.keys(this._state.core.account.currencyConfig).filter(plugin => !filterPlugins.includes(plugin))) {
+      const { currencyCode, metaTokens } = this._state.core.account.currencyConfig[pluginId].currencyInfo
+      if (currencyCodes.includes(currencyCode)) {
+        tokenIds.push({ pluginId, currencyCode })
+      }
+      for (const allowedCode of currencyCodes) {
+        if (metaTokens.find(token => token.currencyCode === allowedCode) != null) {
```
swansontec, line 515: This will not work for custom tokens, but our partners aren't supporting custom tokens anyhow. Going forward, we need to stop using `metaTokens` and use the core's new token structures instead. For now, let's just leave this as-is.

swansontec (approved):

PR #3331, opened 2022-04-29
Jon/feat/text hooks components

Jon-edge:
#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a

---
  - https://app.asana.com/0/0/1202179932115374


src/components/common/text/CryptoText.js
```diff
@@ -0,0 +1,42 @@
...
+  nativeAmount?: string
+}
+
+/**
+ * Returns a cleaned crypto amount string. If no nativeAmount is given, use the
+ * wallet balance.
+ **/
+export const CryptoText = ({ walletId, tokenId, nativeAmount }: Props) => {
+  const dispatch = useDispatch()
+  const currencyInfo = useSelector(state => state.core.account.currencyWallets[walletId].currencyInfo)
```
swansontec, line 20: You cannot use `useSelector` to grab things out of the account like this. You have to do it this way:

```
const wallet = account.currencyWallets[walletId]
const { currencyInfo, fiatCurrencyCode } = wallet
```

This is because the properties of the account are *not* managed by redux, so it won't know when they change or when to re-render. In this particular case, both these items are fixed. In other cases, such as balances, the wallet can update them at any time, so you actually need to use a subscription and `useState`.

src/components/common/text/CryptoText.js
```diff
@@ -0,0 +1,42 @@
...
+  const fiatCurrencyCode = useSelector(state => state.core.account.currencyWallets[walletId].fiatCurrencyCode)
+  const tokenOrNativeCode = tokenId ?? currencyInfo.currencyCode
+  const exchangeRates = useSelector(state => state.exchangeRates)
+  const nativeAmountOrBalance = useSelector(state => nativeAmount ?? state.core.account.currencyWallets[walletId].balances[tokenOrNativeCode] ?? '0')
```
swansontec, line 24: Do not hook up to the balance in this way. There are two problems here:

1. The balances array is indexed by currencyCode, not tokenId. I have an outstanding task to create a `tokenBalances` property on the wallet which is indexed by tokenId. In the meantime, we have to convert the tokenId to a currencyCode before using it.
2. The balances can change dynamically. You have to combine a `useState` hook with a subscription to track it properly.

Therefore, we should just make `nativeAmount` mandatory, and not bother with any of this complexity at this level.

src/components/common/text/CryptoText.js
```diff
@@ -0,0 +1,42 @@
...
+  const rateKey = `${tokenOrNativeCode}_${isoFiatCurrencyCode}`
+  const exchangeRate = !zeroString(exchangeRates[rateKey]) ? exchangeRates[rateKey] : '1'
+
+  return useCryptoText({
+    nativeAmount: nativeAmountOrBalance,
+    exchangeRate,
+    exchangeDenomination,
+    fiatDenomination,
+    denomination,
+    currencyCode: tokenOrNativeCode
```
swansontec, line 40: You cannot pass a tokenId as a currencyCode. The two are not the same. Instead, `useCryptoText` should be fixed to use `tokenId` directly.

src/hooks/useCryptoText.js
```diff
@@ -0,0 +1,52 @@
...
+  zeroString
+} from '../util/utils'
+
+type UseCryptoTextParams = {
+  nativeAmount: string,
+  exchangeRate?: string,
+  exchangeDenomination: EdgeDenomination,
+  fiatDenomination: EdgeDenomination,
+  denomination: EdgeDenomination,
+  currencyCode: string
```
swansontec, line 22: This hook should take a `tokenId`, not a `currencyCode`. Then do `wallet.currencyConfig.allTokens[tokenId]` to get an `EdgeToken` object, which contains the denominations and such.

Edit: Actually, `currencyCode` isn't even used here! It's only displayed as part of an error message. Can we remove it, and use `denomination.name` in the error or something?

src/components/common/ExchangeRate.js
```diff
@@ -69,9 +69,8 @@ class ExchangeRateComponent extends React.Component<Props & ThemeProps> {
         <FiatText
           format="raw"
           nativeCryptoAmount={primaryInfo.displayDenomination.multiplier}
-          cryptoCurrencyCode={primaryInfo.exchangeCurrencyCode}
-          isoFiatCurrencyCode={secondaryInfo.exchangeCurrencyCode}
-          cryptoExchangeMultiplier={primaryInfo.exchangeDenomination.multiplier}
+          tokenId={primaryInfo.exchangeCurrencyCode}
```
swansontec, line 72: You cannot pass a currencyCode as a tokenId. Currency codes are things like "BTC", while tokenId's are either a contract address or `undefined`.

src/components/common/text/FiatText.js
```diff
@@ -11,17 +14,26 @@ import { useFiatText } from '../../../hooks/useFiatText'
...
+  format: FiatTextFormat
 }
 
-export const FiatText = ({ format, nativeCryptoAmount, cryptoCurrencyCode, cryptoExchangeMultiplier, isoFiatCurrencyCode }: Props) => {
+/**
+ * Return a formatted fiat text string representing the exchange rate of a
+ * specific crypto asset and native amount.
+ **/
+export const FiatText = ({ walletId, tokenId, nativeCryptoAmount, format }: Props) => {
+  const fiatCurrencyCode = useSelector(state => state.core.account.currencyWallets[walletId].fiatCurrencyCode)
```
swansontec, line 28: Same thing here with the incorrect `useSelector` usage. Just grab the account, then access everything else from that using normal non-hook code.

src/components/common/text/FiatText.js
```diff
@@ -11,17 +14,26 @@ import { useFiatText } from '../../../hooks/useFiatText'
...
+  const currencyInfo = useSelector(state => state.core.account.currencyWallets[walletId].currencyInfo)
+  const tokenOrNativeCode = tokenId ?? currencyInfo.currencyCode
```
swansontec, line 30: You cannot use a tokenId as a currencyCode.

src/components/common/text/TickerText.js
```diff
@@ -0,0 +1,119 @@
...
+import { truncateDecimals } from '../../../locales/intl'
+import { getExchangeDenominationFromState } from '../../../selectors/DenominationSelectors'
+import { useDispatch, useSelector } from '../../../types/reactRedux'
+import { type GuiExchangeRates } from '../../../types/types'
+import { fixFiatCurrencyCode, zeroString } from '../../../util/utils'
+import { type Theme, cacheStyles, useTheme } from '../../services/ThemeContext'
+import { EdgeText } from '../../themed/EdgeText'
+
+type Props = {
+  walletId: string,
```
swansontec, line 17: Please take `wallet: EdgeCurrencyWallet` instead of `walletId`.

src/components/common/text/TickerText.js
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Returns a text string that displays the crypto-fiat exchange rate and the
+ * daily % change from a wallet asset
+ **/
+export const TickerText = ({ walletId, style, tokenId }: Props) => {
+  const dispatch = useDispatch()
+  const styles = getStyles(useTheme())
+
+  const fiatCurrencyCode = useSelector(state => state.core.account.currencyWallets[walletId].fiatCurrencyCode)
```
swansontec, line 77: Incorrect `useSelector` usage. Just grab the account, then extract the other stuff using regular non-selector code.

src/components/common/text/TickerText.js
```diff
@@ -0,0 +1,119 @@
...
+  const currencyInfo = useSelector(state => state.core.account.currencyWallets[walletId].currencyInfo)
+  const tokenOrNativeCode = tokenId ?? currencyInfo.currencyCode
```
swansontec, line 79: You cannot use a tokenId as a currencyCode.

src/components/themed/WalletListCurrencyRow.js
```diff
@@ -38,166 +24,56 @@ type Props = {
...
   const dispatch = useDispatch()
   const theme = useTheme()
   const styles = getStyles(theme)
 
-  const edgeWallet = useSelector(state => state.core.account.currencyWallets[walletId])
-  const exchangeRates = useSelector(state => state.exchangeRates)
+  const currencyInfo = useSelector(state => state.core.account.currencyWallets[walletId].currencyInfo)
+  const fiatCurrencyCode = useSelector(state => state.core.account.currencyWallets[walletId].fiatCurrencyCode)
+  const balance = useSelector(state => state.core.account.currencyWallets[walletId].balances[currencyCode] ?? '0')
+  const name = useSelector(state => state.core.account.currencyWallets[walletId].name)
```
swansontec, line 39: Incorrect selector usage again. The way it was before was more correct, since it included the necessary subscriptions.

src/components/themed/WalletListCurrencyRow.js
```diff
@@ -38,166 +24,56 @@ type Props = {
...
   const exchangeDenomination = dispatch(getExchangeDenominationFromState(currencyInfo.pluginId, currencyCode))
-  const fiatDenomination = getDenomFromIsoCode(fiatCurrencyCode)
   const isoFiatCurrencyCode = fixFiatCurrencyCode(fiatCurrencyCode)
-  const rateKey = `${currencyCode}_${isoFiatCurrencyCode}`
-  const exchangeRate = !zeroString(exchangeRates[rateKey]) ? exchangeRates[rateKey] : '1'
   const cryptoExchangeMultiplier = exchangeDenomination.multiplier
 
-  const nativeCryptoAmount = getCryptoAmount({ balance, exchangeRate, exchangeDenomination, fiatDenomination, denomination, currencyCode })
-
   const { fiatText: fiatBalanceText } = useFiatText({
```
swansontec, line 47: We were supposed to switch to `<FiatText>` here, rather than using a hook.

Actually, can we not modify the wallet list rows here at this point? I have another PR, #3336, that goes pretty deep into refactoring the wallet list rows, and these two PR's are conflicting. Once we get the new text components done, as well as the refactors to the wallet list, we can do a third PR that brings both together.

swansontec (changes requested):
I found quite a few issues here. Sorry about that.

src/components/common/text/FiatText.js
```diff
@@ -1,52 +1,30 @@
...
-import { DECIMAL_PRECISION, formatFiatString } from '../../../util/utils'
+import { useFiatText } from '../../../hooks/useFiatText'
+
+/**
+ * - raw: Only the numeric fiat text itself
+ * - primary: Fiat text with the fiat symbol
+ * - secondary: Fiat text enclosed in parenthesis, used when the fiat text is
+ * combined with a crypto amount
+ **/
+export type FiatTextFormat = 'raw' | 'primary' | 'secondary'
```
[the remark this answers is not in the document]
Jon-edge: This came down from @thehobbit85 but I like this better as well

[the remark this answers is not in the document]
Jon-edge: @swansontec This isn't going to work. I can't use `FiatText` as a JSX component without adding some ugly `<EdgeText>` around each parenthesis. If I use `FiatText` as a method, everything class-based explodes since I need to add a `useSelector` to get the `account` inside of `FiatText` to satisfy the other comments about handling the `tokenId`.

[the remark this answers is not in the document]
swansontec: Go to `.eslintrc.json` and edit the rule to say `"react-native/no-raw-text": ["error", { "skip": ["FormattedText", "T", "B", "EdgeText"] }]`. Adding `EdgeText` to the list should make the usage legal.

However, I am OK making a boolean prop called `parenthesisEnclosed`. It's a better solution than the enum.

src/hooks/useFiatText.js
```diff
@@ -24,7 +24,6 @@ export const useFiatText = (props: Props) => {
     cryptoExchangeMultiplier = defaultMultiplier,
     appendFiatCurrencyCode,
     nativeCryptoAmount = cryptoExchangeMultiplier,
-    fiatSymbolSpace,
```
Jon-edge, line 27: Note: I found that the fiat symbol is spaced in 2 places right now: WiredBalanceBox and ControlPanel.

Can't really integrate FiatText into WiredBalanceBox because of the new requirement of a wallet so not bothering to touch it for now.

For ControlPanel, however, it looks like there was a bug introduced in ff7595ab0ce81c75abf58ef4da4b973480cb2d52 where the fiat symbol space was hard-coded into the return value of FiatText no matter what.

Don't really see any other way to handle this besides re-introducing the fiatSymbolSpace variable in the FiatText hook and component.

Jon-edge: Alternatively, maybe we can re-visit the design and change the fiat symbol spacing to be consistent across the app.
Personally, I think it's weird that in the Wallet List there's crypto symbol spacing but not fiat symbol spacing.

src/components/common/text/TickerText.js
```diff
@@ -0,0 +1,119 @@
...
+  const exchangeDenomination = dispatch(getExchangeDenominationFromState(currencyInfo.pluginId, tokenOrNativeCode))
+
+  const cryptoExchangeMultiplier = exchangeDenomination.multiplier
+  const isoFiatCurrencyCode = fixFiatCurrencyCode(fiatCurrencyCode)
+  const rateKey = `${tokenOrNativeCode}_${isoFiatCurrencyCode}`
+  const exchangeRates = useSelector(state => state.exchangeRates)
```
thehobbit85, line 85: It's a bad idea to select the entire 'exchangeRates' since that would trigger a render to ALL the TickerTexts every time ANY of them updates. 

Instead, you should be more precise in your selection.

You should do something like this - 

```js

const [fiatCurrencyCode, setFiatCurrencyCode] = useState(edgeWallet.fiatCurrencyCode)
  const isoFiatCurrencyCode = fixFiatCurrencyCode(fiatCurrencyCode)
  const rateKey = `${currencyCode}_${isoFiatCurrencyCode}`
  const yesterdayUsdRateKey = `${currencyCode}_iso:USD_${getYesterdayDateRoundDownHour()}`

  const rawExchangeRate = useSelector(state => state.exchangeRates[rateKey])
  const usdExchangeRate = useSelector(state => state.exchangeRates[`iso:USD_${isoFiatCurrencyCode}`])
  const yesterdayUsdExchangeRate = useSelector(state => state.exchangeRates[yesterdayUsdRateKey])
  
  ```
  
  
  
  Look at this file for reference: https://github.com/EdgeApp/edge-react-gui/blob/cfa44ebcc86e4316be7123357fdecdc8243cd571/src/components/themed/WalletListCurrencyRow.js

Comment (thehobbit85) [inline src/components/common/text/TickerText.js:85]: It's a bad idea to select the entire 'exchangeRates' since that would trigger a render to ALL the TickerTexts every time ANY of them updates. 

Instead, you should be more precise in your selection.

You should do something like this - 

```js

const [fiatCurrencyCode, setFiatCurrencyCode] = useState(edgeWallet.fiatCurrencyCode)
  const isoFiatCurrencyCode = fixFiatCurrencyCode(fiatCurrencyCode)
  const rateKey = `${currencyCode}_${isoFiatCurrencyCode}`
  const yesterdayUsdRateKey = `${currencyCode}_iso:USD_${getYesterdayDateRoundDownHour()}`

  const rawExchangeRate = useSelector(state => state.exchangeRates[rateKey])
  const usdExchangeRate = useSelector(state => state.exchangeRates[`iso:USD_${isoFiatCurrencyCode}`])
  const yesterdayUsdExchangeRate = useSelector(state => state.exchangeRates[yesterdayUsdRateKey])
  
  ```
  
  
  
  Look at this file for reference: https://github.com/EdgeApp/edge-react-gui/blob/cfa44ebcc86e4316be7123357fdecdc8243cd571/src/components/themed/WalletListCurrencyRow.js

File: src/components/common/text/CryptoText.js [hunk 16]
```diff
@@ -0,0 +1,50 @@
...
+  tokenId?: string,
+  nativeAmount: string
+}
+
+/**
+ * Returns a cleaned crypto amount string. If no tokenId is given, use the
+ * wallet's native currency.
+ **/
+export const CryptoText = ({ wallet, tokenId, nativeAmount }: Props) => {
+  const exchangeRates = useSelector(state => state.exchangeRates)
```
thehobbit85, line 21: It's a bad idea to select the entire 'exchangeRates' since that would trigger a render to ALL the TickerTexts every time ANY of them updates. 

Instead, you should be more precise in your selection.

You should do something like this - 

```js

const [fiatCurrencyCode, setFiatCurrencyCode] = useState(edgeWallet.fiatCurrencyCode)
  const isoFiatCurrencyCode = fixFiatCurrencyCode(fiatCurrencyCode)
  const rateKey = `${currencyCode}_${isoFiatCurrencyCode}`
  const yesterdayUsdRateKey = `${currencyCode}_iso:USD_${getYesterdayDateRoundDownHour()}`

  const rawExchangeRate = useSelector(state => state.exchangeRates[rateKey])
  const usdExchangeRate = useSelector(state => state.exchangeRates[`iso:USD_${isoFiatCurrencyCode}`])
  const yesterdayUsdExchangeRate = useSelector(state => state.exchangeRates[yesterdayUsdRateKey])
  
  ```
  
  
  
  Look at this file for reference: https://github.com/EdgeApp/edge-react-gui/blob/cfa44ebcc86e4316be7123357fdecdc8243cd571/src/components/themed/WalletListCurrencyRow.js

thehobbit85, line 21: It's a bad idea to select the entire 'exchangeRates' since that would trigger a render to ALL the TickerTexts every time ANY of them updates. 

Instead, you should be more precise in your selection.

You should do something like this - 

```js

const [fiatCurrencyCode, setFiatCurrencyCode] = useState(edgeWallet.fiatCurrencyCode)
  const isoFiatCurrencyCode = fixFiatCurrencyCode(fiatCurrencyCode)
  const rateKey = `${currencyCode}_${isoFiatCurrencyCode}`
  const yesterdayUsdRateKey = `${currencyCode}_iso:USD_${getYesterdayDateRoundDownHour()}`

  const rawExchangeRate = useSelector(state => state.exchangeRates[rateKey])
  const usdExchangeRate = useSelector(state => state.exchangeRates[`iso:USD_${isoFiatCurrencyCode}`])
  const yesterdayUsdExchangeRate = useSelector(state => state.exchangeRates[yesterdayUsdRateKey])
  
  ```
  
  
  
  Look at this file for reference: https://github.com/EdgeApp/edge-react-gui/blob/cfa44ebcc86e4316be7123357fdecdc8243cd571/src/components/themed/WalletListCurrencyRow.js

thehobbit85 (changes requested):

thehobbit85 (changes requested):

Jon-edge:
I haven't yet removed `react-native-flux` or `GuiWallet` yet in this PR, but I figure if this is good I can make those changes as part of the merge PR along with any last comments you guys might have after these most recent commits.

PR #3337, opened 2022-05-02
Make sure values passed to hexToDecimal will be interpreted as hex

peachbits:


peachbits:
Approved OTS by @thehobbit85

PR #3342, opened 2022-05-05
William/wallet list disconnection

swansontec:
This PR disconnects the top-level wallet list from Redux. It also begins the process of modernizing the props passed down to the row components. We will need these modern props to render the exchange rates efficiently.

#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a


src/components/services/SortedWalletList.js
```diff
@@ -0,0 +1,297 @@
...
+    return () => cleanups.forEach(cleanup => cleanup())
+  }, [account])
+
+  return out
+}
+
+/**
+ * Subscribes to all the enabled token lists in the account.
+ */
+function useEnabledWalletIds(account: EdgeAccount): EnabledTokenIds {
```
peachbits, line 178: Can we use useWatchAccount for this and useAllTokens?

peachbits (changes requested):

src/components/services/SortedWalletList.js
```diff
@@ -0,0 +1,297 @@
...
+    lastList.current = sorted
+  })
+
+  return null
+}
+
+/**
+ * Subscribes to all the tokens across all plugins in an account.
+ */
+export function useAllTokens(account: EdgeAccount): EdgeTokenMaps {
```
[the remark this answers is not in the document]
swansontec: I just moved `useAllTokens` into this file, since nobody else is using it.

src/components/services/SortedWalletList.js
swansontec: Neither of these are simple properties - we need multiple subscriptions, one per item, so we have to build a bit fancier hook.

peachbits (approved):

PR #3352, opened 2022-05-10
Add a `fixCurrencyCodes` filter for GUI plugins

swansontec:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1202255797698875

### DIFF
--- a/src/modules/UI/scenes/Plugins/EdgeProvider.js
+++ b/src/modules/UI/scenes/Plugins/EdgeProvider.js
@@ -145,7 +145,7 @@ export class EdgeProvider extends Bridgeable {
       <WalletListModal
         bridge={bridge}
         showCreateWallet
-        allowedAssets={upgradeExtendedCurrencyCodes(this._state.core.account, this._plugin.filterPlugins, allowedCurrencyCodes)}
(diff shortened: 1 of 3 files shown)


src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -552,7 +556,13 @@ function upgradeExtendedCurrencyCodes(account: EdgeAccount, excludePluginIds?: s
   const out: EdgeTokenId[] = []
   for (const code of currencyCodes) {
     if (typeof code === 'string') {
-      const [parentCode, tokenCode] = code.split('-')
+      const fixed = fixCurrencyCodes[code] ?? code
+      if (typeof fixed === 'object') {
+        //  We have a tokenId for this code
+        out.push(fixed)
+        continue
+      }
```
swansontec, line 564: Let's change this to:

```js
      // See if we have a tokenId override for this currencyCode:
      const fixed = fixCurrencyCodes[code]
      if (fixed != null) {
        out.push(fixed)
        continue
      }

      const [parentCode, tokenCode] = code.split('-')
```

swansontec (commented):
Since we are changing this to a full `EdgeTokenId`, we can simplify the filter logic a bit.

swansontec (commented):
Approved

PR #3354, opened 2022-05-11
Future/jon/feat/update text hooks components

Jon-edge:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/__tests__/__snapshots__/Request.test.js.snap
+++ b/src/__tests__/__snapshots__/Request.test.js.snap
@@ -66,10 +66,18 @@ exports[`Request should render with loaded props 1`] = `
         }
       >
         <FiatText
-          appendFiatCurrencyCode={true}
-          autoPrecision={true}
-          cryptoExchangeMultiplier="100000000"
+          format="primary"
           nativeCryptoAmount="100000000"
+          wallet={
+            Object {
+              "balances": Object {
+                "BTC": "1234",
+              },
+              "currencyInfo": Object {
+                "pluginId": "bitcoin",
+              },
+            }
+          }
         />
       </WithTheme(EdgeTextComponent)>
     </View>
--- a/src/components/modals/FlipInputModal.js
+++ b/src/components/modals/FlipInputModal.js
@@ -1,7 +1,8 @@
 // @flow
+/* eslint-disable react-native/no-raw-text */
 
 import { div, eq } from 'biggystring'
-import { type EdgeDenomination, asMaybeNoAmountSpecifiedError } from 'edge-core-js'
+import { type EdgeCurrencyWallet, type EdgeDenomination, asMaybeNoAmountSpecifiedError } from 'edge-core-js'
 import * as React from 'react'
 import { TouchableWithoutFeedback, View } from 'react-native'
 import { type AirshipBridge } from 'react-native-airship'
@@ -53,11 +54,11 @@ type StateProps = {
   overridePrimaryExchangeAmount: string,
   forceUpdateGuiCounter: number,
   pluginId: string,
+  wallet: EdgeCurrencyWallet,
 
   // Fees
   feeCurrencyCode: string,
   feeDisplayDenomination: EdgeDenomination,
-  feeExchangeDenomination: EdgeDenomination,
   feeNativeAmount: string,
   feeAmount: string,
   feeStyle?: string,
@@ -159,7 +160,7 @@ export class FlipInputModalComponent extends React.PureComponent<Props, State> {
   }
 
   renderBalance = () => {
-    const { balanceCrypto, primaryInfo, currencyCode, secondaryInfo, theme } = this.props
+    const { wallet, balanceCrypto, primaryInfo, theme } = this.props
     const styles = getStyles(theme)
     const { multiplier, name } = primaryInfo.displayDenomination
     const balance = `${formatNumber(div(balanceCrypto, multiplier, DECIMAL_PRECISION))} ${name} `
@@ -168,13 +169,8 @@ export class FlipInputModalComponent extends React.PureComponent<Props, State> {
         <EdgeText style={styles.secondaryTitle}>{s.strings.send_confirmation_balance}</EdgeText>
         <EdgeText style={styles.rateBalanceText}>
           {balance}
-          <FiatText
-            nativeCryptoAmount={balanceCrypto}
-            cryptoCurrencyCode={currencyCode}
-            isoFiatCurrencyCode={secondaryInfo.exchangeCurrencyCode}
-            cryptoExchangeMultiplier={primaryInfo.exchangeDenomination.multiplier}
-            parenthesisEnclosed
-          />
+          (
+          <FiatText wallet={wallet} currencyCode={primaryInfo.displayCurrencyCode} nativeCryptoAmount={balanceCrypto} />)
         </EdgeText>
       </View>
     )
@@ -207,11 +203,12 @@ export class FlipInputModalComponent extends React.PureComponent<Props, State> {
   }
 
   renderFees = () => {
-    const { feeAmount, feeCurrencyCode, feeDisplayDenomination, feeExchangeDenomination, feeNativeAmount, feeStyle, secondaryInfo, theme } = this.props
+    const { wallet, feeAmount, feeCurrencyCode, feeDisplayDenomination, feeNativeAmount, feeStyle, theme } = this.props
     const truncatedFeeAmount = truncateDecimals(feeAmount, DEFAULT_TRUNCATE_PRECISION, false)
     const feeCryptoText = `${truncatedFeeAmount} ${feeDisplayDenomination.name} `
     const styles = getStyles(theme)
     const feeTextStyle = feeStyle === 'dangerText' ? styles.feeTextDanger : feeStyle === 'warningText' ? styles.feeTextWarning : styles.feeTextDefault
+
     return (
       <View style={styles.feeContainer}>
         <View style={styles.feeTitleContainer}>
@@ -220,13 +217,7 @@ export class FlipInputModalComponent extends React.PureComponent<Props, State> {
         </View>
         <EdgeText style={feeTextStyle}>
           {feeCryptoText}
-          <FiatText
-            nativeCryptoAmount={feeNativeAmount}
-            cryptoCurrencyCode={feeCurrencyCode}
-            isoFiatCurrencyCode={secondaryInfo.exchangeCurrencyCode}
-            cryptoExchangeMultiplier={feeExchangeDenomination.multiplier}
-            parenthesisEnclosed
-          />
+          (<FiatText nativeCryptoAmount={feeNativeAmount} currencyCode={feeCurrencyCode} wallet={wallet} />)
         </EdgeText>
       </View>
     )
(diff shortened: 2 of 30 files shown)


thehobbit85 (changes requested):
The first commit is not needed, it's leftover from the original PR.

src/components/modals/FlipInputModal.js
```diff
@@ -21,6 +21,7 @@ import type { GuiCurrencyInfo } from '../../types/types.js'
 import { getAvailableBalance, getWalletFiat, getWalletName } from '../../util/CurrencyWalletHelpers.js'
 import { convertTransactionFeeToDisplayFee, DECIMAL_PRECISION, DEFAULT_TRUNCATE_PRECISION, getDenomFromIsoCode, truncateDecimals } from '../../util/utils.js'
 import { ExchangeRate } from '../common/ExchangeRate.js'
+import { FiatText } from '../text/FiatText.js'
```
swansontec: This first commit adds a bunch of imports that are not used, which fails precommit. The only useful thing I can see in this first commit is the cleanup to the dependency lists in `FioStakingChangesScene.js`.

None of the other changes in this commit should be here.

src/util/CurrencyInfoHelpers.js
```diff
@@ -100,3 +100,31 @@ export const getTokenId = (account: EdgeAccount, pluginId: string, currencyCode:
...
+      return tokenId != null
+    })
+  }
+  return { pluginId, tokenId }
+}
+
+export const getAllTokens = (currencyConfig: EdgeCurrencyConfig): EdgeTokenMap => {
+  const { builtinTokens = {}, customTokens = {} } = currencyConfig
+  return { ...customTokens, ...builtinTokens }
+}
```
swansontec, line 130: This is not necessary. The `edge-core-js` library now includes `EdgeCurrencyConfig.allTokens`, so you can just read that directly.

src/hooks/useTokenDisplayData.js
```diff
@@ -0,0 +1,60 @@
...
+
+import { useSelector } from '../types/reactRedux'
+import { getAllTokens } from '../util/CurrencyInfoHelpers'
+import { fixFiatCurrencyCode, getDenomFromIsoCode, getYesterdayDateRoundDownHour, zeroString } from '../util/utils'
+
+/**
+ * Returns data from tokens relevant for display
+ * */
+
+export const useTokenDisplayData = (props: { currencyConfigMap: EdgePluginMap<EdgeCurrencyConfig>, tokenId?: string, wallet: EdgeCurrencyWallet }) => {
```
swansontec, line 14: You do not need `currencyConfigMap`. You can just access `EdgeWallet.currencyConfig` to get to the right one directly.

src/hooks/useTokenDisplayData.js
```diff
@@ -0,0 +1,60 @@
...
+  // Get currencyCode and deomination from token
+  let currencyCode, denomination
+  if (tokenId == null || nativeCurrencyCode === tokens[tokenId].currencyCode) {
+    currencyCode = nativeCurrencyCode
+    denomination = currencyInfo.denominations[0]
+  } else {
+    const token = tokens[tokenId]
+    currencyCode = token.currencyCode
+    denomination = token.denominations[0]
+  }
```
swansontec, line 29: This can be simplified to:

```js
const { currencyCode, denominations } = tokenId == null ? currencyInfo : allTokens[tokenId]
const [denomination] = denominations
```

We don't need to worry about tokens whose `currencyCode` matches the parent chain. If we have a tokenId, it's a token - period.

src/hooks/useTokenDisplayData.js
```diff
@@ -0,0 +1,60 @@
...
+
+  // Fiat Codes
+  const fiatCurrencyCode = wallet.fiatCurrencyCode
+  const isoFiatCurrencyCode = fixFiatCurrencyCode(fiatCurrencyCode)
```
swansontec, line 33: I think the first one isn't used? So this could just be `const isoFiatCurrencyCode = fixFiatCurrencyCode(wallet.fiatCurrencyCode)`?

The current way you have it is confusing because the wallet's `fiatCurrencyCode` should already have the "iso:" prefix, but the name implies it does not.

src/hooks/useCryptoText.js
```diff
@@ -0,0 +1,51 @@
...
+  truncateDecimals as nonLocalTruncateDecimals,
+  zeroString
+} from '../util/utils'
+
+type UseCryptoTextParams = {
+  denomination: EdgeDenomination,
+  exchangeRate?: string,
+  fiatDenomination: EdgeDenomination,
+  nativeAmount: string,
+  tokenId?: string
```
swansontec, line 21: The `tokenId` property appears to be unused?

src/components/text/TickerText.js
```diff
@@ -0,0 +1,70 @@
...
+  })
+
+  // Drop decimals if over '1000' of any fiat currency
+  if (Math.log10(parseFloat(assetToFiatRate)) >= 3) fiatText = truncateDecimals(fiatText, 0)
+
+  const theme = useTheme()
+  const { percentString, deltaColorStyle } = getPercentDeltaString(currencyCode, assetToFiatRate, assetToYestFiatRate, usdToWalletFiatRate, theme)
+
+  const tickerText = `${fiatText} ${percentString}`
+  return <EdgeText style={[style, { color: deltaColorStyle }]}>{tickerText}</EdgeText>
```
swansontec, line 69: Please return `<Text>` instead of `<EdgeText>`, and remove the `style` prop. Then, the user of this component would do:

```js
return <EdgeText style={whatever}><TickerText /></EdgeText>
```

This means the parent can _fully_ control the layout, while the component _only_ controls the coloring. We have a pretty strong rule about never allowing components to accept style props, and there is no good reason to break that rule here.

src/__tests__/__snapshots__/Request.test.js.snap
```diff
@@ -1,202 +1,16 @@
...
-  hasTabs={false}
->
-  <View
-    style={
-      Object {
-        "flex": 1,
-        "justifyContent": "flex-start",
-        "paddingHorizontal": 22,
-      }
+<ActivityIndicator
```
swansontec, line 4: Something has gone wrong with the unit tests. We are rendering a loading indicator, instead of the actual scene we are testing.

src/components/tiles/FiatAmountTile.js
```diff
@@ -7,25 +8,18 @@ import { EdgeText } from '../themed/EdgeText'
...
   const styles = getStyles(theme)
-  const fiatAmount = (
-    <FiatText
-      nativeCryptoAmount={nativeCryptoAmount}
-      cryptoCurrencyCode={cryptoCurrencyCode}
-      isoFiatCurrencyCode={isoFiatCurrencyCode}
-      cryptoExchangeMultiplier={cryptoExchangeMultiplier}
-    />
-  )
+  const fiatAmount = <FiatText currencyCode={currencyCode} tokenId={tokenId} nativeCryptoAmount={nativeCryptoAmount} wallet={wallet} noFiatSymbol />
```
swansontec, line 22: Could this be put inline below?

src/components/tiles/CryptoFiatAmountTile.js
```diff
@@ -1,28 +1,31 @@
 // @flow
+/* eslint-disable react-native/no-raw-text */
```
swansontec, line 2: Do you remember the edit I made to `.eslintrc.json`? If we do those, we don't need these `eslint-disable` lines.

src/components/themed/WalletListCurrencyRow.js
```diff
@@ -184,14 +66,26 @@ export const WalletListCurrencyRowComponent = (props: Props) => {
...
+            <TickerText currencyConfigMap={currencyConfigMap} wallet={wallet} tokenId={tokenId} style={styles.exchangeRateText} />
+          ) : null}
         </View>
         <EdgeText style={styles.nameText}>{name}</EdgeText>
       </View>
       {showBalance ? (
         <View style={styles.balanceColumn}>
-          <EdgeText>{nativeCryptoAmount}</EdgeText>
-          <EdgeText style={styles.fiatBalanceText}>{fiatBalanceText}</EdgeText>
+          <EdgeText>
```
swansontec, line 77: Hooray! This is exactly how I wanted this component to look.

src/components/themed/WalletListCurrencyRow.js
```diff
@@ -184,14 +66,26 @@ export const WalletListCurrencyRowComponent = (props: Props) => {
       <View style={styles.nameColumn}>
         <View style={styles.currencyRow}>
           <EdgeText style={styles.currencyText}>{currencyCode}</EdgeText>
-          {showRate ? <EdgeText style={[styles.exchangeRateText, styles[exchangeRateType ?? 'neutral']]}>{exchangeRateText}</EdgeText> : null}
+          {showRate && wallet != null ? (
+            <TickerText currencyConfigMap={currencyConfigMap} wallet={wallet} tokenId={tokenId} style={styles.exchangeRateText} />
```
swansontec, line 70: This `<TickerText />` should be wrapped in an `<EdgeText style={styles.exchangeRateText}>` component, assuming we remove the style prop from `TickerText` itself.

src/components/modals/WcSmartContractModal.js
```diff
@@ -156,18 +154,17 @@ export const WcSmartContractModal = (props: Props) => {
           <CryptoFiatAmountTile
             title={s.strings.wc_smartcontract_network_fee}
             nativeCryptoAmount={networkFeeCrypto}
-            cryptoCurrencyCode={feeCurrencyCode}
-            isoFiatCurrencyCode={isoFiatCurrencyCode}
             denomination={feeDenom}
+            walletId={walletId}
+            tokenId={feeCurrencyCode}
```
swansontec, line 159: Should be passed as `currencyCode`, not `tokenId`.

src/components/modals/WcSmartContractModal.js
```diff
@@ -139,9 +137,9 @@ export const WcSmartContractModal = (props: Props) => {
           <CryptoFiatAmountTile
             title={s.strings.string_amount}
             nativeCryptoAmount={amountCrypto}
-            cryptoCurrencyCode={amountCurrencyCode}
-            isoFiatCurrencyCode={isoFiatCurrencyCode}
             denomination={amountDenom}
+            walletId={walletId}
+            tokenId={amountCurrencyCode}
```
swansontec, line 142: Should be passed as `currencyCode`, not `tokenId`.

swansontec (changes requested):
This looks much, much better than before! I found a few more small things, though. Nothing that should require a giant refactor, since it's all pretty solid.

src/components/text/TickerText.js
```diff
@@ -66,5 +65,5 @@ export const TickerText = ({ wallet, style, tokenId }: Props) => {
   const { percentString, deltaColorStyle } = getPercentDeltaString(currencyCode, assetToFiatRate, assetToYestFiatRate, usdToWalletFiatRate, theme)
 
   const tickerText = `${fiatText} ${percentString}`
-  return <EdgeText style={[style, { color: deltaColorStyle }]}>{tickerText}</EdgeText>
+  return <Text style={{ color: deltaColorStyle }}>{tickerText}</Text>
```
swansontec, line 68: The `style` prop still needs to be deleted.

src/components/tiles/CryptoFiatAmountTile.js
swansontec: Ok, we can try some things and submit it as another PR.

swansontec (changes requested):
One thing left. We could merge without it, if we were in a hurry.

PR #3356, opened 2022-05-11
Borrow API spec

samholmes:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [x] n/a

---
  - https://app.asana.com/0/0/1202187333983852

### DIFF
--- a/src/plugins/borrow-plugins/types.js
+++ b/src/plugins/borrow-plugins/types.js
@@ -1,0 +1,107 @@
+// @flow
+import { type EdgeCurrencyWallet, type EdgeNetworkFee } from 'edge-core-js'
+
+// Borrow more:
+export type BorrowRequest = {
+  tokenId?: string,
+  nativeAmount: string,
+
+  // Optional wallet which will deposit on behalf of the borrow engine's currencyWallet
+  fromWallet?: string
+}
+
+// Make payment:
+export type RepayRequest = {
+  tokenId?: string,
+  nativeAmount: string,
+
+  // Optional wallet which will deposit on behalf of the borrow engine's currencyWallet
+  fromWallet?: string
+}
+
+// Deposit collateral:
+export type DepositRequest = {
+  tokenId?: string,
+  nativeAmount: string,
+
+  // Optional wallet which will deposit on behalf of the borrow engine's currencyWallet
+  fromWallet?: string
+}
+
+// Withdraw collateral:
+export type WithdrawRequest = {
+  tokenId?: string,
+  nativeAmount: string,
+
+  // Optional destination for the funds
+  toWallet?: string
+}
+
+// General purpose approvable action
+export type ApprovableAction = {
+  +networkFee: EdgeNetworkFee,
+  +approve: () => Promise<void>
+}
+
+//
+// Engine
+//
+
+export type BorrowCollateral = {
+  tokenId?: string,
+  nativeAmount: string
+}
+
+export type BorrowDebt = {
+  tokenId?: string,
+  nativeAmount: string,
+  apy: number
+}
+
+export type BorrowEngine = {
+  // Currency wallet for the loan account (e.g. Ethereum wallet)
+  currencyWallet: EdgeCurrencyWallet,
+
+  collaterals: BorrowCollateral[],
+  debts: BorrowDebt[],
+
+  loanToValue: number,
+
+  // Returns the APR for borrow a particular token
+  getAprQuote: (tokenId?: string) => Promise<number>,
+
+  // Collateral modification
+  deposit: (request: DepositRequest) => Promise<ApprovableAction>,
+  withdraw: (request: WithdrawRequest) => Promise<ApprovableAction>,
+
+  // Debt modification
+  borrow: (request: BorrowRequest) => Promise<ApprovableAction>,
+  repay: (request: RepayRequest) => Promise<ApprovableAction>,
+
+  // Close loan account
+  close: () => Promise<ApprovableAction>
+}
+
+//
+// Plugin Info
+//
+
+export type BorrowPluginInfo = {
+  pluginId: string,
+  displayName: string,
+
+  // Defines the relationship to the type of currency plugins which the plugin supports
+  currencyPluginId: string,
+
+  // Maximum loan-to-value ratio
+  maxLtvRatio: number
+}
+
+//
+// Plugin
+//
+
+export type BorrowPlugin = {
+  borrowInfo: BorrowPluginInfo,
+  makeBorrowEngine: (wallet: EdgeCurrencyWallet) => Promise<BorrowEngine>
+}


Jon-edge (changes requested):

src/plugins/borrow-plugins/types.js
```diff
@@ -0,0 +1,83 @@
...
+
+//
+// Loan Account – Primary entity
+//
+
+export type LoanAccount = {
+  // Account wallet for the loan (e.g. Ethereum wallet)
+  mainWallet: EdgeCurrencyWallet,
+
+  supplies: Array<{
```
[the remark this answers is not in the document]
samholmes: Change `borrow` and `repay` to `borrowPrinciple` and `repayPrinciple`; Change `deposit` and `withdraw` to `depositCollateral` and  `withdrawCollateral`?

src/plugins/borrow-plugins/types.js
```diff
@@ -0,0 +1,111 @@
+// @flow
+import { type EdgeAccount, type EdgeCurrencyWallet, type EdgeNetworkFee } from 'edge-core-js'
+
+// Proposal for a "service plugin" which is a Higher-Order-Plugin that can be
+// instantiated with some account context and returns a service instance.
+export type EdgeServicePlugin<Info, Api> = {
+  info: Info,
+  init: (account: EdgeAccount) => Promise<Api>
+}
```
swansontec, line 9: Borrow plugins do not get access to the account. This is a hold-over from back when we had the action plugin idea confused with the borrow plugin idea. So, we can just get rid of this indirection, and put `BorrowPluginInfo` right on `BorrowService`. So basically, `BorrowService` gets renamed to `BorrowPlugin`.

src/plugins/borrow-plugins/types.js
```diff
@@ -0,0 +1,111 @@
...
+  }>,
+  debts: Array<{
+    tokenId?: string,
+    nativeAmount: string,
+    apy: number
+  }>,
+
+  loanToValue: number,
+
+  getApr: (tokenId?: string) => Promise<number>,
```
swansontec, line 73: I am not sure what this does. Does this provide an estimate for new debt? Otherwise, we have `apy` in the `debts` array already. Also, why is one compounding and the other not?

src/plugins/borrow-plugins/types.js
```diff
@@ -0,0 +1,111 @@
...
+
+  // Close entire account
+  close: () => Promise<ApprovableAction>
+}
+
+//
+// Plugin Info
+//
+
+export type BorrowPluginInfo = {
```
swansontec, line 91: This should have a `displayName`, like "AAVE". We also need a `supportedCurrencyPluginIds` for which chains this plugin operates over - ETH, FTM, or whatever.

src/plugins/borrow-plugins/types.js
```diff
@@ -0,0 +1,111 @@
...
+  // Maximum loan-to-value ratio
+  maxLtvRatio: number
+}
+
+//
+// Service
+//
+
+export type BorrowService = {
+  fetchLoanAccount: (wallet: EdgeCurrencyWallet) => Promise<LoanAccount>,
```
swansontec, line 103: We use the term `fetch` for network requests. This is why swap plugins use `fetch`. However, here we are just creating objects, so these should be `makeLoanAccount`, or maybe `makeLoanEngine` or `makeBorrowEngine` as other names.

src/plugins/borrow-plugins/types.js
```diff
@@ -0,0 +1,111 @@
...
+  fetchLoanAccounts: () => Promise<LoanAccount[]>
```
swansontec, line 104: We need to think about how loan plugins relate to currency plugins. I do not think the plugins should be responsible for this.

One idea would be to simply create a loan manager for every wallet in an account, for each supported plugin type. If Edge supports 5 loan contracts on ETH, then each ETH wallet would get 5 loan managers. Most would have a 0 balance, which is fine. If the user uses an external app + wallet connect to start a new loan, then Edge we will see that on next login, because we always check all possible contracts.

Another idea is to keep track of which walletId's have which loan types on them. This won't auto-detect external loans, but could save resources.

Yet another idea is to only open the plugin when the user opens the loan scene. This would combine the reliability of the first idea with the efficiency of the second idea.

In any case, it should *not* be the responsibility of this plugin to know which wallets it belongs to. The core / GUI should be in charge of that, since there are different possible strategies, and because it's common logic we shouldn't have to replicate between plugins.

swansontec (changes requested):

src/plugins/borrow-plugins/types.js
samholmes: Without access to `EdgeAccount`, how will the borrow plugin determine which wallets have "loan accounts"?

src/plugins/borrow-plugins/types.js
samholmes: A user may not have any `debts` and the GUI will want the APR/APY for a particular token

src/plugins/borrow-plugins/types.js
samholmes: Good idea. I realize I need to flush out the "info" part of the plugin more.

src/plugins/borrow-plugins/types.js
samholmes: These are network request though. Each "fetch" method will read the state on the blockchain.

src/plugins/borrow-plugins/types.js
samholmes: Ah I see. We should discuss these options together.

src/plugins/borrow-plugins/types.js
samholmes: Resolution: Remove `EdgeServicePlugin` in favor of an explicit `BorrowPlugin` which has a `makeBorrowEngine` method which expects a `EdgeCurrencyWallet` and returns a `BorrowEngine` object which is a rename for `BorrowService`.

src/plugins/borrow-plugins/types.js
samholmes: I'm adding `displayName` and `currencyPluginId` to the info interface. The `currencyPluginId` is a `string` which means the relationship between borrow plugins and currency plugins types is one-to-one.

src/plugins/borrow-plugins/types.js
```diff
@@ -0,0 +1,102 @@
...
+
+// General purpose approvable action
+export type ApprovableAction = {
+  +networkFee: EdgeNetworkFee,
+  +approve: () => Promise<void>
+}
+
+//
+// Borrow Engine – Primary entity
+//
```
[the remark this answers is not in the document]
samholmes: I'll remove the "Primary entity" part. This isn't meant to be a comment which contains a description like a Javadoc comment. It's a section heading.

src/plugins/borrow-plugins/types.js
samholmes: Change to `getAprQuote` and added a comment

PR #3376, opened 2022-05-16
Add default fallback electrum server settings object to asMaybe cleaner

peachbits:
---
  - https://app.asana.com/0/0/1202248461628089

### DIFF
--- a/src/components/scenes/CurrencySettingsScene.js
+++ b/src/components/scenes/CurrencySettingsScene.js
@@ -53,7 +53,7 @@ export function CurrencySettingsComponent(props: Props) {
 
   // Are the electrum servers enabled? What are they?
   const defaults = asMaybe(asElectrumDefaults)(defaultSettings)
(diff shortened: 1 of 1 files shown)


src/components/scenes/CurrencySettingsScene.js
```diff
@@ -53,7 +53,7 @@ export function CurrencySettingsComponent(props: Props) {
 
   // Are the electrum servers enabled? What are they?
   const defaults = asMaybe(asElectrumDefaults)(defaultSettings)
-  const settings = asMaybe(asElectrumSettings)(userSettings)
+  const settings = asMaybe(asElectrumSettings, { electrumServers: [], disableFetchingServers: false })(userSettings)
```
thehobbit85, line 56: This is an OK fix for now. 
But we need to fix this entire 'CurrencySettingsComponent' to not have the settings hard coded in the GUI and instead receive a `userSettingsType` from the plugin.

thehobbit85 (approved):

PR #3380, opened 2022-05-17
Refactor Currency Settings

swansontec:
This is in preparation for the new Bitcoin plugin.

@samholmes, what is the format for the new nodes settings in your plugin? I just guessed when I wrote the `asCustomNodes` cleaner (see the last commit).

@thehobbit85 You can see where I have some commented out lines for "foo" and "bar" settings, which would also use the same structure as the existing custom nodes components.

#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1202278972928633


src/components/themed/MaybeCustomNodesSetting.js
```diff
@@ -0,0 +1,134 @@
...
+    fontFamily: theme.fontFaceDefault,
+    fontSize: theme.rem(1),
+    paddingHorizontal: theme.rem(0.5),
+    textAlign: 'left'
+  }
+}))
+
+const asCustomNodesSetting: Cleaner<CustomNodesSetting> = asObject({
+  enableCustomNodes: asOptional(asBoolean, false),
+  customNodes: asArray(asString)
```
samholmes, line 107: In `edge-currency-plugins`, the name of the field is `serverList`: https://github.com/EdgeApp/edge-currency-plugins/blob/master/src/common/plugin/pluginState.ts#L23

samholmes (approved):
Approved with optional changes. Also the relevant piece in edge-currency-plugins is located https://github.com/EdgeApp/edge-currency-plugins/blob/master/src/common/plugin/pluginState.ts

src/components/themed/MaybeCurrencySetting.js
```diff
@@ -0,0 +1,52 @@
...
+|}
+
+type Props = {|
+  currencyConfig: EdgeCurrencyConfig
+|}
+
+/**
+ * Renders a settings section, but only if the cleaner passes.
+ */
+export function maybeCurrencySetting<T>(
```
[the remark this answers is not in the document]
swansontec: The task was to pave the way for arbitrary settings, such as API keys, polling frequencies, or who knows what. All those sections would just use the HOC to slot into the scene in the same way. I don't know where to put HOC's, but we could agree on a location, I would put this in there.

src/components/themed/MaybeCustomNodesSetting.js
swansontec: OK, so your plugin needs to be using a cleaner. Also, we talked about changing the name so the upgrade goes smoothly. We don't want the new plugin trying to talk to Electrum servers leftover from the old plugin.

I am open to changing the shape of the cleaner, maybe naming the fields `customServers` and `enableCustomServers` if you prefer, but we want to agree before merging.

src/components/themed/MaybeCurrencySetting.js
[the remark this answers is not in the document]
swansontec: Conclusion: `src/components/hoc`

src/components/themed/MaybeCustomNodesSetting.js
swansontec: https://github.com/EdgeApp/edge-currency-plugins/pull/311

samholmes: I concur with the movement to change to your proposal. Very clean and clear semantics! 👍

src/components/hoc/MaybeCurrencySetting.js
```diff
@@ -0,0 +1,64 @@
...
+
+type Props = {|
+  currencyConfig: EdgeCurrencyConfig
+|}
+
+/**
+ * Renders a settings section, but only if the cleaner passes.
+ */
+export function maybeCurrencySetting<T, X>(
+  Component: React$StatelessFunctionalComponent<CurrencySettingProps<T, X>>,
```
samholmes, line 36: Where does this type come from btw `React$StatelessFunctionalComponent`?

samholmes (approved):

src/components/hoc/MaybeCurrencySetting.js
swansontec: It is built into Flow. It's a global, always-on symbol, just like `Array` or `$Call` are built in.

PR #3382, opened 2022-05-17
Add a `useCallbork` hook

swansontec:
#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/components/modals/WalletListModal.js
+++ b/src/components/modals/WalletListModal.js
@@ -3,8 +3,9 @@
 import * as React from 'react'
 import { type AirshipBridge } from 'react-native-airship'
 
+import { useHandler } from '../../hooks/useHandler.js'
 import s from '../../locales/strings.js'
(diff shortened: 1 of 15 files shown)


thehobbit85 (changes requested):
Everything's fine except for the hook's name.
Please find something better then that.

Also, since it's going to basically replace 'useCallback', can you export it from the 'reactHooks' type file so we could just import it as we did 'useCallback'?

swansontec:
The team agreed on the name `useHandler` during the engineering meeting.

The plan is to have the reactHooks file be an exact copy of what you would get from React itself, so that we can migrate cleanly at a later date. We should not add this hook to that file, since React itself does not plan to add a `useHandler` hook - even if they do add it, it will be called by another name.

thehobbit85 (approved):
Approved. Waiting to be merged after we release version 2.15 to avoid any potential conflicts

PR #3387, opened 2022-05-19
Sam/borrow plugin engine

samholmes:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
+++ b/src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
@@ -1,28 +1,69 @@
 // @flow
+import { gt } from 'biggystring'
 import { type EdgeCurrencyWallet } from 'edge-core-js'
 
-import { type ApprovableAction, type BorrowEngine, type BorrowRequest, type DepositRequest, type RepayRequest, type WithdrawRequest } from '../../types'
+import {
+  type ApprovableAction,
+  type BorrowCollateral,
+  type BorrowDebt,
+  type BorrowEngine,
+  type BorrowRequest,
+  type DepositRequest,
+  type RepayRequest,
+  type WithdrawRequest
+} from '../../types'
+import { type AaveNetwork } from './AaveNetwork'
+import { addressToTokenId } from './util/addressToTokenId'
 
-export type BorrowEngineBlueprint = {}
+export type BorrowEngineBlueprint = {
+  aaveNetwork: AaveNetwork,
+  enabledTokens: {
+    [symbol: string]: {
+      isCollateral: boolean,
+      isDebt: boolean
+    }
+  }
+}
 
 export const makeBorrowEngineFactory =
   (blueprint: BorrowEngineBlueprint) =>
   async (wallet: EdgeCurrencyWallet): Promise<BorrowEngine> => {
+    const { aaveNetwork, enabledTokens } = blueprint
+
+    const walletAddress = (await wallet.getReceiveAddress()).publicAddress
+
+    const reserveTokens = await aaveNetwork.getAllReservesTokens()
+    const collateralTokens = reserveTokens.filter(token => enabledTokens[token.symbol]?.isCollateral)
+    const debtTokens = reserveTokens.filter(token => enabledTokens[token.symbol]?.isDebt)
+
+    const whenCollaterals: Promise<BorrowCollateral>[] = collateralTokens.map(async token => {
+      const { aToken } = await aaveNetwork.getReserveTokensAddresses(token.address)
+      const balance = await aToken.balanceOf(walletAddress)
+
+      return {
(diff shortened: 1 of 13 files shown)


src/plugins/borrow-plugins/common/BorrowPluginFactory.js
```diff
@@ -0,0 +1,20 @@
+// @flow
```
swansontec, line 1: We like to keep a blank line between the `// @flow` comment and the imports. Otherwise the import sorting plugin will think the comment goes with the next import, and move it around.

src/plugins/borrow-plugins/plugins/aave/util/addressToTokenId.js
```diff
@@ -0,0 +1,3 @@
+// @flow
+
+export const addressToTokenId = (address: string) => address.slice(2).toLowerCase()
```
swansontec, line 3: This is a problem. The only way to look up a token should be by visiting the core. Yes, we happen to have secret knowledge about how the Ethereum plugin chooses its tokenIds's, but this sort of thing should *never* be happening in the GUI.

We should probably add an `EdgeAccount.currencyConfig[pluginId].getTokenId(token: EdgeTokenInfo)`.

However, I think the real solution is to scan the wallet for available tokens, and just use the tokenId's from those. If the contract returns a token that doesn't exist in the wallet, we can't show it anyhow, and the app is likely to crash if we try.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
```diff
@@ -1,28 +1,59 @@
...
-          tokenId: '2260fac5e5542a773aa44fbcfedf7c193bc2c599',
-          nativeAmount: '100000000'
+export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
+  return async (wallet: EdgeCurrencyWallet): Promise<BorrowEngine> => {
+    const { aaveNetwork } = blueprint
+
+    const walletAddress = (await wallet.getReceiveAddress()).publicAddress
+
+    //
+    // Collaterals and Debts
```
swansontec, line 27: We have been using `// Collaterals and Debts -------------------------------------------` as our section divider in the core and other parts of the GUI. Just add dashes to get to 80 columns or so.

src/plugins/borrow-plugins/plugins/aave/index.js
```diff
@@ -1,35 +1,70 @@
...
+  }
+})
+
+const aaveKovanBlueprint: BorrowPluginBlueprint = {
   borrowInfo: {
-    pluginId: 'aave-kovan',
+    pluginId: 'aavekovan',
     displayName: 'AAVE (Kovan)',
-    currencyPluginId: 'ethereum',
+    currencyPluginId: 'kovan',
```
swansontec, line 52: Aha! I was confused by this in the previous commit, but now I see that this is for another network.

swansontec (changes requested):
The architecture is pretty factory-heavy and recursive, so I wonder if it could be made flatter. However, I do see that it will work (assuming the EVM stuff is right).

The other issue is the way we get tokenId's from addresses - we need to be searching the `allTokens` table, as opposed to just munching strings.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
samholmes: _We agreed to try out the new convention for section headrs_

src/plugins/borrow-plugins/plugins/aave/index.js
samholmes: The name is tentative (Eli and I are discussing the right pluginId name), but yes.

src/plugins/borrow-plugins/common/BorrowPluginFactory.js
samholmes: Good to know. I added the space.

samholmes:
> The architecture is pretty factory-heavy and recursive, so I wonder if it could be made flatter. However, I do see that it will work (assuming the EVM stuff is right).
> 
> The other issue is the way we get tokenId's from addresses - we need to be searching the `allTokens` table, as opposed to just munching strings.

It could be made flatter without the `common` factory. However, what I'm going for is the "info" and "factory" pattern we have in for other plugins, but I'm calling it a "blueprint" instead of an "info". The idea is that the blueprint is the abstraction for the implementation. I'm not entirely sold that this is giving us the architectural malleability that I expect over time. Alternatively, instead of using the factory pattern, I may opt to use some kind of functional composition pattern, but this is what I landed on initially.

swansontec (commented):
Great! It's all clean now.

swansontec (approved):

PR #3389, opened 2022-05-20
Sam/borrow plugin apr

samholmes:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1202303706365537

### DIFF
--- a/src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
+++ b/src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
@@ -1,7 +1,7 @@
 // @flow
 
 import { type Cleaner, asMaybe } from 'cleaners'
-import { type EdgeCurrencyWallet } from 'edge-core-js'
+import { type EdgeCurrencyWallet, type EdgeToken } from 'edge-core-js'
 
 import {
   type ApprovableAction,
@@ -44,6 +44,17 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
         }
       }
     }
+    const getToken = (tokenId?: string): EdgeToken => {
+      if (tokenId == null) throw new Error('Getting wrapped native token not supported yet. ' + 'Explicitly pass in tokenId for the wrapped token.')
+      const token = wallet.currencyConfig.allTokens[tokenId]
+      if (token == null) throw new Error(`Unable to find token on wallet for ${tokenId} tokenId`)
+      return token
+    }
+    const getTokenAddress = (token: EdgeToken) => {
+      const tokenAddress = asMaybe(asTokenContractAddress)(token)
+      if (tokenAddress == null) throw new Error(`Unable to find contract address for ${token.displayName} (${token.currencyCode})`)
+      return tokenAddress
+    }
 
     //
     // Collaterals and Debts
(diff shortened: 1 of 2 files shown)


src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
```diff
@@ -79,7 +79,16 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
       loanToValue: 55,
 
       async getAprQuote(tokenId?: string): Promise<number> {
-        return 0.0381
+        if (tokenId == null) throw new Error('Getting wrapped native token not supported yet. ' + 'Explicitly pass in tokenId for the wrapped token.')
+
+        const edgeToken = wallet.currencyConfig.allTokens[tokenId]
+        const tokenAddress: string | void = edgeToken?.networkLocation?.contractAddress
```
swansontec, line 85: Should the cleaner be used here as well? There seems to be a whole sequence of things that can go wrong here, from `tokenId == null`, then `token == null` (not currently handled), and finally a failure to clean the address.

I wonder if a utility function would help here. Then the utility can just return `string | void`, and we can have one throw like "Token ${tokenId} does not have a contract address for AAVE" if it returns `void`.

swansontec (changes requested):
This will probably work, but needs more paranoia against the rest of the app being weird.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
samholmes: I am now leveraging the new `asTokenContractAddress` cleaner from the upstream branch. Addresses which are not included in AAVE's smart contract state will simply return `0` as a response, so we're safe there.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
```diff
@@ -45,6 +45,25 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
...
+    }
+    const asTokenAddress = (token: EdgeToken) => {
+      const tokenAddress = asMaybe(asTokenContractAddress)(token)
+      if (tokenAddress == null) throw new Error(`Unable to find contract address for ${token.displayName} (${token.currencyCode})`)
+      return tokenAddress
+    }
+    const asTokenIdParam = (tokenId?: string) => {
+      if (tokenId == null) throw new Error('Getting wrapped native token not supported yet. ' + 'Explicitly pass in tokenId for the wrapped token.')
+      return tokenId
+    }
```
swansontec, line 65: These are not cleaners, since they take typed input. These are utility methods, and pretty obfuscated ones at that. Please, do it the way I asked in the previous review:

```js
getContractAddress(wallet: EdgeCurrencyWallet, tokenId?: string): string | void {
  if (tokenId == null) return
  const token = wallet.currencyConfig.allTokens[tokenId]
  if (token == null) return
  const address = asMaybe(asTokenContractAddress)(token)
  return address
}
```

Now that we have a clear utility, using it is easy we can do:

```js
async getAprQuote(tokenId?: string): Promise<number> {
  const address = getContractAddress(wallet, tokenId)
  if (address == null) {
    throw new Error(`Unable to find contract address for token ${tokenId} on AAVE`)
  }

  // ...
}
```

Nobody is supposed to pass bad data to AAVE, so the `throw` won't happen unless we have a bug. Therefore, we don't need to have a bunch of over-engineered error handling logic. Just keep it simple, and return something a dev can understand.

Ok, if you *really* like over-detailed errors, I suppose you could have `getContractAddress` throw when `token == null` or when the cleaner fails, but this wouldn't actually include any additional information, so I don't see the point...

swansontec (changes requested):
This is not how cleaners work.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
samholmes: I agree. I like this better. Originally, I had these utilities named as "resolveEdgeToken" and "resolveTokenAddress" etc. Then incorrectly conflated that with cleaners.

PR #3390, opened 2022-05-20
Sam/borrow plugin ltv

samholmes:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1202303706365542

### DIFF
--- a/src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
+++ b/src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
@@ -80,6 +80,14 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
       })
 
     //
+    // Loan to value
+    //
+
+    const userData = await aaveNetwork.lendingPool.getUserAccountData(walletAddress)
+    const { totalCollateralETH, totalDebtETH } = userData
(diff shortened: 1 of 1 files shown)


src/plugins/borrow-plugins/constants.js
```diff
@@ -0,0 +1,3 @@
+// @flow
+
+export const MAX_FLOAT_PRECISION = 17
```
swansontec, line 3: It seems that these constants should be coin-specific, not in a generic constants file that presumably applies to all plugins on all networks. So, I actually liked it the other way, where the constant was in AaveNetwork.js.

So, just grab it from there and skip the whole second commit?

swansontec (changes requested):

src/plugins/borrow-plugins/constants.js
swansontec: Then don't use bignum math? If the result of a calculation is a `double`, biggystring is just burning CPU for a no reason. The only reason to use biggystring is when you need a precise output, down to the Satoshi or Wei or whatever. If you are going to chop off the answer anyhow, just do `parseFloat(string1) / parseFloat(string2)`.

swansontec (approved):

PR #3391, opened 2022-05-20
Add promo banner

peachbits:


src/components/themed/PromoBanner.js
```diff
@@ -0,0 +1,49 @@
...
+import { PromoCard } from './PromoCard.js'
+
+export const PromoBanner = () => {
+  const dispatch = useDispatch()
+
+  const accountMessages = useSelector(state => state.account.referralCache.accountMessages)
+  const accountReferral = useSelector(state => state.account.accountReferral)
+
+  // Look for active referral promotion
+  const referralPromoCard = useMemo(() => {
```
thehobbit85, line 67: Don't use 'useMemo' to bundle multiple things into an object.

You can call this outside of the memo: 
`const messageSummary = bestOfMessages(accountMessages, accountReferral)`

And you also don't need to memoize the 'message' and 'iconUri' since they are strings.


Which leaves you with just two 'useCallbacks' for `onPress` and `onClose`.

src/components/themed/PromoCard.js
```diff
@@ -5,71 +5,45 @@ import { TouchableOpacity, View } from 'react-native'
...
-            {message.message}
-          </EdgeText>
-          <TouchableOpacity onPress={this.handleClose}>
-            <AntDesignIcon name="close" color={theme.iconTappable} size={theme.rem(1)} style={styles.close} />
-          </TouchableOpacity>
-        </View>
-      </ButtonBox>
-    )
-  }
+export const PromoCard = (props: Props) => {
```
thehobbit85, line 19: Pretty much the same comments as in the PromoBanner file

src/components/themed/PromoBanner.js
```diff
@@ -14,6 +17,51 @@ export const PromoBanner = () => {
...
+  const store = useSelector(state => state.core.account.dataStore)
+
+  const [promotions, setPromotions] = useState<Promotion[]>([])
+  const [localFlags, setLocalFlags] = useState<AppFlags>({})
+
+  // Define user profile
+  const profile = useMemo(() => getProfile(), [])
+
+  // Get active promotions
+  useEffect(() => {
```
thehobbit85, line 29: Use the 'useAsyncEffect' hook we have which already checks for safety and catches errors.

src/components/themed/PromoBanner.js
```diff
@@ -14,6 +17,51 @@ export const PromoBanner = () => {
...
+      setPromotions(promos.filter(promo => testProfile(profile, promo)))
+    })
+
+    return () => {
+      abort = true
+    }
+  }, [profile])
+
+  // Test for wyreLinkedBank
+  useEffect(() => {
```
thehobbit85, line 42: Use the 'useAsyncEffect' hook we have which already checks for safety and catches errors.

src/components/themed/PromoBanner.js
```diff
@@ -35,15 +83,35 @@ export const PromoBanner = () => {
...
+
+  return (
+    <Carousel
+      enabled={enabled}
+      autoFillData={autoFillData}
+      mode="parallax"
+      width={300}
+      height={150}
+      data={promoCards}
+      renderItem={({ item }) => <PromoCard message={item.message} iconUri={item.iconUri} onPress={item.onPress} onClose={item.onClose} />}
```
thehobbit85, line 114: This should be a 'useCallback' and not an inline function

src/util/promoHelpers.js
```diff
@@ -0,0 +1,151 @@
...
+  })
+  if (!response.ok) throw new Error('hasWyreLinkedBank not ok')
+  const json = asWyrePaymentMethodResponse(await response.json())
+
+  // The first item in the array is the most recent and only item that matters
+  if (json.data[0]?.status === 'ACTIVE') return true
+  return false
+}
+
+export const getProfile = (): Profile => {
```
thehobbit85, line 98: This should be a hook 'useProfile' and not a getter.
The user might change 'language' or 'location' so those need to be set as state so they can trigger a render when changed.

src/util/promoHelpers.js
```diff
@@ -0,0 +1,151 @@
...
+  language: string,
+  location: string,
+  platform: 'ios' | 'android'
+}
+
+export type AppFlags = {
+  wyreLinkedBank?: boolean
+}
+
+export const getPromotions = async (): Promise<Promotion[]> => {
```
thehobbit85, line 50: This can be a hook 'usePromotions(profile: Profile)' which returns the filtered tested promoCards instead of a getter.

That way you abstract the entire logic out of the component and into the hook.

thehobbit85 (changes requested):
I know this is still a WIP but I already added this comments to help guide the PR

src/hooks/promotions/useWyrePromoFlags.js
```diff
@@ -0,0 +1,54 @@
...
+
+    // Wyre has two possible keys that hold the secret so we need to look in both places
+    let secret
+    try {
+      secret = await store.getItem('co.edgesecure.wyre', 'wyreSecret')
+    } catch (e) {
+      secret = await store.getItem('co.edgesecure.wyre', 'wyreAccountId')
+    }
+
+    fetchData({
```
peachbits, line 44: How is the result from this returned to the caller of useWyrePromoFlags?

peachbits (commented):
Fixups approved

PR #3393, opened 2022-05-22
Payment protocol support in EdgeProvider 

paullinator:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

### DIFF
--- a/src/locales/strings/enUS.json
+++ b/src/locales/strings/enUS.json
@@ -1,5 +1,5 @@
 {
-  "bitpay_metadata_name": "BitPay Invoice ID: %s",
+  "bitpay_metadata_name": "Invoice ID: %s",
   "bitcoin_received": "%1$s Received",
   "dialog_title": "Set Auto Logoff time",
   "share_subject": "Hey, I think you should try %s",
--- a/src/modules/UI/scenes/Plugins/EdgeProvider.js
+++ b/src/modules/UI/scenes/Plugins/EdgeProvider.js
@@ -7,6 +7,7 @@ import type {
   EdgeCurrencyWallet,
   EdgeMetadata,
   EdgeNetworkFee,
+  EdgeParsedUri,
   EdgeReceiveAddress,
   EdgeSpendTarget,
   EdgeTransaction,
@@ -20,6 +21,7 @@ import SafariView from 'react-native-safari-view'
 import { sprintf } from 'sprintf-js'
 import { Bridgeable, update } from 'yaob'
 
+import { launchBitPay } from '../../../../actions/BitPayActions.js'
 import { trackAccountEvent, trackConversion } from '../../../../actions/TrackingActions.js'
 import { selectWallet } from '../../../../actions/WalletActions'
 import { ButtonsModal } from '../../../../components/modals/ButtonsModal.js'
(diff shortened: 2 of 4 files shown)


src/locales/strings/enUS.json
```diff
@@ -1,5 +1,5 @@
 {
-  "bitpay_metadata_name": "BitPay Invoice ID: %s",
+  "bitpay_metadata_name": "Invoice ID: %s",
```
stevenzeiler, line 2: Thanks the hardcoded reference to Bitpay was confusing our customers :D

src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -349,10 +351,16 @@ export class EdgeProvider extends Bridgeable {
...
     const { currencyWallets } = this._state.core.account
     const edgeWallet = currencyWallets[this._state.ui.wallets.selectedWalletId]
-    const result = await edgeWallet.parseUri(uri)
 
+    const result: EdgeParsedUri & { paymentProtocolURL?: string } = await edgeWallet.parseUri(uri)
     const { currencyCode = result.currencyCode, customNetworkFee, metadata, lockInputs = true, uniqueIdentifier, orderId } = options
 
+    // Check is PaymentProtocolUri
+    if (!!result.paymentProtocolURL && !result.publicAddress) {
+      await launchBitPay(result.paymentProtocolURL, { wallet: edgeWallet, metadata }).catch(showError)
```
stevenzeiler, line 360: Perfect. I will take a look at `launchBitPay` to see if it can be modified to better conform with the full payment protocol spec. Ideally it will eventually be renamed throughout to something like `launchPayProtocol`.

src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -349,10 +351,16 @@ export class EdgeProvider extends Bridgeable {
     console.log(`requestSpendUri ${uri}`)
     const { currencyWallets } = this._state.core.account
     const edgeWallet = currencyWallets[this._state.ui.wallets.selectedWalletId]
-    const result = await edgeWallet.parseUri(uri)
 
+    const result: EdgeParsedUri & { paymentProtocolURL?: string } = await edgeWallet.parseUri(uri)
     const { currencyCode = result.currencyCode, customNetworkFee, metadata, lockInputs = true, uniqueIdentifier, orderId } = options
 
+    // Check is PaymentProtocolUri
+    if (!!result.paymentProtocolURL && !result.publicAddress) {
```
thehobbit85, line 359: Don't use a sketchy-boolean, instead explicitly check for null like this:
`if (result.paymentProtocolURL != null && result.publicAddress == null)`

thehobbit85 (changes requested):

src/modules/UI/scenes/Plugins/EdgeProvider.js
paullinator: This was copy/pasted from the logic in the AddressTile

https://github.com/EdgeApp/edge-react-gui/blob/d8788689a2b6d46fd277750543eece24c64a287b/src/components/tiles/AddressTile.js#L113

paullinator: Done. I also removed the check for `result.publicAddress == null` since that's not necessary to proceed w/payment protocol. The address existing can just be a backup but we can prefer the payment proto.

peachbits (approved):

PR #3397, opened 2022-05-24
Deprecate currency code from FiatAmountTile and FiatText

peachbits:
New hooks have forced the deprecation of this prop in favor tokenId

---
  - https://app.asana.com/0/0/1202312168060953


swansontec (approved):
The last commit looks great! I see the "HACK" comment, and this is fine since we have set the proper expectation.

However, the first two commits are going to conflict with PR #3398. Once that PR gets merged, this will have to be rebased.

PR #3398, opened 2022-05-24
Fix the side menu crash on wallet archive

swansontec:
We were grabbing the selected wallet in an unsafe way,
so pull all that logic into a hook and give it a cleanup.

#### PR Requirements

- [x] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1202325280482200

### DIFF
--- a/src/hooks/useSelectedWallet.js
+++ b/src/hooks/useSelectedWallet.js
@@ -1,0 +1,36 @@
+// @flow
+
+import { type EdgeCurrencyWallet } from 'edge-core-js'
+
+import { useWatchAccount } from '../hooks/useWatch.js'
+import { useSelector } from '../types/reactRedux.js'
+import { getTokenId } from '../util/CurrencyInfoHelpers.js'
+
+export type SelectedWallet = {
+  currencyCode: string,
+  tokenId?: string,
+  wallet: EdgeCurrencyWallet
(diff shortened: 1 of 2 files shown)


thehobbit85 (approved):

src/hooks/useSelectedWallet.js
```diff
@@ -0,0 +1,41 @@
...
+  const currencyWallets = useWatchAccount(account, 'currencyWallets')
+  const wallet = currencyWallets[walletId]
+  if (wallet == null) return
+
+  // We cannot use any hooks after the return statement,
+  // but we don't need to worry about `allTokens` being stale,
+  // because the selected token must exist before being selected,
+  // so the selector above will force us to render in any case:
+  const { allTokens } = wallet.currencyConfig
+  const tokenId =
```
thehobbit85, line 29: You can use the already existing 'getTokenId' method - https://github.com/EdgeApp/edge-react-gui/blob/d8788689a2b6d46fd277750543eece24c64a287b/src/util/CurrencyInfoHelpers.js#L99

thehobbit85 (changes requested):

thehobbit85 (approved):

PR #3401, opened 2022-05-25
Sam/borrow plugin collateral

samholmes:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1202249992745202

### DIFF
--- a/src/plugins/borrow-plugins/common/ApprovableCall.js
+++ b/src/plugins/borrow-plugins/common/ApprovableCall.js
@@ -1,0 +1,67 @@
+// @flow
+
+import { asObject, asOptional, asString } from 'cleaners'
+import { type EdgeCurrencyWallet, type EdgeMetadata, type EdgeSpendInfo, type EdgeToken, type EdgeTransaction } from 'edge-core-js'
+import { ethers } from 'ethers'
+
+import { type ApprovableAction } from '../types'
+import { asBigNumber } from './cleaners/asBigNumber'
+
+export const asTxInfo = asObject({
+  data: asString,
+  to: asString,
+  value: asOptional(asBigNumber),
+  gasPrice: asOptional(asBigNumber),
+  gasLimit: asOptional(asBigNumber)
+})
+export type TxInfo = $Call<typeof asTxInfo>
+
+export type CallInfo = {
+  tx: TxInfo,
+  wallet: EdgeCurrencyWallet,
+  spendToken?: EdgeToken,
+  metadata?: EdgeMetadata
+}
+
+export const makeApprovableCall = async (params: CallInfo): Promise<ApprovableAction> => {
+  const { tx: txInfo, wallet, spendToken, metadata } = params
+  const { gasLimit, gasPrice } = txInfo
+
+  if (gasPrice == null || gasLimit == null) throw new Error('Explicit gas price and limit required for ApprovableAction.')
+
+  const edgeSpendInfo: EdgeSpendInfo = {
+    pluginId: wallet.currencyInfo.pluginId,
+    currencyCode: spendToken?.currencyCode ?? wallet.currencyInfo.currencyCode,
+    spendTargets: [
+      {
+        nativeAmount: txInfo.value ? txInfo.value.toString() : '0',
+        publicAddress: txInfo.to,
+        otherParams: { data: txInfo.data }
+      }
+    ],
+    customNetworkFee: {
+      gasPrice: ethers.utils.formatUnits(gasPrice, 'gwei').toString(),
+      gasLimit: gasLimit.toString()
+    },
+    networkFeeOption: 'custom',
+    metadata
+  }
+  const edgeUnsignedTx: EdgeTransaction = await wallet.makeSpend(edgeSpendInfo)
+
+  return {
+    networkFee: {
+      currencyCode: wallet.currencyInfo.currencyCode,
+      nativeAmount: edgeUnsignedTx.parentNetworkFee ?? '0'
+    },
+
+    approve: async () => {
+      const tx = await wallet.signTx(edgeUnsignedTx)
+      await wallet.broadcastTx(tx)
+      await wallet.saveTx(tx)
+    }
+  }
+}
+
+export const makeTxCalls = async (actionInfos: CallInfo[]): Promise<ApprovableAction[]> => {
+  return await Promise.all(actionInfos.map(makeApprovableCall))
+}
--- a/src/plugins/borrow-plugins/common/util/composeApprovableActions.js
+++ b/src/plugins/borrow-plugins/common/util/composeApprovableActions.js
@@ -1,0 +1,28 @@
+// @flow
+import { add } from 'biggystring'
+
+import { type ApprovableAction } from '../../types'
+
+export const composeApprovableActions = (...actions: ApprovableAction[]): ApprovableAction => {
+  if (
+    !actions.every((action, index, actions) => index === actions.length - 1 || action.networkFee.currencyCode === actions[index + 1].networkFee.currencyCode)
+  ) {
+    // TODO: Maybe we should consider a different type for networkFee
+    throw new Error('Cannot compose ApprovableActions with different networkFee currencyCodes')
+  }
+
+  const currencyCode = actions[0].networkFee.currencyCode
+  const nativeAmount = actions.reduce((sum, action) => add(sum, action.networkFee.nativeAmount), '0')
+
+  return {
+    networkFee: {
+      currencyCode,
+      nativeAmount
+    },
+    approve: async () => {
+      for (const action of actions) {
+        await action.approve()
+      }
+    }
+  }
+}
--- a/src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
+++ b/src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
@@ -2,7 +2,11 @@
 
 import { type Cleaner, asMaybe } from 'cleaners'
 import { type EdgeCurrencyWallet, type EdgeToken } from 'edge-core-js'
+import { BigNumber, ethers } from 'ethers'
 
+import { type CallInfo, asTxInfo, makeApprovableCall, makeTxCalls } from '../../common/ApprovableCall'
+import { asGraceful } from '../../common/cleaners/asGraceful'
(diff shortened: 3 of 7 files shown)


src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
```diff
@@ -122,22 +144,79 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
       },
 
       async deposit(request: DepositRequest): Promise<ApprovableAction> {
-        return {
-          networkFee: {
-            currencyCode: wallet.currencyInfo.currencyCode,
-            nativeAmount: '2100000000000000'
-          },
-          approve: async () => {}
+        const { nativeAmount, tokenId, fromWallet } = asDepositRequest(request)
```
swansontec, line 147: This is a bad way to use cleaners. We aren't tying to go from `any` or `mixed` to typed data - we already know the types. We are just trying to validate constraints. Therefore, it should probably look more like this:

```js
const { nativeAmount, tokenId, fromWallet } = request
validateWalletParam(fromWallet)

const token = getToken(tokenId)
const tokenAddress = getContractAddress(tokenId)
```

This is the same logic you already have, but with the confusing & misleading function names changed to reflect what is actually happening.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
```diff
@@ -122,22 +144,79 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
...
+        const onBehalfOf = (await fromWallet.getReceiveAddress()).publicAddress
+
+        const gasPrice = await aaveNetwork.provider.getGasPrice()
+
+        const tokenContract = await aaveNetwork.makeTokenContract(tokenAddress)
+
+        const txCallInfos: CallInfo[] = []
+
+        const allowance = await tokenContract.allowance(onBehalfOf, aaveNetwork.lendingPool.address)
+        if (!allowance.sub(nativeAmount).gte(0)) {
```
swansontec, line 163: Good, I am glad you check the allowance first before requesting more.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
```diff
@@ -122,22 +144,79 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
...
+            gasLimit: '500000',
+            gasPrice
+          })
+          txCallInfos.push({
+            tx: approveTx,
+            wallet,
+            token,
+            metadata: {
+              name: 'AAVE',
+              category: 'expense:approval',
```
swansontec, line 174: Are these existing categories in the app? I think Edge has `expense:fee`, but not `expense:approval`. I would be very confused if I got `expense:approval` on my tax spreadsheets.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
```diff
@@ -122,22 +144,79 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
...
-          networkFee: {
-            currencyCode: wallet.currencyInfo.currencyCode,
-            nativeAmount: '2100000000000000'
-          },
-          approve: async () => {}
-        }
+        const { nativeAmount, tokenId, toWallet } = asWithdrawRequest(request)
+
+        const token = asEdgeToken(tokenId)
+        const tokenAddress = asTokenAddress(token)
```
swansontec, line 200: Same cleaner confusion here too.

swansontec (changes requested):
I see the same cleaner confusion repeated from the earlier PR in this series. We need to get these validation methods renamed correctly.

I have also left comments on the testbed branch, even though it's not up for PR. Just some stuff I noticed, but obviously doesn't need to be fixed.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
swansontec: We do have `expense:Network Fee` being generated in other parts of the app, so let's use that one.

src/components/scenes/TestScene.js
```diff
@@ -0,0 +1,308 @@
...
+import { useEffect, useState } from '../../types/reactHooks'
+import { useSelector } from '../../types/reactRedux.js'
+import { SceneWrapper } from '../common/SceneWrapper.js'
+import { TextInputModal } from '../modals/TextInputModal'
+import { Airship } from '../services/AirshipInstance'
+import { cacheStyles, getTheme } from '../services/ThemeContext'
+import { EdgeText } from '../themed/EdgeText'
+import { MiniButton } from '../themed/MiniButton'
+
+function usePromise<T>(promise: Promise<T>): [T | void, Error | void] {
```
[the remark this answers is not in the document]
samholmes: This entire file is auxiliary for testing only. Though valuable feedback for code that will be pushed.

src/components/scenes/TestScene.js
```diff
@@ -0,0 +1,308 @@
...
+}.dev
+
+const wbtTokenId = 'd1b98b6607330172f1d991521145a22bce793277'
+const daiTokenId = 'ff795577d9ac8bd7d90ee22b6c1703490b6512fd'
+
+export function TestScene(props: Props) {
+  const theme = getTheme()
+  const styles = getStyles(theme)
+  const account = useSelector(state => state.core.account)
+  const [whenBorrowEngines, setWhenBorrowEngines] = useState(getAaveBorrowEngines(borrowPlugins, account).then(filterActiveBorrowEngines))
```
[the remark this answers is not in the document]
samholmes: Ah yes. Good catch.

src/plugins/helpers/getAaveBorrowPlugins.js
```diff
@@ -0,0 +1,27 @@
...
+
+    const ps = filteredWallets.map(wallet => {
+      return plugin.makeBorrowEngine(wallet).catch(err => {
+        showError(err)
+      })
+    })
+
+    promises.push(...ps)
+  }
+  const out: any = (await Promise.all(promises)).filter(result => result != null)
```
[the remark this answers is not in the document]
samholmes: True, also these type hacks wouldn't be necessary if implemented in TypeScript. Flow is also a constraint to the coding style here.

src/plugins/borrow-plugins/common/util/composeApprovableActions.js
```diff
@@ -0,0 +1,28 @@
...
+  return {
+    networkFee: {
+      currencyCode,
+      nativeAmount
+    },
+    approve: async () => {
+      for (const action of actions) {
+        await action.approve()
+      }
+    }
```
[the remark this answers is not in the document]
samholmes: Thanks!

src/plugins/borrow-plugins/common/ApprovableCall.js
```diff
@@ -0,0 +1,67 @@
...
+import { asBigNumber } from './cleaners/asBigNumber'
+
+const asTxInfo = asObject({
+  data: asString,
+  to: asString,
+  value: asOptional(asBigNumber),
+  gasPrice: asOptional(asBigNumber),
+  gasLimit: asOptional(asBigNumber)
+})
+type TxInfo = $Call<typeof asTxInfo>
```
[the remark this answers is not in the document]
samholmes: It should instead be a static type instead of a cleaner because we should be trusting ethers.js.

src/components/scenes/TestScene.js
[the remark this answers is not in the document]
samholmes: Not sure if the `useState(() => ...)` is valid syntax. Maybe `useMemo` or `useCallback` is the correct hook for this job. Either way, it doesn't matter because this is just a test scene.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
```diff
@@ -65,6 +70,23 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
       if (tokenId == null) throw new Error('Getting wrapped native token not supported yet. ' + 'Explicitly pass in tokenId for the wrapped token.')
       return tokenId
     }
+    const asWalletParam = (walletParam?: EdgeCurrencyWallet) => {
```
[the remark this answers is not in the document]
samholmes: I went with `validateWalletParam`

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
samholmes: I took your suggestion.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
samholmes: Should we just exclude the sub-category?

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
samholmes: Fixed

src/plugins/borrow-plugins/common/ApprovableCall.js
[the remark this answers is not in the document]
swansontec: If we don't have type definitions for a library, we should write our own, and then cast the `any` to the expected thing. That way, if we move to Typescript, of if the library eventually adds types, our cast should become a no-op and we can just delete it.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
swansontec: We should ask @paullinator what he wants, since the categories are important for a good UX. I personally vote for `expense:Network Fee`, which we are already using.

swansontec (approved):
I am approving this. It looks like we are going with the "We don't trust ethers" approach, which is fine, although perhaps a simple typecast would have sufficed.

PR #3402, opened 2022-05-25
Jon/feat/aave/tappable-card

Jon-edge:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

Screen Shot 2022-05-24 at 6 04 57 PM

---
  - https://app.asana.com/0/0/1202334361628187

### DIFF
--- a/src/components/cards/TappableCard.js
+++ b/src/components/cards/TappableCard.js
@@ -1,0 +1,60 @@
+// @flow
+
+import * as React from 'react'
+import { TouchableOpacity, View } from 'react-native'
+import FontAwesome5 from 'react-native-vector-icons/FontAwesome5'
+
+import { memo } from '../../types/reactHooks.js'
+import { type Theme, cacheStyles, useTheme } from '../services/ThemeContext.js'
+import { Card } from './Card'
+
+/**
+ * An (optionally) tappable card that displays its children in up to two left/right
+ * sections. If the card is configured to be tappable, a chevron is drawn on the
+ * right side of the card.
+ */
+const TappableCardComponent = ({
+  children,
+  nonTappable = false,
+  onPress
+}: {
+  children: React.Node,
+  nonTappable?: boolean,
+  onPress?: any => Promise<void> | void
+}) => {
+  const theme = useTheme()
+  const styles = getStyles(theme)
+
+  return (
+    <View style={styles.container}>
+      <TouchableOpacity onPress={onPress}>
+        <Card>
+          <View style={styles.cardContainer}>
+            <View style={styles.spacedContainer}>{children}</View>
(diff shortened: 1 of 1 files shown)


src/components/cards/TappableCard.js
```diff
@@ -0,0 +1,76 @@
...
+ */
+const TappableCardComponent = ({
+  leftChildren,
+  nonTappable = false,
+  onPress,
+  rightChildren
+}: {
+  leftChildren: React.Node,
+  nonTappable?: boolean,
+  onPress?: any => void | (any => Promise<void>),
```
samholmes, line 24: `onPress?: any => void | Promise<void>` would be the same, no?

src/components/cards/TappableCard.js
```diff
@@ -0,0 +1,76 @@
...
+ * right side of the card.
+ */
+const TappableCardComponent = ({
+  leftChildren,
+  nonTappable = false,
+  onPress,
+  rightChildren
+}: {
+  leftChildren: React.Node,
+  nonTappable?: boolean,
```
samholmes, line 23: Remove the negative, just `isTappable` which defaults to `true`.

src/components/cards/TappableCard.js
```diff
@@ -0,0 +1,76 @@
...
+  const styles = getStyles(theme)
+
+  return (
+    <View style={styles.container}>
+      <TouchableOpacity onPress={onPress}>
+        <Card>
+          <View style={styles.cardContainer}>
+            <View style={styles.spacedContainer}>
+              <View style={styles.leftContainer}>{leftChildren}</View>
+              <View style={styles.rightContainer}>{rightChildren}</View>
```
samholmes, line 37: Why not just `children` instead of `leftChildren` and `rightChildren` being separate? The parent component can pass any number of children it can get away with fitting within the card.

samholmes (changes requested):
Squash the last two commits because the final commit is really a fixup for the penultimate commit. We shouldn't be reverting code to a commit which is include in the same PR.

Also, why is this PR dependent on the future of another branch? Seems like the TappableCard stands on its own.

src/components/cards/TappableCard.js
Jon-edge: Set it up this way because our other components seem to do this. 

I believe it allows flags to be easily set when calling the component with `<Component negativeFlag>` instead of `<Component flag={false}>`

src/components/cards/TappableCard.js
Jon-edge: Flow complains

samholmes: `onPress?: any => Promise<void> | void` should work

PR #3408, opened 2022-05-25
Sam/borrow plugin debt

samholmes:
#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1202249992745203

### DIFF
--- a/src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
+++ b/src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
@@ -31,6 +31,7 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
     const walletAddress = (await wallet.getReceiveAddress()).publicAddress
 
     const REFERRAL_CODE = 0 // No referral code is used for AAVE contract calls
+    const INTEREST_RATE_MODE = 2 // Only variable is supported for now
 
     //
     // Private Methods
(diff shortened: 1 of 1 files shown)


src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
```diff
@@ -219,22 +230,60 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
...
       },
       async borrow(request: BorrowRequest): Promise<ApprovableAction> {
-        return {
-          networkFee: {
-            currencyCode: wallet.currencyInfo.currencyCode,
-            nativeAmount: '2100000000000000'
-          },
-          approve: async () => {}
-        }
+        const { nativeAmount, tokenId, fromWallet } = asBorrowRequest(request)
```
swansontec, line 233: Same cleaner issue here as in the previous PR.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
```diff
@@ -219,22 +230,60 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
...
       },
       async repay(request: RepayRequest): Promise<ApprovableAction> {
-        return {
-          networkFee: {
-            currencyCode: wallet.currencyInfo.currencyCode,
-            nativeAmount: '2100000000000000'
-          },
-          approve: async () => {}
-        }
+        const { nativeAmount, tokenId, fromWallet } = asRepayRequest(request)
```
swansontec, line 261: Same cleaner issue here as in the previous PR.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
```diff
@@ -219,22 +230,60 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
...
+        const onBehalfOf = (await fromWallet.getReceiveAddress()).publicAddress
+
+        const gasPrice = await aaveNetwork.provider.getGasPrice()
+
+        const repayTx = await aaveNetwork.lendingPool.populateTransaction.repay(asset, amount, INTEREST_RATE_MODE, onBehalfOf, {
+          gasLimit: '800000',
+          gasPrice
+        })
+
+        return await makeApprovableCall({
```
swansontec, line 277: We don't need an approval for this? How do we return the USDC we borrowed earlier, without sending it back to AAVE?

swansontec (changes requested):
Requesting changes because of the cleaner confusion from earlier. I do have one other question.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.js
samholmes: It didn't mention so in the documentation for this function (https://docs.aave.com/developers/v/2.0/the-core-protocol/lendingpool#repay) though it would make sense that this is required... Perhaps it omitted it because it is assumed that the borrower account has already approved by the `deposit` method. That assumption isn't guaranteed, especially if this is invoked not by the borrower but by an account repaying on behalf of the borrower.

I can add the approve just to be sure.

swansontec (approved):

PR #3415, opened 2022-05-30
Revert currencyCode to UID when returning from chooseCurrencyWallet

peachbits:
---
  - https://app.asana.com/0/0/1202356314982260


swansontec (approved):
Oh, yeah! This would be necessary if the call it by some weird name. I think the method name `revertExtendedCurrencyCodes` could be `unfixCurrencyCode`, to reflect the idea that we are undoing the `fixCurrencyCodes` conversion, but this is up to you.

peachbits:
> 

Good idea! Will change and then merge

PR #3417, opened 2022-05-31
Prevent Double Spends on SendScene

paullinator:
Use new SmartSlider which locks itself upon slide completion

#### PR Requirements

- [ ] Tested on iOS Tablet
- [ ] Tested on small Android
- [ ] n/a

---
  - https://app.asana.com/0/0/1201921891215450


src/components/scenes/SendScene.js
```diff
@@ -329,22 +323,19 @@ class SendComponent extends React.PureComponent<Props, State> {
     }
   }
 
-  submit = async () => {
+  submit = async (doResetSlider: Function) => {
+    this.resetSliderFunc = doResetSlider
```
thehobbit85, line 327: Not needed with the comment from the previous commit

src/components/scenes/SendScene.js
```diff
@@ -53,7 +53,6 @@ type StateProps = {
   error: Error | null,
   exchangeRates: GuiExchangeRates,
   nativeAmount: string | null,
-  pending: boolean,
   pin: string,
   resetSlider: boolean,
```
thehobbit85, line 57: Not needed with the comment from the previous commit

thehobbit85 (changes requested):

thehobbit85 (approved):

PR #3427, opened 2022-06-02
Limit reserve tokens by configuration

samholmes:
Fix the performance issue caused by fetching all the token balances for all supported AAVE reserver tokens for v1. Solution is to limit which tokens are enabled for the network until some caching/persistence is implemented to improve the loading performance post initialization.

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1202369707618481

### DIFF
--- a/src/plugins/borrow-plugins/plugins/aave/AaveNetwork.js
+++ b/src/plugins/borrow-plugins/plugins/aave/AaveNetwork.js
@@ -15,7 +15,8 @@ export type AaveNetworkBlueprint = {
   contractAddresses: {
     lendingPool: string,
     protocolDataProvider: string
-  }
+  },
+  enabledTokens: { [currencyCode: string]: boolean }
 }
 
 export type AaveNetwork = {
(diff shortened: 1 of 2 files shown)


src/plugins/borrow-plugins/plugins/aave/AaveNetwork.js
```diff
@@ -15,7 +15,8 @@ export type AaveNetworkBlueprint = {
   contractAddresses: {
     lendingPool: string,
     protocolDataProvider: string
-  }
+  },
+  enabledTokens: { [sybmbol: string]: boolean }
```
swansontec, line 19: Please add a comment that these are currencyCodes, not tokenId's. Or, change `symbol` to `currencyCode` Ideally we would be using tokenId's for everything, but as long as this is an internal implementation detail, it's ok.

swansontec (changes requested):

src/plugins/borrow-plugins/plugins/aave/AaveNetwork.js
samholmes: Done

PR #3450, opened 2022-06-13
Fix an issue where the it was calling a prop outside of the scope of the animation thread

thehobbit85:
#### PR Requirements

- [ ] Tested on iOS device


thehobbit85 (commented):
Approved OTS by @peachbits

PR #3455, opened 2022-06-14
Automate env.json setup during `yarn prepare`

GoodluckH:
The automation process aims to remove the manual setup process for
`env.json`.

~~The script requires [jq](https://stedolan.github.io/jq/) for merging `env.example.json` and `env.json`. 
Therefore, `brew install jq` is included in the shell script.~~

If `env.json` does not exist in the root project directory, then it will
be created with content mirroring that of `env.example.json`.

Otherwise, `env.json` will be updated to reflect the merge of `env.json`
and `env.example.json`. ~~The merging script~~ `config.js` recognizes 
`env.json` as the canonical file where existing values will remain as-is, if 
the example json file has empty corresponding values.

If new keys/fields are introduced in the example json file, these new
fields will be reflected in `env.json`.

Please advise if `env.example.json` is still relevant.

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1202446369187201


paullinator:
@thehobbit85 isn't this what cleaner-config is for?

thehobbit85:
> cleaner-config

It's funny you're saying that. That was the original idea, to use the cleaner-config.

And after I already explained to both @GoodluckH and @Jon-edge how to use the 'cleaner-config', we realized that it won't work in react-native since the library is currently hard-coded using node modules. 

We can still use the 'configure' script but not the 'makeConfig' function instead of the code.

So unless we want to update the 'cleaner-config' to accept the file system functions (load, save) as props, and then inject them from any environment, which I actually in favor of doing, then we might as well just go with the most basic solution.

But like I said, I am very much in favor of update the 'cleaner-config' and then using it in the GUI as well.

paullinator:
@thehobbit85 you could have a common file that has the cleaner that's run from both a nodejs script upon yarn prepare as well as being included in RN gui code.

scripts/prepare.sh
```diff
@@ -44,3 +44,12 @@ mkdir -p "$core_assets"
 # Bundle currency, swap, & rate plugins:
 echo Webpacking plugins...
 node ./node_modules/.bin/webpack
+
+# Env file setup:
+brew install jq
```
thehobbit85, line 49: This would break Horribly on every system that is not a mac, which is ALL the build systems.

All the automated CI/CD tools (Travis, Jenkins) and even a big portion of our developers uses linux and even god forbid Windows.

thehobbit85 (changes requested):

thehobbit85 (approved):

PR #3456, opened 2022-06-15
Sam/edge currency plugins

samholmes:
# Upgrade to new edge-currency-plugins

### New notice modal regarding migration to Blockbook

Simulator Screen Shot - iPhone 8 - 2022-06-14 at 19 07 39

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1190725190171985


thehobbit85 (changes requested):
Update the modal to behave like we discusses.
Other then that everything looks great!

thehobbit85 (approved):

PR #3460, opened 2022-06-16
Remove dispatch dependency from DeepLinkingManager's useEffect

peachbits:
---
  - https://app.asana.com/0/0/1200051810094156

### DIFF
--- a/src/components/services/DeepLinkingManager.js
+++ b/src/components/services/DeepLinkingManager.js
@@ -5,6 +5,7 @@ import { Linking } from 'react-native'
 
 import ENV from '../../../env.json'
 import { launchDeepLink, retryPendingDeepLink } from '../../actions/DeepLinkingActions.js'
+import { useAsyncEffect } from '../../hooks/useAsyncEffect'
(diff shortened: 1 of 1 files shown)


src/components/services/DeepLinkingManager.js
```diff
@@ -28,23 +29,24 @@ export function DeepLinkingManager(props: Props) {
     requestAnimationFrame(() => dispatch(retryPendingDeepLink()))
   }, [accountReferralLoaded, dispatch, pendingDeepLink, wallets])
 
+  const handleUrl = url => {
+    try {
+      dispatch(launchDeepLink(parseDeepLink(url)))
+    } catch (error) {
+      showError(error)
+    }
+  }
```
thehobbit85, line 38: Why was this moved outside of the 'useAsyncEffect'?
It shouldn't matter too much but I think it's better if you put it back

thehobbit85 (changes requested):

thehobbit85 (approved):

thehobbit85 (approved):

PR #3461, opened 2022-06-17
Deprecate Actions for components

GoodluckH:
Replace existing `Actions` with the mock version of the React Navigation
lib.

This PR limits the deprecation to files under `src/components` only,
along with some necessary changes for test units.

Note that a `getNavigation` was created for `handlePluginBack()` to
consume. This might not be the best practice since it might not be
replicatable when migrating to React Navigation. Please advise how to
treat the said function with the mock navigation.

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1201388479070450

### DIFF
--- a/config.js
+++ b/config.js
@@ -28,13 +28,6 @@ const asConfig = asObject({
     }),
     { apiKey: '' }
   ),
-  x_CHANGELLY_INIT: asOptional(
-    asObject({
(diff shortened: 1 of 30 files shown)


GoodluckH:
Remove this PR due to upstream URL change. This PR is continued here: #3472

PR #3463, opened 2022-06-17
Add `isImportKeySupported` to UFO and EBOOST

samholmes:
#### PR Requirements

- [ ] Tested on iOS device

### DIFF
--- a/src/constants/WalletAndCurrencyConstants.js
+++ b/src/constants/WalletAndCurrencyConstants.js
@@ -547,12 +547,20 @@ export const SPECIAL_CURRENCY_INFO: {
   eboost: {
     initWalletName: s.strings.string_first_eboost_wallet_name,
     chainCode: 'EBST',
+    isImportKeySupported: {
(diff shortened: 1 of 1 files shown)


src/constants/WalletAndCurrencyConstants.js
```diff
@@ -547,12 +547,20 @@ export const SPECIAL_CURRENCY_INFO: {
   eboost: {
     initWalletName: s.strings.string_first_eboost_wallet_name,
     chainCode: 'EBST',
+    isImportKeySupported: {
```
thehobbit85, line 550: No need to add to eboost since it's in 'keysOnlyMode' mode so it won't appear anyway.

thehobbit85 (changes requested):

src/constants/WalletAndCurrencyConstants.js
samholmes: Keeping it just incase we reenable eboost

PR #3464, opened 2022-06-18
Jon/aave/loan-details-scene

Jon-edge:
#### PR Requirements

- [ ] Tested on iOS device


samholmes:
Can we rename the scene `LoanCreateScene` or `LoanCreationScene` so we can use `LoanDetailsScene` as the name for the scene that shows details for an existing loan?

PR #3467, opened 2022-06-20
UI2 Login Approval Scene

itayplav:
#### PR Requirements

- [x] Tested on iOS device

https://user-images.githubusercontent.com/4023066/174657477-70978824-bf5a-4b3e-b375-09d70cbdb3d5.mp4

---
  - https://app.asana.com/0/0/1201611529606705


src/components/scenes/WaitScene.tsx
```diff
@@ -0,0 +1,50 @@
...
+type Props = {
+  message: string
+  route: RouteProp<'waitScene'>
+}
+
+export const WaitScene = (props: Props) => {
+  const { message } = props.route.params
+
+  const theme = useTheme()
+  const styles = getStyles(theme)
```
swansontec, line 18: We have too many of these scenes already. We have `LoadingScene`, `FullScreenLoader`, and `AirshipFullScreenSpinner`, at least. There is really no need for another one of these. If we really care about animations, maybe we can update one of the other scenes. However, Lottie seems like a pretty heavy dependency to take on.

src/components/scenes/EdgeLoginScene.tsx
```diff
@@ -1,242 +1,158 @@
...
+  const error = useSelector(state => state.core.edgeLogin.error)
+  const isProcessing = useSelector(state => state.core.edgeLogin.isProcessing)
+  const lobby = useSelector(state => state.core.edgeLogin.lobby)
+  const dispatch = useDispatch()
+
+  const handlePress = useHandler(() => {
+    navigation.navigate('waitScene', { message: s.strings.edge_login_fetching_message })
+    dispatch(lobbyLogin()).then(() => {
+      navigation.goBack()
+    })
```
swansontec, line 35: Since we already have the `isProcessing` state, we don't need to push this whole other `waitScene` scene on top. The existing spinner is prefectly fine.

src/components/scenes/EdgeLoginScene.tsx
```diff
@@ -1,242 +1,158 @@
...
+export const EdgeLoginScene = (props: Props) => {
+  const { navigation } = props
+  const theme = useTheme()
+  const styles = getStyles(theme)
 
-type Props = StateProps & DispatchProps & OwnProps
+  const error = useSelector(state => state.core.edgeLogin.error)
+  const isProcessing = useSelector(state => state.core.edgeLogin.isProcessing)
+  const lobby = useSelector(state => state.core.edgeLogin.lobby)
+  const dispatch = useDispatch()
```
swansontec, line 29: I would love to pull all the state out of Redux - pass the login token as part of the route props, and then use `React.useState` to manage all this stuff, instead of Redux and actions.

src/components/scenes/EdgeLoginScene.tsx
```diff
@@ -1,242 +1,158 @@
...
-        <View style={styles.buttons}>
-          <SecondaryButton style={styles.cancel} onPress={handleDecline}>
-            <SecondaryButton.Text>{s.strings.string_cancel_cap}</SecondaryButton.Text>
-          </SecondaryButton>
-          <PrimaryButton style={styles.submit} onPress={this.props.accept}>
-            <PrimaryButton.Text>{s.strings.accept_button_text}</PrimaryButton.Text>
-          </PrimaryButton>
-        </View>
+      <View style={styles.buttons}>
+        <MainButton marginRem={[1, 0, 1, 0]} label={s.strings.accept_button_text} onPress={handlePress} />
```
swansontec, line 76: If you pass an async `handlePress` to `MainButton`, it will show a spinner while it does the work. This may be simpler & better looking than the big `if` statement you have here to manually show a spinner.

swansontec (changes requested):

src/components/scenes/EdgeLoginScene.tsx
```diff
@@ -1,242 +1,158 @@
...
+  const renderImage = useHandler(() => {
+    if (lobby && lobby?.loginRequest && lobby?.loginRequest?.displayImageUrl) {
+      return <Image source={theme.primaryLogo} style={styles.logo} resizeMode="contain" />
     }
     return null
-  }
+  })
 
-  renderHeader() {
+  const renderHeader = useHandler(() => {
```
swansontec, line 89: This `useHandler` approach is not necessary, since these are not event handlers, but rather helpers for the render function. You can either drop the `useHandler` wrapper and leave these as plain functions, or you can wrap the blocks with `useMemo` to *actually* get caching logic, so the sections only re-render when real changes happen.

swansontec (changes requested):

src/components/scenes/EdgeLoginScene.tsx
```diff
@@ -1,242 +1,118 @@
...
-}
+export const EdgeLoginScene = (props: Props) => {
+  const { navigation, route } = props
+  const { account } = useSelector(state => state.core)
+  const { loginToken } = route.params
+  const [isLobby, setIsLobby] = React.useState<boolean>(false)
+  const [isApproved, setIsApproved] = React.useState<boolean>(false)
+  const isLoading = React.useMemo(() => {
+    return !isLobby || isApproved
+  }, [isLobby, isApproved])
```
swansontec, line 31: We don't actually need a `useMemo` on this, since it's simple boolean (so `===` works) and it's cheap to calculate. Just do `const isLoading = !isLobby || isApproved`, and you'll have the exact same result.

Well, actually, with the other refactors, we will probably also need `const isLobby = lobby != null`, but yeah, same thimg.

swansontec (changes requested):

swansontec (approved):
Approved, but I have two optional changes for you to do if you like.

swansontec (approved):

PR #3483, opened 2022-06-23
Add DevLink to DeepLink

samholmes:
This fixes YOLO_DEEP_LINK dev env.json config. Before these changes, the redirect would repeat in an infinite loop.

#### PR Requirements

- [ ] Tested on iOS device

### DIFF
--- a/src/util/DeepLinkParser.js
+++ b/src/util/DeepLinkParser.js
@@ -21,6 +21,10 @@ export function parseDeepLink(uri: string, opts: { aztecoApiKey?: string } = {})
 
   const url = new URL(uri)
 
+  if (url.protocol === 'dev:') {
+    return { type: 'dev', sceneName: url.pathname }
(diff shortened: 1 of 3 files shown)


src/util/DeepLinkParser.js
```diff
@@ -21,6 +21,10 @@ export function parseDeepLink(uri: string, opts: { aztecoApiKey?: string } = {})
 
   const url = new URL(uri)
 
+  if (url.protocol === 'dev:') {
```
thehobbit85, line 24: Please block this feature behind the development build flag so it won't work on live builds

thehobbit85 (changes requested):
1. Remove the first commit since it's already fixed in a better way.
2. Rebase the second commit and resolve conflicts if needed.

thehobbit85 (approved):

PR #3485, opened 2022-06-24
Implement CryptoFiatAmountGroup

Jon-edge:
Based on an input native fiat amount, wallet, and token, return a row showing:
- Crypto icon,
- The equivalent formatted/localized crypto amount
- Fiat icon
- Fiat (display) amount

#### PR Requirements

- [x] Tested on iOS device
- [ ] 
Screen Shot 2022-06-23 at 6 46 29 PM

---
  - https://app.asana.com/0/0/1202491779403713

### DIFF
--- a/src/hooks/useFiatText.js
+++ b/src/hooks/useFiatText.js
@@ -14,6 +14,7 @@ type Props = {
   cryptoCurrencyCode: string,
   cryptoExchangeMultiplier?: string,
   fiatSymbolSpace?: boolean,
+  hideFiatSymbol?: boolean,
   isoFiatCurrencyCode?: string,
   nativeCryptoAmount?: string,
   noGrouping?: boolean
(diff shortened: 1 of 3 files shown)


src/hooks/useFiatText.js
```diff
@@ -49,7 +51,7 @@ export const useFiatText = (props: Props): string => {
         })
       : '0'
 
-  const fiatSymbol = `${getSymbolFromCurrency(isoFiatCurrencyCode)}${fiatSymbolSpace ? ' ' : ''}`
+  const fiatSymbol = hideFiatSymbol ? '' : `${getSymbolFromCurrency(isoFiatCurrencyCode)}${fiatSymbolSpace ? ' ' : ''}`
```
thehobbit85, line 54: 'hideFiatSymbol' can be either a boolean of an undefined. 

So either set the default value of 'hideFiatSymbol' to false or explicitly test for `hideFiatSymbol === true'.
This is something typescript would give an error about.

thehobbit85 (changes requested):

PR #3497, opened 2022-07-01
Add DISABLE_WARNINGS to env.example.json

samholmes:
#### PR Requirements

- [ ] Tested on iOS device


env.example.json
```diff
@@ -80,5 +80,6 @@
   "-------- debug options --------": 0,
   "DEBUG_CORE": false,
   "DEBUG_PLUGINS": false,
-  "DEBUG_VERBOSE_ERRORS": false
+  "DEBUG_VERBOSE_ERRORS": false,
+  "DISABLE_WARNINGS": false
```
thehobbit85, line 84: Since we are using cleaner-config now, you need add this to the config file: https://github.com/EdgeApp/edge-react-gui/blob/develop/config.js

thehobbit85 (changes requested):

env.example.json
samholmes: Oh. Fixed it then.

thehobbit85 (approved):

thehobbit85 (changes requested):
Needs to fix the travis file

thehobbit85 (approved):

thehobbit85 (approved):

PR #3498, opened 2022-07-02
Links Edge site to external browser for Android

GoodluckH:
#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1201957615380295

### DIFF
--- a/src/components/modals/HelpModal.js
+++ b/src/components/modals/HelpModal.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as React from 'react'
-import { Image, Keyboard, Linking, View } from 'react-native'
+import { Image, Keyboard, Linking, Platform, View } from 'react-native'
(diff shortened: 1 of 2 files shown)


src/components/modals/HelpModal.js
```diff
@@ -75,6 +75,22 @@ export class HelpModalComponent extends React.Component<Props & ThemeProps> {
...
+            })
+          } else {
+            console.log("Don't know how to open URI: " + HELP_URIS.site)
+          }
+        })
+      } else {
+        showWebViewModal(HELP_URIS.site, helpSiteMoreInfoText)
+      }
+    }
+
```
thehobbit85, line 93: Since the HelpModalComponent is a class component, you should add this function as a class function (like the 'handleClose' above)

thehobbit85 (changes requested):

PR #3499, opened 2022-07-04
Enable MATIC/USDC sell on Wyre

paullinator:
This PR cleans up token handling in EdgeProvider and updates to the latest edge-plugin-wyre which requires the newer EdgeProvider to properly handle MATIC and MATIC-USDC tokens. 

This is pretty valuable as many AAVE users borrow MATIC-USDC and need to sell it to their bank.

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1202542030026805

### DIFF
--- a/src/modules/UI/scenes/Plugins/EdgeProvider.js
+++ b/src/modules/UI/scenes/Plugins/EdgeProvider.js
@@ -43,6 +43,7 @@ type WalletDetails = {
   receiveAddress: {
     publicAddress: string
   },
+  chainCode: string,
   currencyCode: string,
   fiatCurrencyCode: string
 }
@@ -239,6 +240,7 @@ export class EdgeProvider extends Bridgeable {
       name: walletName,
       pluginId: edgeWallet.currencyInfo.pluginId,
       receiveAddress,
+      chainCode: this.selectedChainCode,
       currencyCode,
       fiatCurrencyCode: edgeWallet.fiatCurrencyCode.replace('iso:', ''),
       currencyIcon: icons.symbolImage,
--- a/src/modules/UI/scenes/Plugins/EdgeProvider.js
+++ b/src/modules/UI/scenes/Plugins/EdgeProvider.js
@@ -23,7 +23,6 @@ import { Bridgeable, update } from 'yaob'
 
 import { launchBitPay } from '../../../../actions/BitPayActions.js'
 import { trackAccountEvent, trackConversion } from '../../../../actions/TrackingActions.js'
-import { selectWallet } from '../../../../actions/WalletActions'
 import { ButtonsModal } from '../../../../components/modals/ButtonsModal.js'
 import { type WalletListResult, WalletListModal } from '../../../../components/modals/WalletListModal.js'
 import { Airship, showError, showToast } from '../../../../components/services/AirshipInstance.js'
@@ -108,6 +107,9 @@ export class EdgeProvider extends Bridgeable {
   deepQuery: UriQueryMap | void
   promoCode: string | void
   restartPlugin: () => void
+  selectedWallet: EdgeCurrencyWallet
+  selectedChainCode: string
+  selectedCurrencyCode: string
 
   constructor(
     plugin: GuiPlugin,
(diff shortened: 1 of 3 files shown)


src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -163,18 +163,55 @@ export class EdgeProvider extends Bridgeable {
...
+        } else {
+          // Tokens for all other chains must be specified using a full currency code. ie. "ETH-USDC"
+          returnCurrencyCode = `${chainCode}-${tokenCode}`
+        }
+      } else {
+        // This is a mainnet coin.
+        returnCurrencyCode = chainCode
+      }
+
+      this.selectedChainCode = chainCode
```
peachbits, line 213: These won't be set if the function returns on line 175

src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -163,18 +163,55 @@ export class EdgeProvider extends Bridgeable {
...
+            returnCurrencyCode = (allowedCurrencyCodes ?? []).find(m => m === `ETH-${tokenCode}`)
+          }
+          if (returnCurrencyCode == null) {
+            throw new Error(`Internal error. Tokencode ${tokenCode} selected but not in allowedCurrencyCodes`)
+          }
+        } else {
+          // Tokens for all other chains must be specified using a full currency code. ie. "ETH-USDC"
+          returnCurrencyCode = `${chainCode}-${tokenCode}`
+        }
+      } else {
```
peachbits, line 208: For readability, these blocks could be flattened and each conditional returned immediately.

peachbits (changes requested):
It took a while but I think the logic is sound and I only found a couple issues. That being said, I don't like it. The Wyre plugin could have used existing tokenId functionality to choose a wallet instead of overcrowding this function with additional logic with chainCodes and full currency codes. We really need to add tests to be sure that it's working as intended.

src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -163,18 +163,55 @@ export class EdgeProvider extends Bridgeable {
...
+            // See if tokenCode matches one of the allowedCurrencyCodes
+            returnCurrencyCode = (allowedCurrencyCodes ?? []).find(m => m === tokenCode)
+          }
+
+          // If no match found. Check if a full currency code was used.
+          if (returnCurrencyCode == null) {
+            returnCurrencyCode = (allowedCurrencyCodes ?? []).find(m => m === `ETH-${tokenCode}`)
+          }
+          if (returnCurrencyCode == null) {
+            throw new Error(`Internal error. Tokencode ${tokenCode} selected but not in allowedCurrencyCodes`)
```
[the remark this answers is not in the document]
paullinator: They actually do provide currencyCodes now.

src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -163,18 +163,55 @@ export class EdgeProvider extends Bridgeable {
     if (walletId && currencyCode) {
       this.selectedWallet = this._state.core.account.currencyWallets[walletId]
       if (this.selectedWallet == null) throw new Error(`Missing wallet for walletId`)
+      const chainCode = this.selectedWallet.currencyInfo.currencyCode
+      const tokenCode = currencyCode
       const { pluginId } = this.selectedWallet.currencyInfo
       const tokenId = getTokenId(this._state.core.account, pluginId, currencyCode)
 
-      // If allowedCurrencyCodes is an array of EdgeTokenIdExtended objects
+      let returnCurrencyCode = ''
```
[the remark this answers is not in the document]
paullinator: Done

src/modules/UI/scenes/Plugins/EdgeProvider.js
paullinator: fixed

src/modules/UI/scenes/Plugins/EdgeProvider.js
paullinator: Agreed there's only a little improvement to be had but will do the cleanup in a new PR

PR #3504, opened 2022-07-06
Fix the typing for 'useWatch' and remove the 'useWatch' aliases

thehobbit85:
#### PR Requirements

- [ ] Tested on iOS device

### DIFF
--- a/src/hooks/useWatch.js
+++ b/src/hooks/useWatch.js
@@ -1,16 +1,14 @@
 // @flow
 
-import { type EdgeAccount, type EdgeContext, type EdgeCurrencyConfig, type EdgeCurrencyWallet } from 'edge-core-js'
+import { type Subscriber } from 'yaob'
 
 import { useEffect, useState } from '../types/reactHooks.js'
 
 /**
- * Subscribes to changes in a core object's property.
- *
- * Flow has trouble typing this, so use one of the named aliases below.
+ * Subscribes to changes in a yaob (core) object's property.
  */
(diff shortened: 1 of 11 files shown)


src/hooks/useWatch.js
```diff
@@ -1,16 +1,14 @@
...
 
 /**
- * Subscribes to changes in a core object's property.
- *
- * Flow has trouble typing this, so use one of the named aliases below.
+ * Subscribes to changes in a yaob (core) object's property.
  */
-function useWatch(object: any, name: string): any {
-  const [out, setOut] = useState<any>(object[name])
+export const useWatch = <T: Object, Name: $Keys<T>>(object: { +watch: Subscriber<T> }, name: Name): $ElementType<T, Name> => {
```
swansontec, line 10: I see how this works, and I'm both amazed an horrified at how the inference is working here. We basically ignore the `EdgeAccount` properties altogether, and instead look at the `watch` method itself, and the parameters that guy takes. Since the core has solid type definitions, this works perfectly for `useWatch` consumers.

I never thought to try it this way, since the `object[name]` line means we need the property to actually exist as well. Flow, however, just isn't checking line 14 at all, because of course it isn't. This solution only works because Flow is a slouch.

swansontec (approved):

src/hooks/useWatch.js
thehobbit85: So when we finally migrate to typescript we already have the type fix for this hook.

PR #3510, opened 2022-07-08
Resync the wallet after sweeping keys

thehobbit85:
When scanning to sweep private keys, funds not showing up until after resync, so just force a resync after the sweep.

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1202491251474019


peachbits (commented):
As discussed IRL, we shouldn't merge this until a bug related to resync and file handling is resolved in memlet (affecting edge-currency-plugins). Even after that issue is resolved, it's worth exploring again if this is an appropriate solution for failing to see swept funds right away.

PR #3513, opened 2022-07-10
Harden EdgeProvider

paullinator:
- Change chooseCurrencyWallet to not show wallets for codes that are both mainnet and ETH tokens. This removes similar logic from the code that chooses what currency code to return after selection
- Flatten logic in chooseCurrencyWallet that decides what currency code to return
- Add unit tests for getReturnCurrencyCode and upgradeExtendedCurrencyCodes

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1202578075245361

### DIFF
--- a/src/__tests__/edgeProvider.test.js
+++ b/src/__tests__/edgeProvider.test.js
@@ -1,0 +1,7290 @@
+// @flow
+/* globals describe test expect */
+
+import { getReturnCurrencyCode, upgradeExtendedCurrencyCodes } from '../modules/UI/scenes/Plugins/EdgeProvider.js'
+
+describe('upgradeExtendedCurrencyCodes', () => {
+  describe('string list', () => {
+    test('empty array', () => {
+      // MATIC and BNB are ambiguous since they could be an ETH token or a mainnet coin. Therefore it should not get included
+      // As well, BSC is not an actual currencyCode so it won't be included either. BSC can only be specified using object
+      // params with pluginId = binancesmartchain
+      const result = upgradeExtendedCurrencyCodes(currencyConfig, undefined, [])
+      expect(result).toEqual(undefined)
+    })
+    test('undefined', () => {
+      // MATIC and BNB are ambiguous since they could be an ETH token or a mainnet coin. Therefore it should not get included
+      // As well, BSC is not an actual currencyCode so it won't be included either. BSC can only be specified using object
+      // params with pluginId = binancesmartchain
+      const result = upgradeExtendedCurrencyCodes(currencyConfig, undefined, undefined)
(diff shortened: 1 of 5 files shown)


src/util/utils.js
```diff
@@ -589,10 +589,16 @@ export function tokenIdsToCurrencyCodes(currencyConfig: EdgeCurrencyConfig, toke
   return out
 }
 
+export type MiniCurrencyConfig = {
```
peachbits, line 592: This type doesn't appear to be used anywhere and can be removed

src/modules/UI/scenes/Plugins/EdgeProvider.js
```diff
@@ -171,48 +174,28 @@ export class EdgeProvider extends Bridgeable {
...
       this.selectedChainCode = chainCode
 
-      let returnCurrencyCode
+      const unfixCode = unfixCurrencyCode(this._plugin.fixCurrencyCodes, pluginId, tokenId)
+      if (unfixCode != null) {
+        return unfixCode
+      }
 
-      if (allowedCurrencyCodes != null && allowedCurrencyCodes.length > 0 && allowedCurrencyCodes.every(code => typeof code === 'object')) {
+      if (allObjects) {
```
peachbits, line 182: allObjects is true if allowedCurrencyCodes is an empty array because .every() always returns true on empty arrays. Given the param name, and current usage (ie. Banxa), we should return a currency code when allowedCurrencyCodes is empty. I wrote this exact same Banxa breaking bug previously.

src/__tests__/edgeProvider.test.js
```diff
@@ -0,0 +1,7276 @@
...
+        { pluginId: 'ethereum', tokenId: 'dac17f958d2ee523a2206206994597c13d831ec7' }
+      ])
+    })
+  })
+})
+
+describe('getReturnCurrencyCode', () => {
+  describe('string list', () => {
+    test('mainnets', () => {
+      const result = getReturnCurrencyCode(['BTC', 'ETH', 'MATIC'], 'ETH', 'ETH')
```
peachbits, line 68: Need an empty array test

peachbits (changes requested):

src/modules/UI/scenes/Plugins/EdgeProvider.js
paullinator: I was aware  that empty arrays always return true, but did not realize Banxa passed an empty array. I thought they passed undefined. Now they actually pass an array with values. 

Ok I’ll make the update to check for empty array in case others do that

peachbits (approved):

PR #3525, opened 2022-07-17
Paul/simplex banxa plugin

paullinator:
#### PR Requirements

- [ ] Tested on iOS device


src/constants/plugins/buyPluginList.json
```diff
@@ -909,7 +417,7 @@
     },
     "title": "Faster Payments",
     "paymentTypeLogoKey": "fasterPayments",
-    "description": "Fee: 5.5% / Settlement: Instant - 1 hour\nLimit: £6000",
+    "description": "Fee: 5.5%\nSettlement: Instant - 1 hour",
```
paullinator, line 1178: remove crypto codes

src/util/utils.js
```diff
@@ -623,6 +623,30 @@ export function makeCurrencyCodeTable(currencyConfigMap: CurrencyConfigMap): (cu
   return currencyCode => map.get(currencyCode.toLowerCase()) ?? []
 }
 
+export const shuffleArray = <T>(array: T[]): T[] => {
```
thehobbit85, line 627: When using a "known" algorithm like Fisher-Yates over here, you might want to link to an explanation either the Wiki - https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle or at least to the stackoverflow that has this answer - https://stackoverflow.com/a/2450976

Not a "showstopper" but just thought it would be nice to have

PR #3526, opened 2022-07-18
Action queue (scheduler)

samholmes:
#### PR Requirements

- [ ] Tested on iOS device

### DIFF
--- a/src/controllers/action-queue/cleaners.js
+++ b/src/controllers/action-queue/cleaners.js
@@ -1,0 +1,177 @@
+// @flow
+
+//
+// Action Operations
+//
+import { type Cleaner, asArray, asCodec, asEither, asMaybe, asNumber, asObject, asOptional, asString, asValue } from 'cleaners'
+import { base64 } from 'rfc4648'
+
+import { type ActionEffect, type ActionOp, type ActionProgram, type ActionProgramState } from './types'
+
+const asBase64 = asCodec(
+  raw => base64.parse(asString(raw)),
+  clean => base64.stringify(clean)
+)
+
+// A serializeable error object
+const asJsonError = asObject({
+  name: asString,
+  message: asString,
+  stack: asString
+}).withRest
+
+const asError = asCodec(
+  (raw: any) => {
+    // Handle Error
+    if (raw instanceof Error) return raw
+
+    // Handle JsonError
+    const jsonError = asMaybe(asJsonError)(raw)
+    if (jsonError != null) {
+      const error = new Error(jsonError.message)
+      Object.defineProperty(error, 'name', {
+        configurable: true,
+        value: jsonError.name
+      })
+      return error
+    }
+
+    // Invalid type
+    throw new TypeError('Expected Error')
+  },
+  // Serialize as a JsonError
+  asJsonError
+)
+
+export const asActionOp: Cleaner<ActionOp> = asEither(
+  asObject({
+    type: asValue('seq'),
+    actions: asArray((raw: any) => asActionOp(raw))
+  }),
+  asObject({
+    type: asValue('par'),
+    actions: asArray((raw: any) => asActionOp(raw))
+  }),
+  asObject({
+    type: asValue('broadcast-tx'),
+    pluginId: asString,
+    rawTx: asBase64
+  }),
+  asObject({
+    type: asValue('exchange-buy'),
+    nativeAmount: asString,
+    walletId: asString,
+    tokenId: asOptional(asString),
+    exchangePluginId: asString
+  }),
+  asObject({
+    type: asValue('exchange-sell'),
+    nativeAmount: asString,
+    walletId: asString,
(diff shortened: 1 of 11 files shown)


samholmes:
**We may want to use underscores as word separators for types (`exchange_buy` instead of `exchange:buy`)**

src/controllers/action-queue/cleaners.js
```diff
@@ -0,0 +1,175 @@
...
+    tokenId: asOptional(asString)
+  }),
+  asObject({
+    type: asValue('swap'),
+    fromWalletId: asString,
+    toWalletId: asString,
+    fromTokenId: asOptional(asString),
+    toTokenId: asOptional(asString),
+    nativeAmount: asString,
+    amountFor: asEither(asValue('from'), asValue('to'))
```
swansontec, line 107: You can do `asValue('from', 'to')` to make an enum, without the `asEither`

src/controllers/action-queue/index.js
```diff
@@ -0,0 +1,3 @@
+// @flow
+
+export { makeActionProgram } from './ActionProgram'
```
swansontec, line 3: We have been on a long campaign to get rid of index files that just re-export things. I feel like this is a hold-over pattern from CommonJS, but in an ES6 named-export world they just make things harder to maintain.

src/controllers/action-queue/redux/actions.js
```diff
@@ -0,0 +1,66 @@
...
+  type: 'ACTION_QUEUE/LOAD_QUEUE',
+  data: ActionQueueMap
+}
+
+export const loadActionQueueState = (data: ActionQueueMap): LoadActionQueueStateAction => {
+  return {
+    type: 'ACTION_QUEUE/LOAD_QUEUE',
+    data
+  }
+}
```
swansontec, line 17: We are also trying to get rid of action creator methods like this. Just `dispatch({ ...whatever })` directly whenever possible - it shows that there are no secret side effects, which makes the program easier to reason about.

src/controllers/action-queue/redux/actions.js
```diff
@@ -0,0 +1,66 @@
...
+  const state = getState()
+  const store: ActionQueueStore = makeActionQueueStore(state.core.account)
+
+  const actionProgramStateItemId = `${actionProgramState.programId}:ActionProgramState`
+  await store.set(actionProgramStateItemId, actionProgramState)
+
+  dispatch({ type: 'ACTION_QUEUE/UPDATE_PROGRAM_STATE', state: actionProgramState })
+}
+
+export type ActionQueueAction = LoadActionQueueStateAction | ScheduleProgramAction | UpdateProgramStateAction
```
swansontec, line 66: Please move this to the top of the program with the type definitions.

src/types/reduxActions.js
```diff
@@ -182,3 +183,4 @@ export type Action =
   | { type: 'FIO/SET_LAST_EXPIRED_CHECKS', data: { [fioName: string]: Date } }
   | { type: 'FIO/CHECKING_EXPIRED', data: boolean }
   | { type: 'FIO/WALLETS_CHECKED_FOR_EXPIRED', data: { [walletId: string]: boolean } }
+  | ActionQueueAction
```
swansontec, line 186: We have been trying to gather all action definitions into this one file, rather than spreading them around throughout the program. However, I think your work is clean & focused enough that we can leave the actions where you have them. So, no need to change anything here.

swansontec (changes requested):
This looking really good! I have a few annoying nit-picky comments, but all the big pieces fit nicely into place now.

src/controllers/action-queue/index.js
samholmes: I used it as a divider between what is internal API and what's external API to the "controller"

src/controllers/action-queue/redux/actions.js
samholmes: Hmm, I know what you mean, but you may want to explain this one to me so I "get what you mean".

src/types/reduxActions.js
samholmes: I had originally placed the files in the `plugins/` directory because I felt like this codebase was sort of like a stand-alone package, and because of this, I felt like all or most of the codebase needed to be colocated in the directory. But after moving the code to a new `controllers`, I don't feel like the package should be as "stand-alone-ish" as I originally felt like it needed to be. I'll leave it for now, but I'm also weary of my decision to spread out the redux API like I have done.

samholmes:
```
// Self-Contained Package Actions:
```

src/controllers/action-queue/redux/actions.js
samholmes: Got it after the explanation. 👍

swansontec (approved):

PR #3532, opened 2022-07-21
Action queue runtime

samholmes:
#### Dependencies
https://github.com/EdgeApp/edge-react-gui/pull/3526

#### PR Requirements

- [ ] Tested on iOS device

### DIFF
--- a/src/controllers/action-queue/runtime.js
+++ b/src/controllers/action-queue/runtime.js
@@ -1,10 +1,13 @@
 // @flow
 
+import { add } from 'biggystring'
 import { type EdgeAccount } from 'edge-core-js'
 import * as React from 'react'
 
 import { AirshipToast } from '../../components/common/AirshipToast'
 import { Airship } from '../../components/services/AirshipInstance'
+import { queryBorrowPlugins } from '../../plugins/helpers/borrowPluginHelpers'
+import { getCurrencyCode } from '../../util/CurrencyInfoHelpers'
 import { snooze } from '../../util/utils'
 import { type ActionEffect, type ActionProgram, type ActionProgramState, type ExecutionResult, type ExecutionResults } from './types'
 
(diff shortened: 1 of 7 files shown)


src/plugins/action-queue/cleaners.js
```diff
@@ -0,0 +1,150 @@
...
+import { type Cleaner, asArray, asEither, asNumber, asObject, asOptional, asString, asValue } from 'cleaners'
+
+import { type ActionEffect, type ActionOp, type ActionProgram, type ActionProgramState } from './types'
+
+const asError = (raw: any) => {
+  if (raw instanceof Error) return raw
+  throw new TypeError('Expected an Error')
+}
+
+export const asActionOp: Cleaner<ActionOp> = asEither(
```
swansontec, line 15: We don't need this nesting with the new Cleaners update.

src/plugins/action-queue/cleaners.js
```diff
@@ -0,0 +1,150 @@
...
+      walletId: asString,
+      tokenId: asOptional(asString)
+    }),
+    asObject({
+      type: asValue('swap'),
+      fromWalletId: asString,
+      toWalletId: asString,
+      fromTokenId: asOptional(asString),
+      toTokenId: asOptional(asString),
+      nativeAmount: asString
```
swansontec, line 75: Is this a from amount of a to amount? In the core quote request, we have a direction flag to tell which way it is. We don't need to support both directions here (or maybe we could), but we do need to be clear about which one we are using.

src/plugins/action-queue/cleaners.js
```diff
@@ -0,0 +1,150 @@
+// @flow
+
+//
+// Action Operations
+//
+import { type Cleaner, asArray, asEither, asNumber, asObject, asOptional, asString, asValue } from 'cleaners'
+
+import { type ActionEffect, type ActionOp, type ActionProgram, type ActionProgramState } from './types'
+
+const asError = (raw: any) => {
```
swansontec, line 10: Error object will not round-trip through `JSON.parse`. Maybe we need to be explicit and grab `name` and `message` from the error object. So `{ name: 'TypeError', message: 'Expected a string at .nativeAmount' }` becomes what we write / read. If you use `.withRest`, I think it will also preserve the extra properties you find on core errors like `SwapAboveLimitError`.

You may need `asCodec` to make this round-trip to an actual `Error` instance. Updating the `.name` property requires `Object.defineProperty`, as it is not writable by default.

src/plugins/action-queue/types.js
```diff
@@ -0,0 +1,171 @@
...
+      confirmations: number
+    }
+  | {
+      type: 'price',
+      currencyPair: string,
+      aboveRate?: string,
+      belowRate?: string
+    }
+  | {
+      type: 'end',
```
swansontec, line 112: The name "end" implies "end of code", which is not what we mean. We really mean "done" or "finished" or a word more like that.

src/plugins/action-queue/types.js
```diff
@@ -0,0 +1,171 @@
...
+declare function executeActionOp(actionOp: ActionOp): ActionEffect
+
+//
+// Aciton Display API
+//
+
+export type ActionDisplayInfo = {
+  title: string,
+  message: string,
+  complete: boolean,
```
swansontec, line 163: Doesn't the Figma design support multiple states? not-started / started / done / error

src/plugins/action-queue/ActionQueueStore.js
```diff
@@ -0,0 +1,90 @@
...
+import { type EdgeAccount } from 'edge-core-js'
+
+import { asActionProgram, asActionProgramState } from './cleaners'
+import { type ActionProgram, type ActionProgramState, type ActionQueueMap } from './types'
+
+export const ACTION_QUEUE_DATASTORE_ID = 'actionQueue'
+
+export type ActionQueueStore = {
+  set: (itemId: string, data: any) => Promise<void>,
+  get: (itemId: string) => Promise<any>,
```
swansontec, line 13: Maybe use `mixed` instead of `any`, since that will error if we forget to use a cleaner / uncleaner.

src/plugins/borrow-plugins/common/util/composeApprovableActions.js
```diff
@@ -19,10 +20,14 @@ export const composeApprovableActions = (...actions: ApprovableAction[]): Approv
       currencyCode,
       nativeAmount
     },
+    unsignedTxs: actions.reduce((txs, action) => [...txs, ...action.unsignedTxs], []),
     approve: async () => {
+      const txs: EdgeTransaction[] = []
       for (const action of actions) {
-        await action.approve()
+        const txs = await action.approve()
+        txs.push(...txs)
```
swansontec, line 28: You have a naming conflict in here, so this code doesn't do what you think.

src/plugins/action-queue/util/delay.js
```diff
@@ -0,0 +1,3 @@
+// @flow
+
+export const delay = (ms: number): Promise<void> => new Promise(resolve => setTimeout(resolve, ms))
```
swansontec, line 3: Redundant. Already exists as `snooze` in src/util/utils.js

swansontec (changes requested):
I mainly focused on the types here. I see the various TODO items in the implementation, so continue working on those as needed.

src/plugins/action-queue/types.js
samholmes: Needs to be changed to `status: 'pending' | 'complete' | Error` (something I am going to do today).

The design doesn't include a "incomplete" state, and it doesn't even include designs for error states (yet). Do you think we need to design for an "incomplete" vs "pending" state, where the former means we haven't done the action and the latter is "we have done the action but we're waiting on the the effect"? The "incomplete" state is immediately after the previous action's effect is observed, so the time between "effect observed" and "do next action" is milliseconds, so I wouldn't think this state should be tracked from the GUI. If this state is omitted, "pending" would stand in for both states (an action is "pending" if the program is blocked by its effect or a previous effect before it).

src/plugins/borrow-plugins/common/util/composeApprovableActions.js
samholmes: Good catch! That's an embarrassing one haha.

src/plugins/action-queue/types.js
samholmes: I can live with `"done"`.

src/plugins/action-queue/cleaners.js
samholmes: I went with adding an `amountFor: 'from' | 'to'` field.

src/plugins/action-queue/cleaners.js
samholmes: Let's go with

```
{
  type: 'done',
  error?: {
    message: string
  }
}
```

src/plugins/action-queue/types.js
samholmes: `pending | doing | done | Error`

src/plugins/action-queue/cleaners.js
samholmes: Instead of changing the runtime type and cleaning it after catching the error, I implemented a serialization type for the error in the `asError` cleaner so we only need to clean/unclean the type during disk read/write.

src/controllers/action-queue/util/delay.js
```diff
@@ -0,0 +1,3 @@
+// @flow
+
+export const delay = (ms: number): Promise<void> => new Promise(resolve => setTimeout(resolve, ms))
```
swansontec, line 3: Still duplicates `snooze` in `src/util/utils.js`.

src/controllers/action-queue/runtime.js
```diff
@@ -46,6 +46,39 @@ async function checkActionEffect(account: EdgeAccount, effect: ActionEffect): Pr
...
+    case 'tx-confs': {
+      const { txId, walletId, confirmations } = effect
+      const wallet = account.currencyWallets[walletId]
+
+      // Get transaction
+      const txs = await wallet.getTransactions({
+        // TODO: Add a parameter to limit to one transaction in result
+        searchString: txId
+      })
+      if (txs.length === 0) throw new Error(`Cannot find transaction with txid '${txId}' in '${wallet.id}'`)
```
swansontec, line 67: Instead of failing, shouldn't we conclude that the confirmations are 0? We might be racing with engines that haven't seen it yet.

swansontec (approved):
I have a few small comments. I'm going to approve this because it's close enough, and I trust you to do the changes when you do your rebase off of future!.

src/controllers/action-queue/runtime.js
samholmes: Agree

src/controllers/action-queue/util/delay.js
samholmes: I am using `snooze`, I must of forgot to remove the file.

PR #3535, opened 2022-07-22
Tighten filtering for sweep private key

Jon-edge:
Sweep private key filters only by currency code.

Tighten filtering so that the wallet's native currency is taken into account in the wallet picker for sweep private key.

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202647424989390
  - https://app.asana.com/0/0/1202648122272329

### DIFF
--- a/src/components/themed/ControlPanel.js
+++ b/src/components/themed/ControlPanel.js
@@ -38,6 +38,10 @@ import { DividerLine } from './DividerLine'
 
 type Props = { navigation: NavigationProp<'controlPanel'> }
 
+const SWEEPABLE_CURRENCY_CODES = Object.keys(SPECIAL_CURRENCY_INFO)
+  .filter(pluginId => SPECIAL_CURRENCY_INFO[pluginId].isPrivateKeySweepable)
+  .map(pluginId => SPECIAL_CURRENCY_INFO[pluginId].chainCode)
+
 export function ControlPanel(props: Props) {
   const { navigation } = props
   const state: any = navigation.state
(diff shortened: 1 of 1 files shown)


src/components/themed/ControlPanel.js
```diff
@@ -104,8 +101,21 @@ export function ControlPanel(props: Props) {
   }
 
   const handleSweep = () => {
+    const sweepableCurrencyCodes = Object.keys(SPECIAL_CURRENCY_INFO)
```
paullinator, line 104: why did `sweepableCurrencyCodes` get moved? now the function would need to get defined on every call to `handleSweep`. Move back to original location

src/components/themed/ControlPanel.js
```diff
@@ -52,6 +48,7 @@ export function ControlPanel(props: Props) {
 
   // ---- Redux State ----
 
+  const wallets = useSelector(state => state.core.account.currencyWallets)
```
paullinator, line 51: Can't use useSelector on a value that is not directly in the redux store. `state.core.account` is in the store but `state.core.account.currencyWallets` is not. useSelector will properly re-render when `state.core.account` changes but not when `state.core.account.currencyWallets` changes. You'll want to change this to something like

```
const [wallets, setWallets] = useState()
useEffect(() => account.watch('currencyWallets', setWallets), [account])
```

paullinator (changes requested):

paullinator (approved):

PR #3538, opened 2022-07-26
Action queue dryrun

samholmes:
#### PR Dependencies

[Sam/aq store fix #3561](https://github.com/EdgeApp/edge-react-gui/pull/3561)

#### PR Requirements

- [ ] Tested on iOS device

### DIFF
--- a/src/controllers/action-queue/runtime.js
+++ b/src/controllers/action-queue/runtime.js
@@ -120,7 +120,7 @@ async function checkActionEffect(account: EdgeAccount, effect: ActionEffect): Pr
     case 'address-balance': {
       // TODO: Use effect.address when we can check address balances
       const { aboveAmount, belowAmount, tokenId, walletId } = effect
-      const wallet = account.currencyWallets[walletId]
+      const wallet = await account.waitForCurrencyWallet(walletId)
       const currencyCode = getCurrencyCode(wallet, tokenId)
       const walletBalance = wallet.balances[currencyCode] ?? '0'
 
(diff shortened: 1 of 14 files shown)


src/controllers/action-queue/runtime.js
```diff
@@ -6,15 +6,36 @@ import * as React from 'react'
...
+    const precomputed: ExecutionResult[] = []
+    while (true) {
+      const result = await executeActionOp(account, program, state, true)
+
+      // Stop if action isn't present, meaning we couldn't dryrun
+      if (result.action == null) break
+
+      precomputed.push(result)
+    }
+    // Convert each precomputed operation from the dryrun into an arrya of
```
swansontec, line 31: "arrya" sounds like some super-cool sci-fi thing.

swansontec (commented):
It seems like there are some missing pieces here:

1. No calls to upload these to the server. Makes sense, since I'm still trying to build that for you.
2. No dry-run implementation of any actual actions, like swapping or borrowing. I'm not sure how we solve the nonce / utxo problem for stacked but non-broadcasted transactions (or maybe we don't support stacking for now).

However, what we have here so far looks generally on the right path. On thing that surprised me was the `dryrun` flag. In my mind, the execution routine and the dry-run routine were going to be separate. The return types should be different, with the dry-run routine returning an effect + signed raw transaction, but the actual execution just returning an effect. OTOH, the internals of the two routines would be pretty similar, so maybe the flag makes sense after all.

src/controllers/action-queue/runtime.js
samholmes: Haha, thanks for the spell-check catch.

samholmes:
After discussion with @swansontec, we've concluded what changes are needed:

1. Implement the push HTTP API calls in this PR
2. Include a new `pendingTransactions: Uint8Array[]` field in the `EdgeSpendInfo` for each call to `makeSpend` in order to "stack" non-broadcasted transactions. We'll need to hold pending transactions in the GUI's action queue state and implement the API upgrade in the currency plugins.

PR #3549, opened 2022-07-29
Add extraTab app config option

paullinator:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202687715819276

### DIFF
--- a/src/components/themed/MenuTab.js
+++ b/src/components/themed/MenuTab.js
@@ -9,7 +9,7 @@ import Ionicon from 'react-native-vector-icons/Ionicons'
 import { Fontello } from '../../assets/vector/index.js'
 import s from '../../locales/strings.js'
 import { config } from '../../theme/appConfig'
-import { type NavigationProp, type ParamList, Actions } from '../../types/routerTypes.js'
+import { type NavigationProp, type ParamList } from '../../types/routerTypes.js'
 import { type Theme, type ThemeProps, cacheStyles, withTheme } from '../services/ThemeContext.js'
 import { DividerLine } from './DividerLine.js'
 import { EdgeText } from './EdgeText.js'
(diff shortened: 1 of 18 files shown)


src/components/themed/MenuTab.js
```diff
@@ -35,13 +35,13 @@ export class MenuTabComponent extends React.PureComponent<Props> {
...
+        return navigation.navigate('walletListScene')
       case 'pluginListBuy':
-        return Actions.jump('pluginListBuy', { direction: 'buy' })
+        return navigation.navigate('pluginListBuy', { direction: 'buy' })
       case 'pluginListSell':
-        return Actions.jump('pluginListSell', { direction: 'sell' })
+        return navigation.navigate('pluginListSell', { direction: 'sell' })
       case 'exchange':
-        return Actions.jump('exchange')
+        return navigation.navigate('exchange')
```
thehobbit85, line 44: This is a good change but it already exists in another PR so it might cause conflicts.
So I'm sorry to ask but can you remove this commit? It would get fixed anyway.

thehobbit85 (changes requested):
Not a full review, just something I've noticed.

src/theme/variables/edgeDark.js
```diff
@@ -216,6 +216,8 @@ export const edgeDark: Theme = {
   tabBarTopOutlineColors: [palette.edgeNavy, palette.edgeNavy],
   tabBarIcon: palette.white,
   tabBarIconHighlighted: palette.edgeMint,
+  extraTabBarIconFont: '',
+  extraTabBarIconName: '',
```
swansontec, line 220: Couldn't this just go into the config, as opposed to the theme? This does not change dynamically.

swansontec (approved):
Approving, but I have a concern. Since we aren't going to change the icon dynamically, just once per app, it doesn't need to go into the theme. It can just go into the config.

You can either fix this or not - it certainly works either way.

PR #3550, opened 2022-07-29
Add more white label config options

paullinator:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202687715819276

### DIFF
--- a/src/components/scenes/GuiPluginListScene.tsx
+++ b/src/components/scenes/GuiPluginListScene.tsx
@@ -1,5 +1,6 @@
 import AsyncStorage from '@react-native-async-storage/async-storage'
 import { asObject, asString } from 'cleaners'
+import { Disklet } from 'disklet'
 import { EdgeAccount } from 'edge-core-js/types'
 import * as React from 'react'
 import { FlatList, Image, Platform, TouchableOpacity, View } from 'react-native'
@@ -15,13 +16,14 @@ import s from '../../locales/strings'
 import { getSyncedSettings, setSyncedSettings } from '../../modules/Core/Account/settings'
 import { executePlugin } from '../../plugins/gui/fiatPlugin'
 import { config } from '../../theme/appConfig'
-import { asGuiPluginJson, GuiPluginRow } from '../../types/GuiPluginTypes'
+import { asBuySellPlugins, asGuiPluginJson, BuySellPlugins, GuiPluginRow } from '../../types/GuiPluginTypes'
 import { connect } from '../../types/reactRedux'
 import { AccountReferral } from '../../types/ReferralTypes'
 import { NavigationProp, RouteProp } from '../../types/routerTypes'
 import { PluginTweak } from '../../types/TweakTypes'
 import { getPartnerIconUri } from '../../util/CdnUris'
 import { filterGuiPluginJson } from '../../util/GuiPluginTools'
+import { fetchInfo } from '../../util/network'
 import { bestOfPlugins } from '../../util/ReferralHelpers'
 import { SceneWrapper } from '../common/SceneWrapper'
 import { ButtonsModal } from '../modals/ButtonsModal'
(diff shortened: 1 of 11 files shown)


src/components/scenes/GuiPluginListScene.js
```diff
@@ -107,6 +112,18 @@ class GuiPluginList extends React.PureComponent<Props, State> {
...
+      const response = await fetchInfo(`v1/buySellPlugins/${config.appId ?? 'edge'}`)
+      const reply = await response.json()
+      const plugins = asBuySellPlugins(reply)
+      this.setState({ buySellPlugins: plugins })
+    } catch (e) {
+      console.log(e.message)
+      // This is ok. We just use default values
+    }
+  }
+
```
swansontec, line 126: This works, but we don't have it cached, so we either need to add a TODO or we need to add a cache. Otherwise, the user will see the old list each time they visit the scene, only to have it flash to the new list after the network gets back.

src/types/Theme.js
```diff
@@ -229,6 +229,9 @@ export type Theme = {
   tabBarIcon: string,
   tabBarIconHighlighted: string,
 
+  extraTabBarIconFont: string,
+  extraTabBarIconName: string,
+
```
swansontec, line 234: Please put these in the `extraTab` section of the app config. These icon properties do not need to change dynamically.

We have agreed that the theme is for colors, not things like dimensions or icons or such - otherwise the whole UI would be in the theme.

swansontec (changes requested):

src/types/Theme.js
paullinator: Funny you say that, I actually started with everything in the config file. However I felt that icons were more of a theme concept and the theme actually does contain icons and has for quite some time. This allows for different icons in light vs dark mode. Granted this specific icon isn't a PNG so its color can be changed without specifying a different icon, but for consistency, it felt correct to put it into the theme like the other icons. Also, a dark vs light theme might want a different vector icon for thinner/thicker lines that might contrast better with different colors. I'll admit, that's a bit of a stretch but still a valid thought

I can move the `name` part out to the config as that admittedly should never need a difference between light and dark, but it does make make the two values a bit disconnected.

src/components/scenes/GuiPluginListScene.js
paullinator: That's fair. What's the correct `account.dataStore` equivalent for local (non-synced) storage that I can use?

src/components/themed/VectorIcon.js
```diff
@@ -0,0 +1,101 @@
...
+  Octicons,
+  SimpleLineIcons,
+  Zocial
+]
+
+const fontMap = {}
+
+for (const icon of iconFamily) {
+  const fontFamily = icon.getFontFamily()
+  console.log(`vector icons: ${fontFamily}`)
```
paullinator, line 54: Crud. need to remove!

src/components/scenes/GuiPluginListScene.js
swansontec: We have a task to create this. Since it is not synced, the state.core.disklet should be fine. This is app-global, not tied to any account.

src/types/Theme.js
swansontec: Right, the only reason we have PNG's in there is because PNG's contain color information. However, this is a bad thing. If we could get away from colored PNG's, that would be preferable to stuffing images in the theme.

Of course, the argument about line width would apply to anything. We could put literally every dimension in our app into the theme, and then have the file be 10,000 lines long and unmaintainable for "maximum flexibility". This is obviously this is a bad idea. Therefore, we need a *strong* bias in favor of putting as little stuff as possible into our theme. The default position should be to keep stuff out of the theme, and then only add them when proven absolutely necessary.

paullinator:
This was rebased to incorporate typescript conversion and requested change for extraTabBarIcon* being moved to AppConfig. I should have used a fixup commit (sorry) but the TS conversion was easier with your requested changes. Nothing else changed but those two above.

src/components/scenes/GuiPluginListScene.tsx
```diff
@@ -105,6 +117,56 @@ class GuiPluginList extends React.PureComponent<Props, State> {
...
+        }
+      })
+    }
+  }
+
+  async updatePlugins() {
+    const { disklet } = this.props.account
+    let diskPlugins
+    try {
+      const fileText = await disklet.getText(PLUGIN_LIST_FILE)
```
swansontec, line 140: It is not appropriate to store caches in the account synced settings. Please use state.core.disklet, which is a local disklet, not synced.

swansontec (changes requested):
One tiny change required.

swansontec (approved):

PR #3551, opened 2022-08-01
Add activity log

paullinator:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202506473095599

### DIFF
--- a/src/actions/CryptoExchangeActions.js
+++ b/src/actions/CryptoExchangeActions.js
@@ -320,12 +320,32 @@ export const shiftCryptoCurrency = (swapInfo: GuiSwapInfo, onApprove: () => void
   const { account } = state.core
   dispatch({ type: 'START_SHIFT_TRANSACTION' })
 
-  const { quote, request } = swapInfo
-  const { pluginId, toNativeAmount } = quote
+  const { fromDisplayAmount, quote, request, fee, fromFiat, fromTotalFiat, toDisplayAmount, toFiat } = swapInfo
+  const { isEstimate, fromNativeAmount, toNativeAmount, networkFee, pluginId, expirationDate } = quote
   const { fromWallet, toWallet, fromCurrencyCode, toCurrencyCode } = request
   try {
     logEvent('SwapStart')
     const result: EdgeSwapResult = await quote.approve()
+
+    global.logActivity(`Swap Exchange Executed: ${account.username}`)
+    global.logActivity(`
+    fromDisplayAmount: ${fromDisplayAmount}
+    fee: ${fee}
+    fromFiat: ${fromFiat}
+    fromTotalFiat: ${fromTotalFiat}
+    toDisplayAmount: ${toDisplayAmount}
+    toFiat: ${toFiat}
+    quote:
+      pluginId: ${pluginId}
+      isEstimate: ${isEstimate.toString()}
+      fromNativeAmount: ${fromNativeAmount}
+      toNativeAmount: ${toNativeAmount}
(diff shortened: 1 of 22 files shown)


src/actions/LogActions.js
```diff
@@ -35,16 +35,24 @@ export const showSendLogsModal = () => async (dispatch: Dispatch, getState: GetS
 }
 
 const prepareLogs = (text: string) => async (dispatch: Dispatch, getState: GetState) => {
+  const id = Math.floor(Math.random() * 16777215).toString(16)
```
thehobbit85, line 38: Can you use '0xFFFFFF' instead of decimal? It looks less "magic" in hex ;)

wyrekey.js
```diff
@@ -0,0 +1,13 @@
+function makeSecret(length) {
```
thehobbit85, line 1: What's that? Where is it used?

src/actions/CryptoExchangeActions.js
```diff
@@ -320,12 +320,32 @@ export const shiftCryptoCurrency = (swapInfo: GuiSwapInfo, onApprove: () => void
...
-  const { pluginId, toNativeAmount } = quote
+  const { fromDisplayAmount, quote, request, fee, fromFiat, fromTotalFiat, toDisplayAmount, toFiat } = swapInfo
+  const { isEstimate, fromNativeAmount, toNativeAmount, networkFee, pluginId, expirationDate } = quote
   const { fromWallet, toWallet, fromCurrencyCode, toCurrencyCode } = request
   try {
     logEvent('SwapStart')
     const result: EdgeSwapResult = await quote.approve()
+
+    global.logActivity(`Swap Exchange Executed: ${account.username}`)
+    global.logActivity(`
```
thehobbit85, line 331: Why not pass the information as an object to the logs?

src/util/logger.js
```diff
@@ -78,45 +107,43 @@ async function writeLog(content) {
...
-      return writeLog(d + ': ' + logs)
+      return writeLog(type, d + ': ' + logs)
     })
   } catch (e) {
     global.clog(e)
   }
   global.clog(logs)
 }
 
+export async function log(...info: Array<number | string | null | {}>): Promise<void> {
```
thehobbit85, line 144: Why not create a type `Log: { type: LogType, data: any }` instead of wrapper functions?

thehobbit85 (changes requested):

src/actions/CryptoExchangeActions.js
paullinator: Visual formatting would suck. Also I don't want all object info to be printed as there may be sensitive info in the object so I would have to filter out some items anyway.

src/util/logger.js
paullinator: I'm just trying to give the equivalent of what we do now with console.log but for the activity log. Less for people to "get used to" when trying to use `logActivity`. Any old `console.log` can also easily be ported to use `logActivity`.

wyrekey.js
paullinator: whoops. this snuck in the PR. i'll remove

src/actions/LogActions.js
paullinator: okaaaay

paullinator: Great. now VScode thinks it's a color and colorizes the text 😏

thehobbit85 (approved):

PR #3552, opened 2022-08-01
v2.19.0

peachbits:
---
  - https://app.asana.com/0/0/1202693704899140

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,34 @@
 # edge-react-gui
 
+## 2.19.0 (2022-08-01)
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,32 @@
 # edge-react-gui
 
+## 2.19.0 (2022-08-01)
```
thehobbit85, line 3: We also added the `New Fiat Plugin framework` and fixed another crash when tapping on a wallet 'too' fast after login.

thehobbit85 (changes requested):

PR #3568, opened 2022-08-05
Fix Anypay invoices from Dash Direct

paullinator:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202717710044918


Jon-edge (approved):
Not sure if the intention was to cleanse all mentions of BitPay as there's still some references, but I'm fine with the changes regardless. Approved!

PR #3574, opened 2022-08-09
Fix unnecessary security alert popups

paullinator:
Commit 5d7ae1a1 broke this as it removed the check for hasAlerts and if the current scene is the SecurityAlerts scene.

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202732125262714


Jon-edge (approved):

src/components/services/AccountCallbackManager.js
```diff
@@ -67,7 +67,9 @@ export function AccountCallbackManager(props: Props) {
       }),
 
       watchSecurityAlerts(account, hasAlerts => {
-        navigation.push('securityAlerts')
+        if (hasAlerts && Actions.currentScene !== 'securityAlerts') {
+          navigation.push('securityAlerts')
+        }
```
swansontec, line 72: If we had used `navigation.navigate` instead of `navigation.push`, that would actually include the current-scene check and avoid the need for legacy Actions.

PR #3582, opened 2022-08-12
Add promo cards with restriction by country, os, buildnum, and bank linkage

paullinator:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202126232980944


src/actions/AccountReferralActions.js
```diff
@@ -195,18 +196,27 @@ async function validatePromoCards(cards: MessageTweak[]): Promise<MessageTweak[]
   } catch (e) {}
 
   const out = []
+  let hasLinkedBankBool
```
peachbits, line 204: Not super important but hasLinkedBankBool and notLinkedBank appear to be wyre specific so maybe the variable names should reflect that clarity in the code and for the individual creating the promos

src/actions/AccountReferralActions.js
```diff
@@ -199,8 +200,20 @@ async function validatePromoCards(account: EdgeAccount, cards: MessageTweak[]):
   let hasLinkedBankBool
 
   for (const card of cards) {
-    // Validate Country
+    // Validate app buildnum
+    const buildNum = getBuildNumber()
```
peachbits, line 214: In my stale promo banner PR I had a function to test version rather than build number which would be easier to use. Also, there's also the chance that build numbers may not be the same between iOS and android where version would be more consistent. Here's the function from the other PR:

export const testVersion = (deviceVersion: string, minVersion: string, maxVersion: string): boolean => {
  const [device, min, max] = [deviceVersion, minVersion, maxVersion].map(semver =>
    Number(
      semver
        .split('.')
        .map(id => id.padStart(2, '0'))
        .join()
        .replace(/,/g, '')
    )
  )

  return min <= device && device <= max
}

peachbits (changes requested):

src/actions/AccountReferralActions.js
paullinator: The request was for a generic check on bank account linkage that could be across any provider. However I'd agree that the promos are likely more specific to a provider. I'd prefer not changing the name, but instead specifying a  providerId/pluginId list to look for. I'll make the change.

src/actions/AccountReferralActions.js
paullinator: Build number is more precise. The main purpose is so that we can do testing of a promo card and not hit customers with it, this let's us target an internal build number. also, if we intend to target customers, build numbers are always numerically increasing so it's easy to specify a range. semantic versions, while usable, aren't real numbers so ranges are awkward. I'm happy to also add a buildVersion for flexibility but I didn't think it was needed.

src/actions/AccountReferralActions.js
```diff
@@ -140,25 +141,44 @@ export const ignoreAccountSwap =
   }
 
 export const refreshAccountReferral = () => async (dispatch: Dispatch, getState: GetState) => {
+  console.log(JSON.stringify(config))
```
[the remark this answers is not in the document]
paullinator: Nice catch. will do

src/actions/AccountReferralActions.js
```diff
@@ -173,16 +173,43 @@ export const refreshAccountReferral = () => async (dispatch: Dispatch, getState:
...
   dispatch({ type: 'ACCOUNT_TWEAKS_REFRESHED', data: cache })
   saveReferralCache(getState())
 }
 
+async function validatePromoCards(cards: MessageTweak[]): Promise<MessageTweak[]> {
+  const apiKey = ENV.IP_API_KEY ?? ''
+  const reply = await fetch(`https://pro.ip-api.com/json/?key=${apiKey}`).catch(e => false)
+  let result = { countryCode: '--' }
+  try {
+    if (reply) {
```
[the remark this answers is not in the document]
paullinator: 👍

peachbits (approved):

PR #3586, opened 2022-08-15
Remove Flow hack in balance calculation

swansontec:
The `// $FlowFixMe` should have been a clue that the logic wasn't right here. Flip it around so type type-checker proves correctness for us.

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device


GoodluckH (approved):
LGTM

PR #3587, opened 2022-08-15
Action queue fiat sell

samholmes:
#### PR Dependencies

[Sam/aq store fix #3561](https://github.com/EdgeApp/edge-react-gui/pull/3561)
[Action queue dryrun #3538](https://github.com/EdgeApp/edge-react-gui/pull/3538)
[Action queue push int #3579](https://github.com/EdgeApp/edge-react-gui/pull/3579)

#### PR Requirements

- [ ] Tested on iOS device


src/controllers/action-queue/cleaners.js
```diff
@@ -199,3 +199,17 @@ export const asActionProgramState: Cleaner<ActionProgramState> = asObject({
...
+
+export const asPaymentMethodsResponse = asObject({
+  data: asArray(
+    asObject({
+      id: asString,
+      // ...some other stuff
+      blockchains: asOptional(asObject(asString))
+    })
+  )
+})
```
swansontec, line 215: I feel like this goes in the WyreClient file, since it's an internal type used by that code. This cleaners file is more for common types used in multiple places.

src/controllers/action-queue/WyreClient.js
```diff
@@ -0,0 +1,88 @@
...
+  //
+  // Public
+  //
+
+  return {
+    // Properties:
+    isAccountSetup,
+
+    // Methods:
+    async getCryptoPaymentMethod(currencyCode: string): Promise<string | void> {
```
swansontec, line 58: I understand that `currencyCode` lines up with their API, and is the simplest option for now, but at some point we may need to make a table going from pluginId / tokenId to Wyre-specific currency codes.

Indeed, we already have the beginning of a table in for Wyre in GuiPlugins.js:

```js
 fixCurrencyCodes: {
      ETH: { pluginId: 'ethereum' },
      BTC: { pluginId: 'bitcoin' },
      DAI: { pluginId: 'ethereum', tokenId: '6b175474e89094c44da98b954eedeac495271d0f' },
      USDC: { pluginId: 'ethereum', tokenId: 'a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48' }
    }
```

I worry that somebody trying to swap Fantom-USDC could end up getting an Ethereum address, and therefore send money to the wrong place, if we *just* use the currency code.

swansontec (changes requested):
I have two comments. If you agree with them, please fix them. If you disagree, I will happily approve - it looks good otherwise.

src/controllers/action-queue/WyreClient.js
samholmes: Ah yes. This WyreClient should be replaced with the upcoming FiatPlugin work from Paul. I think we may be able to get away with cutting a corner here unless it may be a real issue in the short-term.

src/controllers/action-queue/cleaners.js
samholmes: I moved this to the bottom of `WyreClient.js` file

src/controllers/action-queue/WyreClient.js
samholmes: I think we can get away without this and fix this later in future work (`fiat` actions need more work).

samholmes: It appears we may need this after all in order to support Polygon network. I will work on some changes to this PR.

samholmes: Fixed to send to the parent currencyCode always.

PR #3593, opened 2022-08-19
Sam/aq support

samholmes:
#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device


swansontec (approved):
I am not sure why we need to do explicit checks in the collateral methods, and then add a flag to skip these checks, but assuming this design makes sense, it seems to be implemented correctly.

PR #3594, opened 2022-08-19
Add check for disabled swap assets from info server

paullinator:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202807600136152


peachbits (changes requested):
I thought it would be better to add the pluginId and tokenId as state porps rather than account and currencywallets so I created the fixup commit. That needs to be reviewed and the one other comment question answered and we're good to go

PR #3608, opened 2022-08-26
Jon/patch/loan-create-token

Jon-edge:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202827562867778

### DIFF
--- a/src/components/themed/WalletListCreateRow.js
+++ b/src/components/themed/WalletListCreateRow.js
@@ -22,6 +22,7 @@ import { EdgeText } from './EdgeText.js'
 export type WalletListCreateRowProps = {|
   currencyCode: string,
   currencyName: string,
+  createWalletId?: string,
   pluginId?: string,
   walletType?: string,
 
@@ -32,6 +33,7 @@ export const WalletListCreateRowComponent = (props: WalletListCreateRowProps) =>
   const {
     currencyCode = '',
     currencyName = '',
+    createWalletId,
     walletType,
     pluginId,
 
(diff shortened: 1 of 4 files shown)


src/components/themed/WalletListCreateRow.js
```diff
@@ -47,7 +49,7 @@ export const WalletListCreateRowComponent = (props: WalletListCreateRowProps) =>
     if (walletType != null) {
       dispatch(createAndSelectWallet({ walletType })).then(handleRes)
     } else if (pluginId != null) {
-      dispatch(createAndSelectToken({ currencyCode, pluginId })).then(handleRes)
+      dispatch(createAndSelectToken({ tokenCode: currencyCode, pluginId, createWalletId })).then(handleRes)
```
swansontec, line 52: Let keep this as `currencyCode`. Consistent naming makes the code easier to work with for everyone, and this is the term we use all through the app to indicate an exchange symbol like "BTC" or "ETH" or "DAI". It doesn't necessarily have to be a token.

Otherwise, if you really want to fix it, you could upgrade this file to use `tokenId`, which is the less ambiguous way to identify things based on contract address.

But yeah, we want to keep the naming consistent on this legacy stuff so it's more obvious where fixes are needed.

swansontec (approved):
I am approving, because the changes look like they will work, but I do have one naming quibble that I wouldn't mind if you fixed.

src/components/themed/WalletListCreateRow.js
Jon-edge: I made this change because the bug came about due to ambiguity of currencyCode vs token(Currency)Code but I can revert

Jon-edge: On second thought, searching for currencyCode would still show this line. Regarding tokenId upgrade, we are already tracking that here: https://app.asana.com/0/1200895675496964/1202827562867788/f

I think there's value to keeping this here in case further changes are made before getting to that task

PR #3609, opened 2022-08-26
Sam/action queue fixes

samholmes:
#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1202859004377836

### DIFF
--- a/src/components/services/ActionQueue.js
+++ b/src/components/services/ActionQueue.js
@@ -61,7 +61,8 @@ export const ActionQueue = () => {
           }
 
           const { nextState } = await executeActionProgram(account, program, state).catch((error: Error) => {
-            console.error('Action Program Exception:', error.message)
+            console.warn(new Error('Action Program Exception: ' + error.message))
+            console.error(error)
             return {
               nextState: {
                 ...state,
--- a/src/components/services/ActionQueue.js
+++ b/src/components/services/ActionQueue.js
@@ -53,7 +53,7 @@ export const ActionQueue = () => {
 
           const { program, state } = queue[programId]
 
-          if (state.mockMode) {
+          if (program.mockMode) {
             const { nextState } = await mockActionProgram(account, program, state)
             dispatch(updateActionProgramState(nextState))
             executing[programId] = false
--- a/src/controllers/action-queue/push.js
+++ b/src/controllers/action-queue/push.js
@@ -2,11 +2,11 @@
 
 import { asMaybe } from 'cleaners'
 import { type EdgeAccount } from 'edge-core-js'
+import { base58 } from 'edge-core-js/lib/util/encoding'
 import { getUniqueId } from 'react-native-device-info'
 
 import ENV from '../../../env'
 import s from '../../locales/strings'
-import { asBase64 } from '../../util/cleaners/asBase64'
 import { asHex } from '../../util/cleaners/asHex'
 import { exhaustiveCheck } from '../../util/exhaustiveCheck'
 import { type ActionEffect, type ActionProgram, type ExecutionOutput } from './types'
@@ -89,11 +89,11 @@ export async function prepareNewPushEvents(
 }
 
 export async function checkPushEvents(account: EdgeAccount, eventIds: string[]): Promise<boolean> {
-  const { rootLoginId: loginId } = account
+  const { rootLoginId } = account
   const requestBody: PushRequestBody = {
     apiKey: AIRBITZ_API_KEY,
     deviceId,
-    loginId: asBase64(loginId)
+    loginId: base58.parse(rootLoginId)
   }
 
   const response = await fetch(`${pushServerUri}/v2/login`, {
@@ -125,7 +125,7 @@ export async function effectCanBeATrigger(account: EdgeAccount, effect: ActionEf
 }
 
 export async function uploadPushEvents(account: EdgeAccount, newPushEvents: NewPushEvent[]): Promise<void> {
-  const { rootLoginId: loginId } = account
+  const { rootLoginId } = account
   const loginUpdatePayload: LoginUpdatePayload = {
     createEvents: newPushEvents,
     removeEvents: []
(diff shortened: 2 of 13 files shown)


src/controllers/action-queue/runtime.js
```diff
@@ -49,7 +49,7 @@ export const executeActionProgram = async (account: EdgeAccount, program: Action
       return { nextState: { ...state, effect: nextEffect } }
     } catch (error) {
       // Silently fail dryrun
-      showError(error)
+      console.error(error)
```
swansontec, line 52: This will redbox in development mode. It might be better to use `console.warn` or `console.log` instead.

src/controllers/action-queue/push.js
```diff
@@ -89,11 +89,11 @@ export async function prepareNewPushEvents(
 }
 
 export async function checkPushEvents(account: EdgeAccount, eventIds: string[]): Promise<boolean> {
-  const { rootLoginId: loginId } = account
+  const { rootLoginId } = account
   const requestBody: PushRequestBody = {
     apiKey: AIRBITZ_API_KEY,
     deviceId,
-    loginId: asBase64(loginId)
+    loginId: asBase58(rootLoginId)
```
swansontec, line 96: Do not use a cleaner here, since `rootLoginId` has a known type. Instead, use `base58.parse` directly.

src/util/cleaners/asBase64.js
```diff
@@ -4,7 +4,7 @@ import { asCodec, asString, uncleaner } from 'cleaners'
 import { base64 } from 'rfc4648'
 
 export const asBase64 = asCodec(
-  raw => base64.parse(asString(raw)),
+  raw => base64.parse(asString(raw), { loose: true }),
```
swansontec, line 7: Do not put `loose` in here! It is going to mask other errors like this in the future.

src/util/cleaners/asBase58.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+
+import { asCodec, asString, uncleaner } from 'cleaners'
+import { base58 } from 'edge-core-js/lib/util/encoding.js'
+
+export const asBase58 = asCodec(
+  (raw): Uint8Array => base58.parse(asString(raw), { loose: true }),
```
swansontec, line 7: There is no `loose` for base58

src/controllers/action-queue/push.js
```diff
@@ -125,15 +125,15 @@ export async function effectCanBeATrigger(account: EdgeAccount, effect: ActionEf
...
+  const { rootLoginId } = account
   const loginUpdatePayload: LoginUpdatePayload = {
     createEvents: newPushEvents,
     removeEvents: []
   }
   const requestBody: PushRequestBody = {
     apiKey: AIRBITZ_API_KEY,
     deviceId,
-    loginId: asBase64(loginId)
+    loginId: asBase58(rootLoginId)
```
swansontec, line 143: Same thing here. Just call `base58.parse`.

src/components/services/ActionQueue.js
```diff
@@ -52,7 +52,8 @@ export const ActionQueue = () => {
 
           const { program, state } = queue[programId]
           const { nextState } = await executeActionProgram(account, program, state).catch((error: Error) => {
-            console.error('Action Program Exception:', error.message)
+            console.warn(new Error('Action Program Exception: ' + error.message))
+            console.error(error)
```
swansontec, line 56: Again with the redbox. We should only use `console.error` when the program crashes, not for generic error logging. Use `warn` to get a yellow box, or log to go to the output stream.

src/controllers/action-queue/types/pushTypes.js
```diff
@@ -54,7 +54,7 @@ export type PushMessage = {
   +data?: { [key: string]: string } // JSON to push to device
 }
 
-export type PushEventState = 'waiting' | 'cancelled' | 'complete'
+export type PushEventState = 'waiting' | 'cancelled' | 'triggered'
```
swansontec, line 57: The "cancelled" and "hidden" states both mean the event will not be returned from the API. So you should either keep both (since they are valid states) or remove both (for consistency). Probably keep both, in case we update the API with options like `includeHidden` or `includeCancelled`.

src/controllers/action-queue/runtime.js
```diff
@@ -54,31 +54,27 @@ export const executeActionProgram = async (account: EdgeAccount, program: Action
...
-
       // Break out of effect check loop if the ActionEffect passes the check
       if (isEffective) break
 
       // Delay next check
       await snooze(delay)
     } catch (err) {
-      checkErrors.push(err)
+      console.warn(`Action effect check failed:\n\t${String(err)}`)
+      console.error(err)
```
swansontec, line 71: console.warn for both?

swansontec (changes requested):
Found some small things.

src/controllers/action-queue/runtime.js
samholmes: As long as warns show stack traces as well, then I agree (I think it does, but for some reason briefly thought it wouldn't).

src/controllers/action-queue/push.js
samholmes: That's right. Here I go again mis-using cleaners. I realized my desire for a kind of function useful for runtime type-checking vs _runtime type casting_ (which I believe is a useful description as to what cleaners do).

src/util/cleaners/asBase64.js
samholmes: Woops, that creeped in from the _original, incorrect fix_. Thanks for catching this; sorry for being sloppy.

src/controllers/action-queue/runtime.js
samholmes: Warn would be good for both. I'd really like to only have one warn here, but I'm not sure how to add some context to the error (`"Action effect check failed:..."`) without mucking with the stack trace.

samholmes: Unfortunately, RN messed this one up for us; console.error is the only log util we have to maintain the original error stack trace. We'll need to figure out a way to manage our log levels and have traceability for these error types. But until then, it's more useful to us to have the stack trace.

I'm hoping that either we can get a patch for this in RN itself, or that the reason the stack traces are not there are by some other mutation to the `console` object by our codebase so we can fix it (hopeful guess).

swansontec (approved):

PR #3610, opened 2022-08-26
Sam/aq mock mode

samholmes:
#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1202859004377829

### DIFF
--- a/src/controllers/action-queue/mock.js
+++ b/src/controllers/action-queue/mock.js
@@ -1,0 +1,343 @@
+// @flow
+
+import { type EdgeAccount, type EdgeCurrencyWallet, type EdgeNetworkFee, type EdgeTransaction } from 'edge-core-js'
+
+import { exhaustiveCheck } from '../../util/exhaustiveCheck'
+import { filterNull } from '../../util/safeFilters'
+import { snooze } from '../../util/utils'
+import {
+  type ActionEffect,
+  type ActionProgram,
+  type ActionProgramState,
+  type BroadcastTx,
+  type ExecutableAction,
+  type ExecutionOutput,
+  type ExecutionResults,
+  type PendingTxMap
+} from './types'
+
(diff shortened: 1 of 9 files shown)


src/controllers/action-queue/mock.js
```diff
@@ -0,0 +1,349 @@
...
+    case 'done': {
+      throw new Error(`No implementation for action type ${actionOp.type}`)
+    }
+    case 'fiat-buy': {
+      throw new Error(`No implementation for action type ${actionOp.type}`)
+    }
+    case 'noop': {
+      throw new Error(`No implementation for action type ${actionOp.type}`)
+    }
+    case 'unixtime': {
```
swansontec, line 304: Didn't you just delete `unixtime` in the previous commit?

src/controllers/action-queue/mock.js
```diff
@@ -0,0 +1,349 @@
...
+      throw new Error(`No implementation for action type ${actionOp.type}`)
+    }
+
+    default: {
+      // $ExpectError
```
swansontec, line 309: Since `exhaustiveCheck` is specifically supposed to create a Flow error if you miss stuff, ignoring it seems weird.

swansontec (approved):
I am approving this, since it is for testing only, but I did notice some things you might fix.

src/controllers/action-queue/mock.js
samholmes: `exhaustiveCheck` is meant to throw an error if I didn't miss things. It is a tricky to remember when revisiting this, but for clarity: It's inverted logic hack for Flow; it expects an `empty` type which means that when `default` is reachable, then we haven't exhausted the cases and we expect Flow to give us a type error.

src/controllers/action-queue/mock.js
samholmes: Yes, I removed those after this commit then rebased. I thought I caught this in the rebase. Thanks for noticing.

PR #3611, opened 2022-08-26
Revert unexpected console.error behavior for RN

samholmes:
#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1202859004377824

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -37,12 +37,20 @@ console.log('***********************')
 global.clog = console.log
 
 // TODO: Remove isMounted from IGNORED_WARNINGS once we upgrade to RN 0.57
-const IGNORED_WARNINGS = ['slowlog', 'Setting a timer for a long period of time', 'Warning: isMounted(...) is deprecated']
+const IGNORED_WARNINGS = [
+  'Module RCTSplashScreen requires main queue setup since it overrides `constantsToExport`',
+  /Warning: \w+ has been renamed, and is not recommended for use. See [^\s]* for details./,
+  'slowlog',
+  'Setting a timer for a long period of time',
+  'Warning: isMounted(...) is deprecated'
+]
 // $FlowExpectedError
 console.ignoredYellowBox = IGNORED_WARNINGS
 
 // Ignore errors and warnings(used for device testing)
-if (ENV.DISABLE_WARNINGS) LogBox.ignoreAllLogs()
+if (ENV.DISABLE_WARNINGS) {
+  LogBox.ignoreLogs(IGNORED_WARNINGS)
+}
 
 global.OS = Platform.OS
 // Disable the font scaling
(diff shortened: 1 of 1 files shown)


src/app.js
```diff
@@ -42,7 +42,13 @@ const IGNORED_WARNINGS = ['slowlog', 'Setting a timer for a long period of time'
 console.ignoredYellowBox = IGNORED_WARNINGS
 
 // Ignore errors and warnings(used for device testing)
-if (ENV.DISABLE_WARNINGS) LogBox.ignoreAllLogs()
+if (ENV.DISABLE_WARNINGS) {
+  const ignoreWarns = [
+    "Module RCTSplashScreen requires main queue setup since it overrides `constantsToExport` but doesn't implement `requiresMainQueueSetup`. In a future release React Native will default to initializing all native modules on a background thread unless explicitly opted-out of.",
```
swansontec, line 47: Can we remove some detail from this first pattern? Maybe just match the first line, like "Module RCTSplashScreen requires main queue setup since it overrides `constantsToExport`", so it's less wordy.

src/app.js
```diff
@@ -64,9 +70,9 @@ if (__DEV__) {
   // $FlowFixMe
   console.info = log
   // $FlowFixMe
-  console.warn = log
+  // console.warn = log
```
swansontec, line 73: The `log` method is/was supposed to write the logs to disk, so dropping this will break that feature. However, I see from the git blame that Paul changed `if (!__DEV__)` condition guarding this block to `if (__DEV__)` in his "Add activity log" PR. This seems like a mistake. So, unless there is some context we are missing here, I think the correct fix is to put `if (!__DEV__)` back as it was, to leave the logs unmodified while debugging the app.

Otherwise, if the `log` feature truly has been replaced by the activity logs, we should just delete this whole block of code, and the `log` function itself as well, since we don't want it either in production or debugging.

swansontec (changes requested):

src/app.js
paullinator: I talked to @samholmes  about this and yes the line should be put back to !__DEV__

It was changed because I needed to debug that condition but it snuck into the PR. What Sam was going to do was understand how the production build was still properly getting logs at all. This change would imply only dev builds would get the uploaded logs

src/app.js
samholmes: Ok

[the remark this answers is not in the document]
samholmes: The flag is useful to be able to see which known warnings are showing up to do housekeeping here later.

src/app.js
samholmes: Though what's confusing is how this "works" in production. Is this block needed at all?

swansontec (approved):
Let's get this in.

PR #3621, opened 2022-09-01
Jon/patch/loan-create-disps

Jon-edge:
Adds 'Collateral Amount' tile to LoanCreateConfirmationScene

Other changes:
-Update network fee tile to show an empty value while loading
-Adjust styling alignment in LoanCreate workflow

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202895674767128

### DIFF
--- a/src/components/scenes/Loans/LoanCreateScene.js
+++ b/src/components/scenes/Loans/LoanCreateScene.js
@@ -187,7 +187,7 @@ export const LoanCreateScene = (props: Props) => {
           <EdgeText style={disabled ? styles.textInitialDisabled : styles.textInitial}>{emptyLabel}</EdgeText>
         ) : (
           <View style={styles.currencyRow}>
-            <CurrencyRow tokenId={tokenId} wallet={wallet} />
+            <CurrencyRow tokenId={tokenId} wallet={wallet} marginRem={[0, 0.5, 0, 0.5]} />
           </View>
         )}
       </TappableCard>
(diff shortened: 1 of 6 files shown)


src/components/scenes/Loans/LoanCreateScene.js
```diff
@@ -187,7 +187,7 @@ export const LoanCreateScene = (props: Props) => {
           <EdgeText style={disabled ? styles.textInitialDisabled : styles.textInitial}>{emptyLabel}</EdgeText>
         ) : (
           <View style={styles.currencyRow}>
-            <CurrencyRow tokenId={tokenId} wallet={wallet} />
+            <CurrencyRow tokenId={tokenId} wallet={wallet} marginRem={[0, 0.5, 0, 0.5]} />
```
samholmes, line 190: Maybe this is an opportunity for me to ellaborate on what I mean when I push for "composablity over extendablity"...

Notice how we extended CurrencyRow. It went from "A component that displays currency information" to "A component that displays currency information with space parameters". It's like we did `class CurrencyRow extends SpaceComponent`. This is extendability/classical-OOP-inheretence. This is fine, but what's the alternative I speak about?

The alternative is to not extend, but rather compose this component with another component that is the `Space` component:

```jsx
<Space veritical={0.5}>
  <CurrencyRow tokenId={tokenId} wallet={wallet} />
</Space>
```

Now the component never extends to also be augmented/extended; it remains unchanged. It's like we kept the word as is and just used another word in our sentence to describe our thing (noun) with some augmentation (adjective). Their may be times when we want to invent a new word where it makes sense, but maybe this is a rarer necessity.

I imagine once a component becomes so common as an idea that we can bake into all of our component, then it makes sense as a base class. With that said, I don't think there needs to be any change here because marginRem is a pretty common idea in throughout our codebase. This is just me sharing design pattern ideas.

`</Rant>`

samholmes (approved):

src/components/scenes/Loans/LoanCreateScene.js
Jon-edge: Yeah I'm not disagreeing with you here, I did find the styling/spacing components intriguing, but not appropriate for a component used elsewhere. Such changes to remove spacing would need to be handled on a per-component basis, for not only the component itself, but all the other components that are even used next to it throughout the app, since unfortunately we do these little spacing tweaks everywhere. Unfortunately, the GUI was slapped together without much oversight from its conception and these are some of the deep rooted design issues we need to think about prioritizing fixing somehow over time.

I think the other argument was that we would potentially end up with a lot of wrapping views when we break it up this way. Not sure what the implications are there, but I did hear complaints from certain persons whenever ANY extra Views were used. Not sure if these concerns were overblown or not.

samholmes: You lead to a good point: I may be overblowing it here. I just felt like sharing some design considerations on the subject of how we design our components. It's a nit really and not much gained by having "layout components" separate from other presentational components. The separation is just a matter of preference or perhaps guideline. In fairness, our guideline could very well be extend view components with layout properties by "using" some functionality (a hook; wink wink).

Jon-edge: I think another way to think about this though is:

"It's a component that displays currency info" -> currencyInfo **extends component**
"All components have spacing attributes X Y Z" -> component(spacing X Y Z)

I'd argue that we already baked in "spacing attributes X Y Z," because they are used universally. There's questionable value to me in now going:
"spacing attribute X" is a component, "spacing attribute Y" is a component...

What I did like though is beginning to use for example "flex space inbetween" as components, as an _intermediary_ step before they ultimately get baked into the components themselves. Maybe we just need to take a hard look at how we use views to group and format layouts between other components.

PR #3627, opened 2022-09-06
Sam/aave info link

samholmes:
#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1202895674767137

### DIFF
--- a/src/components/scenes/FormScene.tsx
+++ b/src/components/scenes/FormScene.tsx
@@ -9,6 +9,7 @@ import { SceneHeader } from '../themed/SceneHeader'
 
 interface OwnProps {
   headerText: string
+  headerTertiary?: React.ReactNode
   children?: React.ReactNode
   onSliderComplete: (reset: () => void) => Promise<void>
   sliderDisabled: boolean
@@ -17,15 +18,13 @@ interface OwnProps {
 type Props = OwnProps
 
 export const FormScene = (props: Props) => {
-  const { headerText, children, onSliderComplete, sliderDisabled } = props
+  const { headerText, headerTertiary, children, onSliderComplete, sliderDisabled } = props
   const theme = useTheme()
   const styles = getStyles(theme)
 
-  const sceneHeader = React.useMemo(() => <SceneHeader style={styles.sceneHeader} title={headerText} underline />, [headerText, styles.sceneHeader])
-
   return (
     <SceneWrapper background="theme">
-      {sceneHeader}
+      <SceneHeader style={styles.sceneHeader} title={headerText} tertiary={headerTertiary} withTopMargin underline />
       <KeyboardAwareScrollView extraScrollHeight={theme.rem(2.75)} enableOnAndroid>
         {children}
         <View style={styles.footer}>
@@ -39,7 +38,6 @@ export const FormScene = (props: Props) => {
 const getStyles = cacheStyles((theme: Theme) => ({
   sceneHeader: {
     flexDirection: 'row',
-    justifyContent: 'flex-start',
     alignItems: 'center'
   },
   footer: {
--- a/src/components/themed/SceneHeader.tsx
+++ b/src/components/themed/SceneHeader.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react'
 import { View, ViewStyle } from 'react-native'
 
+import { Space } from '../layout/Space'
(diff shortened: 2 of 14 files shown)


src/hooks/useUrlHandler.ts
```diff
@@ -0,0 +1,10 @@
+import { Platform } from 'react-native'
+import { CustomTabs } from 'react-native-custom-tabs'
+import SafariView from 'react-native-safari-view'
+
+export const useUrlHandler = (url: string): (() => void) => {
```
Jon-edge, line 5: Why not use/extend/modify WebUtils? If this is the new way we should open URL's, then we ought to apply the change globally to the app and keep these utilities organized in one place.

src/components/scenes/FormScene.tsx
```diff
@@ -9,6 +9,7 @@ import { SceneHeader } from '../themed/SceneHeader'
 
 interface OwnProps {
   headerText: string
+  headerIcon?: React.ReactNode
```
Jon-edge, line 12: Same comment as the other file, headerTooltipIcon. Make this more distinct than just headerIcon to avoid confusion with static left-justified icons.

src/components/scenes/FormScene.tsx
```diff
@@ -17,15 +18,13 @@ interface OwnProps {
...
+  const { headerText, headerIcon, children, onSliderComplete, sliderDisabled } = props
   const theme = useTheme()
   const styles = getStyles(theme)
 
-  const sceneHeader = React.useMemo(() => <SceneHeader style={styles.sceneHeader} title={headerText} underline />, [headerText, styles.sceneHeader])
-
   return (
     <SceneWrapper background="theme">
-      {sceneHeader}
+      <SceneHeader style={styles.sceneHeader} title={headerText} icon={headerIcon} withTopMargin underline />
```
Jon-edge, line 27: Why was withTopMargin added?

src/constants/aaveConstants.ts
```diff
@@ -0,0 +1 @@
+export const AAVE_SUPPORT_ARTICLE_URL = 'https://www.edge.app/aave/%s'
```
Jon-edge, line 1: Let's start using the new wildcard conventions here, AAVE_SUPPORT_ARTICLE_URL_1S

Jon-edge (changes requested):

src/components/scenes/FormScene.tsx
samholmes: Because that's what is "looks like" from the designs.

src/components/themed/SceneHeader.tsx
```diff
@@ -1,27 +1,30 @@
...
 
+import { Space } from '../layout/Space'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 import { DividerLine } from './DividerLine'
 import { EdgeText } from './EdgeText'
 
 interface Props {
   title?: string
   children?: React.ReactNode
+  icon?: React.ReactNode
```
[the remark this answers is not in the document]
samholmes: Wouldn't it make sense to rename the prop when there are other icon props?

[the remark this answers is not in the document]
samholmes: I don't see what left-right justification has to do with the name of props on a component. Wouldn't it make sense to rename the props when there are other icon props because only then does it need more distinction?

src/hooks/useUrlHandler.ts
samholmes: I don't think those utils use the same API to show an in-app browser. Are we not suppose to be using the in-app browser and instead `openBrowserUri`?

src/components/themed/SceneHeader.tsx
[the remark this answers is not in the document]
Jon-edge: condense into one children prop again, make sure no existing users of this component will have its styling affected if we update the children styling to space-inbetween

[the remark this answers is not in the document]
samholmes: Rename to `tertiary`.

src/components/scenes/FormScene.tsx
samholmes: `headerTertiary`

Jon-edge (approved):

PR #3629, opened 2022-09-07
Add SceneWrapper around FillLoader

Jon-edge:
A plain FillLoader was used while waiting for LoanCreateConfirmationScene to load, showing up as white since there was no background.
 
Add SceneWrapper around FillLoader during LoanCreateConfirmationScene init so there's a background.

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202925555994050

### DIFF
--- a/src/components/scenes/Loans/LoanCreateConfirmationScene.js
+++ b/src/components/scenes/Loans/LoanCreateConfirmationScene.js
@@ -16,6 +16,7 @@ import { type NavigationProp, type RouteProp } from '../../../types/routerTypes'
 import { makeAaveBorrowAction, makeAaveDepositAction } from '../../../util/ActionProgramUtils'
 import { translateError } from '../../../util/translateError'
 import { NetworkFeeTile } from '../../cards/LoanDebtsAndCollateralComponents'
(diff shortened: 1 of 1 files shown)


src/components/scenes/Loans/LoanCreateConfirmationScene.js
```diff
@@ -126,7 +127,9 @@ export const LoanCreateConfirmationScene = (props: Props) => {
   if (loanAccountError != null) return <Alert title={s.strings.error_unexpected_title} type="error" message={translateError(loanAccountError)} />
 
   return loanAccount == null ? (
-    <FillLoader />
+    <SceneWrapper background="theme">
+      <FillLoader />
+    </SceneWrapper>
```
samholmes, line 132: **Optional change:**
It looks cleaner and is more consistent to have an early exit if statement for the loading/error render cases before the default rendering case:

```js
if (loanAccountError != null) return <Alert ... />

if (loanAccount == null) 
return (
    <FillLoader />
  </SceneWrapper>
)

return (
  <FormScene...>
```

samholmes (approved):

PR #3634, opened 2022-09-08
Ignore firebase config files

peachbits:
---
  - https://app.asana.com/0/0/1202948311373401


scripts/prepare.sh
```diff
@@ -27,6 +27,16 @@ mkdir -p android/app/build/intermediates/merged_assets/debug/out/saplingtree/mai
...
+# Copy Firebase configs
+if [ ! -f "GoogleService-Info.plist" ]; then
+  cp GoogleService-Info.sample.plist GoogleService-Info.plist
+fi
+cp GoogleService-Info.plist ios/edge/GoogleService-Info.plist
+if [ ! -f "google-services.json" ]; then
+  cp google-services.sample.json google-services.json
+fi
+cp google-services.json android/app/google-services.json
+
```
swansontec, line 39: Since `GoogleService-Info.plist` is for iOS and `google-services.json` is for Android, maybe we should leave them in the original locations, and avoid the double-copy.

So we merely rename the sample files to `ios/edge/GoogleService-Info.sample.plist` and `android/app/google-services.sample.json`, and then copy the samples on top of the real files with the `if [ ! -f ...]` guard. That keeps our root directory clean, and will work with the current Jenkins setup as well.

swansontec (changes requested):
Both commits should be squashed into one, and then the samples should be moved back into the iOS / Android folders, I think. However, I can live with putting the files in the root directory if you have a good reason for that.

README.md
```diff
@@ -111,7 +111,7 @@ The included `deploy.js` is a script to automate building, signing, and deployin
 - Set the env var KEYCHAIN_PASSWORD to the keychain password of the current user
 - Copy the `deploy-config.sample.json` to `deploy-config.json` and edit the parameters accordingly. You'll need a HockeyApp account to get ids and keys
 - Put any Android keystore files into `edge-react-gui/keystores/`
-- If using Firebase, put your account's `google-services.json` and `GoogleService-Info.plist` into `edge-react-gui/`
+- If using Firebase, put your account's `google-services.json` and `GoogleService-Info.plist` into `edge-react-gui/android/app` and `edge-react-gui/ios/edge`
```
swansontec, line 114: These instructions actually stay the same, since deploy.js includes the following lines:

```js
  if (fs.existsSync(`${buildObj.guiDir}/google-services.json`)) {
    call(`cp -a ${buildObj.guiDir}/google-services.json ${buildObj.guiPlatformDir}/app/`)
  }
```

So Jenkins just drops the files in the root directory, and the deploy.js script moves them to the platform-specific folders.

If we want to remove this logic from the deploy script, we *could* do something like this in the prepare script:

```sh
if [ -f GoogleService-Info.plist ]; then
  cp GoogleService-Info.plist ios/edge/GoogleService-Info.plist
else
   cp ios/edge/GoogleService-Info.sample.plist ios/edge/GoogleService-Info.plist
fi
```

That means the app with build with either the sample file or the user-supplied credentials from the root directory. You would have to add the root-directory copies to .gitignore as well.

swansontec (changes requested):

README.md
peachbits: I force-pushed the commit removing the readme change. Let's keep everything else the same to avoid mucking around with too many things

swansontec (approved):

PR #3638, opened 2022-09-09
AAVE: ActionQueue - Implement Cancel Support

samholmes:
#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1202895685301468

### DIFF
--- a/src/components/services/ActionQueueService.js
+++ b/src/components/services/ActionQueueService.js
@@ -8,17 +8,25 @@ import { updateActionProgramState } from '../../controllers/action-queue/redux/a
 import { executeActionProgram } from '../../controllers/action-queue/runtime'
 import { type ActionQueueMap, type ExecutionContext } from '../../controllers/action-queue/types'
 import { useAsyncEffect } from '../../hooks/useAsyncEffect'
-import { useRef } from '../../types/reactHooks'
+import { useEffect, useMemo } from '../../types/reactHooks'
 import { useDispatch, useSelector } from '../../types/reactRedux'
+import { makePeriodicTask } from '../../util/PeriodicTask'
+
+const EXECUTION_INTERVAL = 1000
 
 export const ActionQueueService = () => {
   const dispatch = useDispatch()
   const account: EdgeAccount = useSelector(state => state.core.account)
   const clientId: string = useSelector(state => state.core.context.clientId)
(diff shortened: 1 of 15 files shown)


src/components/services/ActionQueueService.js
```diff
@@ -36,48 +43,72 @@ export const ActionQueueService = () => {
...
 
+        // Update program state
+        await dispatch(updateActionProgramState({ ...nextState, executing: false }))
+      })
       await Promise.all(promises)
     }
-  }, [queue])
 
+    // Setup loop
+    const intervalId = setInterval(execute, EXECUTION_INTERVAL)
```
swansontec, line 106: `setInterval` is dangerous, because it doesn't account for how long the task takes to run. If each run takes 2 seconds, but you schedule the thing to run every second, you end up with infinite overlapping copies. You can use the `makePeriodicTask` from edge-core-js repo to handle this, since it is async-aware and able to work around this problem.

src/components/services/ActionQueueService.js
```diff
@@ -36,48 +43,72 @@ export const ActionQueueService = () => {
...
-          executing[programId] = false
+        const { nextState } = await executeActionProgram(executionContext, program, state).catch((error: Error) => {
+          console.warn(new Error('Action Program Exception: ' + error.message))
+          console.error(error)
+          return {
+            nextState: {
+              ...state,
+              effect: { type: 'done', error },
+              effective: true,
+              nextExecutionTime: NaN
```
swansontec, line 94: NaN is scary. It converts to JSON as `null`, which could mess up your cleaners if you try to read it back (although you happen to use `asOptional`, so you didn't hit it in testing). Maybe -1, as used elsewhere?

swansontec (changes requested):
I see two safety issues here, but everything else makes sense.

src/components/services/ActionQueueService.js
samholmes: Ah, I was going with `-1` but missed this one. Thanks for the catch!

src/components/services/ActionQueueService.js
samholmes: Hmm, I wonder if this will be an issue considering that I'm filtering out tasks which are "executing".

swansontec (approved):
This looks good now.

PR #3639, opened 2022-09-10
Jon/fix/loan-picker-wyre

Jon-edge:
This PR :
-Updates the Wyre implementation in the ActionQueue to handle multiple payment accounts
-Refactors the CurrencyRow component to allow code reuse for implementing PaymentMethodRow
-Implements PaymentMethodRow that displays relevant information and images from the payment plugin and account
-Update bank button conditions on WalletListModal, changes the button to navigate to the Wyre plugin for user signup
-Implements PaymentMethod rendering in the WalletListModal

Screen Shot 2022-09-13 at 5 31 14 PM

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202895674767120
  - https://app.asana.com/0/0/1202895674767124

### DIFF
--- a/src/components/data/row/CurrencyRow.js
+++ b/src/components/data/row/CurrencyRow.js
@@ -1,19 +1,16 @@
 // @flow
 import { type EdgeCurrencyWallet, type EdgeToken } from 'edge-core-js'
 import * as React from 'react'
-import { View } from 'react-native'
 
 import { useWalletBalance } from '../../../hooks/useWalletBalance.js'
 import { useWalletName } from '../../../hooks/useWalletName.js'
 import { memo } from '../../../types/reactHooks.js'
 import { useSelector } from '../../../types/reactRedux.js'
-import { fixSides, mapSides, sidesToMargin } from '../../../util/sides'
 import { CryptoIcon } from '../../icons/CryptoIcon'
-import { type Theme, cacheStyles, useTheme } from '../../services/ThemeContext.js'
 import { CryptoText } from '../../text/CryptoText'
 import { FiatText } from '../../text/FiatText.js'
 import { TickerText } from '../../text/TickerText.js'
-import { EdgeText } from '../../themed/EdgeText.js'
+import { IconDataRow } from './IconDataRow.js'
 
 type Props = {|
   marginRem?: number[] | number,
@@ -28,9 +25,6 @@ type Props = {|
 // -----------------------------------------------------------------------------
 const CurrencyRowComponent = (props: Props) => {
   const { marginRem, showRate = false, token, tokenId, wallet } = props
-  const theme = useTheme()
-  const styles = getStyles(theme)
-  const margin = sidesToMargin(mapSides(fixSides(marginRem, 1), theme.rem))
 
   // Currency code and wallet name for display:
   const allTokens = wallet.currencyConfig.allTokens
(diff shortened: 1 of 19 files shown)


src/components/modals/WalletListModal.js
```diff
@@ -76,11 +80,23 @@ export function WalletListModal(props: Props) {
...
+  useAsyncEffect(async () => {
+    const wyreClient = await makeWyreClient({ account })
+    if (wyreClient.isAccountSetup) {
+      setLinkedBankAccounts(await wyreClient.getCryptoPaymentMethods())
+    }
+  })
+
+  // #endregion Constants
+
+  // #region Init - Upgrade deprecated props
```
samholmes, line 105: I'm getting use to this region thing. Though, notice how an "end" is usually followed by the beginning of another region? I'd like to see an instances where this isn't the case because it seems like "endregion" is redundant (infer the end of one region by the start of another region. Of course, unless we nest regions (but that seems a bit overboard; how many region levels can we have?).

src/components/modals/WalletListModal.js
```diff
@@ -76,11 +80,23 @@ export function WalletListModal(props: Props) {
...
   const account = useSelector(state => state.core.account)
   const [searching, setSearching] = useState(false)
   const [searchText, setSearchText] = useState('')
+  const [linkedBankAccounts, setLinkedBankAccounts] = useState()
+  useAsyncEffect(async () => {
+    const wyreClient = await makeWyreClient({ account })
+    if (wyreClient.isAccountSetup) {
+      setLinkedBankAccounts(await wyreClient.getCryptoPaymentMethods())
+    }
+  })
```
samholmes, line 102: Missing deps array. `useEffect` has a lint-rule which warns you when you do thi, but `useAsyncEffect` does not. I wonder if there is a way to add our rule for our custom `useAsyncEffect`.

src/components/modals/WalletListModal.js
```diff
@@ -100,29 +120,42 @@ export function WalletListModal(props: Props) {
...
   }, [allowKeysOnlyMode, excludeAssets, legacyExcludeAssets])
 
+  // #region Bank Button/Row Display
+  const bankSignupButton = <MainButton label={s.strings.deposit_to_bank} type="secondary" onPress={handleShowBankPlugin} marginRem={[0.5, 0.75, 1.5, 0.75]} />
+  const bankSection = showWithdrawToBank ? (
+    <>
+      {linkedBankAccounts == null || Object.keys(linkedBankAccounts).length === 0 ? bankSignupButton : null /* TODO: Show payment methods, if available */}
+      <EdgeText>{s.strings.deposit_to_edge}</EdgeText>
+    </>
+  ) : null
```
samholmes, line 174: Eh, I don't see much benefit to not just inlining this component. It now looks like this is _state_ rather than _view_ code. I'd be okay with a "render" function (though keep in mind, I wasn't really into render functions either, but followed the convention).

src/components/data/row/IconDataRow.js
```diff
@@ -0,0 +1,96 @@
...
+  leftSubtext: string,
+  rightText?: string | React.Node,
+  rightSubText?: string | React.Node,
+  marginRem?: number[] | number
+|}
+
+// -----------------------------------------------------------------------------
+// A view representing fields of data accompanied by a left-justified icon
+// -----------------------------------------------------------------------------
+const IconDataRowComponent = (props: Props) => {
```
samholmes, line 25: This component doesn't appear to be used anywhere in this commit.

src/components/data/row/IconDataRow.js
```diff
@@ -30,29 +30,29 @@ const IconDataRowComponent = (props: Props) => {
...
     <View style={[styles.container, margin]}>
       {icon}
-      <View style={styles.nameColumn}>
-        <View style={styles.currencyRow}>
-          <EdgeText style={styles.currencyText}>{leftText}</EdgeText>
-          {leftTextExtended != null ? <EdgeText style={styles.exchangeRateText}>{leftTextExtended}</EdgeText> : null}
+      <View style={styles.leftColumn}>
+        <View style={styles.row}>
+          <EdgeText style={styles.leftText}>{leftText}</EdgeText>
+          {leftTextExtended != null ? <EdgeText style={styles.leftTextExtended}>{leftTextExtended}</EdgeText> : null}
```
samholmes, line 37: Shouldn't this commit be a squash?

src/components/data/row/CurrencyRow.js
```diff
@@ -41,81 +35,22 @@ const CurrencyRowComponent = (props: Props) => {
...
-    </View>
+    <IconDataRow
+      icon={icon}
+      leftText={currencyCode}
+      leftTextExtended={tickerText}
+      leftSubtext={name}
+      rightText={cryptoText}
+      rightSubText={fiatText}
+      marginRem={marginRem}
+    />
```
samholmes, line 52: Oh I see. Why do a "partially factor out" and then a "full factor out" commit separate? "Factor out" should include the whole thing in one commit, this way it can be reviewed to be correct on its own.

src/plugins/gui/scenes/EnterAmountScene.js
```diff
@@ -157,7 +157,7 @@ export const FiatPluginEnterAmountScene = memo((props: Props): React.Node => {
 
                   <View style={styles.poweredByContainerColumn}>
                     <View style={styles.poweredByContainerRow}>
-                      <Text style={styles.poweredByText}>{s.strings.plugin_powered_by}</Text>
+                      <Text style={styles.poweredByText}>{s.strings.plugin_powered_by_space}</Text>
```
samholmes, line 160: It appears there is already a space char below? Also why isn't `%s` used here and only one `Text` component?

samholmes (changes requested):
Here's some of the feedback and changes needed. I'll finish review once the commits are reworked to not be fragmented. It's hard to review a new component that used to be inline when it's separate commits.

src/controllers/action-queue/cleaners.js
```diff
@@ -80,6 +80,7 @@ const asFiatBuyActionOp: Cleaner<FiatBuyActionOp> = asObject({
 })
 const asFiatSellActionOp: Cleaner<FiatSellActionOp> = asObject({
   type: asValue('fiat-sell'),
+  fiatAccountId: asString,
```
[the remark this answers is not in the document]
Jon-edge: "FTX Fiat Plugin" may have a separate designation for subaccounts A, B, C (FTX really does have subaccounts).

Though not Wyre-Specific, fair enough, it's possible that a provider might not provide separate account Ids.

src/components/data/row/CurrencyRow.js
Jon-edge: I wanted to separate the renaming. Figured it would be easier to determine sameness if I didn't factor it out and rename everything generically at the same time.

src/plugins/gui/scenes/EnterAmountScene.js
Jon-edge: Good catch

src/controllers/action-queue/cleaners.js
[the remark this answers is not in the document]
Jon-edge: General comment: Rename fiat->wyre in a new commit

src/components/modals/WalletListModal.js
Jon-edge: We can nest regions, and #endregions are required for the folding to work at all. A #region followed by another #region is considered the start of a nested region.

Not sure how far we can nest (don't see why there would be a limit), but at that point it's probably prudent to start just making new constants/functions.

src/components/modals/WalletListModal.js
Jon-edge: I think it really helps with self documentation to keep bigger blocks separate. Keeps related code that helps build these component groups together in one place, too. 

Besides convention and connotations of "render*" , is there any functional benefit to turning this into a function vs just having a constant?

samholmes (approved):

PR #3642, opened 2022-09-12
Add firebase message listeners and price change notification modal

peachbits:
---
  - https://app.asana.com/0/0/1202555663767418

### DIFF
--- a/src/actions/DeepLinkingActions.ts
+++ b/src/actions/DeepLinkingActions.ts
@@ -1,6 +1,7 @@
 import { EdgeCurrencyWallet } from 'edge-core-js'
 import { sprintf } from 'sprintf-js'
 
+import { launchPriceChangeBuySellSwapModal } from '../components/modals/PriceChangeBuySellSwapModal'
 import { showError, showToast } from '../components/services/AirshipInstance'
 import { guiPlugins } from '../constants/plugins/GuiPlugins'
 import s from '../locales/strings'
@@ -126,6 +127,11 @@ function handleLink(dispatch: Dispatch, state: RootState, link: DeepLink): boole
 
     case 'bitPay': {
       launchBitPay(link.uri, { currencyWallets }).catch(showError)
+      return true
+    }
+
+    case 'price-change': {
+      dispatch(launchPriceChangeBuySellSwapModal(link))
       return true
     }
 
--- a/src/components/modals/PriceChangeBuySellSwapModal.tsx
+++ b/src/components/modals/PriceChangeBuySellSwapModal.tsx
@@ -1,0 +1,36 @@
+import * as React from 'react'
+import { sprintf } from 'sprintf-js'
+
+import { PriceChangePayload } from '../../controllers/action-queue/types/pushPayloadTypes'
+import s from '../../locales/strings'
+import { Dispatch, GetState } from '../../types/reduxTypes'
+import { Actions } from '../../types/routerTypes'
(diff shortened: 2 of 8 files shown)


src/controllers/action-queue/types/pushPayloadTypes.js
```diff
@@ -0,0 +1,33 @@
...
+      pluginId: asString
+    })
+  })(message)
+  return { ...clean.notification, ...clean.data }
+}
+
+export const asPushMessagePayloadToEdgeUri = (data: FirebaseMessagingTypes.RemoteMessage): string => {
+  const { type, ...params } = asPriceChangePayload(data) // extend later with asEither
+  return `edge://${type}?${URL.qs.stringify(params)}`
+}
```
swansontec, line 33: These are not cleaners, since they take typed input (`FirebaseMessagingTypes.RemoteMessage`). So, their names should not start with `as`.

You should still make an `asPriceChange` cleaner with the `type: asValue('price-change'), pluginId: asString` bit, but then simply use it in the `pushMessagePayloadToEdgeUri` utility function. We don't need to clean the notification body stuff, since that's already type-safe.

swansontec (changes requested):
We have some utility functions pretending to be cleaners, but otherwise this looks good.

src/controllers/action-queue/types/pushPayloadTypes.js
```diff
@@ -14,20 +14,13 @@ export type PriceChangePayload = {
...
+  pluginId: asString
+})
 
-export const asPushMessagePayloadToEdgeUri = (data: FirebaseMessagingTypes.RemoteMessage): string => {
-  const { type, ...params } = asPriceChangePayload(data) // extend later with asEither
-  return `edge://${type}?${URL.qs.stringify(params)}`
+export const pushMessagePayloadToEdgeUri = (message: FirebaseMessagingTypes.RemoteMessage): string => {
+  const { type, pluginId } = asPriceChangePayloadData(message.data)
+  const body = asString(message.notification.body)
+  return `edge://${type}?${URL.qs.stringify({ body, pluginId })}`
```
swansontec, line 25: Ahh, much more readable! However, I don't think we need the `body` in the query string, since it's a localized string, not something our code can use.

src/actions/NotificationActions.js
```diff
@@ -243,8 +249,8 @@ export const newPriceChangeEvent = (
         )}`
       ],
       pluginId: currencyInfo.pluginId,
-      dailyChange: dailyChangeEnabled ? 10 : undefined,
-      hourlyChange: hourlyChangeEnabled ? 3 : undefined
+      dailyChange: dailyChangeEnabled ? 0.0001 : undefined,
+      hourlyChange: hourlyChangeEnabled ? 0.0003 : undefined
```
swansontec, line 253: What is up with these numbers? For testing? They seem too small.

swansontec (changes requested):

src/controllers/action-queue/types/pushPayloadTypes.js
peachbits: The body is displayed in the modal itself as the first sentence

src/actions/NotificationActions.js
peachbits: Yup for testing and removed in the force pushed branch

samholmes (approved):

PR #3643, opened 2022-09-13
Add loginIds to /device calls

peachbits:
---
  - https://app.asana.com/0/0/1202966393126770


swansontec (changes requested):
I have just published edge-core-js v0.19.29, which has the stuff you need. I will stick it on develop.

src/actions/NotificationActions.js
```diff
@@ -44,7 +46,10 @@ export const registerNotificationsV2 = () => async (dispatch: Dispatch, getState
   try {
     const body = {
       apiKey: ENV.AIRBITZ_API_KEY,
-      deviceId: state.core.context.clientId
+      deviceId: state.core.context.clientId,
```
swansontec, line 49: Same problem here. We need to provide both `loginIds` and `deviceToken` in the top-level body. Without the `deviceToken`, the query will actually de-register notifications for this device.

swansontec (changes requested):

PR #3648, opened 2022-09-15
Jon/aave/wyre-loan-create

Jon-edge:
This PR integrates the Wyre withdraw to bank option for the Loan Create workflow. Supports a withdrawal from BTC to Fiat

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202974109503110

### DIFF
--- a/src/__tests__/modals/__snapshots__/WalletListModal.test.tsx.snap
+++ b/src/__tests__/modals/__snapshots__/WalletListModal.test.tsx.snap
@@ -111,14 +111,6 @@ Array [
       </Text>
     </View>
     <View
-      style={
-        Object {
-          "padding": 0,
-          "width": "100%",
-        }
-      }
-    />
-    <View
       accessible={true}
       focusable={true}
       onClick={[Function]}
--- a/src/components/data/row/PaymentMethodRow.tsx
+++ b/src/components/data/row/PaymentMethodRow.tsx
@@ -1,9 +1,11 @@
 import * as React from 'react'
+import { TouchableOpacity } from 'react-native'
 import FastImage from 'react-native-fast-image'
 import { cacheStyles } from 'react-native-patina'
 
 import { guiPlugins } from '../../../constants/plugins/GuiPlugins'
 import { PaymentMethod } from '../../../controllers/action-queue/WyreClient'
+import { useHandler } from '../../../hooks/useHandler'
 import s from '../../../locales/strings'
 import { asGuiPluginJson } from '../../../types/GuiPluginTypes'
 import { memo, useState } from '../../../types/reactHooks'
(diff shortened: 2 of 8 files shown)


src/controllers/action-queue/WyreClient.js
```diff
@@ -41,6 +42,8 @@ export const makeWyreClient = async (opt: WyreClientOptions): Promise<WyreClient
     Accept: 'application/json'
   }
 
+  logyv(wyreSecret, 'wyreSecret')
```
samholmes, line 45: This looks like it crept in.

src/controllers/action-queue/WyreClient.js
```diff
@@ -83,10 +87,10 @@ export const makeWyreClient = async (opt: WyreClientOptions): Promise<WyreClient
         if (blockchains == null) return
 
         // Special cases for testnets, if they don't yet support (they currently do not)
-        if (blockchains.MUMBAI == null) blockchains.MUMBAI = blockchains.MATIC
+        if (blockchains.MUMBAI == null && blockchains.MATIC != null) blockchains.MUMBAI = blockchains.MATIC
```
samholmes, line 90: This doesn't really have any impact?

src/controllers/action-queue/WyreClient.js
```diff
@@ -83,10 +87,10 @@ export const makeWyreClient = async (opt: WyreClientOptions): Promise<WyreClient
...
         if (blockchains.TESTBTC == null) {
           const address = blockchains.BTC
-          if (['m', 'n', '2'].includes(address[0])) blockchains.TESTBTC = address
+          if (address != null && ['m', 'n', '2'].includes(address[0])) blockchains.TESTBTC = address
```
samholmes, line 93: Ah, this makes sense however.

samholmes (changes requested):
Looks like some cleanup needed to be done at the end. Also rebase onto develop now that the dependency PR is merged.

src/controllers/action-queue/WyreClient.js
Jon-edge: Good catch, this was misguided. The real issue was the test BTC. Still getting used to being allowed to reference nonexistent keys 😂

src/controllers/action-queue/WyreClient.js
Jon-edge: Good catch, missed this when debugging for my fixes.

src/controllers/action-queue/WyreClient.js
Jon-edge: Actually I think still has merit. Don't want to create a MUMBAI key at all if there's no MATIC configured.

Thinking on this a bit more, I don't think there's even a point to setting up MUMBAI at all to a live net address. Plus, MUMBAI really does not behave as reliably as live net

Jon-edge: Going to just leave this for now. We do have a lot of Object.keys usage throughout the gui for map logic.

Jon-edge: I think still has merit. Don't want to even create a MUMBAI key at all if there's no MATIC configured.

Thinking on this a bit more, I don't think there's even a point to setting up MUMBAI at all to a live net address. Plus, MUMBAI really does not behave as reliably as live net

Going to just leave this for now. We do have a lot of Object.keys usage throughout the gui for map logic.

Jon-edge:
Will rebase after these changes are approved

samholmes (approved):
Approved but squash/apply my fixups commits

PR #3640, opened 2022-09-12
Use typescript

paullinator:
#### PR Dependencies

 none

#### PR Requirements

If you have made **any** visual changes to the GUI. Make sure you have:
- [x] Tested on iOS device
- [ ] Tested on Android device
- [x] Tested on small-screen device (iPod Touch)
- [ ] Tested on large-screen device (tablet)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202802426640427

PR #3650, opened 2022-09-16
William/typescript

swansontec:
I have finished rebasing #3640, but it's so different now that I am submitting the work as a new PR.


package.json
```diff
@@ -99,7 +99,7 @@
     "assert": "^2.0.0",
     "async-lock": "^1.1.2",
     "base-x": "^4.0.0",
-    "biggystring": "^3.0.2",
+    "biggystring": "^4.0.0",
```
paullinator, line 99: yarn.lock wasn't updated.

package.json
```diff
@@ -187,6 +188,7 @@
     "@types/react-test-renderer": "^18.0.0",
     "@types/sprintf-js": "^1.1.2",
     "@types/url-parse": "^1.4.8",
+    "@types/uuid": "^7.0.0",
```
paullinator, line 188: no yarn.lock change

src/components/scenes/Loans/LoanStatusScene.js
```diff
@@ -103,7 +103,7 @@ export const LoanStatusScene = (props: Props) => {
           ) : (
             <MainButton label={s.strings.loan_status_cancel_txs} type="secondary" onPress={handleCancelPress} marginRem={[0.5, 1, 2, 1]} />
           )}
-          {isProgramDone ? <ConfettiCannon autostart count={250} origin={{ x: -50, y: -50 }} fallSpeed={4000} /> : null}
+          {isProgramDone ? <ConfettiCannon autoStart count={250} origin={{ x: -50, y: -50 }} fallSpeed={4000} /> : null}
```
paullinator: to be consistent with previous behavior, remove `autoStart` completely

src/modules/UI/components/Buttons/TertiaryButton.ui.js
```diff
@@ -30,7 +30,7 @@ export class TertiaryButton extends React.Component<Props> {
   render() {
     const { children, style, ...props } = this.props
     return (
-      <TouchableHighlight underlayColor={rawStyles.tertiaryButtonUnderlay.color} style={[styles.button, styles.tertiaryButton, style]} {...props}>
+      <TouchableHighlight underlayColor={tertiaryButtonUnderlay.color} style={[styles.button, styles.tertiaryButton, style]} {...props}>
```
paullinator: does this work? you're passing a style parameter into a non-style prop. The previous code looks more correct

paullinator (changes requested):
See comments

src/modules/UI/components/Buttons/TertiaryButton.ui.js
swansontec: The `underlayColor` is a [normal prop](https://reactnative.dev/docs/0.68/touchablehighlight#underlaycolor), not a style for some reason. So we need to pass it as a normal value, which is really weird but correct.

src/components/scenes/Loans/LoanStatusScene.js
swansontec: Will do.

package.json
swansontec: This version is already in yarn.lock, so no change needed.

package.json
swansontec: This version is already in yarn.lock, so no change needed.

paullinator (approved):

src/modules/UI/components/Buttons/style.js
```diff
@@ -5,7 +5,11 @@ import { StyleSheet } from 'react-native'
 import { THEME } from '../../../../theme/variables/airbitz'
 import { scale } from '../../../../util/scaling.js'
 
-export const rawStyles = {
+export const primaryButtonUnderlay = { color: THEME.COLORS.PRIMARY }
+export const secondaryButtonUnderlay = { color: THEME.COLORS.GRAY_1 }
+export const tertiaryButtonUnderlay = { color: THEME.COLORS.GRAY_3 }
+
+export const styles = StyleSheet.create({
```
swansontec: @paullinator See how we have pulled the underlay exports out of the style block, and made them normal exports? This is why the files above work correctly.

PR #3653, opened 2022-09-16
Add loginIds to /device endpoint requests

peachbits:
---
  - https://app.asana.com/0/0/1202966393126770


swansontec (approved):
I am approving, but I do worry about dropping the try / catch block around getting the `deviceToken`. I think we should allow `deviceToken` to be `undefined`, but then fix the server not to update the saved value in that case. We can do this in the next release, after we have had a chance to fix the server. Let's get this out now.

PR #3664, opened 2022-09-20
Do not include inactive payment methods

paullinator:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203008188455322


src/controllers/action-queue/WyreClient.ts
```diff
@@ -85,6 +85,7 @@ export const makeWyreClient = async (opt: WyreClientOptions): Promise<WyreClient
       paymentMethodsResponse.forEach(paymentMethod => {
         const { blockchains, id } = paymentMethod
         if (blockchains == null) return
+        if (paymentMethod.status !== 'ACTIVE') return
```
samholmes, line 88: It would be better to filter out inactive payment methods at the caller. But for now we will do this.

samholmes (approved):

PR #3674, opened 2022-09-23
Update formatFiatString minPrecision->maxPrecision

Jon-edge:
Callers of formatFiatString always use the default minPrecision of 2.
The existing implementation dynamically increased the precision if the fiat amount was less than one decimal place without an upper limit.

Update the implementation to instead take a maxPrecision, so if the need arises it can support higher precision fiat amount display.

#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1202970502058128


src/hooks/useFiatText.ts
```diff
@@ -70,7 +70,7 @@ export const formatFiatString = (props: { autoPrecision?: boolean; fiatAmount: s
     }
     while (tempFiatAmount <= 0.1 && tempFiatAmount > 0) {
       tempFiatAmount *= 10
-      precision++
+      precision += precision < maxPrecision ? 1 : 0
```
samholmes, line 72: I don't quite know what's going on here. But I'll trust that you've got it to work. I did make a fixup commit to fix the types that appear to be incorrect.

samholmes (approved):

src/hooks/useFiatText.ts
Jon-edge: This dynamically increases precision as the fiat value gets smaller. Not really applicable to any of our current fiat cases, so I just changed it to do the same thing, but flip it around to cap at a maxPrecision for future-proofing

PR #3675, opened 2022-09-23
Jon/fix/aave/swap-status-display

Jon-edge:
- Update SwapActionOp type
- Update swap description to be more descriptive about the WBTC network being swapped into.
- Fix sprintf call

Screen Shot 2022-09-23 at 12 40 27 PM

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203020953556924

### DIFF
--- a/src/controllers/action-queue/cleaners.ts
+++ b/src/controllers/action-queue/cleaners.ts
@@ -119,12 +119,13 @@ const asLoanWithdrawActionOp = asObject<LoanWithdrawActionOp>({
 })
 const asSwapActionOp = asObject<SwapActionOp>({
   type: asValue('swap'),
-  fromWalletId: asString,
-  toWalletId: asString,
+  amountFor: asValue('from', 'to'),
+  walletId: asString,
   fromTokenId: asOptional(asString),
-  toTokenId: asOptional(asString),
+  fromWalletId: asString,
   nativeAmount: asString,
-  amountFor: asValue('from', 'to')
+  toTokenId: asOptional(asString),
+  toWalletId: asString
 })
 export const asActionOp: Cleaner<ActionOp> = asEither(
   asSeqActionOp,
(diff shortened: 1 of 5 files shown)


src/controllers/action-queue/cleaners.ts
```diff
@@ -112,12 +112,13 @@ const asLoanWithdrawActionOp: Cleaner<LoanWithdrawActionOp> = asObject({
 })
 const asSwapActionOp: Cleaner<SwapActionOp> = asObject({
   type: asValue('swap'),
-  fromWalletId: asString,
-  toWalletId: asString,
+  amountFor: asValue('from', 'to'),
+  walletId: asString,
```
samholmes, line 116: Why are we adding a third walletId?

src/controllers/action-queue/display.ts
```diff
@@ -100,7 +104,14 @@ async function getActionOpDisplayInfo(account: EdgeAccount, actionOp: ActionOp,
...
         ...baseDisplayInfo,
         title: sprintf(s.strings.action_queue_display_swap_title, fromCurrencyCode, toCurrencyCode),
-        message: sprintf(s.strings.action_queue_display_swap_message, fromCurrencyCode, toCurrencyCode)
+        message: sprintf(
+          s.strings.action_queue_display_swap_message,
+          fromCurrencyCode,
+          config.appName,
+          toCurrencyCode,
+          s.strings.loan_aave_fragment,
+          borrowEngineWallet.currencyInfo.currencyCode
```
samholmes, line 113: You could just change this to `toWallet.currencyInfo.currencyCode` and omit the op code change.

samholmes (changes requested):

src/controllers/action-queue/cleaners.ts
Jon-edge: This is to support adding the AAVE network name to the swap description

src/controllers/action-queue/display.ts
Jon-edge: I think it's fine to keep this explicit. I can think of a use case come up soon about wanting to swap for depositing, in which toWallet is no longer valid.

samholmes: `swap` action op is AAVE agnostic. Unless we want to rename it to `swap-aave` and implement the generic action op later (kinda like what we did with wyre-sell over fiat-sell).

samholmes (approved):

PR #3676, opened 2022-09-23
Fix missing Done row on iOS devices in Request scene

peachbits:
The flip input used to check for an empty string inputAccessoryViewID but that was changed recently to just accept the prop as passed in (which is correct). Therefore the fix needs to be in the request scene which created the prop.

---
  - https://app.asana.com/0/0/1203028877160239


samholmes (approved):
I trust ya

PR #3677, opened 2022-09-23
Jon/fix/aave/dashboard-no-loans

Jon-edge:
- Disallow new loan creation while loading
- Extend SpaceProps to control justifyContent
- Implement display for no loans
- Add a new display for no loans while loans are not yet fully synced with AAVE

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202780692217985


src/components/scenes/Loans/LoanDashboardScene.tsx
```diff
@@ -146,11 +146,12 @@ export const LoanDashboardScene = (props: Props) => {
...
-        <TouchableOpacity onPress={handleAddLoan} style={styles.addButtonsContainer}>
-          <Ionicon name="md-add" style={styles.addItem} size={theme.rem(1.5)} color={theme.iconTappable} />
-          <EdgeText style={[styles.addItem, styles.addItemText]}>{s.strings.loan_new_loan}</EdgeText>
-        </TouchableOpacity>
+        ) : (
+          <TouchableOpacity onPress={handleAddLoan} style={styles.addButtonsContainer}>
+            <Ionicon name="md-add" style={styles.addItem} size={theme.rem(1.5)} color={theme.iconTappable} />
+            <EdgeText style={[styles.addItem, styles.addItemText]}>{s.strings.loan_new_loan}</EdgeText>
+          </TouchableOpacity>
+        )}
```
samholmes, line 156: I see the race condition we're trying to solve here. User can start the loan creation process before user has queried the network which means the user has the chance of completing the create workflow after the query finishes and a loan account is created.

This solve this problem at the tradeoff that the UI is stuck until all wallets have synced with the network. This blocking UX is not going to be reasonable for accounts with many wallets.

An alternative solution would be to check for loan account at the LoanCreateConfirmationScene to determine whether a `createLoanAccount` or `updateLoanAccount` action is to be dispatched. This covers our edge case with a the fallback behavior of "create workflow does update" which we know will work. This will avoid the expensive tradeoff.

src/hooks/useSpaceStyle.ts
```diff
@@ -35,19 +41,22 @@ export const useSpaceStyle = (props: SpaceProps): ViewStyle => {
   const paddingRight = theme.rem(typeof right === 'number' ? right : right ? 1 : 0)
 
   // Direction:
-  const { sideways = false } = props
+  const { isSideways: sideways = false } = props
   const flexDirection = sideways ? 'row' : 'column'
 
   // Alignment:
-  const { start = false, center = false, end = false } = props
```
samholmes, line 42: I realize one of my motivations for naming things was to keep things really short. Now that I see what the names were `start` `center` and `end`, I recall how it was meant to be used with multiple space components leveraging the `sideways` flag to control the direction. If one wanted to do `justifyContent` they'd need an extra component, but I see the value in not needing the extra component to change the alignment for this direction.

I wonder if there is a set of words we could use that are short that also encode the direction?

src/locales/en_US.ts
```diff
@@ -959,10 +959,12 @@ const strings = {
   loan_insufficient_funds_warning:
     'You must have %1$s and sufficient funds in associated %2$s wallet in order to proceed with the loan. If you do not have %3$s, enable this token in your %2$s wallet and perform an exchange in %4$s.',
   loan_interest_rate: 'Interest Rate',
+  loan_loading_loans: 'Loading Loans...',
   loan_loan_breakdown_title: 'Loan Breakdown',
   loan_loan_to_value_ratio: 'Loan to Value Ratio',
   loan_make_payment: 'Make Payment',
   loan_must_be_s_or_less: 'Loan must be less than %1$s of your total collateral value',
+  loan_no_active_loans: 'You have no active loans.\nCreate a loan and start earning with AAVE!',
```
samholmes, line 967: Interesting how the spec says "earning" with AAVE.

src/components/scenes/Loans/LoanDashboardScene.tsx
```diff
@@ -50,10 +50,12 @@ export const LoanDashboardScene = (props: Props) => {
 
   const sortedWalletList = useSelector(state => state.sortedWalletList)
   const account = useSelector(state => state.core.account)
-  const loanAccounts = useSelector(state => state.loanManager.loanAccounts)
+  const loanAccountsMap = useSelector(state => state.loanManager.loanAccounts)
```
samholmes, line 53: Should we rename the field in the redux state?

src/components/scenes/Loans/LoanDashboardScene.tsx
```diff
@@ -169,14 +171,35 @@ export const LoanDashboardScene = (props: Props) => {
...
-        keyboardShouldPersistTaps="handled"
-        renderItem={renderLoanCard}
-        style={margin}
-        ListFooterComponent={renderFooter()}
-        keyExtractor={(loanAccount: LoanAccount) => loanAccount.id}
-      />
+      {Object.keys(loanAccountsMap).length === 0 ? (
+        <>
+          {isLoansLoading ? (
+            <Space isFill isGroupCenter isItemCenter horizontal bottom={2.5}>
```
samholmes, line 177: Suggesting ideas just cause:

```jsx
<Space fill center middle horizontal bottom={2.5}>
```

Maybe `start`, `center`, `end` for item alignment, and `beginning`, `middle`, `ending`. Though if we do this, then what's the point of `sideways`? Hmm, we should drop sideways if we include justification fields.

Alternatively, `alignLeft, alignRight, alignTop, alignBottom` to abstract away the concept of align v justify (which I think has always been confusing). Centering would get interesting, we could have `center` to center both directions, but if another vertical alignment is included then it will only center horizontally, or if a horizontal alignment is included then it'll only center vertically. In other words, `align*` props will always override `center` making things straight forward and expressive:

```jsx
<Space center> // Everything is aligned in the middle (very common usage)
<Space alignLeft center> // Only vertically centered because horizontal alignment is "left"
<Space alignTop center> // Only horizontally centered because vertical alignment is "right"
<Space alignTop alignLeft center> // This should log some warning that the center prop was completely overwritten
```

So with that.

```
<Space fill center horizontal bottom={2.5}>
``` 

Is all you would need, and it is easy to grok at. And no confusing sideways mode.

samholmes (approved):
I approve these changes, but I'd like to perhaps revisit the Space props naming conventions.

src/components/scenes/Loans/LoanDashboardScene.tsx
Jon-edge: Let's think about this more later, because I know for sure at least that having `center` override both directions isn't always what we want to do in the GUI. There are plenty of cases where we want to have `justify` and `align` have different values.

PR #3678, opened 2022-09-23
Jon/fix/aave/loan-create-dest-row

Jon-edge:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202895674767117

### DIFF
--- a/src/components/modals/WalletListModal.tsx
+++ b/src/components/modals/WalletListModal.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { View } from 'react-native'
+import { TouchableOpacity, View } from 'react-native'
 import { AirshipBridge } from 'react-native-airship'
 import { FlatList } from 'react-native-gesture-handler'
 import { sprintf } from 'sprintf-js'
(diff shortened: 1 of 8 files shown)


src/components/modals/WalletListModal.tsx
```diff
@@ -140,7 +140,7 @@ export function WalletListModal(props: Props) {
   const handleCancel = useHandler(() => {
     bridge.resolve({})
   })
-  const handlePaymentMethodPress = useHandler((paymentMethodId: string, pluginId: string) => {
+  const handlePaymentMethodPress = useHandler((paymentMethodId: string) => () => {
```
samholmes, line 143: No longer currying means we must change the call-site.

samholmes (changes requested):

src/components/modals/WalletListModal.tsx
samholmes: Read this backwards.

samholmes (approved):

PR #3682, opened 2022-09-26
Show scam warning prior to address entry

itayplav:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

https://user-images.githubusercontent.com/4023066/192337862-0c0291ff-1f5d-4ad5-b890-ec418fc308f7.mp4

### DIFF
--- a/src/components/scenes/SendScene.tsx
+++ b/src/components/scenes/SendScene.tsx
@@ -381,7 +381,11 @@ class SendComponent extends React.PureComponent<Props, State> {
   }
 
   renderScamWarning() {
-    return <ScamWarningCard marginRem={[1.5, 1]} />
+    const { recipientAddress } = this.state
+    if (recipientAddress === '') {
+      return <ScamWarningCard marginRem={[1.5, 1]} />
+    }
+    return null
   }
 
   renderAmount() {


src/components/scenes/SendScene.tsx
```diff
@@ -381,7 +381,11 @@ class SendComponent extends React.PureComponent<Props, State> {
   }
 
   renderScamWarning() {
-    return <ScamWarningCard marginRem={[1.5, 1]} />
+    const { recipientAddress } = this.state
+    if (!recipientAddress) {
```
paullinator, line 385: Don't use vague boolean checks in IF statements. Do a check that returns a true boolean. ie `recipientAddress != null` or `recipientAddress != ''` or whatever is most appropriate. This should have caused a typescript error.

Also, function doesn't have a defined return value. Not sure how this file is passing typescript checks, but all functions should have return types.

paullinator (changes requested):

src/components/scenes/SendScene.tsx
itayplav: Fixed the ambiguous boolean issue & will note that for the future. For some reason, it doesn't trigger typescript errors in my vscode config, will look into that.

For the null return, I was following the pattern that similar `render*()` functions use within the SendScene, such as [renderError()](https://github.com/EdgeApp/edge-react-gui/blob/05a38e2c862f33a8b377bff5f9b229ab6bacebf7/src/components/scenes/SendScene.tsx#L426-L432),

PR #3693, opened 2022-10-03
Long press on tx row to share blockchair URL

itayplav:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

https://user-images.githubusercontent.com/4023066/193646049-b8809b09-f454-4713-a026-60e921573b5d.mp4

### DIFF
--- a/src/components/themed/TransactionListRow.tsx
+++ b/src/components/themed/TransactionListRow.tsx
@@ -1,5 +1,7 @@
 import { abs, div, log10 } from 'biggystring'
 import * as React from 'react'
+import Share from 'react-native-share'
+import { sprintf } from 'sprintf-js'
 
 import { getSymbolFromCurrency } from '../../constants/WalletAndCurrencyConstants'
 import { displayFiatAmount } from '../../hooks/useFiatText'
@@ -116,6 +118,14 @@ export function TransactionListRow(props: Props) {
     })
   })
 
+  const handleLongPress = useHandler(() => {
+    const url = sprintf(currencyInfo.transactionExplorer, transaction.txid)
+    const shareOptions = {
+      url
+    }
+    Share.open(shareOptions).catch(e => console.log(e))
+  })
+
   return (
     <TransactionRow
       cryptoAmount={cryptoAmountFormat}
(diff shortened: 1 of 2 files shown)


src/components/themed/TransactionListRow.tsx
```diff
@@ -116,13 +118,22 @@ export function TransactionListRow(props: Props) {
     })
   })
 
+  const handleLongPress = useHandler(() => {
+    const url = sprintf(currencyInfo.transactionExplorer, transaction.txid)
+    const shareOptions = {
+      url
+    }
+    Share.open(shareOptions).catch(e => console.log(e))
```
paullinator, line 126: Use `openBrowserUri` instead. This opens in an embedded SafariWebView on iOS

paullinator (changes requested):

src/components/themed/TransactionListRow.tsx
itayplav: looks like `openBrowserUri` opens non-embedded safari tab

https://user-images.githubusercontent.com/4023066/194687864-59112f9b-b036-4b3d-a985-91265c25aae0.mp4

paullinator (changes requested):
I see in the commit  history that you merged develop into your feature branch. This is not a part of our workflow. Instead you should re-base your branch on top of develop. 

Squash this down to one commit, rebase on top of develop, and you should be good

src/components/themed/TransactionListRow.tsx
```diff
@@ -115,13 +117,22 @@ export function TransactionListRow(props: Props) {
...
```
swansontec, line 125: We generally use `.catch(showError)` instead of logging to console, unless we _want_ to hide the error for some reason. Since this is a case of direct user interaction, we really should show the error so they know why their long tap didn't work.

swansontec (changes requested):

src/components/themed/TransactionListRow.tsx
itayplav: Changed to `.catch(e => showError(e))`

PR #3694, opened 2022-10-04
Fix fiat amount in txlist and export

paullinator:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203096402919670

### DIFF
--- a/src/actions/TransactionExportActions.tsx
+++ b/src/actions/TransactionExportActions.tsx
@@ -2,7 +2,12 @@ import { abs, add, div, gt, lt, mul } from 'biggystring'
 import csvStringify from 'csv-stringify/lib/browser/sync'
 import { EdgeCurrencyWallet, EdgeGetTransactionsOptions, EdgeTransaction } from 'edge-core-js'
 
+import { getExchangeDenomination } from '../selectors/DenominationSelectors'
+import { Dispatch, GetState } from '../types/reduxTypes'
+import { getHistoricalRate } from '../util/exchangeRates'
 import { DECIMAL_PRECISION } from '../util/utils'
+
+const UPDATE_TXS_MAX_PROMISES = 10
 
 export async function exportTransactionsToQBO(wallet: EdgeCurrencyWallet, txs: EdgeTransaction[], opts: EdgeGetTransactionsOptions): Promise<string> {
   const { currencyCode = wallet.currencyInfo.currencyCode, denomination } = opts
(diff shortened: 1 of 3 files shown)


src/components/themed/TransactionListRow.tsx
```diff
@@ -91,7 +91,7 @@ export function TransactionListRow(props: Props) {
     })
     maxConversionDecimals = maxPrimaryCurrencyConversionDecimals(log10(displayDenomination.multiplier), precisionAdjustValue)
   }
-  const cryptoAmount = div(abs(transaction.nativeAmount ?? '0'), displayDenomination.multiplier, DECIMAL_PRECISION)
+  const cryptoAmount = div(abs(transaction.nativeAmount ?? '0'), exchangeDenomination.multiplier, DECIMAL_PRECISION)
```
peachbits, line 94: Don't think this should have changed since it's a string displayed to the user.

src/actions/TransactionExportActions.tsx
```diff
@@ -3,8 +3,13 @@ import { abs, add, div, gt, lt, mul } from 'biggystring'
 import csvStringify from 'csv-stringify/lib/browser/sync'
 import { EdgeCurrencyWallet, EdgeGetTransactionsOptions, EdgeTransaction } from 'edge-core-js'
 
+import { getExchangeDenomination } from '../selectors/DenominationSelectors'
+import { Dispatch, GetState } from '../types/reduxTypes'
+import { getHistoricalRate } from '../util/exchangeRates'
 import { DECIMAL_PRECISION } from '../util/utils'
 
+const UPDATE_TXS_MAX_PROMISES = 10
```
peachbits, line 11: Can this match the actual rates server limit of 100? If so let's import that constant

src/actions/TransactionExportActions.tsx
```diff
@@ -21,6 +26,36 @@ export async function exportTransactionsToCSV(wallet: EdgeCurrencyWallet, txs: E
...
 }
 
+export const updateTxsFiat = (wallet: EdgeCurrencyWallet, currencyCode: string, txs: EdgeTransaction[]) => async (dispatch: Dispatch, getState: GetState) => {
+  const state = getState()
+  const exchangeDenom = getExchangeDenomination(state, wallet.currencyInfo.pluginId, currencyCode)
+  const { fiatCurrencyCode } = wallet
+
+  let promises: Array<Promise<void>> = []
+  for (const tx of txs) {
+    if (tx.metadata == null || tx.metadata.amountFiat === 0) {
```
peachbits, line 36: The EdgeMetadata type says amountFiat is optional so we should check for that case as well

peachbits (changes requested):

src/actions/TransactionExportActions.tsx
paullinator: Actually this is more of how many promises we want outstanding in the GUI. Has nothing to do with the 100 limit of the server.

peachbits (approved):

PR #3700, opened 2022-10-06
Sam/eth pow

samholmes:
Simulator Screen Shot - iPhone 14 - 2022-10-06 at 12 23 38

Simulator Screen Shot - iPhone 14 - 2022-10-06 at 12 23 03

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device


src/constants/WalletAndCurrencyConstants.ts
```diff
@@ -131,6 +131,7 @@ type SpecialCurrencyInfo = {
   keysOnlyMode?: boolean
   isPrivateKeySweepable?: boolean
   isBitPayProtocolSupported?: boolean
+  isNotTransactionListSupported?: boolean
```
Jon-edge, line 134: If we are using the is* prefix, best practice is never to do a double negative, makes it hard to read. Especially 'not' because it's what we use to describe the logical '!' operator.




Suggest isHiddenTransactionList

peachbits: I agree with Jon and think we should use `isTransactionListUnsupported`

src/locales/en_US.ts
```diff
@@ -163,7 +163,8 @@ const strings = {
   fragment_transaction_list_confirmation_progress: '%s of %s Confirmations',
   fragment_transaction_list_transaction: 'Transactions',
   transaction_list_buy_crypto_message: 'Buy %s',
-  transaction_list_no_tx_yet: 'No transactions yet!',
+  transaction_list_no_tx_yet: 'No transactions yet.',
+  transaction_list_no_tx_support_yet: 'Transaction history is not yet supported.',
```
peachbits, line 167: I say we avoid using ending punctuation for these one-off statements scattered across the app. Ending punctuation makes sense when you need to separate things but strings like this, to me, look better without.

src/locales/en_US.ts
```diff
@@ -163,7 +163,8 @@ const strings = {
   fragment_transaction_list_confirmation_progress: '%s of %s Confirmations',
   fragment_transaction_list_transaction: 'Transactions',
   transaction_list_buy_crypto_message: 'Buy %s',
-  transaction_list_no_tx_yet: 'No transactions yet!',
+  transaction_list_no_tx_yet: 'No transactions yet.',
```
peachbits, line 166: If you're going to remove the exclamation mark (and with it the implication of future excitement when the tx history list becomes populated) just get rid of it instead of replacing it with the killjoy period.

peachbits (changes requested):

src/constants/WalletAndCurrencyConstants.ts
samholmes: I'll make the change.

Though, it might not entirely solve the problem with double negatives (`!isTransactionListUnsupported` is still a double negative, though a bit easier to read/speak I guess).

Ideally, we could make a convention to never allow negative booleans; we'd have `isTransactionListSupported` and default this to a value of `true` somehow. But I haven't given this approach much thought on how to implement it.

peachbits (approved):

peachbits:
Screen Shot 2022-10-18 at 3 45 56 PM

peachbits:
Screen Shot 2022-10-18 at 5 44 40 PM

peachbits:
Ignore the icon obviously we don't have a proper one yet

PR #3702, opened 2022-10-07
Fix wrong fiat in tx list

paullinator:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203101814782468

### DIFF
--- a/src/actions/TransactionListActions.ts
+++ b/src/actions/TransactionListActions.ts
@@ -102,8 +102,7 @@ const getAndMergeTransactions = async (state: RootState, dispatch: Dispatch, wal
           // then add it
           ...tx,
           dateString: date,
-          time,
-          key
+          time
         })
         key++
       }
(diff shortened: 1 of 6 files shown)


src/types/types.ts
```diff
@@ -116,7 +116,6 @@ export type GuiFiatType = {
 
 export type TransactionListTx = EdgeTransaction & {
```
samholmes, line 117: I'm of the opinion that this type belongs in `src/reducers/scenes/TransactionListReducer.ts`. What do you think?

src/actions/TransactionListActions.ts
```diff
@@ -92,19 +90,16 @@ const getAndMergeTransactions = async (state: RootState, dispatch: Dispatch, wal
     for (const tx of transactions) {
       // for each transaction, add some meta info
       const { date, time } = unixToLocaleDateTime(tx.date)
-      // @ts-expect-error
       if (!transactionIdMap[tx.txid]) {
         // if the transaction is not already in the list
-        // @ts-expect-error
-        transactionIdMap[tx.txid] = key
+        transactionIdMap[tx.txid] = true
         // @ts-expect-error
```
samholmes, line 96: I guess an `unfliteredIndex` is needed to satisfy the static type check, but what is the initial value for this I wonder.

samholmes (approved):
Approved. Optional feedback.

PR #3704, opened 2022-10-10
Implement NumericInput component

paullinator:
Wraps a TextInput but forces international formatting of display value Use in EnterAmountScene through OutlineTextInput

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203136033933192

### DIFF
--- a/src/components/modals/NumericInput.tsx
+++ b/src/components/modals/NumericInput.tsx
@@ -1,0 +1,35 @@
+import React, { forwardRef, useEffect } from 'react'
+import { TextInput, TextInputProps } from 'react-native'
+
+import { useHandler } from '../../hooks/useHandler'
+import { formatNumberInput, formatToNativeNumber, isValidInput } from '../../locales/intl'
+import { useState } from '../../types/reactHooks'
+
+type Props = {
+  minDecimals?: number
+  maxDecimals?: number
+} & TextInputProps
+
+export const NumericInput = forwardRef<TextInput, Props>((props: Props, ref) => {
+  const { onChangeText, minDecimals, maxDecimals, value, ...rest } = props
+  const [innerValue, setInnerValue] = useState<string>(props.value ?? '')
+
+  useEffect(() => {
+    const propVal = props.value ?? ''
+    const displayNum = propVal === '' ? '' : formatNumberInput(propVal, { minDecimals, maxDecimals })
+    setInnerValue(displayNum)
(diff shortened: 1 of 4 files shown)


src/components/modals/NumericInput.tsx
```diff
@@ -0,0 +1,38 @@
+import * as React from 'react'
+import { useRef } from 'react'
+import { TextInput, TextInputProps } from 'react-native'
+
+import { useHandler } from '../../hooks/useHandler'
+import { formatNumberInput, formatToNativeNumber, isValidInput } from '../../locales/intl'
+import { useState } from '../../types/reactHooks'
+
+type Props = {
+  ref: React.RefObject<TextInput>
```
swansontec, line 10: Do not require the ref as a prop like this. The forwardRef thing takes care of all the ref typings.

swansontec (changes requested):
This looks pretty straightforward, but needs a typing fix.

src/components/modals/NumericInput.tsx
paullinator: Ack. I changed from accessing the `ref` as a prop to using `forwardRef` for forgot to remove the prop type. Will do

src/components/modals/NumericInput.tsx
```diff
@@ -0,0 +1,38 @@
...
+        if (onChangeText != null) onChangeText(nativeNum)
+        setInnerValue(displayNum)
+      }
+    }
+  })
+
+  if (props.value !== propValue.current) {
+    propValue.current = props.value ?? ''
+    const displayNum = props.value === '' ? '' : formatNumberInput(propValue.current, { minDecimals, maxDecimals })
+    setInnerValue(displayNum)
```
swansontec, line 33: I was going to approve the PR, but then I realized this line will crash React. You absolutely, 100% cannot call `setState` in the main render body. This logic must be wrapped inside `useEffect` or such, or React will throw an error.

swansontec (changes requested):

src/components/modals/NumericInput.tsx
paullinator: Hmm. Do you have documentation on this limitation. I definitely hit that code and I don’t get a crash or even a warning.

swansontec (approved):

PR #3706, opened 2022-10-11
Replace create wallet with create multiple asset flow

peachbits:
Creating new wallets:
https://user-images.githubusercontent.com/41451710/195939559-9cc475f5-e1ef-429b-9960-071ba83622a9.mp4

Import key:
https://user-images.githubusercontent.com/41451710/195939598-fd841be0-cad1-4e1e-9727-ca75141661cf.mp4

---
  - https://app.asana.com/0/0/1202429146875992


src/components/scenes/CreateWalletImportScene.tsx
```diff
@@ -1,167 +1,172 @@
...
+    const results = await Promise.all(promises)
+
+    const successMap: { [pluginId: string]: JsonObject } = {}
+
+    for (const [i, keys] of results.entries()) {
+      if (typeof keys === 'object') {
+        // Success
+        successMap[pluginIds[i]] = keys
+      }
+    }
```
swansontec, line 66: Oooh, I have learned something new. I have always done the oldschool `for  (let i = 0; i < array.length; ++i)` when I needed to iterate over an array and needed the loop counter too. I did not know you could use `array.entries()` to get access to the loop counter in a `for of` loop. Neato.

src/components/scenes/CreateWalletImportScene.tsx
```diff
@@ -1,167 +1,172 @@
...
+      if (resolveValue === 'cancel' || resolveValue == null) {
+        return
+      }
+    }
+
+    navigation.navigate('createWalletCompletion', { createWalletList, walletNames, fiatCode, importText: cleanImportText })
+  })
+
+  // Scale the icon to match the height of the first MainButton container for consistency
+  const [iconContainerLayout, setIconContainerLayout] = useLayout()
```
swansontec, line 124: We don't have the linting rule switched on officially yet, but `onX` props need to have a callback whose name starts with `handle`. So `setIconContainerLayout` needs to be `handleIconContainerLayout`

src/components/scenes/CreateWalletSelectCryptoScene.tsx
```diff
@@ -1,144 +1,255 @@
...
-type StateProps = {
-  account: EdgeAccount
-}
-type Props = StateProps & OwnProps & ThemeProps
 
-type State = {
-  selectedWalletType: string
-  searchTerm: string
-}
+type Props = OwnProps
```
swansontec, line 28: Same thing here. Just call it `Props` and don't bother with the `OwnProps` and the alias.

src/components/scenes/CreateWalletSelectFiatScene.tsx
```diff
@@ -1,161 +1,225 @@
...
-type StateProps = {
-  supportedFiats: GuiFiatType[]
-}
-type Props = OwnProps & StateProps & ThemeProps
 
-type State = {
-  searchTerm: string
-  selectedFiat: string
-}
+type Props = OwnProps
```
swansontec, line 37: Same thing with naming the type `Props` directly & skipping `OwnProps`

src/components/scenes/CreateWalletSelectFiatScene.tsx
```diff
@@ -1,161 +1,225 @@
...
 
-  getFiatType = (fiatKey: string) => {
-    const fiatTypeIndex = this.props.supportedFiats.findIndex(fiatType => fiatType.value === fiatKey)
+  const account = useSelector(state => state.core.account)
+  const currencyWallets = useWatch(account, 'currencyWallets')
 
-    return this.props.supportedFiats[fiatTypeIndex]
-  }
+  const defaultFiat = useSelector(state => getDefaultFiat(state))
+  const [fiat, setFiat] = React.useState(getSupportedFiats(defaultFiat)[0])
```
swansontec, line 51: The `getSupportedFiats` call will run on every render, which is expensive. If you put the `useState` initializer in a function, it will only run on first render. So `React.useState(() => getSupportedFiats(defaultFiat)[0])`

src/components/scenes/CreateWalletSelectFiatScene.tsx
```diff
@@ -1,161 +1,225 @@
...
-          selectedWalletType: selectedWalletType,
-          selectedFiat: this.getFiatType(this.state.selectedFiat),
-          cleanedPrivateKey
-        })
-      } else {
-        navigation.navigate('createWalletAccountSetup', {
-          selectedWalletType: selectedWalletType,
-          selectedFiat: this.getFiatType(this.state.selectedFiat)
-        })
+  const [walletNames, setWalletNames] = React.useState(defaultWalletNames)
```
swansontec, line 64: For the `defaultWalletNames` thing, you can just put the function inside the `React.useState` and skip the whole `React.useMemo`. The function will run on first render, since that is a built-in `React.useState` feature.

src/components/scenes/CreateWalletSelectFiatScene.tsx
```diff
@@ -1,161 +1,225 @@
...
+      return
     }
 
+    navigation.navigate('createWalletImport', { createWalletList: [...newWalletItemsCopy, ...newTokenItems], walletNames, fiatCode: fiat.value })
+  })
+
+  const renderSelectedFiatRow = useHandler(() => {
+    const fiatCountry = FIAT_COUNTRY[fiat.value]
+    // @ts-expect-error
+    const subTitle = s.strings[`currency_label_${fiat.value}`] ?? s.strings.currency_label_
```
swansontec, line 154: Is `s.strings.currency_label_` an error? Since you have the `// @ts-expect-error` for the first half of the statement, it might also be hiding an error in the second half.

You can usually avoid `// @ts-expect-error` by using `any`, which is preferable because it's more targeted:

```ts
const key: any = `currency_label_${fiat.value}`
const subTitle = s.strings[key] ?? s.strings.currency_label_WHAT
```

swansontec (changes requested):
This PR looks great! Most of the stuff I found was quite small. Also, have we cleaned the code of unused strings, post-refactor?

src/actions/CreateWalletActions.tsx
```diff
@@ -203,3 +204,45 @@ export const createHandleUnavailableModal = (newWalletId: string, accountName: s
   ))
   Actions.pop()
 }
+
+export const PLACEHOLDER_WALLET_ID = 'NEW_WALLET_UNIQUE_STRING'
+export type MainWalletCreateItem = WalletCreateItem & { walletType: string }
+type TokenWalletCreateItem = WalletCreateItem & { tokenId: string; createWalletIds: string[] }
```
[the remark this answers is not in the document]
peachbits: Good point and will take care of it now

src/components/scenes/CreateWalletSelectFiatScene.tsx
peachbits: Still getting a type error when using any. I was able to satisfy TS with the following:

```
const key = `currency_label_${fiat.value}`
const subTitle = s.strings[key as keyof typeof s.strings] ?? s.strings.currency_label_
```

src/__tests__/scenes/CreateWalletImportScene.test.tsx
```diff
@@ -1,22 +1,24 @@
...
 import { createStore } from 'redux'
 
 import { CreateWalletImportScene } from '../../components/scenes/CreateWalletImportScene'
 import { getTheme } from '../../components/services/ThemeContext'
 import { rootReducer } from '../../reducers/RootReducer'
 import { fakeNavigation } from '../../util/fake/fakeNavigation'
-import { fakeUser } from '../../util/fake-user'
+
+// Jest doesn't like direct SVG imports
+jest.mock('../../assets/images/import-key-icon.svg', () => 'ImportKeySvg')
```
swansontec, line 13: Great find!

src/__tests__/scenes/CreateWalletSelectCryptoScene.test.tsx
```diff
@@ -46,21 +47,27 @@ describe('CreateWalletSelectCrypto', () => {
...
     }
   }
   const store = createStore(rootReducer, mockState)
 
   it('should render with loading props', () => {
-    const renderer = createRenderer()
-
     const props: any = {
       navigation: fakeNavigation,
-      account: () => fakeUser,
```
swansontec, line 16: I really don't like this `const props = { ... }` followed by `<TestComponent {...props} />`. We should just put the props on the component, like normal. That way TypeScript will check them individually and warn you about extras or missing ones. However, as long as your tests work (which they seem to be), we don't need to fix this right away.

swansontec (approved):
Approving, but with one optional change if you want to tidy up the tests a bit more than you already have.

src/components/scenes/CreateWalletSelectFiatScene.tsx
swansontec: That works for me.

src/__tests__/scenes/CreateWalletSelectCryptoScene.test.tsx
peachbits: I went ahead and did this and actually caught a bug in the cryptorow test where the icon wasn't appearing in the snapshot.

PR #3719, opened 2022-10-13
White label fixes

paullinator:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203156467868506


android/app/src/main/AndroidManifest.xml
```diff
@@ -41,7 +41,7 @@
     android:usesCleartextTraffic="true"
     android:theme="@style/AppTheme">
     <meta-data android:name="com.bugsnag.android.API_KEY"
-               android:value="5aca2dbe708503471d8137625e092675"/>
+               android:value="BUGSNAG_API_KEY"/>
```
samholmes, line 44: Yikes. Good catch. Has the previous value been appropriately deprecated?

scripts/patchFiles.ts
```diff
@@ -39,6 +39,22 @@ async function main() {
...
+  const env = require('../env.json')
+  const bugsnagFiles = ['./android/app/src/main/AndroidManifest.xml', './ios/edge/Info.plist']
+
+  for (const file of bugsnagFiles) {
+    await searchReplace(file, 'BUGSNAG_API_KEY', env.BUGSNAG_API_KEY)
+  }
+}
+
+async function searchReplace(file: string, search: string, replace: string): Promise<void> {
+  console.log(`${file} ${search} ${replace}`)
```
samholmes, line 53: Should we be logging keys like this? Or have we accepted that the log output for builds are sensitive

scripts/patchFiles.ts
```diff
@@ -39,6 +39,22 @@ async function main() {
...
+  const text = fs.readFileSync(file, { encoding: 'utf8' })
+  const regex = new RegExp(search, 'g')
+  const newText = text.replace(regex, replace)
```
samholmes, line 56:
```ts
text.split(search).join(replace)
```

This avoids unintended regex-injection.

scripts/secretFiles.ts
```diff
@@ -0,0 +1,58 @@
...
+      quietCopy(src, dest)
+    })
+    // Copy keystores directory
+    const keystoreSrc = join(filesDir, repoBranch, 'keystores')
+    if (fs.existsSync(keystoreSrc)) {
+      copySync(keystoreSrc, join(_rootProjectDir, 'keystores'))
+    }
+  }
+  // Always copy the files for the master branch first
+  copyFiles('master')
```
samholmes, line 42: Why must we always copy for the master branch?

samholmes (changes requested):

android/app/src/main/AndroidManifest.xml
paullinator: No. Fortunately it's not much of a security issue.

scripts/patchFiles.ts
paullinator: Yes build logs are sensitive-ish (no private keys though).  Also they never leave the local machine. We already log the deploy-config.json which has some key info

scripts/secretFiles.ts
paullinator: Because it contains the base files that are shared with a lot of other configurations like develop and beta. Otherwise we'd have a lot of duplicates. It's like our env.json which is mostly shared and the deploy-config.json overwrites just some of the values.

scripts/patchFiles.ts
paullinator: Ok. will make the change.

samholmes (approved):

PR #3722, opened 2022-10-14
Refactored ScamWarningCard to a generic WarningCard

itayplav:
ScamWarningCard is now a generic warning card that accepts a list of bullet points (as well as a title/header/footer) and will render the bullet points with correct vertical alignment. 

Should be visually identical to the [original ScamWarningCard](https://github.com/EdgeApp/edge-react-gui/pull/3682)

#### PR Requirements

- [x] Tested on iOS device

image

### DIFF
--- a/src/components/cards/WarningCard.tsx
+++ b/src/components/cards/WarningCard.tsx
@@ -2,16 +2,35 @@ import * as React from 'react'
 import { Platform, View } from 'react-native'
 import IonIcon from 'react-native-vector-icons/Ionicons'
 
-import s from '../../locales/strings'
 import { fixSides, mapSides, sidesToMargin, sidesToPadding } from '../../util/sides'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 import { EdgeText } from '../themed/EdgeText'
 
 type Props = {
+  title: string
+  header?: string | null
+  points?: string[] | null
+  footer?: string | null
   marginRem?: number[] | number
   paddingRem?: number[] | number
 }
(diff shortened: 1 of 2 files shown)


src/components/cards/WarningCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+          size={theme.rem(0.8)}
+        />
+        <EdgeText style={styles.title}>{title}</EdgeText>
+      </View>
+      {header && (
+        <EdgeText style={styles.text} numberOfLines={0}>
+          {header}
+        </EdgeText>
+      )}
+      {points && <View style={styles.bulletpoints}>{points?.map(point => renderBulletpoint(point))}</View>}
```
Jon-edge, line 66: Avoid truthy-ness `points && ...` - I'd prefer explicit boolean expressions here when referencing non-boolean vars. 

While in the context of the GUI and the smaller component size here your intentions are obvious, but in more extreme cases it can get tedious having to lookup more declarations and opens unnecessary risk of bugs from the additional required oversight.

The only time I like to see truthy-ness is if the var is explicitly named `isSomething`

Jon-edge, line 66: Also suggest updating this such that both `undefined` and `[]` `points` result in a hidden bulletpoints view here.

Jon-edge (changes requested):

src/components/cards/WarningCard.tsx
itayplav: Acknowledged, updated with explicit null checking in new commit

Jon-edge (approved):

PR #3725, opened 2022-10-14
Jon/aave/multi-action-manager

Jon-edge:
- Adds chained action support to manage loan: borrow (does NOT implement support for chained actions for loan-deposit)
- Refactors fee calculations to use dryrun
- Removes ManageCollateralScene (renamed LoanManageScene) dependent scenes

#### PR Dependencies

 none

#### PR Requirements

If you have made **any** visual changes to the GUI. Make sure you have:
- [x] Tested on iOS device
- [ ] Tested on Android device
- [ ] Tested on small-screen device (iPod Touch)
- [ ] Tested on large-screen device (tablet)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203146252990888
  - https://app.asana.com/0/0/1203135361770742

PR #3727, opened 2022-10-15
Jon/aave/disable-details-buttons

Jon-edge:
- Disable loan action buttons if an ActionProgram is running
- Disable withdraw/repay if there is nothing to withdraw/repay

Simulator Screen Shot - iPhone 8 Plus - 2022-10-14 at 18 51 42

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202949203074528


src/components/scenes/Loans/ManageCollateralScene.tsx
```diff
@@ -168,87 +151,91 @@ export const ManageCollateralScene = <T extends keyof ParamList>(props: Props<T>
...
+        programId: actionProgram.programId,
+        effective: false,
+        executing: false,
+        lastExecutionTime: 0,
+        nextExecutionTime: 0
+      }
+      const executionContext = { account, clientId }
+      const executionOutputs = await dryrunActionProgram(executionContext, actionProgram, actionProgramState, false)
+      const networkFeeMap = getExecutionNetworkFees(executionOutputs)
+      setFeeNativeAmount(networkFeeMap[borrowEngineWallet.currencyInfo.currencyCode]?.nativeAmount ?? '0')
```
samholmes, line 238: We should store the entire `NetworkFeeMap` as state and then derive the `feeNativeAmount` from that state. This way when we go to fix our fee calculation to include all network fee amounts, we would only need to remove the derived value and directly pass the `NetworkFeeMap` object to the component/function handling the calculation of all fees.

src/components/scenes/Loans/ManageCollateralScene.tsx
```diff
@@ -316,7 +302,7 @@ export const ManageCollateralScene = <T extends keyof ParamList>(props: Props<T>
   // #endregion Handlers
 
   return (
-    <FormScene headerText={headerText} onSliderComplete={handleSliderComplete} sliderDisabled={approvalAction == null}>
+    <FormScene headerText={headerText} onSliderComplete={handleSliderComplete} sliderDisabled={actionProgram == null}>
```
samholmes: We should include an insufficient funds case to the `sliderDisabled` condition in a later commit.

src/components/scenes/Loans/ManageCollateralScene.tsx
```diff
@@ -43,22 +43,42 @@ import { FormScene } from '../FormScene'
...
-  headerText: string
   navigation: NavigationProp<T>
 }
 
 export const ManageCollateralScene = <T extends keyof ParamList>(props: Props<T>) => {
   // -----------------------------------------------------------------------------
   // #region Constants
   // -----------------------------------------------------------------------------
-  const { actionOpType, amountChange = 'increase', loanAccount, showAprChange = false, headerText, navigation } = props
+  const { actionOpType, loanAccount, navigation } = props
```
samholmes: Bravo! I love the reduction in complexity made to this component. It's a very useful abstraction now.

src/components/scenes/Loans/ManageCollateralScene.tsx
```diff
@@ -69,11 +89,13 @@ export const ManageCollateralScene = <T extends keyof ParamList>(props: Props<T>
   const { borrowEngine, borrowPlugin } = loanAccount
   const { currencyWallet: borrowEngineWallet } = loanAccount.borrowEngine
   const { fiatCurrencyCode: isoFiatCurrencyCode, currencyInfo: borrowEngineCurrencyInfo } = borrowEngineWallet
+  const actionOpStrings = ACTIONOP_TYPE_STRING_MAP[actionOpType]
   const collaterals = useWatch(borrowEngine, 'collaterals')
   const debts = useWatch(borrowEngine, 'debts')
   const borrowEnginePluginId = borrowEngineCurrencyInfo.pluginId
   const borrowPluginInfo = borrowPlugin.borrowInfo
   const borrowPluginId = borrowPluginInfo.borrowPluginId
+  const isShowAprChange = actionOpType === 'loan-repay' || actionOpType === 'loan-borrow'
```
samholmes, line 98: Name suggestion: `isAprChangeShown`. I think if the verb is always going to be "is", thne we have to figure out a noun form for any verbs like "show".

src/components/scenes/Loans/ManageCollateralScene.tsx
```diff
@@ -43,22 +43,42 @@ import { FormScene } from '../FormScene'
...
+    headerText: s.strings.loan_make_payment,
+    amountCard: s.strings.loan_fragment_repay,
+    srcDestCard: s.strings.loan_fund_source
+  },
+  'loan-withdraw': {
+    headerText: s.strings.loan_withdraw_collateral,
+    amountCard: s.strings.loan_fragment_withdraw,
+    srcDestCard: s.strings.loan_fund_destination
+  }
+}
```
samholmes, line 67: We can add `as const` here to make certain the type is strictly this constant value.

src/components/scenes/Loans/ManageCollateralScene.tsx
```diff
@@ -41,51 +40,52 @@ import { NetworkFeeTile } from '../../tiles/NetworkFeeTile'
...
+  },
+  'loan-borrow': {
+    headerText: s.strings.loan_borrow_more,
+    amountCard: s.strings.loan_fragment_loan,
+    srcDestCard: s.strings.loan_fund_destination
+  },
+  'loan-repay': {
+    headerText: s.strings.loan_make_payment,
+    amountCard: s.strings.loan_fragment_repay,
+    srcDestCard: s.strings.loan_fund_source
```
samholmes, line 65: This re-ordering change should be squashed into the previous commit.

samholmes (changes requested):
This review is really for https://github.com/EdgeApp/edge-react-gui/pull/3725. I didn't notice the future commit. We should make the PR Dependencies section of the pull-request template more noticeable both for the requester and reviewers.

src/components/cards/TappableCard.tsx
```diff
@@ -24,15 +24,19 @@ const TappableCardComponent = ({ children, disabled = false, onPress, ...cardPro
   const theme = useTheme()
   const styles = getStyles(theme)
 
+  const renderTouchableChildren = (
```
samholmes, line 27: This should be a function as all `render*` functions are.

src/components/scenes/Loans/LoanDetailsScene.tsx
```diff
@@ -174,36 +177,44 @@ export const LoanDetailsScene = (props: Props) => {
         <Space horizontal>
           <Space bottom>
             <SectionHeading>{s.strings.loan_actions_title}</SectionHeading>
+            {isActionProgramRunning ? (
+              <Alert
+                type="warning"
+                title={s.strings.wc_smartcontract_warning_title}
+                message={s.strings.loan_action_program_running}
+                marginRem={[0.5, 0, 0, 0]}
+              />
```
samholmes, line 192: Maybe a separate `title` string for this instead of re-using `wc_smartcontract_warning_title`? It could be set to "Waiting on Status" instead of "Warning". What do you think?

src/components/scenes/Loans/ManageCollateralScene.tsx
Jon-edge: Agreed, actually forgot to make a TODO on that to expand the UI design for that case.

In any case, it will be handled here: https://app.asana.com/0/1200382638405084/1203187329419620/f

samholmes (approved):
Approved _this_ PR with this second review. Though, this is blocked by the upstream PR and that has some feedback. The following comments are optional feedback changes.

src/components/scenes/Loans/ManageCollateralScene.tsx
Jon-edge: Yeah, your suggestion is a nicer **sounding** name, but I would like to caution trying to mold this concept into 'proper English.'

'Proper English' is rife with pitfalls in that the English language itself is just plain weird and has a lot of exceptions and arbitrary rules for tense, plurality, etc. French is similar in this regard. As such, IMO:

**Consistency and hierarchy > Nice sounding English**

Less overhead in naming things too if there's a simple system in place without having to wordsmith.

I know also you are a fan of imperatives, so, `isShowApr` might be the best here in that sense

src/components/cards/TappableCard.tsx
Jon-edge: Doesn't work here, TS complains about not allowing functions as children

src/components/scenes/Loans/LoanDetailsScene.tsx
Jon-edge: Good catch, I threw this in just to satisfy the props generically. "Please wait" perhaps?

src/util/ActionProgramUtils.ts
```diff
@@ -77,10 +84,36 @@ export const makeAaveCreateAction = async (params: AaveCreateActionParams): Prom
     walletId: borrowEngineWallet.id
   })
 
+  //
+  // Borrow and ACH steps
+  //
+  // HACK: par as top level ActionOp type not supported
+  const borrowAction = ((await makeAaveBorrowAction({ borrowEngineWallet, borrowPluginId, destination })) as SeqActionOp).actions[0] as ParActionOp
+  if (borrowAction.actions.length > 0) loanParallelActions.push(...borrowAction.actions)
```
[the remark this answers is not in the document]
Jon-edge: It's not just wrapping. We need to cast the type because `actions` does not exist on all `ActionOp` types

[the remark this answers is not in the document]
Jon-edge: This isn't the first time the typing of these guys caused some snags.... I think we need to separate `SeqActionOp` and `ParActionOp` out from the rest. They're not really operations in themselves but organizational nodes. Redefine the ActionOps as `ActionNodes` and `ActionLeafs`?

Just a thought, maybe you have a better idea.

src/components/scenes/Loans/ManageCollateralScene.tsx
```diff
@@ -168,87 +151,91 @@ export const ManageCollateralScene = <T extends keyof ParamList>(props: Props<T>
...
-      return
+    if (actionProgram != null) {
+      // Get fees
+      const actionProgramState = {
+        clientId,
+        programId: actionProgram.programId,
+        effective: false,
+        executing: false,
+        lastExecutionTime: 0,
+        nextExecutionTime: 0
```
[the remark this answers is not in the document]
Jon-edge: I like constants better if it's plain static data

src/components/scenes/Loans/ManageCollateralScene.tsx
Jon-edge: Force pushed these changes

src/components/scenes/Loans/ManageCollateralScene.tsx
Jon-edge: Force pushed these changes

src/components/scenes/Loans/ManageCollateralScene.tsx
samholmes: It's all arguable I suppose. With that said, the claim `Consistency and hierarchy > Nice sounding English` is conjecture without some predicate to back it the inequality. If you don't want to change it, that's A-OK.

PR #3728, opened 2022-10-17
Fix the logout crash on TransactionDetails

swansontec:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203111210852322

### DIFF
--- a/src/components/scenes/TransactionListScene.tsx
+++ b/src/components/scenes/TransactionListScene.tsx
@@ -26,7 +26,7 @@ type StateProps = {
   numTransactions: number
   selectedWalletId: string
   selectedCurrencyCode: string
-  wallet: EdgeCurrencyWallet
+  wallet?: EdgeCurrencyWallet
   tokenId?: string
   transactions: TransactionListTx[]
 }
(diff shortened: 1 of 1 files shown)


src/components/scenes/TransactionListScene.tsx
```diff
@@ -94,12 +93,13 @@ class TransactionListComponent extends React.PureComponent<Props, State> {
...
 
   handleSearchTransaction = (searchString: string) => {
-    const { getTransactions, selectedCurrencyCode, transactions } = this.props
+    const { wallet, selectedCurrencyCode, transactions } = this.props
     this.setState({ loading: true })
-    getTransactions({
-      currencyCode: selectedCurrencyCode,
-      searchString
-    })
+    wallet
```
paullinator, line 98: This would still crash if the user was in search mode at the time of logout.

paullinator: check for `wallet == null` and early return. Then you should be good.

paullinator (changes requested):

paullinator (approved):

paullinator:
Ah crud. Forgot to rebase. My bad!

PR #3730, opened 2022-10-18
Remove Wyre for new users

paullinator:
Only show Wyre when a user has a linked bank account with them

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203186994552049

### DIFF
--- a/src/components/scenes/GuiPluginListScene.tsx
+++ b/src/components/scenes/GuiPluginListScene.tsx
@@ -14,7 +14,7 @@ import { COUNTRY_CODES } from '../../constants/CountryConstants'
 import { customPluginRow, guiPlugins } from '../../constants/plugins/GuiPlugins'
 import s from '../../locales/strings'
 import { getSyncedSettings, setSyncedSettings } from '../../modules/Core/Account/settings'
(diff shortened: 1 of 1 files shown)


src/components/scenes/GuiPluginListScene.tsx
```diff
@@ -333,6 +342,10 @@ class GuiPluginList extends React.PureComponent<Props, State> {
     const activePlugins = bestOfPlugins(accountPlugins, accountReferral, undefined)
     plugins = plugins.filter(plugin => !activePlugins.disabled[plugin.pluginId])
 
+    if (!this.state.hasWyreAccountHack) {
+      plugins = plugins.filter(plugin => plugin.pluginId !== 'wyre')
+    }
+
```
swansontec, line 348: I see. So the scene launches without Wyre, and once the check passes (async), we update the list to include it. This seems right to me.

swansontec (approved):

src/components/scenes/GuiPluginListScene.tsx
paullinator: Yep!

PR #3732, opened 2022-10-18
ACH buy w/moonpay

paullinator:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203186994552053


src/plugins/gui/fiatPluginTypes.ts
```diff
@@ -1,11 +1,11 @@
-import { asValue } from 'cleaners'
+import { asEither, asValue } from 'cleaners'
 import { EdgeAccount } from 'edge-core-js'
 
 import { EdgeTokenId } from '../../types/types'
 import { EnterAmountPoweredBy } from './scenes/EnterAmountScene'
 
-export const asFiatPaymentTypes = asValue('credit', 'applepay', 'googlepay')
-export type FiatPaymentType = 'credit' | 'applepay' | 'googlepay'
+export const asFiatPaymentType = asEither(asValue('credit'), asValue('applepay'), asValue('googlepay'), asValue('iach'))
```
swansontec, line 7: Please keep using the multi-parameter form of `asValue`, instead of using `asEither` like this. The `asValue` run-time behavior is much better, since it knows to check all the enum variants, and then print a meaningful error message if the value doesn't match any of them.

src/plugins/gui/providers/simplexProvider.ts
```diff
@@ -120,7 +120,7 @@ const SIMPLEX_ID_MAP: { [pluginId: string]: { [currencyCode: string]: string } }
 
 const allowedCurrencyCodes: FiatProviderAssetMap = { crypto: {}, fiat: {} }
 const allowedCountryCodes: { [code: string]: boolean } = {}
-const allowedPaymentTypes = { applepay: true, credit: true, googlepay: false }
+const allowedPaymentTypes: { [payment: string]: boolean } = { applepay: true, credit: true, googlepay: false }
```
swansontec, line 123: This type annotation stomps away useful information, since it prevents TypeScript from validating the key. It would be better to delete the type annotation and fix the revealed errors (adding in 'iach' as required).

Otherwise, if you really want a type annotation here, you could type it as `{[K in FiatPaymentType]: boolean}`, which will define one property per enum member.

Or, if the property is optional, you can do `{[K in FiatPaymentType]?: boolean}` instead.

src/plugins/gui/fiatPluginTypes.ts
swansontec: Also, should "iach" just be "ach"? The payment type is the payment type, and the settlement speed is the settlement speed. Mixing the two concepts seems like it will cause problems.

What if MoonPay does instant transfers for most customers, but then decides to wait a day for certain jurisdictions or risk profiles? Do we need two MoonPay plugins at that point, to handle the different speeds?

I think the payment type should be "ach", and the "instant" part plays into how we prioritize things.

src/plugins/gui/providers/moonpayProvider.ts
```diff
@@ -21,7 +21,7 @@ const pluginDisplayName = 'Moonpay'
 
 const allowedCurrencyCodes: FiatProviderAssetMap = { crypto: {}, fiat: {} }
 const allowedCountryCodes: { [code: string]: boolean } = {}
-const allowedPaymentTypes = { applepay: true, credit: true, googlepay: true }
+const allowedPaymentTypes: { [payment: string]: boolean } = { applepay: true, credit: true, googlepay: false, iach: true }
```
swansontec, line 24: Same thing with the type definition hiding info here.

src/constants/plugins/GuiPlugins.ts
```diff
@@ -15,46 +15,6 @@ export const guiPlugins: { [pluginId: string]: GuiPlugin } = {
     originWhitelist: ['https://libertyx.com'],
     permissions: ['location']
   },
-  moonpay: {
```
swansontec, line 18: I take it this "moonpay" plugin is unused, and has been ever since we added the explicit credit card flow?

swansontec (changes requested):

src/plugins/gui/fiatPluginTypes.ts
paullinator: Having iACH is based on the fact that discussions with several partners show they are looking to offer both and have iACH be more costly than regular ACH so it's not just settlement time but also price.

paullinator: I switched away from the multi param model because it didn't properly map to a type when using `ReturnType <typeof ...>` The new syntax does map correctly so I don't have to duplicate the type and cleaner

src/constants/plugins/GuiPlugins.ts
paullinator: Yes, not used anymore although added back to the sell side

src/plugins/gui/providers/moonpayProvider.ts
paullinator: Fixed

src/plugins/gui/providers/simplexProvider.ts
paullinator: fixed

paullinator: Although I fixed using `{[payment: FiatPaymentType] boolean}` syntax. Is that effectively an equivalent?

swansontec (approved):

src/plugins/gui/providers/simplexProvider.ts
swansontec: We found the issue: it was the `asEither` above. Fixing that causes the `in` syntax to work again.

src/plugins/gui/fiatPluginTypes.ts
swansontec: After further experiments, it turns out that `asValue` does work correctly, and `asEither` actually guesses the type as `string`, which is too loose.

PR #3737, opened 2022-10-20
Sam/aave perf

samholmes:
#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203185303900659

### DIFF
--- a/src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.ts
+++ b/src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.ts
@@ -118,10 +118,9 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
         instance.loanToValue = loanToValue
         instance.syncRatio = 1
       } catch (error: any) {
+        // TODO: Handle error cases such as rate limits
         console.warn(`Failed to load BorrowEngine for wallet '${wallet.id}': ${String(error)}`)
         console.error(error)
-        await snooze(2000)
-        return await startNetworkSyncLoop()
       } finally {
         // Re-sync after delay
         await snooze(15000)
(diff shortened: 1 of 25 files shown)


src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.ts
```diff
@@ -119,11 +121,11 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
         console.warn(`Failed to load BorrowEngine for wallet '${wallet.id}': ${String(error)}`)
         console.error(error)
         await snooze(2000)
```
Jon-edge, line 123: Remove retry

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.ts
```diff
@@ -136,8 +138,17 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
       collaterals: [] as BorrowCollateral[],
       debts: [] as BorrowDebt[],
       loanToValue: 0,
+      isRunning: false,
       syncRatio: 0,
 
+      async startEngine() {
+        instance.isRunning = true
+        startNetworkSyncLoop()
```
Jon-edge, line 146: I think this entire block needs to be wrapped in `if (!instance.isRunning)` to avoid potentially starting multiple loops? Then that way the callers don't need to remember to check for `isRunning` before calling startEngine()

samholmes: Good idea

Jon-edge (approved):

PR #3744, opened 2022-10-21
Sam/aave perf fix

samholmes:
#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

### DIFF
--- a/src/controllers/loan-manager/redux/actions.ts
+++ b/src/controllers/loan-manager/redux/actions.ts
@@ -201,7 +201,7 @@ export function resyncLoanAccounts(account: EdgeAccount): ThunkAction<Promise<vo
 
             await waitForBorrowEngineSync(borrowEngine)
 
-            if (checkLoanHasFunds(borrowEngine)) {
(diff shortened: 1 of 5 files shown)


src/controllers/loan-manager/redux/actions.ts
```diff
@@ -106,6 +108,7 @@ export function loadLoanAccounts(account: EdgeAccount): ThunkAction<Promise<void
       // Instantiate loan account from loanAccountEntry
       const loanAccount = await makeLoanAccount(borrowPlugin, wallet)
       loanAccount.closed = loanAccountEntry.closed
+      loanAccount.shown = loanAccountEntry.shown
```
samholmes, line 111: Make sure to not startEngine if `shown` is false. Also rename to `active` because of this logic.

samholmes: We went with a simpler approach which didn't require a new field.

samholmes:
Approved OTS

PR #3746, opened 2022-10-22
Sam/aave perf2

samholmes:
#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203210359195166


src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.ts
```diff
@@ -47,6 +47,24 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
         }
       }
     }
+    const adjustCollateral = (tokenId: string | undefined, amount: string) => {
+      const index = instance.collaterals.findIndex(collateral => collateral.tokenId === tokenId)
+      if (index === -1) return // Silently fail and let the network sync reflect the change
```
Jon-edge, line 54: Consider log level output for these -1 cases, if appropriate

Jon-edge (changes requested):

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.ts
```diff
@@ -271,8 +279,17 @@ export const makeBorrowEngineFactory = (blueprint: BorrowEngineBlueprint) => {
...
+
+        // Get the outstanding debt for it's total
+        const debt = instance.debts.find(debt => debt.tokenId === tokenId)
+        if (debt == null) throw new Error(`No debt to repay for ${tokenId}`)
+
+        // Let the debt total be the upper boundary for nativeAmount
+        const nativeAmount = min(request.nativeAmount, debt.nativeAmount)
+
+        // nativeAmount can't be zero
+        if (nativeAmount === '0') throw new Error('BorrowEngine: repay request contains no nativeAmount.')
```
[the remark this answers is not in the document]
samholmes: The condition was narrowed because `zeroString` coverage was too broad. I don't believe any of our usages of `repay` will enter this code path.

src/plugins/borrow-plugins/plugins/aave/BorrowEngineFactory.ts
samholmes: The comment I made is misleading. This is actually just an assertion that the provided tokenId is valid (if it is not, it fails silently). Perhaps we should fail loudly?

Jon-edge (approved):

PR #3747, opened 2022-10-24
FlipInputModal2

paullinator:
#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device


src/components/common/ExchangeRate2.tsx
```diff
@@ -0,0 +1,33 @@
+// import { log10 } from 'biggystring'
```
swansontec, line 1: Leftover comment.

src/components/common/ExchangeRate2.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import { useDisplayDenom } from '../../hooks/useDisplayDenom'
+import { getCurrencyCode } from '../../util/CurrencyInfoHelpers'
+import { FiatText } from '../text/FiatText'
+import { EdgeText } from '../themed/EdgeText'
+
+interface Props {
+  wallet: EdgeCurrencyWallet
+  tokenId?: string
+  style?: TextStyle
```
swansontec, line 15: We generally disallow passing style props to components. Text nodes do nest, you can always do `<EdgeText style={...}><ExchangeRate /></EdgeText>` to control font. sizing, colors, &c.

I am open to leaving the style if there is a good reason, though.

src/components/modals/FlipInputModal2.tsx
```diff
@@ -0,0 +1,301 @@
+/* eslint-disable react-native/no-raw-text */
```
swansontec, line 1: Per Sam's suggestion, we generally prefer to put these inline, right next to the affected area.

src/components/modals/FlipInputModal2.tsx
```diff
@@ -0,0 +1,301 @@
...
+import { FiatText } from '../text/FiatText'
+import { EdgeText } from '../themed/EdgeText'
+import { ExchangedFlipInput2, ExchangedFlipInputAmounts, ExchangedFlipInputRef, ExchangeFlipInputFields } from '../themed/ExchangedFlipInput2'
+import { MiniButton } from '../themed/MiniButton'
+import { ThemedModal } from '../themed/ThemedModal'
+
+export type FlipInputModalResult = {
+  nativeAmount: string
+  exchangeAmount: string
+}
```
swansontec, line 34: Nice. However, we should include `{...} | undefined` to cover the case where the user cancels the modal. That way, the user wants "no result" and the return type reflects that. In the modals where we haven't followed this convention, things have tended to get a lot more messy.

src/components/modals/FlipInputModal2.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+export interface SetFeesParams {
+  feeTokenId?: string
+  feeNativeAmount: string
+}
+export interface FlipInputModalRef {
+  setFees: (params: SetFeesParams) => void
+  setAmount: (field: ExchangeFlipInputFields, value: string) => void
+  setError: (errorMessage: string | null) => void
+}
```
swansontec, line 44: Wonderful. 👍

src/components/modals/FlipInputModal2.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+type FeeStyleTypes = 'dangerText' | 'warningText'
+
+type Props = {
+  bridge: AirshipBridge<FlipInputModalResult>
+  walletId: string
```
swansontec, line 50: Please pass the wallet directly, as `wallet: EdgeCurrencyWallet`, instead of by ID.

src/components/modals/FlipInputModal2.tsx
```diff
@@ -0,0 +1,301 @@
...
+  const [errorMessage, setErrorMessage] = useState<string | null>(null)
+  const wallet = currencyWallets[walletId]
+  const { currencyInfo } = wallet
+  const balances = useWatch(wallet, 'balances')
+  const { pluginId } = currencyInfo
+  const currencyCode = getCurrencyCode(wallet, tokenId)
+  const displayDenom = useDisplayDenom(pluginId, currencyCode)
+  const exchangedFlipInputRef = React.useRef<ExchangedFlipInputRef>(null)
+  const feeCurrencyCode = getCurrencyCode(wallet, feeTokenId)
+  const feeDisplayDenom = useDisplayDenom(pluginId, feeCurrencyCode)
```
swansontec, line 95: Optional cleanup: Can we separate the `useState` hooks from the others, which are more subscriptions? That way it's more clear what we depend on vs. what we store.

src/components/modals/FlipInputModal2.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+  const [amounts, setAmounts] = useState<ExchangedFlipInputAmounts>({
+    nativeAmount: '',
+    exchangeAmount: '',
+    fiatAmount: '',
+    fieldChanged: forceField ?? 'fiat'
+  })
+
+  const walletName = wallet?.name ?? s.strings.string_no_name
```
swansontec, line 104: `string_no_name` has been exorcised from the codebase. The new hotness is `useWalletName` / `getWalletName`.

src/components/modals/FlipInputModal2.tsx
```diff
@@ -0,0 +1,301 @@
...
+  const flipInputHeaderText = headerText ?? sprintf(s.strings.send_from_wallet, walletName)
+  const theme = useTheme()
+  const styles = getStyles(theme)
+
+  const handleAmountsChanged = useHandler((amounts: ExchangedFlipInputAmounts) => {
+    setAmounts(amounts)
+
+    if (onAmountsChanged) onAmountsChanged(amounts)
```
swansontec, line 112: Please use `onAmountsChanged != null`. While we don't currently have the strict boolean lint rule switched on, we want to move towards getting it enabled one day.

src/components/modals/FlipInputModal2.tsx
```diff
@@ -0,0 +1,301 @@
...
+      onFeesChange()
+    }
+  })
+
+  const handleCloseModal = useHandler(() => {
+    let { nativeAmount, exchangeAmount, fiatAmount } = amounts
+    nativeAmount = nativeAmount === '' ? '0' : nativeAmount
+    exchangeAmount = exchangeAmount === '' ? '0' : exchangeAmount
+    fiatAmount = fiatAmount === '' ? '0' : fiatAmount
+    bridge.resolve(Promise.resolve({ nativeAmount, exchangeAmount, fiatAmount }))
```
swansontec, line 127: The `Promise.resolve` is not needed here.

src/components/modals/FlipInputModal2.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+  const handleAmountsChanged = useHandler((amounts: ExchangedFlipInputAmounts) => {
+    setAmounts(amounts)
+
+    if (onAmountsChanged) onAmountsChanged(amounts)
+  })
+
+  const handleFeesChange = useHandler(() => {
+    if (onFeesChange != null) {
+      handleCloseModal()
```
swansontec, line 117: Why does editing this close the modal?

src/components/modals/FlipInputModal2.tsx
```diff
@@ -0,0 +1,301 @@
...
+  })
+
+  const renderErrorMessage = useHandler(() => {
+    const opacity = errorMessage == null ? 0 : 1
+    return (
+      <EdgeText numberOfLines={1} style={[styles.exchangeRateErrorText, { opacity }]}>
+        {errorMessage == null ? ' ' : errorMessage.split('\n')[0]}
+      </EdgeText>
+    )
+  })
```
swansontec, line 144: Do not use `useHandler` for render helpers like this. They never get passed anywhere, and therefore we do not need to preserve their `===` identity changes.

src/components/modals/FlipInputModal2.tsx
```diff
@@ -0,0 +1,301 @@
...
+    setFees: (params: SetFeesParams) => {
+      setFeeTokenId(params.feeTokenId)
+      setFeeNativeAmount(params.feeNativeAmount)
+    },
+    setAmount: (field: ExchangeFlipInputFields, value: string) => exchangedFlipInputRef.current?.setAmount(field, value),
+    setError: (errorMessage: string | null) => setErrorMessage(errorMessage)
+  }))
+
+  return (
+    <ThemedModal bridge={bridge} onCancel={handleCloseModal}>
```
swansontec, line 223: `onCancel` should resolve to `undefined`, not an amount.

swansontec (changes requested):
Overall this component looks pretty good. Other than the modal return value, most of the stuff here is pretty minor.

src/components/modals/FlipInputModal2.tsx
paullinator: I use `useHandler` to prevent redeclaring the function on every component function call. My understanding is that this  makes the functions equivalent to class methods which are never redeclared. It's not as big of an optimization as functions that are passed as props, but still something

src/components/common/ExchangeRate2.tsx
paullinator: This was really copy/paste from the original ExchangeRate component. Looks like it's only used for text size so I just moved the style into ExchangeRate2

src/components/modals/FlipInputModal2.tsx
paullinator: Another copy/paste from FlipInputModal. Will fix

src/components/modals/FlipInputModal2.tsx
paullinator: Ok. did my best but many of the lines are dependent on each other forcing a particular order

src/components/modals/FlipInputModal2.tsx
paullinator: This was retained from original FlipInputModal. I wasn't trying to change its UX. But admittedly it may be nicer to not close.

src/components/modals/FlipInputModal2.tsx
paullinator: Again, retained from FlipInputModal but I see why. The modal is actually updating the parent scene on every keystroke, so even if the user cancels, the parent was already being updated. Basically the modal has no cancel functionality since so much happens while the user is typing. ie. recreating the tx, makeSpend, etc. It would be hard to "unwind" all the user input.

src/components/modals/FlipInputModal2.tsx
paullinator: This was a copy/paste from FlipInputModal but good catch. Removing actually exposed a type error!

src/components/modals/FlipInputModal2.tsx
paullinator: done

src/components/modals/FlipInputModal2.tsx
paullinator: done

src/components/modals/FlipInputModal2.tsx
paullinator: done

src/components/common/ExchangeRate2.tsx
paullinator: deleted

src/components/modals/FlipInputModal2.tsx
paullinator: Looking at it closer, this is because `handleFeesChange` is called when the user taps the Fee button. That only returns control to the parent which will launch the Change Fee scene so we need to modal gone to properly render the scene.

src/components/modals/FlipInputModal2.tsx
paullinator: done

src/components/modals/FlipInputModal2.tsx
swansontec: The `useHandler` hook does not prevent that. The function still gets re-declared, but then `useHandler` wraps it in a "bouncer" function that retains it's `===` identity. So you still pay for the function to get re-declared, and then pay again for this bouncer we don't need.

src/components/modals/FlipInputModal2.tsx
swansontec: I see. This seems like a bug in the send scene, which we are now paying for in the modal. I was hoping the send scene refactor would clean up stuff like this.

paullinator: I think it's actually better to do all the tx calculation on the parent scene. To do it in the modal would require the modal to be "smart" which makes it unusable in other scenes that aren't simply doing spends. This was the cause of a lot of hacky redux code. It's not impossible to unwind but quite difficult and I don't think the result would be much better. I think of the FlipInputModal as somewhat like the FlipInput itself.. it's mostly dumb, and the parent controls it. It just happens to be in a modal vs a component and it has more UI to control.

swansontec (approved):

PR #3748, opened 2022-10-24
Add Swapuz

peachbits:
---
  - https://app.asana.com/0/0/1203223435791098

### DIFF
--- a/src/components/services/ContactsLoader.ts
+++ b/src/components/services/ContactsLoader.ts
@@ -160,6 +160,18 @@ const merchantPartners = [
     recordID: ''
   },
   {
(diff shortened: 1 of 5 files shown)


src/components/services/ContactsLoader.ts
```diff
@@ -159,6 +159,18 @@ const merchantPartners = [
     familyName: '',
     recordID: ''
   },
+  {
+    givenName: 'Swapuz',
+    hasThumbnail: true,
```
swansontec, line 164: Is the image on the CMS? If so, good to go.

swansontec (approved):

PR #3750, opened 2022-10-24
Sam/eth domain fix

samholmes:
#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203219179340430

### DIFF
--- a/src/components/modals/AddressModal.tsx
+++ b/src/components/modals/AddressModal.tsx
@@ -1,4 +1,5 @@
 import { EdgeAccount, EdgeCurrencyConfig, EdgeCurrencyWallet } from 'edge-core-js'
+import { ethers } from 'ethers'
 import * as React from 'react'
 import { ActivityIndicator, FlatList, Image, TouchableWithoutFeedback, View } from 'react-native'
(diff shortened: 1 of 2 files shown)


peachbits (changes requested):
Essentially approved I just have the one question about changing the check in the FIO code.

src/components/modals/AddressModal.tsx
```diff
@@ -144,7 +145,7 @@ export class AddressModalComponent extends React.Component<Props, State> {
 
   checkIfEnsDomain = (name: string): boolean => ENS_DOMAINS.some(domain => name.endsWith(domain))
 
-  fetchDomain = async (domain: string, currencyTicker: string): Promise<string> => {
+  fetchUnstoppableDomainAddress = async (domain: string, currencyTicker: string): Promise<string> => {
```
peachbits, line 148: Actually one other thing. Since these are separate functions now you can remove the ENS functionality from this function and leave it as UD only. Lines 150 through 154.

peachbits (changes requested):

peachbits (approved):

PR #3751, opened 2022-10-25
Jon/aave/loan-status-redesign

Jon-edge:
This PR updates the Action display implementation for Loan Create workflows to support the new spec requirement of supporting multiple display string options for the same ActionOp type

Simulator Screen Shot - iPhone 8 ios 16 - 2022-10-24 at 19 49 00

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203187813657970


src/components/scenes/Loans/LoanCreateConfirmationScene.tsx
```diff
@@ -107,7 +107,7 @@ export const LoanCreateConfirmationScene = (props: Props) => {
 
       // Target mainnet native balance should be double the fees estimate to be
       // extra generous when accounting for fee volatility.
-      const feeNativeAmount = sub(mul(mainNetworkFeeAmount, '2'), borrowWalletNativeBalance)
+      const feeNativeAmount = sub(mul(mainNetworkFeeAmount, '1.25'), borrowWalletNativeBalance)
```
samholmes, line 110: Swaps can take a very long time to complete. If we cover 2x we're ensuring the product works even if the fees double during the time we're waiting on swaps. If they don't double, then the user has a bit more ETH left over; so what, that can be used for future contract calls to manage their loan. Much better to overshoot and have extra ETH then to undershoot and fail the program asking the user to recover the broken loan state, don't you think?

samholmes (changes requested):
Looks like what's missing in this change is persisting the new display data to disk. The cleaners haven't been touched.

The change does mix concerns (display data with program data), which still irks me. I'd be much happier with the display data being separate from the program or ActionOp object.

One idea I had (but had not really mentioned because I didn't want to bike-shed), was to simply add a `programIntent` field to the `ActionProgram`. The display implementation can use the `programIntent` to determine how to translate the `(ActionProgram, ActionProgramState)` data into `ActionProgramDsiplayInfo` or any other dynamic type; we can completely re-write the display API at any time. The only think we're committing to as frozen API is a single intent field and even if we deprecate a value for this (or mess up usage), we can always fallback to some generic display (keeping our guarantee that the set of all action programs is included in the domain for the `getActionProgramDisplayInfo` function).

src/components/scenes/Loans/LoanCreateConfirmationScene.tsx
Jon-edge: That's fair, I did not consider the possible length of time. However, if I didn't consider it, another user wouldn't either, and would be understandably confused about this.

src/controllers/action-queue/types.ts
```diff
@@ -94,6 +97,26 @@ export type ActionOp =
...
+// | 'borrow-bank'
+// | 'repay-balance'
+// | 'repay-collateral'
+// | 'close-balance'
+// | 'close-collateral'
+// | 'withdraw'
+
+export type ActionOpDisplay = {
+  displayKey?: ActionOpDisplayKey
+}
```
samholmes, line 116: This can now be removed.

src/controllers/action-queue/types.ts
```diff
@@ -94,6 +97,26 @@ export type ActionOp =
...
+// TODO: implement
+// | 'deposit'
+// | 'deposit-swap'
+// | 'borrow'
+// | 'borrow-bank'
+// | 'repay-balance'
+// | 'repay-collateral'
+// | 'close-balance'
+// | 'close-collateral'
+// | 'withdraw'
```
samholmes, line 112: I'd much rather put this action item in a task rather than a TODO comment.

src/controllers/action-queue/types.ts
```diff
@@ -81,10 +82,12 @@ export type SwapActionOp = {
   nativeAmount: string
   toTokenId?: string
   toWalletId: string
+  displayKey?: ActionOpDisplayKey
```
samholmes, line 85: What if the type for `SwapActionOp['displayKey']` was different from the type for `SeqActionOp['displayKey']` and `ParActionOp['displayKey']`? The reason for this is to treat the keys for a particular action op distinctly from the other action ops. This way a display key for the SwapActionOp is specific to the `SwapActionOp`. This can avoid _some of_ the runtime type error problems we could run into if a key is used on an ActionOp that isn't not suppose to be used on.

Here's an example of what I mean:

```ts
export type SwapActionOp = {
  type: 'swap'
  amountFor: 'from' | 'to'
  fromTokenId?: string
  fromWalletId: string
  nativeAmount: string
  toTokenId?: string
  toWalletId: string
  displayKey?: SwapActionOpDisplayKey
}
export type SwapActionOpDisplayKey = 'swap-deposit' | 'swap-deposit-fees'
```

And here's an example for what we can do with `ParActionOp`:

```ts
export type ParActionOp = {
  type: 'par'
  actions: ActionOp[]
  displayKey?: ParActionOpDisplayKey
}
export type ParActionOpDisplayKey = 'create'
```

Making the set of valid keys specific to the ActionOp avoid the category of error of putting say `swap-deposit` key on ` ParActionOp`. Also it make it very clear to the developer about which keys apply to which `ActionOp`; self-documenting code.

src/components/scenes/Loans/LoanCreateConfirmationScene.tsx
```diff
@@ -107,7 +105,7 @@ export const LoanCreateConfirmationScene = (props: Props) => {
...
```
samholmes, line 108: Can we hold off on this change just because we don't want it impacting our launch. We can create a task to reconsider this post-beta?

src/controllers/action-queue/cleaners.ts
```diff
@@ -63,13 +64,16 @@ const asError = asCodec(
...
+
 const asSeqActionOp = asObject<SeqActionOp>({
   type: asValue('seq'),
   actions: asArray((raw: any) => asActionOp(raw))
 })
 const asParActionOp = asObject<ParActionOp>({
   type: asValue('par'),
-  actions: asArray((raw: any) => asActionOp(raw))
+  actions: asArray((raw: any) => asActionOp(raw)),
+  displayKey: asOptional(asActionOpDisplayKey)
```
samholmes, line 76: Just a trick to use for if we were to have distinct `*ActionOpDisplayKey` types (my final ask in this review):

```ts
const asParActionOp = asObject<ParActionOp>({
  type: asValue('par'),
  actions: asArray((raw: any) => asActionOp(raw)),
  displayKey: (raw: unknown) => asOptional(asParActionOpDisplayKey)(raw)
})
const asParActionOpDisplayKey = asValue('create')
```

This extra anonymous arrow-function is how we can reference a cleaner before it's defined.

src/controllers/action-queue/ActionProgram.ts
```diff
@@ -1,10 +1,19 @@
...
 
   return {
     programId,
-    actionOp
+    actionOp,
+    completeMessage: completeMessage ?? {
+      title: s.strings.action_display_title_complete_default,
+      message: s.strings.action_display_message_complete_default
+    },
+    mockMode: ENV.ACTION_QUEUE.mockMode
```
samholmes, line 17: This field isn't necessary here because it's injected by the ActionQueueStore

src/controllers/action-queue/display.ts
```diff
@@ -11,7 +22,17 @@ import { checkEffectIsDone } from './util/checkEffectIsDone'
 import { getEffectErrors } from './util/getEffectErrors'
 
 export async function getActionProgramDisplayInfo(account: EdgeAccount, program: ActionProgram, programState: ActionProgramState): Promise<ActionDisplayInfo> {
-  return await getActionOpDisplayInfo(account, program.actionOp, programState.effect)
+  // Assume that ActionPrograms always have a NodeActionOp at top level, for purposes of retreiving title and message as the completion strings
+  const programType = program.actionOp.type
+  if (programType !== 'seq' && programType !== 'par') throw new Error('getActionProgramDisplayInfo only supports NodeActionOps as root program type')
```
samholmes, line 27: Can't we support `LeafActionOp` types with the pure inference method?

samholmes (changes requested):
Some minor changes, and one main ask. Feel free to rebase-autosquash the branch so far.

src/controllers/action-queue/display.ts
Jon-edge: Ah yeah that was the point in the first place, good catch

src/controllers/action-queue/ActionProgram.ts
Jon-edge: I like the redundancy regardless, for debugging prior to the ActionQueueStore or if implementation changes

src/controllers/action-queue/display.ts
Jon-edge: Actually I did try that. Right now, we can't go by inference because getActionOpDisplayInfo drills down into the leaves. When we fully isolate the display implementation we can rely more on static typing

Jon-edge: Actually this check isn't needed if I instead restrict the ActionProgram's root actionOp type as a NodeActionOp

samholmes: I don't see why we would ever have a `LeafActionOp` in an action program... This is just one action which doesn't require the action queue, and even if we did want a program with one action for whatever reason, adding the `NodeActionOp` could just be required overhead. I'm okay with making this change as a breakaway task.

src/controllers/action-queue/types.ts
```diff
@@ -173,7 +161,7 @@ export type ActionProgramCompleteMessage = { title: string; message: string }
 
 export type ActionProgram = {
   programId: string
-  actionOp: ActionOp
+  actionOp: NodeActionOp
```
samholmes, line 164: Oh you did it. Awesome; cool cool

samholmes: Erm, needs revert cause it cascades type changes for the rest of the runtime implementation. Let's postpone this work item.

samholmes (approved):
Great job here! Approved.

Note, autosquash needed and perhaps double check this all works before merging.

PR #3753, opened 2022-10-26
Refactor Details Scene Loan Action cards

Jon-edge:
Refactor the page-long inlined JSX into a function to avoid confusion when updating conditions.

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202949203074528

### DIFF
--- a/src/components/scenes/Loans/LoanDetailsScene.tsx
+++ b/src/components/scenes/Loans/LoanDetailsScene.tsx
@@ -61,17 +61,10 @@ export const LoanDetailsScene = (props: Props) => {
 
   const collaterals = useWatch(borrowEngine, 'collaterals')
   const debts = useWatch(borrowEngine, 'debts')
-  const openCollaterals = React.useMemo(() => collaterals.filter(collateral => !zeroString(collateral.nativeAmount)), [collaterals])
-  const isCollateralAvailable = openCollaterals.length > 0
-
-  const openDebts = React.useMemo(() => debts.filter(debt => !zeroString(debt.nativeAmount)), [debts])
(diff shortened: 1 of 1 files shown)


src/components/scenes/Loans/LoanDetailsScene.tsx
```diff
@@ -139,6 +117,90 @@ export const LoanDetailsScene = (props: Props) => {
     } else return null
   }
 
+  // #region Loan Action Cards
+
+  const collateralPositions = React.useMemo(() => collaterals.filter(collateral => !zeroString(collateral.nativeAmount)), [collaterals])
+  const debtPositions = React.useMemo(() => debts.filter(debt => !zeroString(debt.nativeAmount)), [debts])
+
+  const actionCards = React.useMemo(() => {
```
samholmes, line 125: It seems appropriate to use a render function here: `renderActionCards` instead of memoizing React elements.

I think as soon as you have this pattern:

```tsx
const foo = useMemo(() => <JSX />, [a, b, c])
```

Then you really should just have a new component:

```tsx
const Foo = ({ a, b, c}) => {
  return <JSX />
}
```

React will only render if the a, b, and c props change, just like the a, b, and c deps in the hook.

But with all that said, we have this pattern of "render functions" which are really just impure-components and get around the overhead of creating a new component and file for the component. Maybe we should consider "local components" which can exist in the same file as the component which uses it, and it isn't exported from the file. I'd prefer that pattern over "render functions" or memoized JSX.

samholmes (approved):
Approve with a some nit-picky feedback 🙃

PR #3756, opened 2022-10-26
Allow new user to select initial wallets

peachbits:
Reuse the CreateWalletSelectCryptoScene component after the account is created and before the wallet list loads for the first time.

---
  - https://app.asana.com/0/0/1203152643001448

### DIFF
--- a/src/actions/LoginActions.tsx
+++ b/src/actions/LoginActions.tsx
@@ -56,8 +56,8 @@ export function initializeAccount(account: EdgeAccount, touchIdInfo: GuiTouchIdI
   return async (dispatch, getState) => {
     // Log in as quickly as possible, but we do need the sort order:
     const syncedSettings = await getSyncedSettings(account)
-    const { walletSort } = syncedSettings
-    dispatch({ type: 'LOGIN', data: { account, walletSort } })
+    const { walletsSort } = syncedSettings
+    dispatch({ type: 'LOGIN', data: { account, walletSort: walletsSort } })
     Actions.push('edge', {})
 
     // Show a notice for deprecated electrum server settings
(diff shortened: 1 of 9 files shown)


peachbits:
https://user-images.githubusercontent.com/41451710/198151943-e30cf252-c73c-4e4e-83b4-d50ead8be1cd.mp4

src/actions/LoginActions.tsx
```diff
@@ -305,45 +295,67 @@ async function safeCreateWallet(account: EdgeAccount, walletType: string, wallet
...
 
-/**
- * Creates the custom default wallets inside a new account.
- * The `currencyCodes` are in the format "ETH:DAI",
- * so we may need to enable tokens on some of the created wallets.
- */
-async function createCustomWallets(account: EdgeAccount, fiatCurrencyCode: string, currencyCodes: string[], dispatch: Dispatch) {
-  const currencyInfos: EdgeCurrencyInfo[] = []
+// The `currencyCodes` are in the format "ETH:DAI",
+const currencyCodesToEdgeTokenIds = (account: EdgeAccount, currencyCodes: string[]): EdgeTokenId[] => {
```
swansontec, line 293: We already have a `upgradeCurrencyCodes` helper in WalletListModal which is extremely similar to this, but not exactly the same. I wonder if we should make a `src/util/currencyCodes.ts` file to gather these conversion routines together into one place, and document the differences.

Your actual `currencyCodesToEdgeTokenIds` implementation seems great - I just don't think this is the right place for it.

src/types/reduxActions.ts
```diff
@@ -172,7 +173,7 @@ export type Action =
   | { type: 'UI/WALLETS/UPSERT_WALLETS'; data: { wallets: EdgeCurrencyWallet[] } }
   | { type: 'UPDATE_SORTED_WALLET_LIST'; data: WalletListItem[] }
   | { type: 'UPDATE_SWAP_QUOTE'; data: GuiSwapInfo }
-  | { type: 'UPDATE_SHOW_PASSWORD_RECOVERY_REMINDER_MODAL'; data: number }
+  | { type: 'UPDATE_SHOW_PASSWORD_RECOVERY_REMINDER_MODAL'; data: keyof typeof PASSWORD_RECOVERY_REMINDERS_SHOWN }
```
swansontec, line 176: Maybe we could export `type PasswordReminderTime =   keyof typeof PASSWORD_RECOVERY_REMINDERS_SHOWN` or something like that, so we don't have to import the actual table, just the type information.

swansontec (approved):
Approved with optional fixes.

src/actions/LoginActions.tsx
peachbits: I thought there was but didn't look hard enough to find it I guess. I like the idea of consolidating but will save it for a future task

PR #3757, opened 2022-10-27
Update Scam Warning

Jon-edge:
If the SendScene is opened with a pre-populated address,

1. Either from swiping the wallet list row OR
2. From the ControlPanel (side menu)

Then, show a scam warning in a ConfirmAndContinue modal once.

Subsequent initializations of the SendScene will then never show the scam warning modal.

Simulator Screen Shot - iPhone 8 Plus - 2022-10-26 at 17 34 13

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203082480130376

### DIFF
--- a/src/actions/ScamWarningActions.tsx
+++ b/src/actions/ScamWarningActions.tsx
@@ -1,0 +1,32 @@
+import { Disklet } from 'disklet'
+import * as React from 'react'
+
+import { ConfirmContinueModal } from '../components/modals/ConfirmContinueModal'
+import { Airship } from '../components/services/AirshipInstance'
+import { ModalMessage } from '../components/themed/ModalParts'
+import { SCAM_WARNING } from '../constants/constantSettings'
+import s from '../locales/strings'
+
+let isWarningChecked = false
+
+export const dismissScamWarning = async (disklet: Disklet) => {
+  if (isWarningChecked) return
+
+  try {
+    await disklet.getText(SCAM_WARNING)
+  } catch (error: any) {
+    Airship.show<boolean>(bridge => {
+      const warningMessage = `\u2022 ${s.strings.warning_scam_message_financial_advice}\n\n\u2022 ${s.strings.warning_scam_message_irreversibility}\n\n\u2022 ${s.strings.warning_scam_message_unknown_recipients}`
+
+      return (
+        <ConfirmContinueModal bridge={bridge} title={s.strings.warning_scam_title}>
+          <ModalMessage isWarning>{warningMessage}</ModalMessage>
+        </ConfirmContinueModal>
+      )
+    }).then(async () => {
+      await disklet.setText(SCAM_WARNING, '')
+
+      isWarningChecked = true
+    })
+  }
+}
--- a/src/components/cards/WarningCard.tsx
+++ b/src/components/cards/WarningCard.tsx
@@ -39,7 +39,7 @@ export function WarningCard({ title, header, points, footer, marginRem, paddingR
 
   const renderBulletpoint = (message: string) => {
     return (
-      <View style={styles.bulletpointRow}>
+      <View style={styles.bulletpointRow} key={message}>
         <EdgeText style={styles.bulletpointText}>{'\u2022 '}</EdgeText>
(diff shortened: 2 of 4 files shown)


src/components/scenes/SendScene.tsx
```diff
@@ -151,14 +153,23 @@ class SendComponent extends React.PureComponent<Props, State> {
     }
   }
 
-  updateSendConfirmationTx = (guiMakeSpendInfo: GuiMakeSpendInfo) => {
+  updateSendConfirmationTx = (guiMakeSpendInfo: GuiMakeSpendInfo, isCheckScamWarning?: boolean) => {
```
peachbits, line 156: Nitpick: Since you're calling dismissScamWarning on component mount then it should be in componentDidMount. updateSendConfirmationTx won't ever be called with this additional isCheckScamWarning argument anywhere else and it seems the only reason to add to updateSendConfirmationTx was for the convenience of using the existing ugly block of recipientAddress identifying code. 

I would hate to see that awful ternary repeated again so a compromise would be to rename isCheckScamWarning to reflect its actual use as a one time on-mount flag.

src/actions/ScamWarningActions.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import { ConfirmContinueModal } from '../components/modals/ConfirmContinueModal'
+import { Airship } from '../components/services/AirshipInstance'
+import { ModalMessage } from '../components/themed/ModalParts'
+import { SCAM_WARNING } from '../constants/constantSettings'
+import s from '../locales/strings'
+
+export const dismissScamWarning = async (disklet: Disklet) => {
+  try {
+    await disklet.getText(SCAM_WARNING)
```
peachbits, line 12: Optional mini optimization: disk reads are fast enough already but you can limit reads to one time per app instance by using a local boolean. 

```
let warningChecked = false
export const dismissScamWarning = async (disklet: Disklet) => {
  if (warningChecked) return
  try {
    await disklet.getText(SCAM_WARNING)
  } catch (error: any) {
    Airship.show<boolean>(bridge => {
      const warningMessage = `\u2022 ${s.strings.warning_scam_message_financial_advice}\n\n\u2022 ${s.strings.warning_scam_message_irreversibility}\n\n\u2022 ${s.strings.warning_scam_message_unknown_recipients}`
      return (
        <ConfirmContinueModal bridge={bridge} title={s.strings.warning_scam_title}>
          <ModalMessage isWarning>{warningMessage}</ModalMessage>
        </ConfirmContinueModal>
      )
    }).then(async () => {
      await disklet.setText(SCAM_WARNING, '')
      warningChecked = true
    })
  }
}
```

peachbits (changes requested):

src/components/scenes/SendScene.tsx
Jon-edge: I was trying to avoid re-checking the recipientAddress but I can move things around

Jon-edge: Yeah initially I tried to factor out that ternary that was parsing the address so I could do as you described, and found that it was similar to handleAddressFromScan but not quite (missing the check for the GuiMakeSpendInfo publicAddress). Ugly, so I went with what I thought was the lesser of the evils.

I'll just rename

src/components/cards/WarningCard.tsx
```diff
@@ -39,7 +39,7 @@ export function WarningCard({ title, header, points, footer, marginRem, paddingR
 
   const renderBulletpoint = (message: string, index: number) => {
     return (
-      <View style={styles.bulletpointRow} key={index}>
+      <View style={styles.bulletpointRow} key={`${index}: ${message}`}>
```
peachbits, line 42: Goal is to avoid using the index at all. The key should unique and stable and array indexes are not stable. In this particular use case it is stable but it's an anti-pattern and we should avoid doing it.

peachbits (changes requested):

peachbits (approved):

PR #3759, opened 2022-10-27
Logout crash fixes

peachbits:
---
  - https://app.asana.com/0/0/1203218445109629

### DIFF
--- a/src/components/scenes/TransactionListScene.tsx
+++ b/src/components/scenes/TransactionListScene.tsx
@@ -236,7 +236,7 @@ export const TransactionList = connect<StateProps, DispatchProps, OwnProps>(
     // getTransactions
     const { currencyWallets } = state.core.account
     const currencyWallet = currencyWallets[walletId]
-    const tokenId = getTokenId(state.core.account, currencyWallet.currencyInfo.pluginId, currencyCode)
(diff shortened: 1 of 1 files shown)


src/components/scenes/TransactionDetailsScene.tsx
```diff
@@ -580,7 +580,13 @@ const getStyles = cacheStyles((theme: Theme) => ({
 }))
 
 export const TransactionDetailsScene = connect<StateProps, DispatchProps, OwnProps>(
+  //  @ts-expect-error
   (state, { route: { params } }) => {
+    if (Object.keys(state.core.account.currencyWallets).length === 0) {
+      // HACK: We've logged out and there are no wallets. This is a patch fix until we can address this appropriately.
+      return
+    }
+
```
swansontec, line 589: Thanks, but we don't need this with the scene rewrite (which will make it in time for the release deadline now that the blockers are solved).

swansontec (changes requested):

src/components/scenes/TransactionDetailsScene.tsx
peachbits: Sweet I'll remove it then

swansontec (approved):

PR #3762, opened 2022-10-28
Use git to get version file

paullinator:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203256232564054

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -39,6 +39,13 @@ pipeline {
 
     stage ("Get secret files") {
       steps {
+        // Import the settings files
+        withCredentials([
+          file(credentialsId: "githubSshKey", variable: "id_github"),
+        ]) {
+          sh "cp ${id_github} ./id_github"
+        }
+
         sh "node -r sucrase/register ./scripts/secretFiles.ts ${BRANCH_NAME} ${SECRET_FILES}"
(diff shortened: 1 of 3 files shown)


swansontec (changes requested):
Now seeing this change, I think we need to have a sidebar. I was under the impression that we would leave the build scripts as-is, but simply fetch the *existing* `release-version.json` as part of the Jenkinsfile, and then push it back up at completion time. Basically, using git instead of the built-in `copyArtifacts` and `archiveArtifacts` steps.

This is important, since we want people who are not us to be able to use the build scripts. Partners should be able to assemble an app-store release from the command-line without reference to any of our infrastructure.

For people not using Jenkins, the instructions should be "make sure these files exist", including the API keys, version file, and so forth. For us, we use a Jenkins-aware script to perform that step, rather than manually writing files.

If we just want to fix the duplicate build numbers in a hurry, I suggest moving the `archiveArtifacts` line right after the `// Pick the new build number and version:` block in the Jenkins file. That will immediately fix the problems related to duplicate build numbers in our single-machine setting. Then we can discuss these other goals with less urgency.

swansontec (approved):

PR #3778, opened 2022-11-03
Fix FIO-specific chain code handling

peachbits:
FIO uses BSC for Binance smart chain. Edge uses BNB so we need to convert back and forth to make them compatible.

This was technically started some time ago to make connected wallets work but these other features didn't really work and displayed data was messed up.

---
  - https://app.asana.com/0/0/1202301192532726

### DIFF
--- a/src/components/scenes/FioRequestListScene.tsx
+++ b/src/components/scenes/FioRequestListScene.tsx
@@ -5,6 +5,7 @@ import { ActivityIndicator, SectionList, View } from 'react-native'
 import { sprintf } from 'sprintf-js'
 
 import { refreshAllFioAddresses } from '../../actions/FioAddressActions'
+import { SPECIAL_CURRENCY_INFO } from '../../constants/WalletAndCurrencyConstants'
 import { formatDate } from '../../locales/intl'
 import s from '../../locales/strings'
 import { addToFioAddressCache, cancelFioRequest, convertFIOToEdgeCodes, FIO_NO_BUNDLED_ERR_CODE } from '../../modules/FioAddress/util'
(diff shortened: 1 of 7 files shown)


src/components/themed/FioRequestRow.tsx
```diff
@@ -199,17 +200,20 @@ export const FioRequestRow = connect<StateProps, {}, OwnProps>(
         fiatAmount: ''
       }
     }
-    const tokenCode = fioRequest.content.token_code.toUpperCase()
+    let tokenCode = fioRequest.content.token_code.toUpperCase()
     try {
       const { allCurrencyInfos } = state.ui.settings.plugins
-      const plugin = allCurrencyInfos.find(plugin => {
-        const { currencyCode: pluginCurrencyCode } = plugin
+      const currencyInfo = allCurrencyInfos.find(info => {
```
swansontec, line 206: We have been trying to move away from `allCurrencyInfos`, since it's just a copy of what the core already has. Instead, please use `Object.keys(account.currencyConfig).find(pluginId => {...})` to directly iterate over the currencies in the account.

There are only two places left that still use `allCurrencyInfos`, so soon we can finally delete it.

src/components/scenes/FioRequestListScene.tsx
```diff
@@ -352,14 +354,18 @@ class FioRequestList extends React.Component<Props, LocalState> {
...
   renderDropUp = async (selectedFioPendingRequest: FioRequest) => {
-    const { onSelectWallet } = this.props
+    const { account, onSelectWallet } = this.props
     const { content } = selectedFioPendingRequest
-    const chainCode = content.chain_code.toUpperCase()
-    const tokenCode = content.token_code.toUpperCase()
-    const allowedFullCurrencyCode: string[] = chainCode !== tokenCode && tokenCode && tokenCode !== '' ? [`${chainCode}-${tokenCode}`] : [chainCode]
+    const pluginId = Object.keys(SPECIAL_CURRENCY_INFO).find(
+      pluginId => SPECIAL_CURRENCY_INFO[pluginId].fioChainCode ?? SPECIAL_CURRENCY_INFO[pluginId].chainCode === content.chain_code.toUpperCase()
+    ) as string
```
swansontec, line 361: This might return `undefined`, which is why you have an `as string` on the end. However, I think it might be better to do some sort of `if (pluginId == null) showError("Oy, mate! We don't got this one.")`

swansontec (changes requested):
This looks good, but I did find two small-ish things that deserve to be fixed.

src/components/scenes/FioRequestListScene.tsx
peachbits: Oof yeah good catch

PR #3782, opened 2022-11-07
Add payoutNativeAmount to the 'swap' action

Jon-edge:
Support a manual payoutNativeAmount that will cause the swap op to be effective.

#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203305657654497

### DIFF
--- a/src/controllers/action-queue/types.ts
+++ b/src/controllers/action-queue/types.ts
@@ -79,10 +79,16 @@ export interface SwapActionOp {
   fromTokenId?: string
   fromWalletId: string
   nativeAmount: string
-  payoutNativeAmount?: string
   toTokenId?: string
   toWalletId: string
   displayKey?: SwapActionOpDisplayKey
+
+  // Optional parameter for special handling, for example when swap providers
+  // are not giving accurate quotes or we want to control the payout trigger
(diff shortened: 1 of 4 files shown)


samholmes (changes requested):
I don't understand this change. it may be that I don't understand the problem being addressed and/or how it translates into this change as the solution. I think a discussion would help.

If I understand correctly, the problem with the BTC-MATIC swap is with a discrepancy the MATIC amount in effect versus on-chain. I'd imagine the change would address the discrepancy or account for discrepancy within the implementation. I'm puzzled about why the API was changed.

src/controllers/action-queue/types.ts
```diff
@@ -79,6 +79,7 @@ export interface SwapActionOp {
   fromTokenId?: string
   fromWalletId: string
   nativeAmount: string
+  payoutNativeAmount?: string
```
samholmes, line 82: Rename to `expectedPayoutNativeAmount`, move it to the end of the type's fields and add a comment explaining to the user when they should use this field.

samholmes (approved):
Approved, but make the minor nit change.

PR #3783, opened 2022-11-07
Fix asset logic on LoanCreateScene

Jon-edge:
In the use case described by the bug report, the user does not have USDC enabled. After finalizing all input fields in the LoanCreateScene, useTokenDisplayData returns 0. This causes a divide by 0 crash.

Additionally, this part of the implementation was operating on the assumption that the borrowEngineWallet will always == collateral source wallet, which is no longer the case with the support of native BTC collateral.

Fix the asset logic around this area to resolve the two points mentioned.

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203313944862165

### DIFF
--- a/src/components/scenes/Loans/LoanCreateScene.tsx
+++ b/src/components/scenes/Loans/LoanCreateScene.tsx
@@ -12,7 +12,7 @@ import { makeWyreClient, PaymentMethod } from '../../../controllers/action-queue
 import { useAllTokens } from '../../../hooks/useAllTokens'
 import { useAsyncEffect } from '../../../hooks/useAsyncEffect'
 import { useHandler } from '../../../hooks/useHandler'
-import { useTokenDisplayData } from '../../../hooks/useTokenDisplayData'
+import { useCurrencyFiatRate } from '../../../hooks/useTokenDisplayData'
 import { useWalletBalance } from '../../../hooks/useWalletBalance'
 import { useWalletName } from '../../../hooks/useWalletName'
 import { useWatch } from '../../../hooks/useWatch'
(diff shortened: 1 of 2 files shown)


src/components/scenes/Loans/LoanCreateScene.tsx
```diff
@@ -181,10 +181,11 @@ export const LoanCreateScene = (props: Props) => {
   // Convert required collateral in fiat -> required collateral in crypto
   const { assetToFiatRate: collateralToFiatRate } = useTokenDisplayData({
     tokenId: collateralTokenId,
-    wallet: borrowEngineWallet
+    wallet: srcWallet ?? borrowEngineWallet // borrowEngineWallet given just so hook doesn't puke
```
peachbits, line 184: I think replacing the usage of `useTokenDisplayData` with something that doesn't depend on a wallet would be more appropriate. You're only actually using it to grab an exchange rate so I think this would work:

const collateralToFiatRate = useSelector(state => state.exchangeRates[`${hardCollateralCurrencyCode}_${isoFiatCurrencyCode}`]) ?? '0'

peachbits (changes requested):

src/components/scenes/Loans/LoanCreateScene.tsx
Jon-edge: IsoFiatCurrencyCode does come from the wallet. 
We had duplication of this sort of thing all over the place and everyone was doing it differently - that was the drive of the initial refactor of all that into useTokenDisplayData. Even though it's just the exchange rate, the motivation of creating this hook is to stop doing exactly that in the scenes and components.

peachbits: I get the appeal of removing duplicate code (and ensuring it's always done the same)but the helper function isn't appropriate for the situation it's used where the wallet could be null. I think it's better to use safe defaults rather than useless replacement data which could negatively impact something further down the line. This would work and matches most of the srcWallet declarations above it:

const srcWalletFiat = srcWallet != null ? getWalletFiat(srcWallet).isoFiatCurrencyCode : ''
const collateralToFiatRate = useSelector(state => state.exchangeRates[${hardCollateralCurrencyCode}_${isoFiatCurrencyCode}]) ?? '0'

Any calculation further down would use a safer '0' instead of a bogus exchange rate returned by using the the borrowEngineWallet fiat.

peachbits (changes requested):

peachbits (approved):

PR #3790, opened 2022-11-09
Add Push Notification for Liquidation Threshold

samholmes:
#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203274072103450


src/components/services/LoanManagerService.ts
```diff
@@ -27,18 +54,95 @@ export const LoanManagerService = () => {
...
+          const getCurrencyCode = (tokenId?: string): string => {
+            const tokens = allTokens[borrowEngine.currencyWallet.currencyInfo.pluginId]
+            const token = tokenId != null ? tokens[tokenId] : undefined
+            return token?.currencyCode ?? borrowEngine.currencyWallet.currencyInfo.currencyCode
+          }
+
+          const debtFiatAmounts = filteredDebts.map(debt => {
+            const debtCurrencyCode = getCurrencyCode(debt.tokenId)
+            const debtExchangeAmount = getExchangeAmount(debt.tokenId, debt.nativeAmount)
+            return convertCurrency(state, debtCurrencyCode, 'iso:USD', debtExchangeAmount) //, 2
```
paullinator, line 101: Use `convertCurrencyFromExchangeRates` and subscribe to `state.exchangeRates` instead of `state` on line 39

paullinator: Do you really even need to get the exchange rate. Might as well just assume 1-1 with USD

src/components/services/LoanManagerService.ts
```diff
@@ -27,18 +54,95 @@ export const LoanManagerService = () => {
...
+          })
+          const debtFiatTotal = debtFiatAmounts.reduce((sum, debtFiatAmount) => add(sum, debtFiatAmount), '0')
+
+          const collateral = filteredCollaterals[0]
+          const collateralCurrencyCode = getCurrencyCode(collateral.tokenId)
+          const collateralExchangeAmount = getExchangeAmount(collateral.tokenId, collateral.nativeAmount)
+
+          const numerator = (ALERT_THRESHOLD_LTV * 100).toString()
+          const thresholdRate = parseFloat(div(mul(debtFiatTotal, '100'), mul(collateralExchangeAmount, numerator), DECIMAL_PRECISION))
```
paullinator, line 110: Remove '100' and in numerator

src/components/services/LoanManagerService.ts
```diff
@@ -27,18 +54,95 @@ export const LoanManagerService = () => {
...
+          const pushMessage: PushMessage = {
+            title: s.strings.loan_notification_ltv_threshold_alert_title,
+            body: s.strings.loan_notification_ltv_threshold_alert_message
+          }
+          const newPushEvent: NewPushEvent = { eventId, trigger, pushMessage }
+          const loginUpdatePayload: LoginUpdatePayload = {
+            createEvents: [newPushEvent]
+          }
+
+          clog({ thresholdRate, loginUpdatePayload })
```
paullinator, line 129: remove log

paullinator (changes requested):

src/components/services/LoanManagerService.ts
samholmes: I think because we have to assume an exchange rate to USD on the push-server that the correct thing is to remove the stable-coin exchange rate calculation like you said. Our assumption is already being made on the push-server and changing this assumption locally wont change that.

src/components/services/LoanManagerService.ts
```diff
@@ -105,8 +105,7 @@ export const LoanManagerService = (props: Props) => {
           const collateralCurrencyCode = getCurrencyCode(collateral.tokenId)
           const collateralExchangeAmount = getExchangeAmount(collateral.tokenId, collateral.nativeAmount)
 
-          const numerator = (ALERT_THRESHOLD_LTV * 100).toString()
-          const thresholdRate = parseFloat(div(mul(debtFiatTotal, '100'), mul(collateralExchangeAmount, numerator), DECIMAL_PRECISION))
+          const thresholdRate = parseFloat(div(debtFiatTotal, mul(collateralExchangeAmount, ALERT_THRESHOLD_LTV.toString()), DECIMAL_PRECISION))
```
paullinator, line 111: `ALERT_THRESHOLD_LTV` is already a string. no need to `toString`

src/components/services/LoanManagerService.ts
```diff
@@ -24,7 +24,11 @@ const USD_BASED_TOKEN_IDS: string[] = [
   // USDC on Ethereum
   'a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48',
   // USDC on Polygon
-  '2791Bca1f2de4661ED88A30C99A7a9449Aa84174'
+  '2791Bca1f2de4661ED88A30C99A7a9449Aa84174',
+  // USDT on Polygon
+  '0xdAC17F958D2ee523a2206206994597C13D831ec7',
+  // DAI on Polygon
+  '0xdAC17F958D2ee523a2206206994597C13D831ec7'
```
paullinator, line 31: wrong contract for DAI. should be 

`0x6B175474E89094C44Da98b954EedeAC495271d0F`

paullinator (changes requested):

samholmes:
Approved OTS yesterday at the office

PR #3802, opened 2022-11-11
Fix crash from non-numeric ltvValue

Jon-edge:
#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203348627056673

### DIFF
--- a/src/components/tiles/LtvRatioTile.tsx
+++ b/src/components/tiles/LtvRatioTile.tsx
@@ -66,5 +66,5 @@ export const LtvRatioTile = (props: {
 }
 
 export const getLtvColorValue = (ltvValue: string, theme: Theme): string => {
-  return gt(ltvValue, '0.7') ? theme.dangerText : gt(ltvValue, '0.6') ? theme.warningText : theme.positiveText
+  return gt(ltvValue, '0.7') ? theme.dangerText : gt(ltvValue, '0.6') ? theme.warningText : gt(ltvValue, '0') ? theme.positiveText : theme.deactivatedText
 }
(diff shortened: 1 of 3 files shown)


src/components/tiles/LtvRatioTile.tsx
```diff
@@ -66,5 +66,10 @@ export const LtvRatioTile = (props: {
 }
 
 export const getLtvColorValue = (ltvValue: string, theme: Theme): string => {
-  return gt(ltvValue, '0.7') ? theme.dangerText : gt(ltvValue, '0.6') ? theme.warningText : theme.positiveText
+  try {
+    return gt(ltvValue, '0.7') ? theme.dangerText : gt(ltvValue, '0.6') ? theme.warningText : theme.positiveText
```
samholmes, line 70: Why is it possible `ltvValue` can be `NaN`? Seems like the right solution is to fix the data being passed into the function rather than handling invalid data types.

samholmes (approved):
This fixes the crash, but not the issue. I approve it because it does the fix the crash which is the issue.

peachbits (approved):
Approved, assuming the discussed changes to BorrowEngineFactory.ts line 87 at added

src/components/tiles/LtvRatioTile.tsx
Jon-edge: Due to dividing by 0 with numbers in BorrowEngineFactory. Fix has been updated to fix the data while still retaining a separate display value/color for fresh loans where LTV cannot yet be calculated.

PR #3803, opened 2022-11-11
Jon/aave/welcome-message

Jon-edge:
image

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203327829237131

### DIFF
--- a/src/actions/LoanWelcomeActions.tsx
+++ b/src/actions/LoanWelcomeActions.tsx
@@ -1,0 +1,12 @@
+import { Disklet } from 'disklet'
+
+import { AAVE_WELCOME } from '../constants/constantSettings'
+
+export const isShowLoanWelcomeModal = async (disklet: Disklet) => {
+  try {
+    await disklet.getText(AAVE_WELCOME)
+  } catch (error: any) {
+    await disklet.setText(AAVE_WELCOME, '')
+    return true
+  }
+}
--- a/src/components/modals/LoanWelcomeModal.tsx
+++ b/src/components/modals/LoanWelcomeModal.tsx
@@ -1,0 +1,43 @@
+import * as React from 'react'
+import { AirshipBridge } from 'react-native-airship'
+import FastImage from 'react-native-fast-image'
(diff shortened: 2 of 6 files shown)


src/components/modals/WalletListModal.tsx
```diff
@@ -157,7 +157,7 @@ export function WalletListModal(props: Props) {
         title={s.strings.deposit_to_bank}
         message={sprintf(s.strings.wallet_list_modal_confirm_s_bank_withdrawal, config.appName)}
         buttons={{
-          continue: { label: s.strings.legacy_address_modal_continue }
+          continue: { label: s.strings.continue_fragment }
```
paullinator, line 160: Do not rename string keys even if they seem inappropriate. This will cause a re-translation to be needed.

src/actions/AaveWelcomeActions.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { Airship } from '../components/services/AirshipInstance'
+import { EdgeText } from '../components/themed/EdgeText'
+import { AAVE_WELCOME } from '../constants/constantSettings'
+import s from '../locales/strings'
+import { borrowPlugins } from '../plugins/helpers/borrowPluginHelpers'
+import { config } from '../theme/appConfig'
+import { Theme } from '../types/Theme'
+import { getBorrowPluginIconUri } from '../util/CdnUris'
+
+export const showAaveWelcomeModal = async (disklet: Disklet, theme: Theme) => {
```
paullinator, line 18: Our coding convention is to make the file the modal component which is launched by the scene. Move the logic for when the modal is launched into the scene within the useEffect. Then launch the modal via Airship.show() inside. This file should be called LoanWelcomeModal.tsx much like our other modals. This way the LoanWelcomeModal is an actual component which can access useTheme() on it's own and not needing it passed in.

paullinator (changes requested):

paullinator (approved):

PR #3807, opened 2022-11-16
AAVE: Asset Selection Filter Fixes

Jon-edge:
Fix filtering rules for wallet pickers on various AAVE scenes.

#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203358690101909

### DIFF
--- a/src/components/scenes/Loans/LoanManageScene.tsx
+++ b/src/components/scenes/Loans/LoanManageScene.tsx
@@ -14,6 +14,7 @@ import { ActionOp, ActionProgram } from '../../../controllers/action-queue/types
 import { makeInitialProgramState } from '../../../controllers/action-queue/util/makeInitialProgramState'
 import { makeWyreClient, PaymentMethodsMap } from '../../../controllers/action-queue/WyreClient'
 import { runLoanActionProgram } from '../../../controllers/loan-manager/redux/actions'
+import { LoanProgramType } from '../../../controllers/loan-manager/store'
 import { useAsyncEffect } from '../../../hooks/useAsyncEffect'
 import { useAsyncValue } from '../../../hooks/useAsyncValue'
 import { useExecutionContext } from '../../../hooks/useExecutionContext'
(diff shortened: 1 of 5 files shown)


src/components/scenes/Loans/LoanManageScene.tsx
```diff
@@ -122,7 +122,8 @@ export const LoanManageScene = (props: Props) => {
     [account, borrowEnginePluginId]
   )
   const excludeWalletIds = Object.keys(wallets).filter(walletId => walletId !== borrowEngineWallet.id)
```
samholmes, line 124: Use the new fancy function

src/components/scenes/Loans/LoanManageScene.tsx
```diff
@@ -122,7 +122,8 @@ export const LoanManageScene = (props: Props) => {
...
-  const hardAllowedCollateralAsset = [{ pluginId: borrowEnginePluginId, tokenId: hardCollateralAddr }, { pluginId: 'bitcoin' }]
+  const hardAllowedCollateralAssets = [{ pluginId: borrowEnginePluginId, tokenId: hardCollateralAddr }]
+  if (actionOpType === 'loan-deposit') hardAllowedCollateralAssets.push({ pluginId: 'bitcoin', tokenId: undefined })
```
samholmes, line 126: Rename actionOpType to `userAction` and the type for this value `LoanManageUserAction` as an extra commit.

src/util/borrowUtils.ts
```diff
@@ -26,3 +26,22 @@ export const useTotalFiatAmount = (wallet: EdgeCurrencyWallet, borrowArray: Borr
     }, '0')
   }, [allTokens, borrowArray, currencyInfo, exchangeRates, isoFiatCurrencyCode])
 }
+
+export const getWalletPickerExcludeWalletIds = (
+  wallets: { [wallet: string]: EdgeCurrencyWallet },
+  actionOpType: string,
```
samholmes, line 32: Use the same stricter type from the LoanManageScene component

src/util/borrowUtils.ts
```diff
@@ -26,3 +26,22 @@ export const useTotalFiatAmount = (wallet: EdgeCurrencyWallet, borrowArray: Borr
...
+  return Object.keys(wallets).filter(walletId => {
+    switch (actionOpType) {
+      case 'loan-deposit':
+        return walletId !== borrowEngineWallet.id && wallets[walletId].currencyInfo.pluginId !== 'bitcoin'
+      case 'loan-borrow':
+      case 'loan-repay':
+      case 'loan-withdraw':
+        return walletId !== borrowEngineWallet.id
+      default:
+        return false
```
samholmes, line 44: Remove this case once you've exhausted the static type.

samholmes (changes requested):
Minor changes.

src/util/borrowUtils.ts
Jon-edge: Wait I take back what I said earlier. This default is still a good safegaurd for runtime in the case the type is extended. If the type is extended, there will be no errors.

src/components/scenes/Loans/LoanManageScene.tsx
```diff
@@ -236,24 +243,40 @@ export const LoanManageScene = (props: Props) => {
...
+                borrowPluginId,
+                nativeAmount: actionNativeAmount,
+                walletId: borrowEngineWallet.id,
+                tokenId: selectedAsset.tokenId
+              }
+            ]
+          }
+          break
+        default:
+          throw new Error(`LoanManageScene cannot generate actionOps for ${loanManageAction}`)
```
samholmes, line 275: The reason I don't like default cases like this is that _today_ they're not needed and we're making assumptions about _future behavior_. I don't like it one bit because we should give the developer the opportunity to make the design decision when (**or if**) that future code change comes.

The right thing to do is use the right tool to help us with this; the static type checker.

src/components/scenes/Loans/LoanManageScene.tsx
```diff
@@ -284,12 +307,12 @@ export const LoanManageScene = (props: Props) => {
...
   })
 
-  const handleInfoIconPress = useUrlHandler(actionOpStrings.supportUrl)
+  const handleInfoIconPress = useUrlHandler(manageActionStrings.supportUrl)
 
   const handleSliderComplete = useHandler(async (resetSlider: () => void) => {
     if (actionProgram != null) {
       try {
-        await dispatch(runLoanActionProgram(loanAccount, actionProgram, actionOpType))
+        await dispatch(runLoanActionProgram(loanAccount, actionProgram, MANAGE_ACTION_DATA_MAP[loanManageAction].programType))
```
samholmes, line 315: Use `manageActionStrings`

src/util/borrowUtils.ts
samholmes: If the type is extended, the developer should consider the implications to this function when the type is extended. Why are we assuming that the correct behavior for some unknown future change to this type is for this function to return false?

src/components/scenes/Loans/LoanManageScene.tsx
```diff
@@ -331,11 +331,11 @@ export const LoanManageScene = (props: Props) => {
       <WalletListModal
         bridge={bridge}
         headerTitle={s.strings.select_wallet}
-        showCreateWallet={isSceneTypeDebts}
+        showCreateWallet
         createWalletId={isSceneTypeDebts ? borrowEngineWallet.id : undefined}
         showBankOptions={loanManageAction === 'loan-manage-borrow'}
-        excludeWalletIds={isSceneTypeDebts ? excludeWalletIds : undefined}
-        allowedAssets={isSceneTypeDebts ? hardAllowedDebtAsset : hardAllowedCollateralAsset}
+        excludeWalletIds={getWalletPickerExcludeWalletIds(wallets, loanManageAction, borrowEngineWallet)}
```
samholmes, line 337: I like that you inlined this. 👍

samholmes (changes requested):
I feel like we're almost there with this one.

src/util/borrowUtils.ts
Jon-edge: Ah missed this change

src/components/scenes/Loans/LoanManageScene.tsx
```diff
@@ -48,33 +49,39 @@ import { NetworkFeeTile } from '../../tiles/NetworkFeeTile'
 import { TotalDebtCollateralTile } from '../../tiles/TotalDebtCollateralTile'
 import { FormScene } from '../FormScene'
 
-export type LoanManageActionOpType = 'loan-deposit' | 'loan-withdraw' | 'loan-borrow' | 'loan-repay'
+export type LoanManageType = 'loan-manage-deposit' | 'loan-manage-withdraw' | 'loan-manage-borrow' | 'loan-manage-repay'
```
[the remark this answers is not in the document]
Jon-edge: Missed the sync between these after bouncing between name decisions

src/util/borrowUtils.ts
Jon-edge: This function restricts selectable wallets. If it was not properly updated, allow selection of any wallet so testing is not blocked

samholmes (approved):

PR #3809, opened 2022-11-17
Jon/aave/fix/close-withdraw

Jon-edge:
- Fix fiat conversion calculations
- Fix error checking formula
- Fix withdrawals with no principals

#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203374649830377

### DIFF
--- a/src/util/ActionProgramUtils.ts
+++ b/src/util/ActionProgramUtils.ts
@@ -1,4 +1,4 @@
-import { div, gt, gte, lt, max, mul, sub } from 'biggystring'
+import { add, div, gt, gte, log10, max, mul, sub, toFixed } from 'biggystring'
 import { EdgeCurrencyWallet } from 'edge-core-js/types'
 import { sprintf } from 'sprintf-js'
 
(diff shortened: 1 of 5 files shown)


src/util/ActionProgramUtils.ts
```diff
@@ -261,66 +266,97 @@ export const makeAaveCloseAction = async ({
...
+
+    // Find maximum amount of collateral that can be swapped while remaining
+    // above the liquidation threshold (74%), where:
+    // Pr: Principal remaining after initial payment above
+    // C: Collateral
+    // Cs: Collateral to swap (maxSwappableCollateral)
+    //
+    // Pr / (C - Cs) = 0.74
+    // Solved for Cs:
+    // Cs = -((50Pr - 37C) / 37)
```
samholmes, line 313: I learnt in my limited exposure to formal maths that the ideal form is a polynomial:

`Cs = -1/.74 * Pr + C`

This is nice because now you only have one constant to deal with.

Full solve:
```
Pr / (C -Cs) = 0.74
1 / (C-Cs) = 0.74/Pr
C -Cs = Pr / .74
-Cs = Pr / .74 - C
Cs = -(Pr / .74 - C)
Cs = -Pr / .74 + C
Cs = Pr * -1/.74 + C
Cs = -1/.74 * Pr + C
```

src/util/ActionProgramUtils.ts
```diff
@@ -239,12 +239,17 @@ export const makeAaveCloseAction = async ({
...
+    actions: []
+  }
+
   // #region Validation
 
   // Only accept this request if the user has only singular debt/collateral assets
   const collaterals = borrowEngine.collaterals.filter(collateral => !zeroString(collateral.nativeAmount))
   const debts = borrowEngine.debts.filter(debt => !zeroString(debt.nativeAmount))
-  if (debts.length === 0) return null
+  if (debts.length === 0 && collaterals.length === 0) return null
```
samholmes, line 252: Don't early exit. Check that collaterals/debts are not >1 and just return the null case at the bottom return statement.

samholmes (changes requested):

src/util/ActionProgramUtils.ts
```diff
@@ -238,109 +238,153 @@ export const makeAaveCloseAction = async ({
...
+    // Cs: Collateral to swap (maxSwappableCollateral)
+    //
+    // Pr / (C - Cs) = 0.74
+    // Solved for Cs:
+    // Cs = (-1/.74) * Pr + C
+    const maxSwappableCollateralFiat = add(mul(div('-1', '0.74', DECIMAL_PRECISION), postRepayPrincipalFiat), collateralFiat)
+
+    // Check if swap and repay amount necessary to cover the remaining principal
+    // would cause liquidation
+    if (gt(postRepayPrincipalFiat, maxSwappableCollateralFiat)) {
```
Jon-edge, line 322: The amount shown here is not right, it's greater than what's actually needed. Not sure what the right maths are to get that value yet... Might be a calculus thing

Jon-edge: https://app.asana.com/0/1200382638405084/1203374650739508/f

src/util/ActionProgramUtils.ts
samholmes: Looks like you left the early exit in. I'm okay with short-circuiting as a pattern when it makes sense. I've made some fixup commits for you to review or discuss and one of them is to switch from short-circuiting to nested-code. I prefer the short-circuit, but I wanted to give the nested logic a chance.

samholmes (commented):
Made fixups as part of my review.

samholmes (approved):

PR #3816, opened 2022-11-22
Jon/aave/fix/amount-denom

Jon-edge:
Fix display denoms not showing, also fix some styling

image

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203145107958053

### DIFF
--- a/src/components/scenes/Loans/LoanCreateScene.tsx
+++ b/src/components/scenes/Loans/LoanCreateScene.tsx
@@ -1,7 +1,7 @@
 import { div, lt, max, mul } from 'biggystring'
 import { EdgeCurrencyWallet } from 'edge-core-js'
 import * as React from 'react'
-import { ActivityIndicator, TouchableOpacity, View } from 'react-native'
+import { ActivityIndicator, TouchableOpacity } from 'react-native'
 import { AirshipBridge } from 'react-native-airship'
 import { KeyboardAwareScrollView } from 'react-native-keyboard-aware-scroll-view'
 import Ionicon from 'react-native-vector-icons/Ionicons'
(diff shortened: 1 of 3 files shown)


src/components/scenes/Loans/LoanCreateScene.tsx
```diff
@@ -343,7 +344,7 @@ export const LoanCreateScene = (props: Props) => {
         }
       />
       <KeyboardAwareScrollView extraScrollHeight={theme.rem(2.75)} enableOnAndroid>
-        <View style={styles.sceneContainer}>
+        <Space top={0} left={0.5} right={0.5} bottom={1}>
```
samholmes, line 347: Optionally: `<Space horizontal={0.5} bottom={1}>`

src/components/text/CryptoText.tsx
```diff
@@ -48,7 +47,7 @@ export const useCryptoTextSimple = ({ wallet, tokenId, nativeAmount, withSymbol
     exchangeRate: assetToFiatRate,
     fiatDenomination,
     nativeAmount,
-    currencyCode: withSymbol ? undefined : currencyCode
+    currencyCode: withSymbol ? undefined : displayDenomination.name
```
samholmes, line 50: I would have expected the field to be `displayDenomination.currencyCode` or `displayDenomination.symbol`. Are you sure this is right?

samholmes (approved):
Approved. 

The first two comments about the use of the spread operator should be considered because we're introducing more tech-debt then we have to.

The comment about Space arguments is optional, but I'd take it over the verbose props.

The last comment is just me questioning things, but if it works and you know what you're doing then ignore it.

PR #3821, opened 2022-11-23
Fix AAVE crashes

samholmes:
#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203414798785237

### DIFF
--- a/src/components/hoc/withLoanAccount.tsx
+++ b/src/components/hoc/withLoanAccount.tsx
@@ -1,0 +1,32 @@
+import * as React from 'react'
+
+import { LoanAccount } from '../../controllers/loan-manager/types'
+import { useSelector } from '../../types/reactRedux'
+
+interface NavigationProps {
+  navigation: { popToTop: () => void }
+  route: { params: { loanAccountId: string } }
+}
+
+type WithoutLoanAccount<Props> = Omit<Props, 'loanAccount'>
+
+/**
+ * Looks up a loanAccount for a scene.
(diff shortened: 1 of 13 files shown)


src/components/hoc/withLoanAccount.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+    const loanAccounts = useSelector(state => state.loanManager.loanAccounts)
+    const loanAccount = loanAccounts[route.params.loanAccountId]
+
+    React.useEffect(() => {
+      if (loanAccount == null) navigation.popToTop()
+    }, [navigation, loanAccount])
+
+    if (loanAccount == null) return <LoadingScene />
+    return <Component {...(props as any)} loanAccount={loanAccount} />
```
Jon-edge, line 32: Redirect only when 'loanAccount == null' && 'loanAccounts != null'
Loading state when 'loanAccounts == null'

Jon-edge (approved):
Approved after one comment addressed

src/components/hoc/withLoanAccount.tsx
samholmes: I realized that `loanAccounts === null` is never a case. Instead, we should introduce some state to indicate that the loanAccounts are still initializing or loading. I think that's a break away task though, and luckily not necessary for this all to work until we have deep linking to specific loan accounts (maybe when we improve push notification routing).

PR #3823, opened 2022-11-27
Implement Thorchain Savers staking

paullinator:
#### PR Dependencies

This is a hard requirement and needs a publish and upgrade before merging this PR
https://github.com/EdgeApp/edge-currency-plugins/pull/349

This is a soft requirement. APY values will show n/a until this is merged and published
https://github.com/EdgeApp/edge-info-server/pull/85

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203237465587634


src/components/themed/TransactionListTop.tsx
```diff
@@ -90,14 +92,18 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
...
-        this.setState({ stakePolicies: filteredStatePolicies })
+      makeStakePlugins().then(stakePlugins => {
+        let stakePolicies: StakePolicy[] = []
+        for (const stakePlugin of stakePlugins) {
+          const filteredStatePolicies = stakePlugin.policies.filter(stakePolicy => {
+            return [...stakePolicy.rewardAssets, ...stakePolicy.stakeAssets].some(asset => asset.pluginId === pluginId && asset.currencyCode === currencyCode)
+          })
+          stakePolicies = [...stakePolicies, ...filteredStatePolicies]
+        }
+        this.setState({ stakePolicies, stakePlugins })
```
samholmes, line 104: I believe this invocation needs to be react to the change in component props because the walletId and pluginId can change for the scene via the wallet list modal. Perhaps create a new method and call it in this life-cycle event and also in the handler for the modal bridge response.

src/components/scenes/Staking/StakeOptionsScene.tsx
```diff
@@ -46,6 +47,10 @@ export const StakeOptionsScene = (props: Props) => {
     return currencyWallets[walletId]
   })
 
+  const pluginId = wallet?.currencyInfo.pluginId
+  const tokenId = pluginId ? getTokenId(account, pluginId, currencyCode) : undefined
+  const icon = React.useMemo(() => <CryptoIcon marginRem={[0, 0.5, 0, 0]} walletId={walletId} tokenId={tokenId} sizeRem={1.5} />, [])
```
samholmes, line 39: Missing walletId, and tokenId dependencies for the memo

samholmes: I'm also okay with inlining this icon JSX in the return like normal (CrypoIcon isn't an expensive function).

src/plugins/stake-plugins/stakePlugins.ts
```diff
@@ -1,6 +1,7 @@
+import { makeTcSaversPlugin } from './thorchainSavers/tcSavers'
```
samholmes, line 2: Can we follow the patter of naming files which export a class of object the same name as the class: e.g. `TcSaversPlugin.ts` and `UniV2StakePlugin.ts`?

src/plugins/stake-plugins/thorchainSavers/tcSavers.ts
```diff
@@ -0,0 +1,168 @@
...
+  {
+    stakePolicyId: 'tcsavers/avalanche:avax=avalanche:avax',
+    stakeProviderInfo,
+    apy: 0,
+    rewardAssets: [{ pluginId: 'avalanche', currencyCode: 'AVAX' }],
+    stakeAssets: [{ pluginId: 'avalanche', currencyCode: 'AVAX' }]
+  }
+]
+
+// type ChangeQuoteFunc = (request: ChangeQuoteRequest, policy: StakePolicy) => Promise<ChangeQuote>
```
samholmes, line 189: Cleanup on isle 58

src/plugins/stake-plugins/types.ts
```diff
@@ -41,8 +41,9 @@ export interface StakePolicy {
 // -----------------------------------------------------------------------------
 // Change Quote
 // -----------------------------------------------------------------------------
+export type ChangeQuoteAction = 'stake' | 'unstake' | 'claim'
```
samholmes, line 75: Keep it inline until it needs to be factored out.

src/plugins/stake-plugins/thorchainSavers/tcSavers.ts
```diff
@@ -0,0 +1,168 @@
...
+    ],
+    approve: async () => {}
+  }
+}
+
+const changeQuoteFuncs = {
+  stake: stakeRequest,
+  unstake: unstakeRequest,
+  claim: unstakeRequest
+}
```
samholmes, line 638: Hmm. I would like to move these private methods into the scope of the instance, but I understand this means testing these functions will requires async calls in order to instantiate the plugin object. Maybe we could keep the constrain that JavaScript class constructors have: sync only. This of course means reverting the method call to get stake policies, but I see `loadPolicies` method could take-place-of initializing at instantiation time.

I'm open to discussing this more. My motivation is to keep code more contiguous and maintainable by using things like closures. I think it's easier to reason about having class methods close over the instance state, even when writing unit tests (instantiating before testing methods is more true to real-world usage, so more reliable testing, IMO).

samholmes (approved):
Half-way review so far.

samholmes (changes requested):
Previous review was accidentally marked as Approve when it really was intended to be a Request Changes

src/components/scenes/Staking/StakeOptionsScene.tsx
paullinator: This was actually copy/paste from line 29 but good catch. I'll move it inline

src/components/themed/TransactionListTop.tsx
paullinator: I've always hated how componentDidMount and componentDidUpdate are two separate funcs. I'll create a common routine that both of them call

src/plugins/stake-plugins/stakePlugins.ts
paullinator: Sure. But to prevent massive conflicts, I'll add as a final commit

src/plugins/stake-plugins/thorchainSavers/tcSavers.ts
paullinator: 👍

src/plugins/stake-plugins/thorchainSavers/tcSavers.ts
paullinator: My desire to keep these funcs outside the instance is to make them independent of instance variables so that I could move these function to another file in the future. Instance functions cause bloat as you can't move functions to another file so unless they are heavily dependent on instance params, I like to keep them out and make them pure functions which as you mentioned are easier to unit test.

src/plugins/stake-plugins/types.ts
paullinator: done

samholmes (commented):
Some thoughts/comments

src/plugins/stake-plugins/thorchainSavers/tcSavers.ts
```diff
@@ -128,18 +300,124 @@ const getPolicyFromId = (policyId: string): StakePolicy => {
...
+    }
+  }
+
+  if (needsFundingPrimary) {
+    // Estimate the total cost to create the two transactions
+    // 1. Fund the primary address with the requestedAmount + fees for tx #2
+    // 2. Send the requested amount to the pool address
+
+    const estimateTx = await wallet.makeSpend({
+      spendTargets: [{ publicAddress: primaryAddress, nativeAmount: minAmount }]
```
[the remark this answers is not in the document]
paullinator: Changing to `nativeAmount`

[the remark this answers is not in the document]
paullinator: Cannot shorted `publicAddress: primaryAddres -> publicAddress`. There's only `primaryAddress` declared

src/plugins/stake-plugins/thorchainSavers/tcSavers.ts
```diff
@@ -79,28 +186,93 @@ export const makeTcSaversPlugin = async (opts?: EdgeCorePluginOptions): Promise<
...
+        // Convert from exchangeAmount to nativeAmount
+        stakedAmount = await wallet.denominationToNative(stakedAmount, currencyCode)
+        earnedAmount = await wallet.denominationToNative(earnedAmount, currencyCode)
+
+        // Truncate decimals
+        stakedAmount = toFixed(stakedAmount, 0, 0)
+        earnedAmount = toFixed(earnedAmount, 0, 0)
+
+        // Cap negative value to 0
+        earnedAmount = lt(earnedAmount, '0') ? '0' : earnedAmount
```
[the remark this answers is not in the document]
paullinator: done

src/plugins/stake-plugins/thorchainSavers/tcSavers.ts
```diff
@@ -1,9 +1,96 @@
...
 import { ChangeQuote, ChangeQuoteRequest, StakePlugin, StakePolicy, StakePosition, StakePositionRequest, StakeProviderInfo } from '../types'
 import { asInfoServerResponse, InfoServerResponse } from '../util/internalTypes'
 
+const EXCHANGE_INFO_UPDATE_FREQ_MS = 10 * 60 * 1000 // 2 min
+const INBOUND_ADDRESSES_UPDATE_FREQ_MS = 10 * 60 * 1000 // 2 min
+const MIDGARD_SERVERS_DEFAULT = ['https://midgard.thorchain.info']
+const THORNODE_SERVERS_DEFAULT = ['https://thornode.ninerealms.com']
+const TC_SAVERS_WITHDRAWAL_SCALE_UNITS = '10000'
+const DIVIDE_PRECISION = 18
+export const THOR_LIMIT_UNITS = '100000000'
```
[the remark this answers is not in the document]
paullinator: done

src/plugins/stake-plugins/thorchainSavers/tcSavers.ts
```diff
@@ -1,9 +1,96 @@
...
+import { asArray, asBoolean, asNumber, asObject, asOptional, asString } from 'cleaners'
+import { EdgeCorePluginOptions, EdgeCurrencyWallet, InsufficientFundsError } from 'edge-core-js'
 
-import { fetchInfo } from '../../../util/network'
+import { cleanMultiFetch, fetchInfo, fetchWaterfall } from '../../../util/network'
 import { ChangeQuote, ChangeQuoteRequest, StakePlugin, StakePolicy, StakePosition, StakePositionRequest, StakeProviderInfo } from '../types'
 import { asInfoServerResponse, InfoServerResponse } from '../util/internalTypes'
 
+const EXCHANGE_INFO_UPDATE_FREQ_MS = 10 * 60 * 1000 // 2 min
+const INBOUND_ADDRESSES_UPDATE_FREQ_MS = 10 * 60 * 1000 // 2 min
```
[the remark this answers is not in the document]
paullinator: done

src/plugins/stake-plugins/thorchainSavers/tcSavers.ts
```diff
@@ -97,46 +97,48 @@ const stakeProviderInfo: StakeProviderInfo = {
   stakeProviderId: 'tcsavers'
 }
 
+const policyDefault = {
+  apy: 0,
+  stakeProviderInfo,
+  stakeWarning: null,
+  unstakeWarning: null,
+  claimWarning: null
+}
```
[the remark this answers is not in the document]
paullinator: todo added

src/plugins/stake-plugins/thorchainSavers/tcSavers.ts
```diff
@@ -2,13 +2,14 @@ import { add, div, lt, mul, sub, toFixed } from 'biggystring'
 import { asArray, asBoolean, asNumber, asObject, asOptional, asString } from 'cleaners'
 import { EdgeCorePluginOptions, EdgeCurrencyWallet, InsufficientFundsError } from 'edge-core-js'
 
+import { showError } from '../../../components/services/AirshipInstance'
```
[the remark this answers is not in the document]
paullinator: Can do that for fetchStakePosition since an error means we really need to throw. However for makeTcSaversPlugin, need to find a better solution since we can't throw as that would kill the whole plugin even though the errors would only prevent APY display.

src/types/routerTypes.tsx
```diff
@@ -396,9 +396,8 @@ export function withNavigation<Props>(Component: React.ComponentType<Props>): Re
       push(name, params) {
         props.navigation.push(name, { route: { name, params } })
       },
-      replace(...a) {
-        // prev, next, params) {
-        props.navigation.replace(...a) // prev, next, next, { route: { params } })
+      replace(name, params) {
+        props.navigation.replace(name, { route: { name, params } })
```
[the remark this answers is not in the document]
swansontec: Yes, this is wrong. Jon accidentally broke the navigation shims, but we got that reverted, and now you just pass the params as the 2nd argument, not this structure.

src/components/scenes/Staking/StakeOptionsScene.tsx
```diff
@@ -46,6 +47,10 @@ export const StakeOptionsScene = (props: Props) => {
     return currencyWallets[walletId]
   })
 
+  const pluginId = wallet?.currencyInfo.pluginId
+  const tokenId = pluginId ? getTokenId(account, pluginId, currencyCode) : undefined
```
swansontec, line 38: Do you mean `pluginId != null` or `pluginId != ''` here? Please be explicit.

swansontec (changes requested):
You need to update your `navigation.replace` call, since we fixed that guy last week while this was still in review. I found one other small thing you could optionally fix. Otherwise you have addressed all Sam's feedback, so this looks good now.

src/types/routerTypes.tsx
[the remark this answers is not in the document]
paullinator: I don't understand what I need to fix here? My original changes to this file match what Jon made in commit 2c8b52e6587be59894e519bfaaf5f77a92a23f18

swansontec (approved):

src/types/routerTypes.tsx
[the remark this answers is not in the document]
swansontec: It would be wrong because `navigation.replace` does the work, so doing it again here would double-nest the params. However, I see that this line gets eliminated anyhow, so it's moot.

swansontec:
src/plugins/stake-plugins/stakePlugins.ts(12,26): error TS2322: Type 'boolean' is not assignable to type 'JsonObject'.

src/plugins/stake-plugins/stakePlugins.ts
samholmes: I think a better convention moving forward would be to choose camelCase for names of files (regardless of file type/extension) and PascalCase for any file which exports a main class/component. This way you can tell by the name that the purpose of the file by this case rather than something redundant like it's file extension.

samholmes (approved):
I did a final pass-through on this and it all looks good to me.

PR #3830, opened 2022-12-01
AAVE: Explicit dashboard navigation on null LoanAccount

Jon-edge:
navigation.popToTop() does not reliably navigate back to the dashboard scene in all scenarios.

In any case where the loan account is not available, always navigate to the dashboard scene.

#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203444815739776

### DIFF
--- a/src/components/hoc/withLoanAccount.tsx
+++ b/src/components/hoc/withLoanAccount.tsx
@@ -2,10 +2,11 @@ import * as React from 'react'
 
 import { LoanAccount } from '../../controllers/loan-manager/types'
 import { useSelector } from '../../types/reactRedux'
+import { NavigationProp, ParamList } from '../../types/routerTypes'
 import { LoadingScene } from '../scenes/LoadingScene'
 
 interface NavigationProps {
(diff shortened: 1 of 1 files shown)


src/components/hoc/withWallet.tsx
```diff
@@ -25,10 +26,15 @@ export function withWallet<Props extends { wallet: EdgeCurrencyWallet }>(
     const wallet = account.currencyWallets[route.params.walletId]
 
     React.useEffect(() => {
-      if (wallet == null) navigation.goBack()
+      if (wallet == null) navigation.navigate('loanDashboard', {})
```
peachbits, line 29: This isn't a good spot to add loan-specific navigation since this HOC could be used anywhere. Also, this component isn't used in any loans scenes I suspect you meant to edit withLoanAccount instead?

peachbits (changes requested):

peachbits (approved):

PR #3837, opened 2022-12-03
Fix cutoff text in wallet list

Jon-edge:
- Use EdgeText in TickerText, change API for IconDataRow
- Adjust styling in IconDataRow, WalletListCurrencyRow

image
image

#### Merge Message
-fix: cutoff text in wallet list

#### PR Dependencies

 none

#### PR Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203356149915792

### DIFF
--- a/src/components/data/row/IconDataRow.tsx
+++ b/src/components/data/row/IconDataRow.tsx
@@ -50,21 +50,19 @@ const getStyles = cacheStyles((theme: Theme) => ({
   // Layout:
   rightColumn: {
     alignItems: 'flex-end',
-    flexDirection: 'column',
-    paddingRight: theme.rem(1)
+    flexDirection: 'column'
   },
   leftColumn: {
     flexDirection: 'column',
     flexGrow: 1,
     flexShrink: 1,
-    marginRight: theme.rem(0.5),
+    marginRight: theme.rem(0.25),
     marginLeft: theme.rem(1)
   },
   container: {
     alignItems: 'center',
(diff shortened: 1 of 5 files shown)


src/components/data/row/IconDataRow.tsx
```diff
@@ -31,7 +31,7 @@ const IconDataRowComponent = (props: Props) => {
       <View style={styles.leftColumn}>
```
Jon-edge, line 31: The row-based layout discussed actually won't work because another caller of IconDataRow (PaymentMethodRow) is reliant on the two column base layout style. Changed it so that the top left and top right texts are fixed in size. This way also we retain the left justified ticker.

src/components/text/TickerText.tsx
```diff
@@ -54,5 +54,19 @@ export const TickerText = ({ wallet, tokenId }: Props) => {
...
+        marginLeft: theme.rem(0.75),
+        alignSelf: 'center'
+      }}
+    >
+      {tickerText}
+    </EdgeText>
+  )
 }
+
+export const TickerTextString = () => {}
```
swansontec, line 72: What is this for?

swansontec (approved):
Approved, but with one question / concern.

src/components/text/TickerText.tsx
Jon-edge: Removing

PR #3838, opened 2022-12-03
Fix localization handling on FiatAmountInputCard

Jon-edge:
Fix localization handling on FiatAmountInputCard 
Set 2 decimal max precision on FiatAmountInputCard

#### PR Dependencies

 none

#### PR Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203475087427001

### DIFF
--- a/src/components/cards/FiatAmountInputCard.tsx
+++ b/src/components/cards/FiatAmountInputCard.tsx
@@ -67,7 +67,11 @@ const FiatAmountInputCardComponent = ({ wallet, iconUri, inputModalMessage, titl
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [sanitizedFiatAmount, inputModalMessage, onAmountChanged, title])
 
-  const formattedFiatAmount = React.useMemo(() => formatFiatString({ fiatAmount: sanitizedFiatAmount ?? '0', autoPrecision: true }), [sanitizedFiatAmount])
+  const formattedFiatAmount = React.useMemo(
+    () => formatFiatString({ fiatAmount: sanitizedFiatAmount ?? '0', autoPrecision: true, maxPrecision: 2 }),
+
+    [sanitizedFiatAmount]
+  )
 
   return (
     <UnderlinedNumInputCard currencyCode="USD" formattedAmount={formattedFiatAmount} iconUri={iconUri} title={title} onPress={handleEditActionfiatAmount} />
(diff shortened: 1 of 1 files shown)


src/components/cards/FiatAmountInputCard.tsx
```diff
@@ -41,29 +41,31 @@ const FiatAmountInputCardComponent = ({ wallet, iconUri, inputModalMessage, titl
...
 
   const handleEditActionfiatAmount = React.useCallback(() => {
     Airship.show<string | undefined>(bridge => <TextInputModal title={title} message={inputModalMessage} bridge={bridge} keyboardType="decimal-pad" />).then(
       inputAmount => {
-        if (inputAmount != null && !zeroString(inputAmount)) {
-          setFiatAmount(inputAmount)
-          onAmountChanged(fiatAmount, nativeCryptoAmount)
+        if (inputAmount != null && !zeroString(sanitizeDecimalAmount(inputAmount, 2))) {
+          setInputFiatAmount(inputAmount)
+          onAmountChanged(inputFiatAmount, nativeCryptoAmount)
```
peachbits, line 61: Passing an unsantized user input to onAmountChanged isn't safe. This value is passed to a biggystring div function in LoanCreateScene so the bug still exists. 

You can remove a lot of risk for bugs like this by sanitizing immediately upon user input before doing anything else. Take the return value from Airship and sanitize it before entering this if block. Then every usage of it zeroString, onAmountChanged, and beyond, are using 
a safe and nice decimal amount. If the value needs to be displayed to the user you can convert it back where it gets rendered using utils from locales.intl.ts

src/components/cards/FiatAmountInputCard.tsx
```diff
@@ -41,29 +41,31 @@ const FiatAmountInputCardComponent = ({ wallet, iconUri, inputModalMessage, titl
...
+    const fiatAmount = sanitizeDecimalAmount(inputFiatAmount, 2)
     const calculatedNativeCryptoAmount = truncateDecimals(
       destToFiatRate == null || destToFiatRate === '0' ? '0' : mul(destExchangeMultiplier, div(fiatAmount ?? '0', destToFiatRate, DECIMAL_PRECISION)),
       0
     )
 
     setNativeCryptoAmount(calculatedNativeCryptoAmount)
-    if (!zeroString(calculatedNativeCryptoAmount) && !zeroString(fiatAmount)) onAmountChanged(fiatAmount, calculatedNativeCryptoAmount)
-  }, [allTokens, destToFiatRate, fiatAmount, onAmountChanged, tokenId, wallet.currencyInfo])
+    if (!zeroString(calculatedNativeCryptoAmount) && !zeroString(fiatAmount)) onAmountChanged(inputFiatAmount, calculatedNativeCryptoAmount)
```
peachbits, line 53: Same bug here.

peachbits (changes requested):

peachbits (approved):

PR #3842, opened 2022-12-07
Add big little number handling to toBigNumberString

peachbits:


samholmes:
> Is 0.000000000553564816816 a little big number or a big little number??

It's a little big-number.

src/util/toBigNumberString.ts
```diff
@@ -8,7 +8,7 @@ export function toBigNumberString(n: number | string): string {
   const match = out.match(expoRegex)
 
   if (match != null) {
-    const exponent = parseInt(match[1])
+    const exponent = parseInt(match[0].replace('e', ''))
```
samholmes, line 11: You can keep the `match[1]` and change the matching group in the regex from `/e(?:\+|-)(\d+)$/` to `/e((?:\+|-)\d+)$/`.

samholmes (changes requested):

PR #3845, opened 2022-12-07
Implement "Clear Logs" feature in settings

samholmes:
### CHANGELOG

- Added: Ability to delete log files on device from within the Settings scene

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203503178062177


src/actions/LogActions.tsx
```diff
@@ -76,6 +77,46 @@ export function showSendLogsModal(): ThunkAction<Promise<void>> {
   }
 }
 
+export function showClearLogsModal(): ThunkAction<Promise<void>> {
+  return async (dispatch, getState) => {
+    const state = getState()
+    const { isConnected } = state.network
+    if (!isConnected) return showError(s.strings.network_alert_title)
+
+    Airship.show<string | number | undefined>(bridge => (
```
swansontec, line 86: This creates a dangling promise. Totally fine in reality, but we eventually want to enable the lint rule that prohibits this sort of thing. The solution is to either await the promise, or put a `.catch(showError)` on the end.

swansontec (approved):
Approved, with one question and one optional fix.

src/actions/LogActions.tsx
```diff
@@ -76,6 +77,46 @@ export function showSendLogsModal(): ThunkAction<Promise<void>> {
   }
 }
 
+export function showClearLogsModal(): ThunkAction<Promise<void>> {
+  return async (dispatch, getState) => {
+    const state = getState()
+    const { isConnected } = state.network
+    if (!isConnected) return showError(s.strings.network_alert_title)
```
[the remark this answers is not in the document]
samholmes: I removed these lines. This was from copy-pasta.

src/actions/LogActions.tsx
samholmes: I went with the .catch method because I'm not certain how errors would be caught in a onPress handler if I were to ensure the error propagates up to the event react dispatcher.

PR #3852, opened 2022-12-08
Add incoming spam transaction filter

peachbits:
This adds a filter to prevent worthless transactions from appearing in the transaction list and from showing an incoming transaction notification. A worthless transaction is defined as having a value less than the smallest unit of the wallet's fiat currency (ie 1 cent for USD, 1 pence for GBP, etc.)

The feature is controlled by a local setting that is enabled by default and can be disabled in the settings scene. 

---
  - https://app.asana.com/0/0/1203499867589460


src/locales/en_US.ts
```diff
@@ -352,6 +352,7 @@ const strings = {
   settings_exchange_settings: 'Exchange Settings',
   settings_exchange_instruction: 'Enable or disable the exchange providers available to you',
   settings_notifications_switch: 'Enable Price Alerts',
+  settings_spam_filter: 'Spam Transaction Filter',
```
swansontec, line 355: We should call the setting "Hide spam transactions". That makes it more clear what this feature does. We should also use sentence case, since most settings do that (with a small number of exceptions that might be worth fixing.)

swansontec (changes requested):
One tiny wording change needed, but this seems reasonable.

PR #3858, opened 2022-12-13
Always show wallet list

swansontec:
### CHANGELOG

- changed: Always show the wallet list, even if there are no wallets in the account.

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203532879970168


samholmes (approved):
All code changes look good, however one thing I don't understand is how we can get away introducing the value `''` for `walletid` and `currencyCode` without fully auditing this wont create regressions. If we introduce `undefined` into the type, then we could start auditing that accessors to these fields have not regressed because we'd handle the cases. I'd think that some cases would be to show a message to the user that no wallets are available in places like the SendScene after deep-linking perhaps.

swansontec:
Yes, it is a regression, and it does cause crashes. Visiting a buy / sell partner with no wallets = boom. However, the problem is that it was *already crashing*. So it's certainly better, but still not great. I can eliminate the currently selected wallet from more parts of the app if you prefer, and put guards around the rest, but it will be a big refactor.

PR #3883, opened 2022-12-24
Jon/aave/fix/liq-notif-spam

Jon-edge:
The original implementation would continuously upload liquidation event triggers since the effect had a state dependency.

This would result in repeated liquidation warning notifications on every state update.

This fix addresses the issue with a set of rules and dependency map changes:
1. A cached 'copy' of the collaterals and debts (cachedLoanAssetsMap) are maintained as a basis to detect if a change was made to an account (borrow, withdraw, etc), or if the account was initializing.
2. The cachedLoanAssetsMap is used as the dependency trigger for the liquidation warning event upload. It's typed as a string valued map to ensure delta comparisons and dependency triggers work as expected. Writing to this cache waits for all assets to load on all accounts, as well as all associated exchange rates for those assets, in order to discern what data is initial vs what has been changed by the user's loan actions during app runtime.
3. When the app is initialized, to handle newly observed loan accounts not created in or modified outside of Edge, the liquidation warning is uploaded only if the account does not exceed the warning threshold. This avoids a scenario where the user taps a notification to open Edge, and is immediately given another notification after AAVE finishes initializing.
4. When a change in LoanAccounts is observed by comparing against cachedLoanAssetsMap, it can be assumed a change has been made to the LoanAccount by the user. In this case, no matter what a new liquidation warning event is uploaded, even if the warning threshold has already been exceeded.
5. To address the use case where a notification is triggered and presented while the user is in the app, and the user remains in the app while price rises above the trigger price and falls back below, a 10 minute periodic task is created that will always upload a liquidation warning event regardless of price.

### CHANGELOG

-AAVE: Fix liquidation notification spam

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203536450138339

### DIFF
--- a/src/components/services/LoanManagerService.ts
+++ b/src/components/services/LoanManagerService.ts
@@ -11,7 +11,10 @@ import { waitForBorrowEngineSync } from '../../controllers/loan-manager/util/wai
 import { useAllTokens } from '../../hooks/useAllTokens'
 import { useAsyncEffect } from '../../hooks/useAsyncEffect'
 import s from '../../locales/strings'
+import { BorrowEngine } from '../../plugins/borrow-plugins/types'
+import { useState } from '../../types/reactHooks'
 import { useDispatch, useSelector } from '../../types/reactRedux'
+import { makePeriodicTask } from '../../util/PeriodicTask'
 import { makePushClient } from '../../util/PushClient/PushClient'
 import { DECIMAL_PRECISION, zeroString } from '../../util/utils'
 
@@ -41,9 +44,10 @@ export const LoanManagerService = (props: Props) => {
   const dispatch = useDispatch()
   const clientId: string = useSelector(state => state.core.context.clientId)
   const loanAccountMap: LoanAccountMap = useSelector(state => state.loanManager.loanAccounts)
-
-  const state = useSelector(state => state)
+  const exchangeRates = useSelector(state => state.exchangeRates)
   const allTokens = useAllTokens(account)
+
+  const [cachedLoanAssetsMap, setCachedLoanAssetsMap] = useState<{ [loanAccountId: string]: string }>({})
 
   //
   // Initialization
@@ -64,6 +68,9 @@ export const LoanManagerService = (props: Props) => {
       for (const loanAccount of Object.values(loanAccountMap)) {
         if (!checkLoanHasFunds(loanAccount) && loanAccount.closed) {
           dispatch(deleteLoanAccount(loanAccount))
+
+          cachedLoanAssetsMap[loanAccount.id] = ''
+          setCachedLoanAssetsMap({ ...cachedLoanAssetsMap })
         }
       }
     }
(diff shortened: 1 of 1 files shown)


src/components/services/LoanManagerService.ts
```diff
@@ -41,10 +44,12 @@ export const LoanManagerService = (props: Props) => {
   const dispatch = useDispatch()
   const clientId: string = useSelector(state => state.core.context.clientId)
   const loanAccountMap: LoanAccountMap = useSelector(state => state.loanManager.loanAccounts)
-
-  const state = useSelector(state => state)
+  const exchangeRates = useSelector(state => state.exchangeRates)
   const allTokens = useAllTokens(account)
 
+  const ltvEventTrigger = React.useMemo(() => Object.values(loanAccountMap), [loanAccountMap])
```
samholmes, line 50: What's the point of this `ltvEventTrigger`?

src/components/services/LoanManagerService.ts
```diff
@@ -64,88 +69,158 @@ export const LoanManagerService = (props: Props) => {
...
+        // Only trigger push events after the exchange rates are available for
+        // all loan assets on this account
+        const loanAssetTokenIds = [...filteredDebts, ...filteredCollaterals].map(loanAsset => loanAsset.tokenId)
+        if (loanAssetTokenIds.every(tokenId => tokenId == null || !zeroString(exchangeRates[`${getCurrencyCode(borrowEngine, tokenId)}_iso:USD`]))) {
+          if (debts.length > 0) {
+            // If it's the first time the account has been seen, we want to
+            // avoid a notification if it already exceeds the liquidation price
+            // right when the app is booted. Newly created loans can never
+            // exceed the liquidation price.
+            if (Object.keys(cachedLoanAssetsMap).find(accountAssetId => loanAccountId === accountAssetId) == null) {
```
samholmes, line 122: Use a regular property accessor instead of `.find`.

src/components/services/LoanManagerService.ts
```diff
@@ -64,88 +69,158 @@ export const LoanManagerService = (props: Props) => {
...
+              cachedLoanAssetsMap[loanAccountId] = JSON.stringify([...debts, ...collaterals])
+              await uploadLiquidationPushEvents(loanAccountId, false)
```
samholmes, line 124: Do this here instead and get rid of the need for the `isSkipPriceCheck` flag param:
```ts
      // TODO: Implement a true way to clear the push event
      if (cachedLoanAssetsMap[loanAccountId] === '') {
        uploadLiquidationEvent('WBTC', 0)
        return
      }
```

samholmes: Also, there is a way to remove the price-change notification, so we should do that as instead of this hack. Let's discuss how.

src/components/services/LoanManagerService.ts
```diff
@@ -64,88 +69,158 @@ export const LoanManagerService = (props: Props) => {
...
+            // upload a push notification even if the liquidation price has been
+            // exceeded
+            else if (JSON.stringify([...debts, ...collaterals]) !== cachedLoanAssetsMap[loanAccountId]) {
+              cachedLoanAssetsMap[loanAccountId] = JSON.stringify([...debts, ...collaterals])
+              await uploadLiquidationPushEvents(loanAccountId, true)
+            }
+          } else {
+            // Ensure push event is cleared if account is closed or debts no
+            // longer exist
+            await uploadLiquidationPushEvents(loanAccountId, false)
```
samholmes, line 136: This is just removing the push event too, so it should be collapsed with the other case above I think.

samholmes (changes requested):
This will do the trick, though I think there are some things we can improve to make the code more straight-forward/obvious and also fix the tech-debt we introduced (TODO comment).

src/components/services/LoanManagerService.ts
samholmes: Let's look into if we can remove this.

src/components/services/LoanManagerService.ts
samholmes: After speaking about the different cases, I'll retract this one because there is a lot of coupling of cases in that function.

src/components/services/LoanManagerService.ts
samholmes: Nevermind, misunderstood.

src/components/services/LoanManagerService.ts
samholmes: Because this code may eventually be replaced with server-side handling, I'm not going to push hard on a refactor now.

src/components/services/LoanManagerService.ts
Jon-edge: No longer needed

samholmes (approved):

PR #3889, opened 2022-12-29
Sam/multiple addresses

samholmes:
### CHANGELOG

- Allow switching between address types from the request scene (Legacy, Segwit, etc...)

### Dependencies

- https://github.com/EdgeApp/edge-currency-plugins/pull/356

### Requirements

- [x] Tested on iOS device

Simulator Screen Recording - iPhone 14 - 2022-12-28 at 17 08 26
Simulator Screen Shot - iPhone 14 - 2022-12-28 at 10 53 20

Simulator Screen Shot - iPhone 14 - 2022-12-28 at 10 53 23

Simulator Screen Shot - iPad Air (5th generation) - 2022-12-28 at 10 52 11

Simulator Screen Shot - iPad Air (5th generation) - 2022-12-28 at 10 52 08

Simulator Screen Shot - iPad Air (5th generation) - 2022-12-28 at 10 52 01

Simulator Screen Shot - iPad Air (5th generation) - 2022-12-28 at 10 51 57

---
  - https://app.asana.com/0/0/1200150483153797


src/components/themed/AddressQr.tsx
```diff
@@ -0,0 +1,42 @@
...
+  onPress?: (encodedUri?: string) => void
+}
+
+export const AddressQr = (props: AddressQrProps) => {
+  const theme = useTheme()
+  const styles = getStyles(theme)
+  const { address, currencyCode, wallet, nativeAmount, onPress = () => {} } = props
+
+  const [encodedUri] = useAsyncValue(async () => {
+    const encodedUri = await wallet.encodeUri({ publicAddress: address, currencyCode, nativeAmount })
```
swansontec, line 24: Why not `return await`, or maybe even drop the curly braces?

src/components/themed/Carousel.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  const itemCount = items.length
+  const [itemWidth, setItemWidth] = useState(1) // 1 avoids division by zero
+  const trackWidth = itemWidth * itemCount
+  const handleLayout = (event: LayoutChangeEvent) => {
+    // The check prevents layout jittering
+    if (Math.abs(event.nativeEvent.layout.width - itemWidth) >= 1) {
+      setItemWidth(event.nativeEvent.layout.width)
+    }
+  }
```
swansontec, line 61: This whole calculation seems sketchy. If I were doing it, I think I would measure the outer `Animated.View` container, to see how wide / tall he ends up being in the context of the scene. Then, I would set the children's `maxHeight / maxWidth` to match the container, so no matter what aspect ratio the container has, a whole child will always appear. Finally, I would measure where the children actually land in terms of aspect ratio, and use those locations for detent positions.

However, I see that your logic does work, given the band-aids (orientation detection and de-bounce protection). So we can keep the current logic if you prefer, accepting it's limitations.

yarn.lock
```diff
@@ -7265,7 +7265,7 @@ envinfo@^7.7.2, envinfo@^7.7.3:
   resolved "https://registry.yarnpkg.com/envinfo/-/envinfo-7.8.1.tgz#06377e3e5f4d379fea7ac592d5ad8927e0c4d475"
   integrity sha512-/o+BXHmB7ocbHEAs6F2EnG0ogybVVUdkRunTT2glZU9XAaGmhqskrvKwqXuDfNjEO0LZKWdejEEpnq8aM0tOaw==
 
-"eosjs-api@git+https://github.com/EdgeApp/eosjs-api.git#3676348b6a7443389b28d21c98067760c51e02fb", "eosjs-api@https://github.com/EdgeApp/eosjs-api.git#prepare-script":
+"eosjs-api@https://github.com/EdgeApp/eosjs-api.git#3676348b6a7443389b28d21c98067760c51e02fb", "eosjs-api@https://github.com/EdgeApp/eosjs-api.git#prepare-script":
```
swansontec, line 7268: Irrelevant changes?

swansontec (changes requested):
This looks really good overall. However, I noticed a few smaller things and one optional thing.

yarn.lock
samholmes: Looks like this change was resolved in a commit on develop.

src/components/themed/Carousel.tsx
samholmes: I think for the sake of time, we can keep the impl. I have which does work and is thoroughly dev-tested, and if I have some time during the review time of the sprint, I can try a second attempt before we merge this PR.

swansontec (approved):

PR #3891, opened 2023-01-02
Add opt-out setting for marketing notifications

samholmes:
### CHANGELOG

- Added: Opt-out setting for marketing notifications

### Dependencies

- https://github.com/EdgeApp/edge-push-server/pull/53
- https://github.com/EdgeApp/edge-login-ui-rn/pull/31

### Requirements

- [x] Tested on iOS device

Screen Shot 2023-01-02 at 10 37 47

Screen Shot 2023-01-06 at 14 22 07

---
  - https://app.asana.com/0/0/1203203434290655


src/actions/NotificationActions.ts
```diff
@@ -167,7 +173,7 @@ export function setDeviceSettings(data: DeviceUpdatePayload): ThunkAction<Promis
   return async (dispatch, getState) => {
     const state = getState()
 
-    const deviceToken = await messaging().getToken()
+    const deviceToken = 'foo' // await messaging().getToken()
```
swansontec, line 176: Debugging code?

Idea: We make this field optional on the server for most request, and if it's not provided, we don't update the row. That way we don't need to have it working simply to update the user's preferences.

src/components/scenes/NotificationScene.tsx
```diff
@@ -19,21 +19,23 @@ interface OwnProps {
...
   const dispatch = useDispatch()
 
-  const settings = useSelector(state => state.priceChangeNotifications)
+  const settings = useSelector(state => state.notificationSettings)
   const account = useSelector(state => state.core.account)
 
   const currencyConfigs = useWatch(account, 'currencyConfig')
 
-  const toggleNotifications = useHandler(async () => {
+  const handlePressToggleSetting = useHandler(async (toggleSetting: NotificationSettingToggleSetting) => {
```
swansontec, line 33: There is no point doing `useHander` here since you destroy the optimization by passing inline closures to the `SettingsSwitchRow` components. Maybe make two handlers, each hard-coded for one setting? Or just ditch the `useHandler` wrapper at this point?

swansontec (changes requested):

swansontec:
Also, this PR doesn't seem to address the second requirement, to present an opt-out when asking for notification permissions. That way, users who agree to receive security alerts can choose not to get marketing:

```
Edge needs your permission to send notifications. We will use this to send security alerts.

[x] Also tell me about price changes
[x] Also send me news about the app

[OK]
```

We can spit this requirement off into a second task / PR if we want to keep the scope limited on this PR.

src/actions/NotificationActions.ts
samholmes: Oops, yes debugging code somehow snuck in there. I'll fix this, and as for the idea to make it optional, that sounds fine, but it would need to be scoped in a separate task.

src/components/scenes/NotificationScene.tsx
samholmes: The useHandler is because the handler is a dep for the useMemo. The only other alternative to which makes eslint happy is to remove the useMemo and inline the JSX (which always seemed cleaner to me). I'll add as a new commit to either accept or reject in re-review.

samholmes:
@swansontec Published Login UI PR which is now a dependency for this GUI PR: https://github.com/EdgeApp/edge-login-ui-rn/pull/31

swansontec (approved):
I see what happened. You already squashed your fixes into the branch, so I didn't realize they had already been made. Yes, these fixes look fine.

PR #3899, opened 2023-01-05
William/swift core

swansontec:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203229336962285


scripts/makeNativeHeaders.ts
```diff
@@ -2,18 +2,26 @@ import fs from 'fs'
...
+  // Grab the API key:
   let apiKey = 'Error: Set up env.json & re-run scrips/makeNativeHeaders.js'
   try {
     apiKey = require('../env.json').AIRBITZ_API_KEY
   } catch (e) {
     console.log(apiKey)
   }
 
+  // Grab the push notification server:
+  let pushServer = 'https://push2.edge.app'
```
paullinator, line 14: Do not allow push2.edge.app to be hard coded. We want to ensure we only use the push server specified in the appConfig. Fallback should just be no push servers.

ios/Base58.swift
```diff
@@ -0,0 +1,60 @@
+import Foundation
```
paullinator, line 1: Is this routine borrowed from somewhere? If so, cite the reference or import the library. If this is written from scratch it needs unit tests.

paullinator (changes requested):

ios/Base58.swift
swansontec: I could not figure out how to integrate native tests into the React Native workflow, but I did use them when I implemented the routines: https://github.com/swansontec/native-base58

ios/EdgeCore.swift
```diff
@@ -0,0 +1,104 @@
...
+
+      // List the logins folder:
+      let fs = FileManager.default
+      guard let documentUrl = fs.urls(for: .documentDirectory, in: .userDomainMask).first else {
+        return completion(nil)
+      }
+      let loginsUrl = documentUrl.appendingPathComponent("logins").absoluteURL
+      let files = try fs.contentsOfDirectory(
+        at: loginsUrl, includingPropertiesForKeys: nil, options: [])
+
```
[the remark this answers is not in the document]
swansontec: Line 163 actually deals with errors, not an empty directory. I can still add the check, though, since it's easy to do & makes sense.

paullinator (approved):

PR #3901, opened 2023-01-06
Fix future unstake estimate

paullinator:
Force use of the exact unstake amount when simulating an unstake. The 'unstake' directive also includes the earned amount of the sample Saver which will cause much more slippage.

### CHANGELOG

none

### Dependencies

none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203655596713996

### DIFF
--- a/src/plugins/stake-plugins/types.ts
+++ b/src/plugins/stake-plugins/types.ts
@@ -77,7 +77,7 @@ export interface StakePolicy {
 // Change Quote
 // -----------------------------------------------------------------------------
 export interface ChangeQuoteRequest {
-  action: 'stake' | 'unstake' | 'claim'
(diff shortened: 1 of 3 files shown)


src/plugins/stake-plugins/types.ts
```diff
@@ -77,7 +77,7 @@ export interface StakePolicy {
 // Change Quote
 // -----------------------------------------------------------------------------
 export interface ChangeQuoteRequest {
-  action: 'stake' | 'unstake' | 'claim'
+  action: 'stake' | 'unstake' | 'claim' | 'unstakeexact'
```
samholmes, line 80: 'unstake-exact' or 'unstakeExact' to keep it consistent with other enumeration string naming.

samholmes (approved):

PR #3902, opened 2023-01-06
No Segwit DGB/LTC in GUI

samholmes:
### CHANGELOG

- Added: Non-segwit wallet options for Litecoin and Digibyte to support Jaxx wallet imports

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203651461107953

### DIFF
--- a/src/util/CurrencyInfoHelpers.ts
+++ b/src/util/CurrencyInfoHelpers.ts
@@ -61,22 +61,22 @@ export function getCreateWalletTypes(account: EdgeAccount, filterActivation: boo
 
   const out: CreateWalletType[] = []
   for (const currencyInfo of infos) {
-    const { currencyCode, pluginId } = currencyInfo
+    const { currencyCode, displayName, pluginId, walletType } = currencyInfo
     // Prevent plugins that are "watch only" from being allowed to create new wallets
     if (keysOnlyModePlugins.includes(pluginId)) continue
     // Prevent currencies that needs activation from being created from a modal
     if (filterActivation && activationRequiredCurrencyCodes.includes(currencyCode.toUpperCase())) continue
     // FIO disable changes
-    if (pluginId === 'bitcoin') {
+    if (['bitcoin', 'litecoin', 'digibyte'].includes(pluginId)) {
       out.push({
-        currencyName: 'Bitcoin (Segwit)',
-        walletType: 'wallet:bitcoin-bip49',
+        currencyName: `${displayName} (Segwit)`,
+        walletType: `${walletType}-bip49`,
         pluginId,
         currencyCode
       })
       out.push({
-        currencyName: 'Bitcoin (no Segwit)',
-        walletType: 'wallet:bitcoin-bip44',
+        currencyName: `${displayName} (no Segwit)`,
+        walletType: `${walletType}-bip44`,
         pluginId,
         currencyCode
       })


src/components/scenes/RequestScene.tsx
```diff
@@ -131,17 +129,13 @@ export class RequestSceneComponent extends React.Component<Props, State> {
...
     if (wallet != null) {
       const receiveAddress = await wallet.getReceiveAddress()
-      // @ts-expect-error
-      legacyAddress = receiveAddress.legacyAddress
       publicAddress = receiveAddress.publicAddress
     }
     this.setState({
-      publicAddress,
-      legacyAddress
+      publicAddress
```
paullinator: I'm ok with this cleanup but note that we will have to add some type of state to track legacy vs public, vs segwit address that the user is looking at. good for now

src/components/scenes/RequestScene.tsx
```diff
@@ -184,15 +176,12 @@ export class RequestSceneComponent extends React.Component<Props, State> {
...
-        legacyAddress = s.strings.loading
       }
 
       // eslint-disable-next-line react/no-did-update-set-state
       this.setState({
         encodedURI,
-        publicAddress: publicAddress,
-        // @ts-expect-error
-        legacyAddress: legacyAddress
+        publicAddress: publicAddress
```
paullinator: simplify to just `publicAddress`

src/__tests__/scenes/__snapshots__/RequestScene.test.tsx.snap
```diff
@@ -164,9 +164,7 @@ exports[`Request should render with loaded props 1`] = `
         </TouchableOpacity>
       </View>
     </InputAccessoryView>
-    <QrCode
-      onPress={[Function]}
-    />
+    <QrCode />
```
paullinator: Why is this being removed? On small phones this is useful to show a full size qr code.

src/components/themed/WalletListSortable.tsx
```diff
@@ -44,6 +44,7 @@ export function WalletListSortable(props: Props) {
       order={walletOrder}
       onRowMoved={handleRowMoved}
       renderRow={(wallet: EdgeCurrencyWallet | undefined) => <WalletListSortableRow wallet={wallet} />}
+      disableAnimatedScrolling
```
paullinator, line 46: Can you describe what this fixes? Or show a before/after video?

src/components/themed/ShareButtons.tsx
```diff
@@ -32,7 +32,7 @@ function ShareButton(props: { text: string; onPress: () => void; icon: string })
   const styles = getStyles(theme)
 
   return (
-    <TouchableOpacity style={styles.button} onPress={onPress}>
+    <TouchableOpacity style={styles.button} onPress={() => onPress()}>
```
paullinator, line 35: This would cause unnecessary re-renders. How about

```
const handleOnPress = useHandler(() => onPress())

return (
    <TouchableOpacity style={styles.button} onPress={handleOnPress}>	    <TouchableOpacity style={styles.button} onPress={() => onPress()}>
```

paullinator (approved):

src/components/themed/WalletListSortable.tsx
paullinator: 👍

PR #3903, opened 2023-01-06
Fix sendScene to update amount when last character is deleted

paullinator:
Prevents spending with the previous amount

### CHANGELOG

none

### Dependencies

none

### Requirements

- [ ] Tested on iOS device


swansontec (approved):
Ah, yes, this will work. The routine will drop down to the bottom, fail the `makeSpend` call, and then all the error logic will kick in to disable the slider.

PR #3905, opened 2023-01-06
Add close arrow button to drawer

paullinator:
### CHANGELOG

Add close arrow to side menu

https://user-images.githubusercontent.com/6079354/211218589-3b5458fa-56bd-4726-ac9f-d8d6dc58aae8.mp4

### Dependencies

none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203571252349535


swansontec (changes requested):
The code structure is good, but visually I think the X icon needs to be lower down. If you compare the drawer with a modal, you can see how the position differs. Also, putting the X lower down would put it in a place where the gradient is thicker, so it won't seem to be floating as much.

swansontec (approved):
This looks fantastic now!

PR #3906, opened 2023-01-06
Show 'Delete Account' in the settings scene when locked

itayplav:
### CHANGELOG

Show 'Delete Account' in the settings scene when locked

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

|                                                      Locked                                                     |                                                      Unlocked                                                     |
|:---------------------------------------------------------------------------------------------------------------:|:-----------------------------------------------------------------------------------------------------------------:|
| locked | unlocked |

---
  - https://app.asana.com/0/0/1203451475282025

### DIFF
--- a/src/components/scenes/SettingsScene.tsx
+++ b/src/components/scenes/SettingsScene.tsx
@@ -1,7 +1,7 @@
 import { EdgeAccount, EdgeContext, EdgeLogType } from 'edge-core-js'
 import { getSupportedBiometryType } from 'edge-login-ui-rn'
 import * as React from 'react'
-import { ScrollView, Text } from 'react-native'
+import { ScrollView } from 'react-native'
 import FontAwesomeIcon from 'react-native-vector-icons/FontAwesome'
 import IonIcon from 'react-native-vector-icons/Ionicons'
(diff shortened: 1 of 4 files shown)


src/components/scenes/SettingsScene.tsx
```diff
@@ -24,15 +24,14 @@ import { config } from '../../theme/appConfig'
 import { connect } from '../../types/reactRedux'
 import { NavigationProp } from '../../types/routerTypes'
 import { secondsToDisplay } from '../../util/displayTime'
-import { Collapsable } from '../common/Collapsable'
```
swansontec, line 27: This component isn't used anywhere else, so you could delete `Collapsable.tsx` as well.

swansontec (approved):
Approved, but with one optional clean-up before you merge.

src/components/scenes/SettingsScene.tsx
itayplav: I disagree, this component is well implemented as a simple roll-in, roll-out transition for short-lived components; eg the scam warning card in the Send Scene, which to me currently seems too solidified a part of the scene... before vanishing after pasting an address.

PR #3907, opened 2023-01-06
Add max swaps

peachbits:
---
  - https://app.asana.com/0/0/1203660431077319


src/components/scenes/CryptoExchangeScene.tsx
```diff
@@ -212,37 +212,39 @@ export class CryptoExchangeComponent extends React.Component<Props, State> {
     return false
   }
 
-  getQuote = () => {
+  onNext = () => {
```
swansontec, line 204: This should be called `handleNext`. There is a lint rule we eventually want to enable that requires this naming like `<X onFoo={handleWhatever} />` for all JSX callbacks.

Otherwise, this is a fantastic cleanup!

src/components/scenes/CryptoExchangeScene.tsx
```diff
@@ -400,9 +415,7 @@ export class CryptoExchangeComponent extends React.Component<Props, State> {
             focusMe={this.focusFromWallet}
             onNext={this.onNext}
           >
-            {this.props.hasMaxSpend && (
-              <MiniButton alignSelf="center" label={s.strings.string_max_cap} marginRem={[1.2, 0, 0]} onPress={this.props.exchangeMax} />
-            )}
+            {this.props.hasMaxSpend && <MiniButton alignSelf="center" label={s.strings.string_max_cap} marginRem={[1.2, 0, 0]} onPress={this.onMax} />}
```
swansontec, line 392: Optional: Since you are editing this line, it might be good to fix it to `hasMaxSpend ? <MiniButton ... > : null`. That way we are being explicit about rendering or not rendering, vs abusing the `&&` operator.

src/components/scenes/CryptoExchangeScene.tsx
```diff
@@ -408,7 +382,7 @@ export class CryptoExchangeComponent extends React.Component<Props, State> {
             secondaryCurrencyInfo={fromSecondaryInfo}
             fiatPerCrypto={this.props.fromFiatToCrypto}
             overridePrimaryExchangeAmount={this.state.fromExchangeAmount}
-            forceUpdateGuiCounter={this.state.forceUpdateGuiCounter}
+            forceUpdateGuiCounter={0}
```
swansontec, line 385: This prop is now unused. You can delete it from `CryptoExchangeFlipInputWrapper` and even `ExchangedFlipInput` itself.

src/components/scenes/CryptoExchangeScene.tsx
```diff
@@ -179,25 +172,6 @@ export class CryptoExchangeComponent extends React.Component<Props, State> {
...
-        forceUpdateGuiCounter: props.forceUpdateGuiCounter
-      }
-    } else {
-      // Check which wallet we are currently editing.
-      // Only change the exchangeAmount of the opposite wallet to prevent feedback loops
-      if (state.whichWalletFocus === 'from') {
-        return { toExchangeAmount: props.toExchangeAmount }
-      } else {
-        return { fromExchangeAmount: props.fromExchangeAmount }
-      }
```
swansontec, line 197: Deleting this `else` logic worries me. It doesn't seem related to the max send feature, but to general flip input instability.

src/constants/WalletAndCurrencyConstants.ts
```diff
@@ -300,7 +300,6 @@ export const SPECIAL_CURRENCY_INFO: {
     initWalletName: s.strings.string_first_monero_wallet_name,
     chainCode: 'XMR',
     dummyPublicAddress: '46qxvuS78CNBoiiKmDjvjd5pMAZrTBbDNNHDoP52jKj9j5mk6m4R5nU6BDrWQURiWV9a2n5Sy8Qo4aJskKa92FX1GpZFiYA',
-    noMaxSpend: true,
```
swansontec, line 303: Assuming we merge the other Monero PR's, this commit message needs to change.

swansontec (changes requested):

src/components/scenes/CryptoExchangeScene.tsx
peachbits: Good catch, reverted that part

src/components/scenes/CryptoExchangeScene.tsx
```diff
@@ -212,6 +212,21 @@ export class CryptoExchangeComponent extends React.Component<Props, State> {
     return false
   }
 
+  onMax = () => {
```
swansontec, line 197: `onMax` should be called `handleMax`.

swansontec (changes requested):

swansontec (approved):
Thanks, this looks great now!

PR #3915, opened 2023-01-10
Fix the legacy FlipInput component

swansontec:
This guy relies on some bad hacks to work correctly. We removed those hacks in 56135da548dee880be7a21e2786d270363f6b901, but now we are putting them back in a more explicit way.

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203661324582201


paullinator (approved):
Given that we will remove the old FlipInput, this hack works for now.

PR #3923, opened 2023-01-12
Allow patch file apply failure to disrupt the build

peachbits:


scripts/patchFiles.ts
```diff
@@ -9,7 +9,7 @@ const _rootProjectDir = join(__dirname, '../')
 
 let _currentPath = __dirname
 
-main().catch(e => console.log(e.message))
+main()
```
swansontec, line 12: I think you need it to be like this:

```
main().catch(error => {
  console.error(error)
  process.exit(1)
})
```

Otherwise, it will simply be an "unhandled rejection" and not actually exit the process with an error code like we want.

swansontec (changes requested):

swansontec (approved):

PR #3927, opened 2023-01-14
Replace Actions with navigation in components

Jon-edge:
### CHANGELOG

- Deprecate react-native-router-flux in most components

### Dependencies

sam/navigation-actions

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203659728447756

### DIFF
--- a/src/actions/DeepLinkingActions.ts
+++ b/src/actions/DeepLinkingActions.ts
@@ -136,7 +136,7 @@ function handleLink(navigation: NavigationBase, dispatch: Dispatch, state: RootS
     }
 
     case 'price-change': {
-      dispatch(launchPriceChangeBuySellSwapModal(link))
+      dispatch(launchPriceChangeBuySellSwapModal(navigation, link))
       return true
     }
 
--- a/src/components/Main.ui.tsx
+++ b/src/components/Main.ui.tsx
@@ -43,7 +43,7 @@ import { withServices } from './hoc/withServices'
 import { BackButton } from './navigation/BackButton'
 import { CurrencySettingsTitle } from './navigation/CurrencySettingsTitle'
 import { EdgeLogoHeader } from './navigation/EdgeLogoHeader'
-import { handlePluginBack, renderPluginBackButton } from './navigation/GuiPluginBackButton'
+import { handlePluginBack, PluginBackButton } from './navigation/GuiPluginBackButton'
 import { HeaderTextButton } from './navigation/HeaderTextButton'
 import { HeaderTitle } from './navigation/HeaderTitle'
 import { SideMenuButton } from './navigation/SideMenuButton'
@@ -385,8 +385,7 @@ export class MainComponent extends React.Component<Props> {
                 component={withNavigation(ifLoggedIn(GuiPluginViewScene))}
                 navTransparent
                 renderTitle={props => <HeaderTitle title={props.route.params.plugin.displayName} />}
-                // @ts-expect-error
-                renderLeftButton={renderPluginBackButton()}
+                renderLeftButton={PluginBackButton}
                 // @ts-expect-error
                 renderRightButton={<HeaderTextButton type="exit" placement="right" />}
                 hideTabBar
@@ -395,8 +394,7 @@ export class MainComponent extends React.Component<Props> {
                 key="guiPluginEnterAmount"
                 component={withNavigation(ifLoggedIn(FiatPluginEnterAmountScene))}
                 navTransparent
-                // @ts-expect-error
-                renderLeftButton={renderPluginBackButton()}
+                renderLeftButton={PluginBackButton}
                 hideTabBar
               />
             </Stack>
@@ -417,8 +415,7 @@ export class MainComponent extends React.Component<Props> {
                 component={withNavigation(ifLoggedIn(GuiPluginViewScene))}
                 navTransparent
                 renderTitle={props => <HeaderTitle title={props.route.params.plugin.displayName} />}
-                // @ts-expect-error
-                renderLeftButton={renderPluginBackButton()}
+                renderLeftButton={PluginBackButton}
                 // @ts-expect-error
                 renderRightButton={<HeaderTextButton type="exit" placement="right" />}
                 hideTabBar
@@ -715,8 +712,7 @@ export class MainComponent extends React.Component<Props> {
               component={withNavigation(ifLoggedIn(GuiPluginViewScene))}
               navTransparent
               renderTitle={props => <HeaderTitle title={props.route.params.plugin.displayName} />}
-              // @ts-expect-error
-              renderLeftButton={renderPluginBackButton()}
+              renderLeftButton={PluginBackButton}
               // @ts-expect-error
               renderRightButton={<HeaderTextButton type="exit" placement="right" />}
             />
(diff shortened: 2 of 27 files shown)


src/actions/DeepLinkingActions.ts
```diff
@@ -147,7 +147,8 @@ function handleLink(dispatch: Dispatch, state: RootState, link: DeepLink): boole
 
       // If we don't know what this is, fake a barcode scan:
       if (currencyCode == null) {
-        dispatch(parseScannedUri(link.uri))
+        dispatch(parseScannedUri(navigation, link.uri))
+        dispatch(parseScannedUri(navigation, link.uri))
```
samholmes, line 151: Added a dup?

src/components/scenes/Loans/LoanCreateConfirmationScene.tsx
```diff
@@ -39,7 +39,7 @@ const FEE_VOLATILITY_MULTIPLIER: { [network: string]: string } = {
 }
 
 interface Props {
-  navigation: NavigationProp<'loanCreateConfirmation'>
+  navigation: NavigationBase
```
samholmes, line 42: Revert using `NavigationBase` generic for scene components. The scene is the only component where `NavigationProp<Name>` makes sense because it's the type for that named scene. Where `NavigationBase` makes sense is in components which will take any ol' navigation, and wont expect scene specific information in the navigation object it gets.

src/components/scenes/FioStakingOverviewScene.tsx
```diff
@@ -81,10 +83,10 @@ export const FioStakingOverviewSceneComponent = (props: Props) => {
   }, [stakingStatus, currencyDenomination])
 
   const handlePressStake = () => {
-    Actions.jump('fioStakingChange', { change: 'add', currencyCode, walletId })
+    navigation.jump('fioStakingChange', { change: 'add', currencyCode, walletId })
```
samholmes, line 86: No such thing as `navigation.jump`, replace with `navigation.navigate`. Do a search and replace for anywhere you missed (so far there are just two in FioStatkingOverviewScene, but this comment stands to address all of the occurances).

src/components/scenes/Loans/LoanCreateConfirmationScene.tsx
samholmes: Address this for all the scenes you changeed.

src/components/services/AccountCallbackManager.tsx
```diff
@@ -63,8 +64,8 @@ export function AccountCallbackManager(props: Props) {
       }),
 
       watchSecurityAlerts(account, hasAlerts => {
-        if (hasAlerts && Actions.currentScene !== 'securityAlerts') {
-          Actions.push('securityAlerts', {})
+        if (hasAlerts && navigation.state.currentScene !== 'securityAlerts') {
```
samholmes, line 67: `navigation.state.currentScene` is not going to be a thing. Sorry.. This was one of the things that we decided we're going to keep as `Actions.` API left-over until the flip to the new lib.

src/components/services/DeepLinkingManager.ts
```diff
@@ -27,13 +31,13 @@ export function DeepLinkingManager(props: Props) {
     if (pendingDeepLink == null) return
 
     // Wait a bit, since logging in can sometimes stomp us:
-    requestAnimationFrame(() => dispatch(retryPendingDeepLink()))
-  }, [accountReferralLoaded, dispatch, pendingDeepLink, wallets])
+    requestAnimationFrame(() => dispatch(retryPendingDeepLink(navigation)))
+  }, [accountReferralLoaded, dispatch, navigation, pendingDeepLink, wallets])
```
samholmes, line 35: Ooh, this looks like a fixup for me to do in the base PR.

src/components/themed/ControlPanel.tsx
```diff
@@ -5,6 +5,7 @@ import * as React from 'react'
 import { Image, Platform, Pressable, ScrollView, TouchableOpacity, View } from 'react-native'
 import LinearGradient from 'react-native-linear-gradient'
 import Animated, { Easing, useAnimatedStyle, useSharedValue, withTiming } from 'react-native-reanimated'
+import { Actions } from 'react-native-router-flux'
```
samholmes, line 8: Should import from `./routerTypes`

src/components/themed/ControlPanel.tsx
```diff
@@ -42,7 +43,7 @@ import { TitleText } from '../text/TitleText'
 import { DividerLine } from './DividerLine'
 
 interface Props {
-  navigation: NavigationProp<'controlPanel'>
+  navigation: NavigationBase
```
samholmes, line 46: This is not a scene component, but it might make sense to be more strict with the type here. Keep it as it was?

src/components/themed/TransactionListTop.tsx
```diff
@@ -156,10 +156,10 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
   handleOpenWalletListModal = () => {
     triggerHaptic('impactLight')
     const { navigation } = this.props
-    Airship.show<WalletListResult>(bridge => <WalletListModal bridge={bridge} headerTitle={s.strings.select_wallet} />).then(
+    Airship.show<WalletListResult>(bridge => <WalletListModal bridge={bridge} headerTitle={s.strings.select_wallet} navigation={navigation} />).then(
       ({ walletId, currencyCode }: WalletListResult) => {
         if (walletId != null && currencyCode != null) {
-          navigation.setParams({ currencyCode, walletId })
+          navigation.replace('transactionList', { currencyCode, walletId })
```
samholmes, line 162: Did you test that `.replace` behaves the same as `.setParams`? I wonder if this is equivelent. It'd be safer to keep the navigation prop type as it is so you can just use `setParams` for the scene that TransactionListTop expects to be mounted on.

src/components/themed/WalletListFooter.tsx
```diff
@@ -6,15 +6,15 @@ import { SPECIAL_CURRENCY_INFO } from '../../constants/WalletAndCurrencyConstant
...
+import { NavigationBase } from '../../types/routerTypes'
 import { EdgeTokenId } from '../../types/types'
 import { WalletListModal, WalletListResult } from '../modals/WalletListModal'
 import { Airship, showError } from '../services/AirshipInstance'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 import { EdgeText } from './EdgeText'
 
 interface Props {
-  navigation: NavigationProp<'walletList'>
+  navigation: NavigationBase
```
samholmes, line 17: The more I review, the more I feel uncomfortable widening the type for these components. It's not a breaking API change, but it may be a breaking implementation change, and what's worse is that we wont see the bug until a dev mis-uses the component... what's worse is this bug may not crop up during static analysis, but instead during runtime which may even be in production. See what I mean?

src/components/Main.ui.tsx
```diff
@@ -388,9 +384,8 @@ export class MainComponent extends React.Component<Props> {
                 key="pluginViewBuy"
                 component={withNavigation(ifLoggedIn(GuiPluginViewScene))}
                 navTransparent
-                renderTitle={props => <HeaderTitle title={props.route.params.plugin.displayName} />}
-                // @ts-expect-error
-                renderLeftButton={renderPluginBackButton()}
+                renderTitle={props => <HeaderTitle navigation={props.navigation} title={props.route.params.plugin.displayName} />}
+                renderLeftButton={renderPluginBackButton}
```
samholmes, line 388: If `renderPluginBackButton` is a component, it should be moved to the `components/` dir and be renamed using pascal-casing.

samholmes (changes requested):

src/components/themed/TransactionListTop.tsx
samholmes: Still needs fixing.

samholmes (approved):
Awesome job!

PR #3929, opened 2023-01-14
Add Coin rank scene

paullinator:
### CHANGELOG

Add a coin rank "Markets" tab that shows coins by market cap

simulator_screenshot_91E0BB65-5D14-4D43-9BCD-FAA9846E90DF

### Dependencies

none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203327829237148


src/components/scenes/CoinRankingScene.tsx
```diff
@@ -0,0 +1,290 @@
...
+      if (searchText === '') {
+        out.push(i)
+      } else {
+        if (cr.currencyCode.toLowerCase().includes(searchText.toLowerCase()) || cr.currencyName.toLowerCase().includes(searchText.toLowerCase())) {
+          out.push(i)
+        }
+      }
+    }
+    return out
+  }, [dataSize, searchText])
```
Jon-edge, line 173: Address exhaustive dep warning

```suggestion
    }, [coinRankings, dataSize, searchText])
```

src/components/scenes/CoinRankingScene.tsx
```diff
@@ -0,0 +1,290 @@
...
+
+    return <CoinRankRow index={item} key={key} coinRowData={coinRowData} percentChangeTimeFrame={percentChangeTimeFrame} priceSubText={priceSubText} />
+  })
+
+  const handleEndReached = useHandler(() => {
+    debugLog(LOG_COINRANK, `handleEndReached. setRequestDataSize ${requestDataSize + QUERY_PAGE_SIZE}`)
+    setRequestDataSize(requestDataSize + QUERY_PAGE_SIZE)
+  })
+
+  const handlePercentChange = useHandler(() => {
```
Jon-edge, line 80: This control challenges common assumptions. Tapping a column title without any other affordances I would expect to sort the list based on the column. 

I understand that we have a convention in our app that certain colors are tappable, and colors are good as secondary indicators, but colors alone are bad at conveying information.

I propose following a more common convention such as the one shown on CoinGecko's app with two rows

1. A radio dropdown when tapping this title allowing the user to select the timeframe, along with a visual down triangle indicator there are more options on the first row:
7d▼

2. The column title on the 2nd row along with an indicator displaying the sort status to the right of the column title, either up triangle, down triangle, or both (ascending sort, descending sort, none).
7d△

src/components/scenes/CoinRankingScene.tsx
```diff
@@ -0,0 +1,290 @@
...
+const percentChangeStrings: { [pc: string]: string } = {
+  hours1: '1hr',
+  hours24: '24hr',
+  days7: '7d',
+  days30: '30d',
+  year1: '1y'
+}
+const priceSubTextStrings: { [pc: string]: string } = {
+  marketCap: 'MCap',
+  volume24h: 'Vol 24h'
```
Jon-edge, line 44: Do we need to add these to enus for translation?

src/components/scenes/CoinRankingScene.tsx
```diff
@@ -0,0 +1,290 @@
...
+      console.error(msg)
+      showError(msg)
+      return
+    }
+    const newIndex = index + 1 >= percentChangeOrder.length ? 0 : index + 1
+    const newTimeFrame = percentChangeOrder[newIndex]
+    setPercentChangeTimeFrame(newTimeFrame)
+  })
+
+  const handlePriceSubText = useHandler(() => {
```
Jon-edge, line 93: Similar to the comment on handlePercentChange but more confusing. Vol and mcap are not similar categories yet they occupy the same  column. 

Since volume is given only in 24h terms, we potentially can replace the 2nd line of the 'Asset' space with the volume. I think it's acceptable to drop the name because it's common across all many crypto apps to hide names in an overview list. This proposal also aligns with Coinbase's 'trade' view.

Alternatively(1), we add a price column where we have subtext there that shows the volume, but this I'd prefer seeing percent change along with price instead.

Jon-edge: In order of importance IMO:
1. Currency Code OR Currency Name, not both
2. Price
3. % Change
4. 24h Vol
5. Mcap

More details like the name of the asset, market cap, can be hidden away until tapped

Jon-edge: If we go with a 2-row tile per item design, we can't really utilize two rows without the content contradicting the column title labels. The only exception to this is if we do a 2 row asset cell with name/CC, which I'm not really a fan of.
To make a 2-row design work, we'd need to move the title labels to the elements themselves (i.e. 'BTC\nVol 31.26k'), but then sorting would need to be condensed into one monolithic control. IMO this isn't very intuitive.

An argument could be made that if this is to be an overview, we can trim down the info and fit more items on the list by utilizing a 1 row design.

src/components/scenes/CoinRankingScene.tsx
paullinator: Let's give it a shot. I'm afraid translators will botch this.

Jon-edge (approved):

PR #3931, opened 2023-01-16
Use "biometric" verbiage for Android devices

samholmes:
### CHANGELOG

- Fixed: Settings scene correctly shows "Use Biometric" on Android devices

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203428625085680

### DIFF
--- a/src/components/scenes/SettingsScene.tsx
+++ b/src/components/scenes/SettingsScene.tsx
@@ -1,7 +1,7 @@
 import { EdgeAccount, EdgeContext, EdgeLogType } from 'edge-core-js'
 import { getSupportedBiometryType } from 'edge-login-ui-rn'
 import * as React from 'react'
-import { ScrollView } from 'react-native'
+import { Platform, ScrollView } from 'react-native'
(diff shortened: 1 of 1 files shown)


src/components/scenes/SettingsScene.tsx
```diff
@@ -105,17 +105,14 @@ export class SettingsSceneComponent extends React.Component<Props, State> {
     if (!this.props.supportsTouchId) return
 
     const biometryType = await getSupportedBiometryType()
+    if (Platform.OS === 'android') this.setState({ touchIdText: s.strings.settings_button_use_biometric })
```
swansontec, line 108: This is the opposite of what we talked about, which is to make iOS the special case. You could fix this by doing `if (Platform.OS !== 'ios') ...biometric`, or something like that. Any platform other than iOS (Desktop?) should just use "biometric"

swansontec (changes requested):
This seems right, but we need to flip the logic around a bit.

swansontec (approved):

PR #3933, opened 2023-01-17
Add note support to Tron

peachbits:
---
  - https://app.asana.com/0/0/1203757781046863


peachbits:
Depends on updated edge-currency-accountbased with https://github.com/EdgeApp/edge-currency-accountbased/pull/491

src/constants/WalletAndCurrencyConstants.ts
```diff
@@ -381,6 +381,11 @@ export const SPECIAL_CURRENCY_INFO: {
       privateKeyLabel: s.strings.create_wallet_import_input_key_or_seed_prompt,
       privateKeyInstructions: s.strings.create_wallet_import_input_key_or_seed_instructions
     },
+    uniqueIdentifierInfo: {
+      addButtonText: s.strings.unique_identifier_dropdown_option_note,
+      identifierName: s.strings.unique_identifier_note,
+      keyboardType: 'default'
+    },
```
swansontec, line 388: I see how this is needed, since different coins call the memo different things, but I would have assumed `EdgeCurrencyInfo.memoType` would play into this more, making the keyboard type unnecessary. I know Jon spent a bunch of time working on this; it might be worth talking to him to see if this can be removed.

swansontec (approved):

src/constants/WalletAndCurrencyConstants.ts
peachbits: Only the maxLength param is used in the gui currently.

PR #3934, opened 2023-01-17
Disable the launch explorer functionality and hide the chevron if the address explorer url is an empty string

peachbits:
---
  - https://app.asana.com/0/0/1203704965022950

### DIFF
--- a/src/components/scenes/RequestScene.tsx
+++ b/src/components/scenes/RequestScene.tsx
@@ -292,6 +292,7 @@ export class RequestSceneComponent extends React.Component<Props, State> {
     const requestAddress = selectedAddress?.addressString ?? s.strings.loading
     const flipInputHeaderText = sprintf(s.strings.send_to_wallet, getWalletName(wallet))
     const { keysOnlyMode = false } = getSpecialCurrencyInfo(wallet.currencyInfo.pluginId)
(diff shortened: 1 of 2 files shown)


src/components/scenes/RequestScene.tsx
```diff
@@ -370,10 +371,10 @@ export class RequestSceneComponent extends React.Component<Props, State> {
               />
             )}
           />
-          <TouchableOpacity onPress={this.handleAddressBlockExplorer}>
+          <TouchableOpacity disabled={addressExplorerDisabled} onPress={this.handleAddressBlockExplorer}>
             <View style={styles.rightChevronContainer}>
               <EdgeText>{selectedAddress?.label ?? s.strings.request_qr_your_wallet_address}</EdgeText>
-              <IonIcon name="chevron-forward" size={theme.rem(1.5)} color={theme.iconTappable} />
+              {!addressExplorerDisabled ? <IonIcon name="chevron-forward" size={theme.rem(1.5)} color={theme.iconTappable} /> : null}
```
swansontec, line 377: Optional: I generally like to put the `null` case first, so `addressExplorerDisabled ? null : <IonIcon .../>` would be easier to read.

swansontec (approved):

PR #3935, opened 2023-01-17
Add rbf text option to transaction row

peachbits:
### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203444039875736

### DIFF
--- a/src/components/themed/TransactionRow.tsx
+++ b/src/components/themed/TransactionRow.tsx
@@ -4,15 +4,16 @@ import FastImage from 'react-native-fast-image'
 import Ionicons from 'react-native-vector-icons/Ionicons'
 import { sprintf } from 'sprintf-js'
 
+import { useHandler } from '../../hooks/useHandler'
 import s from '../../locales/strings'
 import { TransactionListTx } from '../../types/types'
 import { formatCategory, splitCategory } from '../../util/categories'
 import { triggerHaptic } from '../../util/haptic'
-import { cacheStyles, Theme, ThemeProps, withTheme } from '../services/ThemeContext'
+import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 import { ClickableRow } from './ClickableRow'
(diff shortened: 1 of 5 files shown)


peachbits:
Simulator Screen Shot - iPod touch (7th generation) - 2023-01-17 at 00 21 54 #2

src/components/themed/TransactionRow.tsx
```diff
@@ -45,6 +47,11 @@ const TransactionRowComponent = (props: Props) => {
   const theme = useTheme()
   const styles = getStyles(theme)
 
+  const account = useSelector(state => state.core.account)
+  const currencyWallets = useWatch(account, 'currencyWallets')
+  const wallet = currencyWallets[transaction.walletId]
+  const canReplaceByFee = !!wallet.currencyInfo.canReplaceByFee
```
swansontec, line 53: Maybe `const { canReplaceByFee = false } = wallet.currencyInfo`?

src/locales/en_US.ts
```diff
@@ -192,6 +192,7 @@ const strings = {
   fragment_transaction_list_tx_dropped: 'Dropped',
   fragment_transaction_list_tx_synchronizing: 'Synchronizing',
   fragment_transaction_list_confirmation_progress: '%s of %s Confirmations',
+  fragment_transaction_list_unconfirmed_rbf: 'Awaiting confirmation, may be replaced',
```
swansontec, line 195: I think Michael wants wording like "may be cancelled". We want normal people to understand that "you may not get the money". The word "cancelled" does this better than "replace".

swansontec (changes requested):

src/locales/en_US.ts
peachbits: Description says cancelled, his comment says replaced. I agree cancelled is better so I changed it

swansontec (approved):

PR #3938, opened 2023-01-19
Use SendScene2 for EdgeProvider 

paullinator:
### CHANGELOG

none

### Dependencies

https://github.com/EdgeApp/edge-react-gui/pull/3932

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203527749920572

### DIFF
--- a/src/modules/UI/scenes/Plugins/EdgeProvider.tsx
+++ b/src/modules/UI/scenes/Plugins/EdgeProvider.tsx
@@ -1,6 +1,16 @@
 import { abs } from 'biggystring'
 import { asArray, asEither, asObject, asOptional, asString } from 'cleaners'
-import { EdgeCurrencyWallet, EdgeMetadata, EdgeNetworkFee, EdgeParsedUri, EdgeReceiveAddress, EdgeSpendTarget, EdgeTransaction, JsonObject } from 'edge-core-js'
+import {
+  EdgeCurrencyWallet,
+  EdgeMetadata,
+  EdgeNetworkFee,
+  EdgeParsedUri,
+  EdgeReceiveAddress,
+  EdgeSpendInfo,
+  EdgeSpendTarget,
+  EdgeTransaction,
+  JsonObject
+} from 'edge-core-js'
 import * as React from 'react'
 import { Linking, Platform } from 'react-native'
 import { CustomTabs } from 'react-native-custom-tabs'
@@ -10,7 +20,7 @@ import { sprintf } from 'sprintf-js'
 import { Bridgeable, update } from 'yaob'
 
 import { launchPaymentProto } from '../../../../actions/PaymentProtoActions'
-import { trackAccountEvent, trackConversion } from '../../../../actions/TrackingActions'
+import { trackConversion } from '../../../../actions/TrackingActions'
 import { ButtonsModal } from '../../../../components/modals/ButtonsModal'
 import { WalletListModal, WalletListResult } from '../../../../components/modals/WalletListModal'
 import { Airship, showError, showToast } from '../../../../components/services/AirshipInstance'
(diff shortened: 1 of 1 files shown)


src/modules/UI/scenes/Plugins/EdgeProvider.tsx
```diff
@@ -398,19 +403,78 @@ export class EdgeProvider extends Bridgeable {
...
+    spendInfo
+  }: {
+    lockInputs: boolean
+    orderId?: string
+    spendInfo: EdgeSpendInfo
+  }): Promise<EdgeTransaction | undefined> {
+    return new Promise((resolve, reject) => {
+      const lockTilesMap = lockInputs ? { address: true, amount: true, wallet: true } : undefined
+
+      this._navigation.push('send2', {
```
swansontec, line 440: We want to use `navigation.navigate` whenever possible. The `navigate` action "just works", since it figures out what to do based on the current context. Most of our navigation bugs come about because we used one of the other actions, like `navigation.push`, in a context where that may have been incorrect.

For instance, if a plugin scene gets mounted inside the tab navigator, the correct navigation may be a `jump` rather than a `push`. Using `navigate` will automatically figure this out for you.

src/modules/UI/scenes/Plugins/EdgeProvider.tsx
```diff
@@ -398,19 +403,78 @@ export class EdgeProvider extends Bridgeable {
...
+        walletId: this.selectedWallet.id,
+        spendInfo,
+        lockTilesMap,
+        onBack: () => resolve(undefined),
+        onDone: (error, transaction) => {
```
swansontec, line 445: I don't see `onDone` call `resolve`, so this would just leave the promise dangling.

src/modules/UI/scenes/Plugins/EdgeProvider.tsx
```diff
@@ -398,19 +403,78 @@ export class EdgeProvider extends Bridgeable {
...
+    spendInfo: EdgeSpendInfo
+  }): Promise<EdgeTransaction | undefined> {
+    return new Promise((resolve, reject) => {
+      const lockTilesMap = lockInputs ? { address: true, amount: true, wallet: true } : undefined
+
+      this._navigation.push('send2', {
+        walletId: this.selectedWallet.id,
+        spendInfo,
+        lockTilesMap,
+        onBack: () => resolve(undefined),
```
swansontec, line 444: Do we need to call `navigation.pop()` in the `onBack` handler? Or does the scene do this on its own?

src/modules/UI/scenes/Plugins/EdgeProvider.tsx
```diff
@@ -398,19 +403,78 @@ export class EdgeProvider extends Bridgeable {
...
+            reject(error)
+            return
+          }
+          if (transaction == null) {
+            reject(new Error('Missing transaction'))
+            return
+          }
+          this._navigation.pop()
+
+          this.selectedWallet.nativeToDenomination(transaction.nativeAmount, transaction.currencyCode).then(exchangeAmount => {
```
swansontec, line 456: This might cause an unhandled promise rejection. Can we put a `.catch` on the end in case something goes wrong with the conversion tracking? It could be a no-op if we really don't care, like `catch(()=>{})`, but we should at least acknowledge it.

swansontec (changes requested):
This came out pretty well.

src/modules/UI/scenes/Plugins/EdgeProvider.tsx
paullinator: nice insight. will do, thanks

src/modules/UI/scenes/Plugins/EdgeProvider.tsx
paullinator: good catch. fixed

src/modules/UI/scenes/Plugins/EdgeProvider.tsx
paullinator: `onBack` is called when the scene is unmounted so yes it's auto called.

src/modules/UI/scenes/Plugins/EdgeProvider.tsx
paullinator: 👍

swansontec (approved):
This looks good now.

PR #3944, opened 2023-01-20
Sam/dev-firebase-error

samholmes:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203768363890533


src/components/scenes/LoginScene.tsx
```diff
@@ -82,7 +82,7 @@ class LoginSceneComponent extends React.PureComponent<Props, State> {
       // @ts-expect-error
       .catch(e => this.setState({ backgroundImage: theme.backgroundImage }))
     const { YOLO_USERNAME, YOLO_PASSWORD, YOLO_PIN } = ENV
-    if (YOLO_USERNAME != null && (YOLO_PASSWORD != null || YOLO_PIN != null) && firstRun) {
+    if (YOLO_USERNAME != null && (Boolean(YOLO_PASSWORD) || Boolean(YOLO_PIN)) && firstRun) {
```
swansontec, line 85: Why not get the username too? Also, this is a pretty cool pattern we might use in more places.

swansontec (approved):
Approved, but with question.

src/components/scenes/LoginScene.tsx
samholmes: Fair point; Username could also be given this treatment. The type-casting pattern is nice, and I'm happy to promote it.

swansontec (approved):

src/components/scenes/LoginScene.tsx
```diff
@@ -82,7 +82,7 @@ class LoginSceneComponent extends React.PureComponent<Props, State> {
       // @ts-expect-error
       .catch(e => this.setState({ backgroundImage: theme.backgroundImage }))
     const { YOLO_USERNAME, YOLO_PASSWORD, YOLO_PIN } = ENV
-    if (YOLO_USERNAME != null && (YOLO_PASSWORD != null || YOLO_PIN != null) && firstRun) {
+    if (!!YOLO_USERNAME && (!!YOLO_PASSWORD || !!YOLO_PIN) && firstRun) {
```
swansontec, line 85: Not approved. Do the `Boolean()` thing instead. We eventually want to enable a lint rule that disallows treating strings and other data types as booleans, such as using `!` or `if` or `?`. The `Boolean()` function encapsulates the `!= null && !== ''` logic in a way ESLint will be happy with.

swansontec (changes requested):
Oh, good call on finally fixing up env.json

swansontec (approved):

PR #3956, opened 2023-01-25
Fix multiple wallet picker popups on deeplink payments

paullinator:
Mark deep link as HANDLED in redux before calling handleLink.

handleLink may take time to handle the link as it may need the user to choose a wallet or send a transaction before the link is "handled". This prevents repeated calls into handleLink while the first call is being handled.

### CHANGELOG

none

### Dependencies

none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203822895117897


swansontec (approved):
Yup, makes sense.

PR #3960, opened 2023-01-29
Add TextDropdown component for confiming account deletion

itayplav:
### CHANGELOG

Show a dropdown indicating that account deletion succeeded. 

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

Screenshot Simulator Screen Recording - iPhone X - 2023-01-29 at 15 07 33 mp4 IINA 2023-01-31 at 02 22 24@2x

---
  - https://app.asana.com/0/0/1203717474770600

### DIFF
--- a/src/components/common/TextDropdown.tsx
+++ b/src/components/common/TextDropdown.tsx
@@ -1,0 +1,32 @@
+import * as React from 'react'
+import { View } from 'react-native'
+import { AirshipBridge } from 'react-native-airship'
+
+import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
+import { EdgeText } from '../themed/EdgeText'
+import { AirshipDropdown } from './AirshipDropdown'
+
+interface Props {
+  bridge: AirshipBridge<void>
+  message: string
+}
+
+export const TextDropdown = ({ bridge, message }: Props) => {
+  const theme = useTheme()
+  const styles = getStyles(theme)
+  return (
+    <AirshipDropdown bridge={bridge} backgroundColor={theme.modal}>
+      <View style={styles.container}>
+        <EdgeText>{message}</EdgeText>
+      </View>
+    </AirshipDropdown>
+  )
+}
+
+const getStyles = cacheStyles((theme: Theme) => ({
+  container: {
+    height: theme.rem(4),
+    justifyContent: 'center',
+    alignItems: 'center'
+  }
+}))
--- a/src/components/scenes/SettingsScene.tsx
+++ b/src/components/scenes/SettingsScene.tsx
@@ -25,12 +25,13 @@ import { connect } from '../../types/reactRedux'
 import { NavigationBase, NavigationProp } from '../../types/routerTypes'
 import { secondsToDisplay } from '../../util/displayTime'
 import { SceneWrapper } from '../common/SceneWrapper'
+import { TextDropdown } from '../common/TextDropdown'
 import { CryptoIcon } from '../icons/CryptoIcon'
 import { Space } from '../layout/Space'
 import { AutoLogoutModal } from '../modals/AutoLogoutModal'
 import { ConfirmContinueModal } from '../modals/ConfirmContinueModal'
 import { TextInputModal } from '../modals/TextInputModal'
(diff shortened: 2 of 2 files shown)


src/components/common/TextDropdown.tsx
```diff
@@ -0,0 +1,30 @@
...
+    <AirshipDropdown bridge={bridge} backgroundColor={theme.modal}>
+      <Text style={styles.text}>{message}</Text>
+    </AirshipDropdown>
+  )
+}
+
+const getStyles = cacheStyles((theme: Theme) => ({
+  text: {
+    // @ts-expect-error
+    ...nightText('row-center'),
```
swansontec, line 27: The `nightText` stuff is all deprecated. If you splat out its contents, you end up with:

```js
text: {
 color: THEME.COLORS.WHITE,
 fontSize: THEME.rem(1),
 textAlign: 'center',
 flexShrink: 1,
 padding: theme.rem(1)
}
```

We are using v1 theme stuff in here, but I believe the `<AirshipDropdown>` itself may be stuck on using v1 theming as well, in which case we kinda have to match it.

src/components/scenes/SettingsScene.tsx
```diff
@@ -175,15 +176,17 @@ export class SettingsSceneComponent extends React.Component<Props, State> {
...
           await this.props.context.deleteLocalAccount(username)
-          showToast(sprintf(s.strings.delete_account_feedback, username))
+
           return true
         }}
       />
-    ))
+    )).then(() => {
+      const message = sprintf(s.strings.delete_account_feedback, username)
+      Airship.show(bridge => <TextDropdown bridge={bridge} message={message} />)
```
swansontec, line 188: The way we launch the dropdown looks great. However, I am worried that it will show even if the user cancels the modal, since we aren't checking the modal's return value.

swansontec (changes requested):

src/components/common/TextDropdown.tsx
itayplav: Yep let's not do that, now uses `<EdgeText>`

image

src/components/scenes/SettingsScene.tsx
itayplav: I have moved the `.then(... <TextDropdown>` after `this.props.context.deleteLocalAccount(username)`. This is more nested but guarantees the modal only shows up after all deletion logic has succeeded.

src/components/scenes/SettingsScene.tsx
```diff
@@ -178,15 +178,14 @@ export class SettingsSceneComponent extends React.Component<Props, State> {
           if (text !== username) return s.strings.delete_account_verification_error
           await this.props.account.deleteRemoteAccount()
           await this.props.logoutRequest()
-          await this.props.context.deleteLocalAccount(username)
-
+          await this.props.context.deleteLocalAccount(username).then(() => {
```
swansontec, line 181: Great, but since this function is already `async`, you can ditch the whole `.then(() =>` bit and just show the modal on the next line.

swansontec (changes requested):

src/components/scenes/SettingsScene.tsx
```diff
@@ -178,14 +178,12 @@ export class SettingsSceneComponent extends React.Component<Props, State> {
...
-            const message = sprintf(s.strings.delete_account_feedback, username)
-            Airship.show(bridge => <TextDropdown bridge={bridge} message={message} />)
-          })
+          await this.props.context.deleteLocalAccount(username)
+          Airship.show(bridge => <TextDropdown bridge={bridge} message={sprintf(s.strings.delete_account_feedback, username)} />)
           return true
         }}
       />
-    ))
+    )).then(() => {})
```
swansontec: This `.then` is useless.

swansontec (approved):
I have made my own fixup, to avoid doing another back & forth on this one.

src/components/scenes/SettingsScene.tsx
itayplav: Thanks, my bad on leaving this one in!

PR #3963, opened 2023-01-30
Audit use of array indices as component keys

peachbits:
---
  - https://app.asana.com/0/0/1203848674296948


swansontec (changes requested):
One tiny change needed.

swansontec (approved):

PR #3964, opened 2023-01-31
Add UI to SendScene for self-transfers

itayplav:
### CHANGELOG

Added the option to send to self when the user has 2 or more wallets/tokens of the same type.

### Dependencies

none

### Requirements

- [x] Tested on iOS device

https://user-images.githubusercontent.com/4023066/215663708-b0082e4f-ad82-49ae-abdb-2cbd99130691.mp4

---
  - https://app.asana.com/0/0/1203374649830361

### DIFF
--- a/src/components/tiles/AddressTile2.tsx
+++ b/src/components/tiles/AddressTile2.tsx
@@ -3,6 +3,7 @@ import { EdgeAccount, EdgeCurrencyConfig, EdgeCurrencyWallet, EdgeParsedUri } fr
 import { ethers } from 'ethers'
 import * as React from 'react'
 import { AppState, TouchableOpacity, View } from 'react-native'
+import AntDesign from 'react-native-vector-icons/AntDesign'
 import FontAwesome from 'react-native-vector-icons/FontAwesome'
 import FontAwesome5 from 'react-native-vector-icons/FontAwesome5'
 
@@ -13,9 +14,11 @@ import { checkPubAddress } from '../../modules/FioAddress/util'
 import { PaymentProtoError } from '../../types/PaymentProtoError'
 import { connect } from '../../types/reactRedux'
 import { NavigationBase } from '../../types/routerTypes'
+import { getTokenId } from '../../util/CurrencyInfoHelpers'
 import { parseDeepLink } from '../../util/DeepLinkParser'
 import { AddressModal } from '../modals/AddressModal'
 import { ScanModal } from '../modals/ScanModal'
+import { WalletListModal, WalletListResult } from '../modals/WalletListModal'
 import { Airship, showError } from '../services/AirshipInstance'
 import { cacheStyles, Theme, ThemeProps, withTheme } from '../services/ThemeContext'
 import { EdgeText } from '../themed/EdgeText'
@@ -166,7 +169,6 @@ export class AddressTileComponent extends React.PureComponent<Props, State> {
 
       // Will throw in case uri is invalid
       await coreWallet.parseUri(uri, currencyCode)
-
       this.setState({
         clipboard: uri,
         loading: false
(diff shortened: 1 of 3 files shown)


src/components/tiles/AddressTile2.tsx
```diff
@@ -40,6 +43,7 @@ interface OwnProps {
 }
 interface StateProps {
   account: EdgeAccount
+  currencyWallets: { [walletId: string]: EdgeCurrencyWallet }
```
swansontec, line 46: We already have the account, so just use `props.account.currencyWallets` instead of `props.currencyWallets`, and delete this.

src/components/tiles/AddressTile2.tsx
```diff
@@ -218,11 +242,18 @@ export class AddressTileComponent extends React.PureComponent<Props, State> {
...
+    const { fioToAddress, recipientAddress, lockInputs, theme, title, currencyWallets, coreWallet } = this.props
+    const { pluginId, currencyCode } = coreWallet.currencyInfo
     const { loading } = this.state
     const styles = getStyles(theme)
     const copyMessage = this.state.clipboard ? `${s.strings.string_paste}: ${this.state.clipboard}` : null
     const tileType = loading ? 'loading' : !!recipientAddress && !lockInputs ? 'delete' : 'static'
+    const tokenId: string | undefined = getTokenId(this.props.account, pluginId, currencyCode)
+    const canSelfTransfer: boolean = Object.keys(currencyWallets).some(
+      walletId =>
+        (walletId !== coreWallet.id && currencyWallets[walletId].type === coreWallet.type) ||
```
swansontec, line 254: After staring carefully at this logic, I think it works, but it's not as readable as I would like. I also think there is a danger that two chains might end up with the same tokenId for different assets (I'm thinking of EOS clones, specifically), and we might show the button when it doesn't make sense.

So, I would propose:

```ts
walletId => {
  if (walletId === coreWallet.id) return false
  if (currencyWallets[walletId].type !== coreWallet.type) return false
  if (tokenId == null) return true
  return currencyWallets[walletId].enabledTokenIds.includes(tokenId)
}
```

That way we ensure we are on the same chain first, before we do anything about tokens.

If you still want a one-liner:


```ts
walletId => 
  walletId !== coreWallet.id &&
  currencyWallets[walletId].type === coreWallet.type && (
    tokenId == null ||
    currencyWallets[walletId].enabledTokenIds.includes(tokenId)
  )
```

src/components/tiles/AddressTile2.tsx
```diff
@@ -232,6 +263,12 @@ export class AddressTileComponent extends React.PureComponent<Props, State> {
                 <FontAwesome name="edit" size={theme.rem(2)} color={theme.iconTappable} />
                 <EdgeText style={styles.buttonText}>{s.strings.enter_as_in_enter_address_with_keyboard}</EdgeText>
               </TouchableOpacity>
+              {canSelfTransfer && (
```
swansontec, line 266: We are trying to avoid the `&&` syntax for this case, since we want to enable the strict-boolean lint rule eventually. Use the `?:` syntax, like `canSelfTransfer ? < ... /> : null `

src/components/tiles/AddressTile2.tsx
```diff
@@ -273,12 +310,16 @@ const getStyles = cacheStyles((theme: Theme) => ({
   buttonText: {
     fontSize: theme.rem(0.75),
     color: theme.textLink
+  },
+  loopbackIcon: {
+    transform: [{ rotate: '90deg' }]
```
swansontec, line 315: Let's use the other icon.

swansontec (changes requested):

src/components/tiles/AddressTile2.tsx
itayplav: Tentatively selected the AntDesign wallet icon.

PR #3968, opened 2023-02-02
Sam/navigation

samholmes:
### CHANGELOG

- Changed: Replaced react-native-router-flux router library with react-navigation

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1201388479070458


src/components/Main.tsx
```diff
@@ -0,0 +1,687 @@
...
+        }}
+      />
+      <Stack.Screen
+        name="transactionDetails"
+        component={ifLoggedIn(TransactionDetailsScene)}
+        options={{
+          headerTitle: () => <TransactionDetailsTitle />
+        }}
+        listeners={{
+          focus: async () => requestPermission('contacts')
```
swansontec, line 213: We don't await here, so why `async`? Maybe we want a `.catch(showError)` on the end, and then the function can return `void` instead of `Promise<void>`.

src/components/Main.tsx
```diff
@@ -0,0 +1,687 @@
...
+      />
+
+      <Stack.Screen
+        name="transactionList"
+        component={ifLoggedIn(TransactionList)}
+        listeners={{
+          focus: async () => requestPermission('contacts')
+        }}
+        options={{
+          title: ' '
```
swansontec, line 259: Single space? A blank string is the default already.

src/components/Main.tsx
```diff
@@ -0,0 +1,687 @@
...
+        component={ifLoggedIn(TransactionsExportScene)}
+        options={{
+          title: s.strings.title_export_transactions,
+          headerRight: () => null
+        }}
+      />
+      <Stack.Screen
+        name="pluginViewBuy"
+        // @ts-expect-error-error
+        component={ifLoggedIn(GuiPluginViewScene)}
```
swansontec, line 295: Go to GuiPluginViewScene, line 119 in `OwnProps`, and change the route prop to `route: RouteProp<'pluginView' | 'pluginViewBuy' | 'pluginViewSell'>`. Now your `// @ts-expect-error` goes away!

src/components/Main.tsx
```diff
@@ -0,0 +1,687 @@
...
+      <Stack.Screen name="stakeOptions" component={ifLoggedIn(StakeOptionsScene)} />
+      <Stack.Screen name="stakeOverview" component={ifLoggedIn(StakeOverviewScene)} />
+      <Stack.Screen name="fioStakingOverview" component={ifLoggedIn(FioStakingOverviewScene)} />
+      <Stack.Screen name="fioStakingChange" component={ifLoggedIn(FioStakingChangeScene)} />
+
+      <Stack.Screen
+        name="manageTokens"
+        component={ifLoggedIn(ManageTokensScene)}
+        options={{
+          title: '',
```
swansontec, line 273: This title may not be needed too.

src/components/Main.tsx
```diff
@@ -0,0 +1,687 @@
...
+        component={ifLoggedIn(FioDomainRegisterSelectWalletScene)}
+        options={{
+          title: s.strings.title_register_fio_domain,
+          headerRight: () => null
+        }}
+      />
+      <Stack.Screen
+        name="fioDomainConfirm"
+        // @ts-expect-error
+        component={ifLoggedIn(FioNameConfirmScene)}
```
swansontec, line 521: Visit the FioNameConfirmScene, and change the props to take `'fioNameConfirm' | "fioDomainConfirm"` to get rid of this.

src/components/Main.tsx
```diff
@@ -0,0 +1,687 @@
...
+          title: s.strings.title_fio_domain_settings
+        }}
+      />
+
+      <Stack.Screen name="fioRequestList" component={ifLoggedIn(FioRequestListScene)} />
+      <Stack.Screen
+        name="fioRequestApproved"
+        // @ts-expect-error
+        component={ifLoggedIn(TransactionDetailsScene)}
+        listeners={{ focus: async () => requestPermission('contacts') }}
```
swansontec, line 580: Also this `async` should go away, and just do `.catch(showError)`.

src/components/Main.tsx
```diff
@@ -0,0 +1,687 @@
...
+        name="loanCreateConfirmation"
+        component={ifLoggedIn(LoanCreateConfirmationScene)}
+        options={{
+          headerTitle: () => <EdgeLogoHeader />
+        }}
+      />
+      <Stack.Screen
+        name="loanDetails"
+        // @ts-expect-error
+        component={ifLoggedIn(LoanDetailsScene)}
```
swansontec, line 622: The `withLoanAccount` HOC is messed up. It took a huge amount of effort to get the `withWallet` HOC typed right, so we either need to copy that type strategy, or just leave this `// @ts-expect-error` and fix it later.

src/components/Main.tsx
```diff
@@ -0,0 +1,687 @@
...
+        options={{
+          title: s.strings.title_fio_domain_settings
+        }}
+      />
+
+      <Stack.Screen name="fioRequestList" component={ifLoggedIn(FioRequestListScene)} />
+      <Stack.Screen
+        name="fioRequestApproved"
+        // @ts-expect-error
+        component={ifLoggedIn(TransactionDetailsScene)}
```
swansontec, line 579: This is a huge abuse of the transaction details scene, but it's been this way for a while. Let's leave this `// @ts-expect-error` and fix it later.

swansontec (changes requested):
Wow, this is a solid PR. There are some things to fix, but compared to the size of the PR, it's really clean.

src/components/Main.tsx
samholmes: This was added by the linter.

src/components/Main.tsx
samholmes: This was copied over from Main.ui.tsx.

src/components/Main.tsx
samholmes: Ah.

src/components/Main.tsx
samholmes: Thanks for this catch!

src/components/navigation/BackButton.tsx
```diff
@@ -2,56 +2,56 @@ import * as React from 'react'
...
+const Icon = () => {
+  const theme = useTheme()
+  const styles = getIconStyles(theme)
+
+  return isIos ? (
+    <IonIcon size={theme.rem(1.5)} color={theme.icon} name="chevron-back-outline" style={styles.backIconStyle} />
+  ) : (
+    <IonIcon size={theme.rem(1.25)} color={theme.icon} name="md-arrow-back" style={styles.backIconAndroid} />
+  )
+}
```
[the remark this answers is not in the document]
samholmes: I gave it a try. I'm liking it too 🙂

src/actions/CryptoExchangeActions.tsx
```diff
@@ -64,7 +64,7 @@ export function getQuoteForTransaction(navigation: NavigationBase, info: SetNati
 
       const swapInfo = await fetchSwapQuote(state, request)
 
-      navigation.push('exchangeQuote', {
+      navigation.replace('exchangeQuote', {
```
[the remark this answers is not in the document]
samholmes: Seems to work when I tested it. 👍

src/components/scenes/LoginScene.tsx
```diff
@@ -234,5 +244,7 @@ export const LoginScene = connect<StateProps, DispatchProps, OwnProps>(
     logout() {
       dispatch(logoutRequest())
     }
-  })
-)(withTheme(LoginSceneComponent))
+  }
+
+  return <LoginSceneComponent {...props} {...dispatchProps} {...stateProps} {...themeProps} />
```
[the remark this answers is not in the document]
samholmes: Fair enough. Though because it's not too impactful of a change, I'm going to leave it for the sake of fixup-brevity here.

src/types/routerTypes.tsx
```diff
@@ -395,6 +395,7 @@ export interface NavigationProp<Name extends keyof AppParamList> extends Navigat
  * The of the `route` prop passed to each scene.
  */
 export interface RouteProp<Name extends keyof AppParamList> {
+  key: string
```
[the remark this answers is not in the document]
samholmes: Yes. My approach was to skip straight to the router conversion which admittedly skips some steps.

src/components/scenes/Loans/LoanCloseScene.tsx
```diff
@@ -117,7 +117,8 @@ export const LoanCloseSceneComponent = (props: Props) => {
       // Update the loan program marking it as close
       await dispatch(saveLoanAccount({ ...loanAccount, closed: true }))
       // Navigate to loan dashboard scene if no action program is necessary
-      navigation.popToTop()
+      // @ts-expect-error I'm not sure why TS is complaining here...
+      navigation.navigate('loadDashboard', {})
```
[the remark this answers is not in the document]
samholmes: Oh jeez. How tired was at this moment... 🤦‍♂️ 

Thanks for this one.

src/components/Main.tsx
samholmes: I went ahead and remove these redundencies.

src/components/Main.tsx
samholmes: I removed it.

src/components/Main.tsx
samholmes: I'm a bit lost on the type strategy difference here. I'm down for a sidebar.

src/components/Main.tsx
samholmes: I looked at it, and I would assume we could just navigate to `transactionDetails` key, but then it's missing the `walletId` param; how did this ever work without the `walletId` param. The `FioRequestListScene` is lengthy with many wallet IDs being thrown around, so without much context on that component I wouldn't know which ID is the correct one to pass along. Can we save this for a sidebar and follow up task?

swansontec: Ok. My guess is that this is actually a genuine bug, but the `withWallet` HOC identifies that the `null` wallet doesn't exist, and backs out before we can properly crash. So, the only symptom is that you don't go to the transaction details scene.

jestSetup.js
```diff
@@ -2,6 +2,9 @@ import { jest } from '@jest/globals'
 
 jest.useFakeTimers()
 
+let _nonce = 0
+global.nonce = () => _nonce++
+
```
swansontec, line 7: Let's create a common utility function and then import that:

```typescript
function makeNonce(start: number = 0): () => number {
  return () => ++start
}
```

swansontec (changes requested):
So close. Just need to fixup the nonce thing, and then rock & roll.

swansontec (approved):

PR #3973, opened 2023-02-02
Add Tron staking

peachbits:
### CHANGELOG

Add Tron resource staking

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203557533664627


peachbits:
Blocked by https://github.com/EdgeApp/edge-currency-accountbased/pull/506

src/components/scenes/LoginScene.tsx
```diff
@@ -81,6 +82,7 @@ class LoginSceneComponent extends React.PureComponent<Props, State> {
       .then(backgroundImage => this.setState({ backgroundImage }))
       // @ts-expect-error
       .catch(e => this.setState({ backgroundImage: theme.backgroundImage }))
+    makeStakePlugins().catch()
```
swansontec, line 85: The empty `catch` does nothing. You need to do `catch(() =>{})` to throw the error away, plus maybe a comment that `// This is just here to load the cache`.

src/plugins/stake-plugins/stakePlugins.ts
```diff
@@ -3,20 +3,31 @@ import { makeTcSaversPlugin } from './thorchainSavers/tcSaversPlugin'
...
+    Promise.all(promises)
+    return loadedPlugins
+  }
+  const results = await Promise.all(promises)
+
+  loadedPlugins = []
   results.forEach(result => {
-    if (result != null) out.push(result)
+    if (result != null) loadedPlugins.push(result)
   })
```
swansontec, line 34: Is this the same as `loadedPlugins = results.filter(result => result != null)`?

src/components/scenes/LoginScene.tsx
swansontec: Also, I do think this would be better in the services, but it's fine either place.

src/components/cards/StakingReturnsCard.tsx
```diff
@@ -29,6 +32,12 @@ export function StakingReturnsCard({ fromCurrencyLogos, toCurrencyLogos, text, s
     )
   }
 
+  const renderEstimatedReturn = () => {
+    if (apy === 0) return null
+    const estimatedReturnMsg = apy > 0 ? toFixed(apy.toString(), 1, 1) + '% APR' : 'N/A'
```
swansontec, line 37: Does the `if` statement on the line above mean the `apy > 0 ? ` ternary is no longer needed? Maybe the `if` statement could be changed to `if (apy <= 0) return null`, to hit both cases.

src/components/scenes/Staking/StakeModifyScene.tsx
```diff
@@ -407,7 +407,9 @@ export const StakeModifyScene = (props: Props) => {
         }
         {
           // Render claim amount tile
-          modification === 'claim' || modification === 'unstake' ? stakePolicy.rewardAssets.map(asset => renderEditableQuoteAmountRow('claim', asset)) : null
+          !stakePolicy.rewardsNotClaimable && (modification === 'claim' || modification === 'unstake')
```
swansontec, line 420: This double-negative is pretty trippy. We could flip the whole thing around, `stakePolicy.rewardsNotClaimable || (modification !== 'claim' && modification !== 'unstake') ? null : stakePolicy.rewardAssets.map(asset => renderEditableQuoteAmountRow('claim', asset))`

swansontec (changes requested):
Only tiny changes needed here. This has some really good cleanups! I am also liking that this is just a normal stake plugin, that still talks to the currency plugin. Very slick, the way the architecture came out.

src/components/scenes/LoginScene.tsx
peachbits: I agree so I'll move it there

src/plugins/stake-plugins/stakePlugins.ts
peachbits: It is, I just struggled with the typing and ignored it. Fixed.

src/components/cards/StakingReturnsCard.tsx
peachbits: Correct, fixed.

src/components/scenes/Staking/StakeModifyScene.tsx
peachbits: Took me a minute but yeah your replacement makes way more sense

swansontec (approved):

PR #3975, opened 2023-02-02
Jon/event-tracking-update

Jon-edge:
### CHANGELOG

- Add new anonymous event analytics

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203762233206860


src/actions/LoginActions.tsx
```diff
@@ -271,23 +272,28 @@ export function logoutRequest(username?: string): ThunkAction<Promise<void>> {
...
+      20000,
+      new Error(s.strings.error_creating_wallets)
+    )
+    if (account.activeWalletIds.length <= 1) {
+      dispatch({
+        type: 'UI/WALLETS/SELECT_WALLET',
+        data: { currencyCode: wallet.currencyInfo.currencyCode, walletId: wallet.id }
+      })
+    }
+    logActivity(`Create Wallet (login): ${account.username} -- ${walletType} -- ${fiatCurrencyCode ?? ''} -- ${walletName}`)
```
paullinator, line 290: Looks like you're missing a `Signup_Wallets_Created_Success` here

src/components/scenes/GuiPluginListScene.tsx
```diff
@@ -256,12 +256,12 @@ class GuiPluginList extends React.PureComponent<Props, State> {
       }
     }
 
-    const regionCode = { countryCode }
+    const isBuy = route.params.direction === 'buy'
     if (plugin.nativePlugin != null) {
-      await executePlugin({ guiPlugin: plugin, regionCode, paymentType, navigation, account })
+      await executePlugin({ isBuy, guiPlugin: plugin, regionCode: { countryCode }, paymentType, navigation, account })
```
paullinator, line 261: Keep with naming `direction` for consistency and make it an enum of 'buy' vs 'sell''. I believe @peachbits has a similar change in a PR that would conflict otherwise.

src/util/tracking.ts
```diff
@@ -64,7 +66,7 @@ export async function logEvent(event: TrackingEventName, values: TrackingValues
  * Send a raw event to Firebase.
  */
 async function logToFirebase(name: TrackingEventName, values: TrackingValues) {
-  const { accountDate, currencyCode, dollarValue, installerId, pluginId, error } = values
+  const { accountDate, currencyCode, dollarValue, installerId, pluginId, error, referralId } = values
```
paullinator, line 69: Looks like we were already have the installerId as a param. Can you just use that without adding referralId?

paullinator (changes requested):

paullinator (approved):

PR #3976, opened 2023-02-02
Activate tokens

paullinator:
### CHANGELOG

Allow token activation for chains that require it (XRP)

### Dependencies

none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1203860324833462


src/actions/WalletActions.tsx
```diff
@@ -21,6 +26,104 @@ export interface SelectWalletTokenParams {
...
+const activateWalletName: MapObject<{ name: string; notes: string }> = {
+  ripple: {
+    name: s.strings.activate_wallet_token_transaction_name_xrp,
+    notes: s.strings.activate_wallet_token_transaction_notes_xrp
+  }
+}
+
+const ACTIVATION_TOAST_AUTO_HIDE_MS = 5000
+
+const activateWalletTokens = async (
```
swansontec, line 38: We don't need to fix this, but the Edge convention is to put un-exported helper functions at the bottom of the file.

src/components/scenes/ConfirmScene.tsx
```diff
@@ -0,0 +1,93 @@
...
+interface Props {
+  navigation: NavigationProp<'confirmScene'>
+  route: RouteProp<'confirmScene'>
+}
+
+export interface ConfirmSceneParams {
+  titleText: string
+  bodyText: string
+  infoTiles?: Array<{ label: string; value: string }>
+  onConfirm: (resetSlider: () => void) => void
```
swansontec, line 24: Optional: I think this callback should be `async`. In the old send scene with it's insane redux logic, we needed the `resetSlider` solution to indicate that `onConfirm` had completed, and the slider could be released. Since this is fresh code, we can just reset the slider when the promise resolves.

This is the pattern we use for `MainButton`, `ButtonsModal`, and a bunch of other stuff in the GUI, and it works pretty well when you are just doing something simple.

src/components/scenes/ConfirmScene.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+const ConfirmComponent = (props: Props) => {
+  const { navigation, route } = props
+  const theme = useTheme()
+  const styles = getStyles(theme)
+
+  const { titleText, bodyText, infoTiles, onConfirm, onBack } = route.params
+
+  const renderInfoTiles = () => {
+    if (!infoTiles || !infoTiles.length) return null
```
swansontec, line 36: We don't have the lint rule turned on yet, but you have to be explicit with your boolean checks. `if (infoTiles == null) return null`. You don't need to worry about length === 0, since the empty array is no problem for React.

src/components/scenes/ConfirmScene.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const handleBackButton = useHandler(() => {
+    navigation.pop()
+    if (onBack != null) onBack()
+  })
+
+  React.useEffect(() => {
+    return () => {
+      if (onBack != null) onBack()
+    }
+  }, [onBack])
```
swansontec, line 53: Something about the navigation upgrade breaks this, but I don't know all the details. We need to make @samholmes aware of this so he can fix it in his upgrade branch.

src/components/scenes/ConfirmScene.tsx
```diff
@@ -0,0 +1,93 @@
...
+  titleText: {
+    fontSize: theme.rem(2)
+  },
+  body: {
+    margin: theme.rem(1),
+    justifyContent: 'center',
+    alignItems: 'center'
+  },
+  bodyText: {
+    fontSize: theme.rem(1)
```
swansontec, line 86: Isn't this already the default, by definition?

src/actions/WalletActions.tsx
```diff
@@ -21,6 +26,104 @@ export interface SelectWalletTokenParams {
...
+  tokenIds?: string[]
+): Promise<void> => {
+  if (tokenIds == null) throw new Error('Activating mainnet wallets unsupported')
+  const { account } = state.core
+  const { assetOptions } = await account.getActivationAssets({ activateWalletId: wallet.id, activateTokenIds: tokenIds })
+  const { pluginId } = wallet.currencyInfo
+  const { fiatCurrencyCode } = wallet
+
+  // See if there is only one wallet option for activation
+  if (assetOptions.length === 1 && assetOptions[0].paymentWalletId != null) {
```
swansontec, line 52: There is no `else` case on this guy. Should we at least put an `} else { // TODO }` block, or maybe show a warning?

src/actions/WalletActions.tsx
```diff
@@ -21,6 +26,104 @@ export interface SelectWalletTokenParams {
...
+            <ButtonsModal
+              bridge={bridge}
+              title={s.strings.create_wallet_account_unfinished_activation_title}
+              message={sprintf(msg, feeString)}
+              buttons={{ ok: { label: s.strings.string_ok } }}
+            />
+          ))
+          navigation.pop()
+          return
+        }
```
swansontec, line 101: If we can detect the error here, why not do it earlier and not even show the scene in the first place?

src/actions/WalletActions.tsx
```diff
@@ -21,6 +26,104 @@ export interface SelectWalletTokenParams {
...
+            }
+          })
+          .then(result => {
+            showToast(s.strings.activate_wallet_token_success, ACTIVATION_TOAST_AUTO_HIDE_MS)
+            navigation.pop()
+          })
+          .catch(e => {
+            navigation.pop()
+            console.error(e)
+            showError(e)
```
swansontec, line 120: `showError` already includes the `console.log`, so don't do that.

swansontec (changes requested):
This PR looks pretty good overall. If we get into a hurry, we could just merge it as-is without fixing anything, but obviously we'd prefer to get it tidier if possible.

src/components/scenes/ConfirmScene.tsx
paullinator: Ah yes, this was a super copy-pasta bug. Came from old SendScene -> SendScene2 -> ConfirmScene but I should have been better. Will fix

src/components/scenes/ConfirmScene.tsx
paullinator: This also exists in the SendScene2 so I assume he'll have to fix it there as well.

src/components/scenes/ConfirmScene.tsx
paullinator: Yes. good catch

src/actions/WalletActions.tsx
paullinator: Sure. will do

src/actions/WalletActions.tsx
paullinator: I didn't want to recreate all the content shown to the user about wallet activation. By showing the scene first, they can see all the content and the cost to activate.

src/actions/WalletActions.tsx
paullinator: will fix

src/actions/WalletActions.tsx
swansontec: Makes sense.

src/components/scenes/ConfirmScene.tsx
swansontec: Right, that's why I know about it. Sam had to come up with a fix there, but I don't remember the details.

swansontec (approved):
This will work!

PR #3990, opened 2023-02-10
CoinRanking (Markets) UI fixes/updates

Jon-edge:
Add CoinRanking scenes back to Main.tsx
Fix CoinRankScene styling
Add USD to CoinRankingDetailsScene

Screenshot 2023-02-10 172805

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203924091043564

### DIFF
--- a/src/components/scenes/CoinRankingDetailsScene.tsx
+++ b/src/components/scenes/CoinRankingDetailsScene.tsx
@@ -2,6 +2,7 @@ import * as React from 'react'
 import { ScrollView, View } from 'react-native'
 import FastImage from 'react-native-fast-image'
 
+import { USD_FIAT } from '../../constants/WalletAndCurrencyConstants'
 import { formatFiatString } from '../../hooks/useFiatText'
 import { toLocaleDate, toPercentString } from '../../locales/intl'
 import s from '../../locales/strings'
@@ -104,7 +105,7 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
         return data == null ? 'N/A' : toPercentString(Number(data) / 100)
       case 'price':
       case 'priceChange24h':
-        return formatFiatString({ fiatAmount: baseString })
+        return `${formatFiatString({ fiatAmount: baseString })} ${USD_FIAT.replace('iso:', '')}`
       case 'rank':
         return `#${baseString}`
       case 'marketCapChange24h':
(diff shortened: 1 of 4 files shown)


itayplav:
Just a suggestion: may want to add a 1 or even 2 decimal precision to the market cap below the symbol:

image

Otherwise it's ambiguous how larger the marketcap difference is in these cases:

image

src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -104,7 +105,7 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
         return data == null ? 'N/A' : toPercentString(Number(data) / 100)
       case 'price':
       case 'priceChange24h':
-        return formatFiatString({ fiatAmount: baseString })
+        return `${formatFiatString({ fiatAmount: baseString })} ${USD_FIAT.replace('iso:', '')}`
```
peachbits: Maybe there's a good reason but I couldn't figure out why this feature is hardcoded to USD. I added a small fixup commit that uses the user's default fiat instead of USD for both components. If you use it the commit message should be changed before merging. 

If there is a good reason to keep USD only feel free to drop the fixup and merge since it would be approved anyway.

peachbits (changes requested):

src/components/scenes/CoinRankingDetailsScene.tsx
Jon-edge: That's legitimate - it used to be hardcoded to USD before because we were sourcing an API that only reliably had US quotes.

Jon-edge: Avoiding scope creep since this PR is meant to be a fix for the rank and USD display. Also the proposed change would break the feature for fiat types not supported by the API.

Tracked this new idea in task: https://app.asana.com/0/1200382638405084/1203963276097127/f

Jon-edge:
> Just a suggestion: may want to add a 1 or even 2 decimal precision to the market cap below the symbol:
> 
> image
> 
> Otherwise there is ambiguity if the values are both the same integer:
> 
> image

We actually already display decimal places in those fields. Since there's potentially less space here for smaller devices, and to keep the possibility of adding new columns, I'm willing to live with the discrepancy of Bn vs B, Mn vs M between the two CoinRank scenes.

PR #3999, opened 2023-02-14
Jon/fix/address-modal-cutoff-text

Jon-edge:
- Remove keyboard autofocus on address input TextInputModal activation
- Add a scrollable message body to the address input modal, move the title into the scrollable message body.

Simulator Screen Shot - iPhone 8 ios 16 - 2023-02-14 at 11 48 41
Simulator Screen Shot - iPhone 8 ios 16 - 2023-02-14 at 11 48 19

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203927724059430


src/components/themed/ModalParts.tsx
```diff
@@ -27,13 +27,13 @@ export function ModalTitle(props: ModalTitleProps) {
...
-export function ModalMessage(props: { children: React.ReactNode; paddingRem?: number[] | number; isWarning?: boolean }) {
+export function ModalMessage(props: { children: React.ReactNode | string; paddingRem?: number[] | number; isWarning?: boolean }) {
   const { children, isWarning, paddingRem } = props
   const theme = useTheme()
   const styles = getStyles(theme)
   const padding = sidesToPadding(mapSides(fixSides(paddingRem, 0), theme.rem))
 
-  return <Text style={[styles.messageText, padding, isWarning && styles.warningText]}>{children}</Text>
+  return typeof children === 'string' ? <Text style={[styles.messageText, padding, isWarning && styles.warningText]}>{children}</Text> : <>{children}</>
 }
```
swansontec, line 37: I think this is the wrong layering. The `TextInputModal` component should be should be doing this instead, like `typeof message === 'string' ? <ModalMessage>{message}</ModalMessage> : message`. That way the `ModalMessage` component retains its original purpose, which is to display text.

Next, instead of passing custom CSS to `EdgeText`, you would just use `ModalMessage` and `ModalTitle` as normal, like:

```tsx
<TextInputModal
  messge={
    <ScrollView>
      <ModalTitle>{...}</ModalTitle>
      <ModalMessage>{...}</ModalMessage>
    </ScrollView>
  }
/>
```

swansontec (changes requested):

src/components/themed/ModalParts.tsx
swansontec: I have tested this change locally, and using the `ModalTitle` and `ModalMessage` inside a scroll view does look better, and requires no additional CSS.

src/components/modals/TextInputModal.tsx
```diff
@@ -93,7 +93,7 @@ export function TextInputModal(props: Props) {
   return (
     <ThemedModal warning={warning} bridge={bridge} onCancel={() => bridge.resolve(undefined)}>
       {title != null ? <ModalTitle>{title}</ModalTitle> : null}
-      {message != null ? <ModalMessage>{message}</ModalMessage> : null}
+      {message != null ? typeof message === 'string' ? <ModalMessage>{message}</ModalMessage> : <>{message}</> : null}
```
swansontec, line 96: This can be simplified to `{typeof message === 'string' ? <ModalMessage>{message}</ModalMessage> : message}`. The `null` will pass through, and the fragment isn't needed in this context. Fragments are mainly just needed to make `return` statements happy.

swansontec (approved):
It would be good to clean this one thing up, but otherwise the visuals look good & will work.

PR #4009, opened 2023-02-17
Fix non-sticky exchange settings

paullinator:
### CHANGELOG

Fix exchange settings not saving after logout

### Dependencies

none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203411283954272

### DIFF
--- a/src/modules/Core/Account/settings.ts
+++ b/src/modules/Core/Account/settings.ts
@@ -102,14 +102,14 @@ export const setDefaultFiatRequest = async (account: EdgeAccount, defaultFiat: s
 
 export const setPreferredSwapPluginId = async (account: EdgeAccount, pluginId: string | undefined) => {
   return getSyncedSettings(account).then(async settings => {
-    const updatedSettings = updateSettings(settings, { preferredSwapPluginId: pluginId == null ? '' : pluginId })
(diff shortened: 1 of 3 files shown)


src/modules/Core/Account/settings.ts
```diff
@@ -102,14 +102,14 @@ export const setDefaultFiatRequest = async (account: EdgeAccount, defaultFiat: s
...
-    const updatedSettings = updateSettings(settings, { preferredSwapPluginId: pluginId == null ? '' : pluginId })
+    const updatedSettings = updateSettings(settings, { preferredSwapPluginId: pluginId == null ? '' : pluginId, preferredSwapPluginType: undefined })
     return setSyncedSettings(account, updatedSettings)
   })
 }
 
 export const setPreferredSwapPluginType = async (account: EdgeAccount, swapPluginType: EdgeSwapPluginType | undefined) => {
   return getSyncedSettings(account).then(async settings => {
-    const updatedSettings = updateSettings(settings, { preferredSwapPluginType: swapPluginType })
+    const updatedSettings = updateSettings(settings, { preferredSwapPluginType: swapPluginType, preferredSwapPluginId: '' })
```
swansontec, line 112: ~Shouldn't this be `undefined` and not a blank string?~

Edit: Nope, checked the code, and we do actually save a blank string to disk, then when we load it back, we use a `=== '' ?:` to convert it back to `undefined`. So this is right.

swansontec (approved):

PR #4010, opened 2023-02-17
Sam/getting-started

samholmes:
### CHANGELOG

- Add new GettingStartedScene to present to new users.

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

### Screenshots

Screen Shot 2023-02-20 at 10 38 10 AM

Screen Shot 2023-02-20 at 10 38 15 AM

Screen Shot 2023-02-20 at 10 38 19 AM

Screen Shot 2023-02-20 at 10 38 22 AM

Screen Shot 2023-02-20 at 10 38 25 AM

---
  - https://app.asana.com/0/0/1203858268077209


swansontec (changes requested):
I feel like we need a phone call about this PR. I literally can't read the last commit without getting a PTSD know in my stomach. The main emotion is fear, actually. This looks too much like the bad code we've spent the last several years as a company trying to clean up.

I know there is a styled-components movement out there, and that a lot of people are happy with this approach. I get that. Perhaps this technique makes sense to some people.

Maybe I am over-reacting. Maybe this is different form the mess we got into earlier on in the company's history, but I would need to be convinced that this is the case. Please help me understand why this technique makes sense.

samholmes:
After discussing this PR with @swansontec, the big point of feedback is with the styled-component pattern. The main concern is with the inheritance pattern it introduces by virtue that it allows for components to be too easily extended ad infinitum. So, a compromise we agreed upon is to provide a crucial constraint to these styled-components, which is to not allow themselves to be extended further; they don't have a style prop. This stops the inheritance at one level deep and communicates to the developer that they should only be used as an alternative coding style/pattern for local components to a scene. 

I'll ad further, that in an ideal world, there should be a minimal number of local styled-components (or even styles for that matter), because we'd have a set of components apart from our style-guide which are flexible enough to meet the design requirements for a particular feature. Having a lot of one off styles or styled-components is a clear indication of one of two things:

1. Our designs are deviating from the style-guide
2. Our style-guide is lacking in its ability to accomplish the design requirements and may need additional considerations (new components, more flexible components, etc).

src/components/scenes/LoginScene.tsx
```diff
@@ -180,6 +182,7 @@ class LoginSceneComponent extends React.PureComponent<Props, State> {
           appId={config.appId}
           accountOptions={{ pauseWallets: true }}
           context={context}
+          initialAction={loginUiInitialAction}
```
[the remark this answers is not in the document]
samholmes: That would make sense if the type wasn't the redux action. Perhaps we should abstract away the redux action entirely with this prop and export a new type for this prop? I'm open to design direction on this one.

src/types/routerTypes.tsx
```diff
@@ -1,6 +1,7 @@
 import * as NavigationCore from '@react-navigation/core'
 import { ParamListBase, StackActionHelpers } from '@react-navigation/native'
 import { EdgeCurrencyInfo, EdgeCurrencyWallet, EdgeSpendInfo, EdgeTransaction, JsonObject, OtpError } from 'edge-core-js'
+import { Action as LoginUiAction } from 'edge-login-ui-rn/lib/types/ReduxActions'
```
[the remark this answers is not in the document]
samholmes: Fair enough. I'll echo my other comment here and suggest we design a type for the purpose of this prop. Or, we could simply export the Action type from the library.

src/util/parseMarkedText.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * Converts formatted text into ReactNodes. The currently supported formatting
+ * rule-set is based on markdown:
+ *
+ * * Use asterisks around text to emphasize the text: \*emphasized\*
+ *
+ * @param str Formatted text with special formatting rules.
+ * @returns An array of ReactNodes which can be inserted directly into JSX
+ */
+export function parseMarkedText(str: string): React.ReactNode[] {
+  const regex = /\*(.*?)\*/g
```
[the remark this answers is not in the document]
samholmes: I expect `\*` as the escaped version. I'll give this a go.

src/util/parseMarkedText.tsx
```diff
@@ -0,0 +1,38 @@
...
+  let lastIndex = 0
+  const parsedArr: React.ReactNode[] = []
+
+  while ((match = regex.exec(str)) !== null) {
+    const startIndex = match.index
+    const endIndex = regex.lastIndex - 2 // adjust for the asterisks
+    if (startIndex > lastIndex) {
+      parsedArr.push(str.substring(lastIndex, startIndex))
+    }
+    parsedArr.push(<Em>{str.substring(startIndex + 1, endIndex + 1)}</Em>)
```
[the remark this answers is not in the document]
samholmes: See https://github.com/EdgeApp/edge-react-gui/pull/4010#issuecomment-1457350083

src/util/parseMarkedText.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * Converts formatted text into ReactNodes. The currently supported formatting
+ * rule-set is based on markdown:
+ *
+ * * Use asterisks around text to emphasize the text: \*emphasized\*
+ *
+ * @param str Formatted text with special formatting rules.
+ * @returns An array of ReactNodes which can be inserted directly into JSX
+ */
+export function parseMarkedText(str: string): React.ReactNode[] {
+  const regex = /(?<!\\)\*(.*?)(?<!\\)\*/g
```
swansontec, line 16: I see. It's "not-backlash, star, then save what comes next, not-backslash, star". Your save-what-comes-next has an extra '?' in it, since `(.*)` does the exact same thing as `(.*?)`.

This regexp also doesn't handle '\\*', but I think we can live with that for now. Writing a full parser is quite tricky, but at least this pattern covers the basic syntax anyone will ever need.

src/components/scenes/GettingStartedScene.tsx
```diff
@@ -0,0 +1,349 @@
...
+      </SwipeOffsetDetector>
+    </SceneWrapper>
+  )
+}
+
+// -----------------------------------------------------------------------------
+// Local Components
+// -----------------------------------------------------------------------------
+
+const Container = styled(View)(_ => ({
```
swansontec, line 154: Instead of `_ ->`, would `() -> ` be better? It makes it clear that we don't care about the props, as opposed to wondering if the `_` might get used somewhere (even though convention says it shouldn't).

src/components/scenes/GettingStartedScene.tsx
```diff
@@ -0,0 +1,349 @@
...
+      opacity: interpolate(props.swipeOffset.value, [0, 1], [0, 1])
+    }
+  })
+)
+
+//
+// Hero
+//
+
+const HeroContainer = styled(Animated.View)(_ => ({
```
swansontec, line 174: This guy isn't animated.

src/components/scenes/GettingStartedScene.tsx
```diff
@@ -0,0 +1,349 @@
...
+}))
+
+//
+// Skip Button
+//
+
+const SkipButton = styled(Animated.View)<{ swipeOffset: SharedValue<number> }>(props =>
+  useAnimatedStyle(() => {
+    return {
+      opacity: interpolate(props.swipeOffset.value, [0, 1], [0, 1])
```
swansontec, line 165: This looks like a no-op. Do you mean `Extrapolation.CLAMP`?

src/components/scenes/GettingStartedScene.tsx
```diff
@@ -0,0 +1,349 @@
...
+
+const HeroImageContainer = styled(View)(_ => ({
+  alignItems: 'stretch',
+  aspectRatio: 1,
+  backgroundColor: 'white',
+  borderRadius: 1000,
+  maxHeight: '100%',
+  overflow: 'hidden',
+  width: '100%'
+}))
```
swansontec, line 241: Since this guy exists to render a white circle, maybe call him `HeroImageCircle`.

I had a long thing typed up about a simpler way to position the image, but then I realized that you were actually doing it this way to get a circle, not to control the image layout.

swansontec (changes requested):
Thanks for the hard work on this. It makes me much happier now. Only small things remain.

src/util/parseMarkedText.tsx
samholmes: `(.*?)` is not greedy, but `(.*)` is greedy. The `(?<!\\)` is a negative look-back for the `\*`; it says that star should not be proceeded by a backslash. We do the same thing after the non-greedy match-all. This means that we only match the non-greedy match-all group if surrounded by non-escaped stars. See the test file to prove the coverage.

src/components/scenes/GettingStartedScene.tsx
samholmes: If the design was a rounded rectangle, then we'd have to rename it to HeroImageRoundedRectangle. But if we keep it HeroImageContainer, no renaming is required because it will always remain the container of the hero image semantically. Does this seem appropriate?

PR #4016, opened 2023-02-19
Fix safe area on iPhone 14

paullinator:
### CHANGELOG

Fix broken headers and footers on iPhone 14 models

### Dependencies

none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203980222901146


src/__tests__/components/MenuTabs.test.tsx
```diff
@@ -12,7 +13,11 @@ describe('MenuTabs', () => {
 
     const props: BottomTabBarProps = { navigation: fakeNavigation, state: { index: 0, routes: [] } } as any
 
-    const actual = renderer.render(<MenuTabs {...props} />)
+    const actual = renderer.render(
+      <SafeAreaProvider>
```
swansontec, line 17: Once we merge #4033, you can simply revert the entire `__tests__` folder, since all the deep tests will now use `FakeProvider`. Then `FakeProvider` can be edited as follows:

```ts
return (
    <SafeAreaProvider initialMetrics={initialMetrics}>
      <Provider store={store}>{children}</Provider>
    </SafeAreaProvider>
  )
}

// These match what our old react-native-safe-area-view library returned:
const initialMetrics: Metrics = {
  frame: { height: 1334, width: 750, x: 0, y: 0 },
  insets: { bottom: 0, left: 0, right: 0, top: 20 }
}
```

With this in place, even the snapshots remain unchanged.

src/components/scenes/CreateWalletAccountSelectScene.tsx
```diff
@@ -15,7 +16,6 @@ import s from '../../locales/strings'
 import { PrimaryButton } from '../../modules/UI/components/Buttons/PrimaryButton.ui'
 import { FormattedText as Text } from '../../modules/UI/components/FormattedText/FormattedText.ui'
 import { Gradient } from '../../modules/UI/components/Gradient/Gradient.ui'
-import { SafeAreaViewComponent as SafeAreaView } from '../../modules/UI/components/SafeAreaView/SafeAreaView.ui'
```
swansontec, line 18: With this gone, you can fully delete `src/modules/UI/components/SafeAreaView/SafeAreaView.ui.tsx`.

src/components/themed/WalletList.tsx
```diff
@@ -314,6 +314,12 @@ export const filterWalletCreateItemListBySearchText = (createWalletList: WalletC
       out.push(item)
     }
   }
+  out.push({
```
swansontec, line 317: It is not necessary to add a new row. Instead, you can apply `contentContainerStyle={{ paddingBottom: theme.rem(4.25) }}` to create one row's worth of extra space at the bottom of the scroll area. So basically, this entire commit becomes a single-line style addition.

swansontec (changes requested):
Assuming we merge #4033, only 3 other changes would need to be performed.

src/components/themed/WalletList.tsx
paullinator: will do

PR #4019, opened 2023-02-22
Add `LogsModal` component for sending & exporting logs

itayplav:
### CHANGELOG

Add `LogsModal` component for sending & exporting logs

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

|                                                             iPhone 8                                                             |                                                             iPhone X                                                             | iPad Pro                                                                                                                         |
|:--------------------------------------------------------------------------------------------------------------------------------:|:--------------------------------------------------------------------------------------------------------------------------------:|----------------------------------------------------------------------------------------------------------------------------------|
| <video width="300" src="https://user-images.githubusercontent.com/4023066/221332360-35933be0-6c59-43e4-bf0d-bb158894e26b.mp4" /> | <video width="300" src="https://user-images.githubusercontent.com/4023066/221332364-323c518b-e5f2-417e-8d5e-4017d30945ad.mp4" /> | <video width="300" src="https://user-images.githubusercontent.com/4023066/221332370-97d5a14b-913c-4668-b853-9a2bc6414196.mp4" /> |

---
  - https://app.asana.com/0/0/1203658636813906
  - https://app.asana.com/0/0/1202565150456515


src/actions/LogActions.tsx
```diff
@@ -61,23 +65,37 @@ export function showSendLogsModal(): ThunkAction<Promise<void>> {
     const { isConnected } = state.network
     if (!isConnected) return showError(s.strings.network_alert_title)
 
-    Airship.show<string | undefined>(bridge => (
-      <TextInputModal
+    await Airship.show(bridge => <LogsModal bridge={bridge} />)
+  }
+}
+export function showClearLogsModal(): ThunkAction<Promise<void>> {
```
swansontec, line 78: Functions should have blank lines between them, generally.

src/actions/LogActions.tsx
```diff
@@ -122,7 +138,7 @@ function prepareLogs(text: string): ThunkAction<Promise<void>> {
     const logOutput: LogOutput = {
       isoDate: new Date().toISOString(),
       uniqueId: infoId,
-      userMessage: text,
+      userMessage: userMessage,
```
swansontec, line 149: This could just be `userMessage,` since the left and right are the same name.

src/components/modals/LogsModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+      bridge.resolve()
+    } catch (e: any) {
+      showError(e.message)
+    }
+    setSpinning(false)
+  }
+
+  const handleChangeText = (text: string) => {
+    setUserMessage(text)
+  }
```
swansontec, line 65: You can ditch this callback, and just pass `setUserMessage` straight to the `onChangeText` prop.

src/components/modals/LogsModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+  bridge: AirshipBridge<void>
+}
+
+export const LogsModal = ({ bridge }: LogsModalProps) => {
+  const theme = useTheme()
+  const dispatch = useDispatch()
+  const styles = getStyles(theme)
+  const [spinning, setSpinning] = React.useState(false)
+  const [userMessage, setUserMessage] = React.useState('')
+  const dir = Platform.OS === 'android' ? RNFS.ExternalCachesDirectoryPath : RNFS.DocumentDirectoryPath
```
swansontec, line 29: Put this inside `handleShare`, since it isn't used anywhere else.

src/actions/LogActions.tsx
```diff
@@ -16,6 +16,13 @@ import { sendLogs } from '../modules/Logs/api'
 import { ThunkAction } from '../types/reduxTypes'
 import { clearLogs, logWithType, readLogs } from '../util/logger'
 
+const SENSITIVE_KEY_REGEX = /(allKeys|displayPrivateSeed|displayPublicSeed|otpKey|loginKey|recoveryKey|dataKey|syncKey)/gm
```
swansontec, line 19: You don't need the parentheses or the 'm' flag. They do nothing in this particular regexp. So just: `/allKeys|displayPrivateSeed|displayPublicSeed|otpKey|loginKey|recoveryKey|dataKey|syncKey/g`

src/actions/LogActions.tsx
```diff
@@ -213,38 +221,47 @@ export function getLogOutput({ userMessage, send = true }: LogOutputOptions): Th
...
-App build: ${getVersion()}.${getBuildNumber()}
-os: ${Platform.OS} ${Platform.Version}
-device: ${getBrand()} ${getDeviceId()}
-`
+    const appInfo = `
+      App version: ${packageJson.version}
+      App build: ${getVersion()}.${getBuildNumber()}
+      os: ${Platform.OS} ${Platform.Version}
+      device: ${getBrand()} ${getDeviceId()}
+    `
```
swansontec, line 229: This indentation gets copied into the string literally, so it has to look like it did before, or the logs you upload to the server are going to be messed up. Please put it back.

src/actions/LogActions.tsx
```diff
@@ -104,16 +129,15 @@ export function showClearLogsModal(): ThunkAction<Promise<void>> {
...
+export async function clearAllLogs(): Promise<void> {
+  try {
+    await Promise.all([clearLogs('activity'), logWithType('activity', 'CLEARED ACTIVITY LOGS'), clearLogs('info'), logWithType('info', 'CLEARED INFO LOGS')])
+  } catch (error) {
+    console.error('Error while clearing logs:', error)
   }
 }
 
-function prepareLogs(text: string): ThunkAction<Promise<void>> {
+export function saveLogOutput({ userMessage, send = true }: LogOutputOptions): ThunkAction<Promise<MultiLogOutput>> {
```
swansontec, line 140: The layering is all messed up in this routine. Let's rename this back to `getLogOutput`, and have it simply return `Promise<MultiLogOutput>`, without any sending, sanitizing, or anything else. Just read from disk, and assemble the payload.

Then, the modal can call `saveLogOutput`, do the sanitatization, and report to the user if the sanitatization fails. That way you don't have `throw new Error()` with emojis  & translated strings inside the error message - you don't need to `throw` in the first place if you are already inside the modal. Just show whatever UI you need directly.

Finally, your modal can call either `sendLog` on the on the `activity` and `info` properties, or put them in the share modal, depending on the button press.

src/actions/LogActions.tsx
```diff
@@ -197,40 +221,54 @@ function prepareLogs(text: string): ThunkAction<Promise<void>> {
...
+      activityOutput.data += activityLogs != null ? activityLogs : ''
+      const activityPayload = JSON.stringify(activityOutput, null, 2)
+      const safeToSend = !SENSITIVE_KEY_REGEX.test(activityPayload)
+      if (send) {
+        if (safeToSend) await sendLogs(activityPayload)
+        else throw new Error('⚠ ' + s.strings.settings_modal_send_unsafe)
+      }
+    } catch (error: any) {
+      throw new Error(`${s.strings.settings_modal_send_logs_failure} (activity logs):\n\n${error.message}`)
+    }
```
swansontec, line 249: This try / catch logic would move to the modal, since that's where translated strings belong.

src/actions/LogActions.tsx
```diff
@@ -197,40 +221,54 @@ function prepareLogs(text: string): ThunkAction<Promise<void>> {
...
+      const logPayload = JSON.stringify(activityOutput, null, 2)
+      const safeToSend = !SENSITIVE_KEY_REGEX.test(logPayload)
+
+      if (send) {
+        if (safeToSend) await sendLogs(logPayload)
+        else throw new Error('⚠ ' + s.strings.settings_modal_send_unsafe)
+      }
+    } catch (error: any) {
+      throw new Error(`${s.strings.settings_modal_send_logs_failure} (info logs):\n\n${error.message}`)
+    }
```
swansontec, line 262: This try / catch logic would also move to the modal.

src/components/modals/LogsModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  const handleSend = async () => {
+    setSpinning(true)
+    try {
+      await dispatch(saveLogOutput({ userMessage, send: true }))
+      bridge.resolve()
+    } catch (e: any) {
+      if (e.message.includes('⚠')) setSafeToSend(false)
+    }
```
swansontec, line 59: We should not be checking error messages for emojis. So yeah, we definitely need to do the safety checks in the modal itself.

src/components/modals/LogsModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+  bridge: AirshipBridge<void>
+}
+
+export const LogsModal = ({ bridge }: LogsModalProps) => {
+  const theme = useTheme()
+  const dispatch = useDispatch()
+  const styles = getStyles(theme)
+  const [spinning, setSpinning] = React.useState(false)
+  const [userMessage, setUserMessage] = React.useState('')
+  const [safeToSend, setSafeToSend] = React.useState(true)
```
swansontec, line 28: Optional idea: We could call `saveLogOutput` in a `useAsyncEffect` as soon as the modal loads. Then we do the safety check & update the UI to either un-grey the buttons or show the error message. That way you know right away whether the logs are good or not, and can provide that feedback immediately.

swansontec (changes requested):

src/components/modals/LogsModal.tsx
itayplav: Going ahead w/ this approach.

itayplav: MultiLogOutput now gets passed as a prop to LogsModal and safety-checked inside the modal, providing immediate feedback.

src/actions/LogActions.tsx
itayplav: try / catch logic moved to the modal so each <MainButton> will spin while the promise(s) resolve.

src/actions/LogActions.tsx
itayplav: Moved there now.

src/actions/LogActions.tsx
itayplav: The implementation of `getLogOutput` has been modified to return a `Promise<MultiLogOutput>` without performing any sanitation or sending. `handleSend` is disabled if `SENSITIVE_KEY_REGEX` matches either `logs.activity.data`. or `log.info.data`.

src/actions/LogActions.tsx
itayplav: Indentation has been reverted

src/components/modals/LogsModal.tsx
itayplav: Moved inside `handleShare`

src/components/modals/LogsModal.tsx
itayplav: Done

src/actions/LogActions.tsx
```diff
@@ -103,17 +121,15 @@ export function showClearLogsModal(): ThunkAction<Promise<void>> {
...
-  return async (_dispatch, _getState) => {
-    await clearLogs('activity')
-    await logWithType('activity', 'CLEARED ACTIVITY LOGS')
-    await clearLogs('info')
-    await logWithType('info', 'CLEARED INFO LOGS')
+async function clearAllLogs(): Promise<void> {
+  try {
+    await Promise.all([clearLogs('activity'), logWithType('activity', 'CLEARED ACTIVITY LOGS'), clearLogs('info'), logWithType('info', 'CLEARED INFO LOGS')])
+  } catch (error) {
+    console.error('Error while clearing logs:', error)
```
swansontec, line 133: Please put back the `await` chain. While using `Promise.all` lets us do work in parallel for better performance, these tasks can't be done at the same time. If you write something to a file that's being deleted at the same time, it's a race which one will get there first, and you may not log your message at all.

Turning this thunk into a normal function is fine, since we don't actually use the `dispatch` anywhere inside.

src/actions/LogActions.tsx
```diff
@@ -61,23 +65,37 @@ export function showSendLogsModal(): ThunkAction<Promise<void>> {
...
+            }
+          },
+          no: {
+            label: 'Cancel',
+            type: 'escape'
+          }
         }}
       />
-    ))
+    )).catch(showError)
```
swansontec, line 98: This new function seems to be a copy of the function below. We don't need two copies.

src/components/modals/LogsModal.tsx
```diff
@@ -0,0 +1,134 @@
...
+    await RNFS.writeFile(path, payload, 'utf8').catch((e: any) => {
+      showError(e.message)
+    })
+    return await Share.open(shareOptions)
+      .catch((e: any) => {
+        showError(e.message)
+      })
+      .then(() => {
+        bridge.resolve()
+      })
```
swansontec, line 59: Since you are already `await`'ing the `Share.open`, you can ditch the `then` and just put this on its own line.

src/components/modals/LogsModal.tsx
```diff
@@ -0,0 +1,134 @@
...
+      failOnCancel: false
+    }
+
+    await RNFS.writeFile(path, payload, 'utf8').catch((e: any) => {
+      showError(e.message)
+    })
+    return await Share.open(shareOptions)
+      .catch((e: any) => {
+        showError(e.message)
+      })
```
swansontec, line 56: Get rid of the catch statements. If this function fails, `MainButton` will actually catch the error and call `showError` on its own.

src/components/modals/LogsModal.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const theme = useTheme()
+  const styles = getStyles(theme)
+  const [userMessage, setUserMessage] = React.useState('')
+
+  const isDangerous = React.useMemo(() => {
+    return SENSITIVE_KEY_REGEX.test(logs.activity.data) || SENSITIVE_KEY_REGEX.test(logs.info.data)
+  }, [logs])
+
+  const handleShare = async () => {
+    const payload = JSON.stringify(logs, null, 2)
```
swansontec, line 36: Add

```ts
logs.info.userMessage = userMessage
logs.activity.userMessage = userMessage
```

src/components/modals/LogsModal.tsx
```diff
@@ -0,0 +1,134 @@
...
+    logs.info.userMessage = userMessage
+    logs.activity.userMessage = userMessage
+
+    return await Promise.all([
+      sendLogs(logs.activity).catch((e: any) => {
+        showError(`${s.strings.settings_modal_send_logs_failure} activity logs code ${e.message}`)
+      }),
+      sendLogs(logs.info).catch((e: any) => {
+        showError(`${s.strings.settings_modal_send_logs_failure} info logs code ${e.message}`)
+      })
```
swansontec, line 72: These `.catch` statements can stay, since they are actually providing a translated message. But do `e?.message` in both of them, for extra safety. We don't want to crash if somebody does `throw null`.

src/components/modals/LogsModal.tsx
```diff
@@ -0,0 +1,134 @@
...
+    ]).then(() => {
+      bridge.resolve()
```
swansontec, line 74: Since you are already `await`'ing the `Promise.all`, you can ditch the `then` and just put this on its own line.

swansontec (changes requested):
Much better! We are really close to the end here, with only little things left.

src/actions/LogActions.tsx
itayplav: Removed the duplicate below since it was the one not being used.

src/components/modals/LogsModal.tsx
itayplav: If we move the `bridge.resolve()` outside of the `Promise.all(...).then(...)`, the modal will close if the users internet connection is momentarily offline whilst sending logs. Keeping `bridge.resolve()` as part of a successfull `Promise.all(...)` ensures the modal remains open for another try.

swansontec: OK, then you have it in the wrong order. The `.catch` lines will mop up any errors and allow the `.then` to run in any case. So, you need to put the `.catch` *after* the `.then` for it to do what you said.

itayplav: Right I see. I have changed the `.catch` statements to instead be `throw new Error(...)` that are then `.catch((e: Error)`'d outside the `Promise.all(...)` which prevents a needless double `showError` whilst only hiding the modal if both `sendLogs` succeed.

src/actions/LogActions.tsx
```diff
@@ -28,6 +32,9 @@ interface LogOutput {
   loggedInUser?: LoggedInUser
   data: string
 }
+export interface LogOutputOptions {
+  userMessage: string
+}
```
swansontec, line 37: Unused now?

src/components/modals/LogsModal.tsx
```diff
@@ -0,0 +1,134 @@
...
+      sendLogs(logs.info).catch((e: any) => {
+        throw new Error(`${s.strings.settings_modal_send_logs_failure} info logs code ${e?.message}`)
+      })
+    ])
+      .then(() => {
+        bridge.resolve()
+      })
+      .catch((e: Error) => {
+        showError(e.message)
+      })
```
swansontec, line 75: You don't need this. The MainButton will catch the error for you.

src/modules/Logs/api.ts
```diff
@@ -34,5 +36,7 @@ const request = async (name, path, method, data) => {
     })
 }
 
-// @ts-expect-error
-export const sendLogs = async logs => request('SENDING LOGS', 'log/', 'PUT', logs)
+export const sendLogs = async (logs: LogOutput) => {
+  const payload = JSON.stringify(logs)
```
swansontec, line 40: Just checked, and we don't need the extra stringify. The `request` handles that.

swansontec (changes requested):

swansontec (approved):
I approve my own fixups.

src/components/modals/LogsModal.tsx
```diff
@@ -0,0 +1,128 @@
...
+    ])
+    bridge.resolve()
+  }
+
+  const handleCancel = () => {
+    bridge.resolve(undefined)
+  }
+  const minHeight: ViewStyle = { minHeight: isDangerous ? theme.rem(7) : theme.rem(9) }
+  return (
+    <ThemedModal bridge={bridge} onCancel={handleCancel}>
```
itayplav, line 77: Switched the ternary around since `isSafe..` is now `isDangerous`

src/actions/LogActions.tsx
itayplav: Yup, removed along with the redundant `textStyle` and associated `<View>...</View>`

PR #4023, opened 2023-02-25
Add Migrate Wallets scenes

peachbits:
---
  - https://app.asana.com/0/0/1204051225230840


samholmes:
I haven't fully reviewed, but something that I think should be considered is to re-use a wallet that was created but failed to send to. This way if the user tries multiple times on a particular wallet and it fails multiple times, then we're not spamming with a bunch of new wallets. @peachbits

peachbits:
> I haven't fully reviewed, but something that I think should be considered is to re-use a wallet that was created but failed to send to. This way if the user tries multiple times on a particular wallet and it fails multiple times, then we're not spamming with a bunch of new wallets. @peachbits

Agreed I'll add that

src/components/scenes/MigrateWalletCompletionScene.tsx
```diff
@@ -0,0 +1,253 @@
...
+interface Props {
+  navigation: NavigationProp<'migrateWalletCompletion'>
+  route: RouteProp<'migrateWalletCompletion'>
+}
+
+interface MigrateWalletTokenItem extends MigrateWalletItem {
+  tokenId: string
+}
+
+const makeSpendSignAndBroadcast = async (wallet: EdgeCurrencyWallet, spendInfo: EdgeSpendInfo): Promise<EdgeTransaction> => {
```
swansontec, line 32: Utility functions go to the bottom of the file, right after the component but before the final `export`. That way people reading the file see the component itself first thing (with its props).

src/components/scenes/MigrateWalletCalculateFeeScene.tsx
```diff
@@ -0,0 +1,268 @@
...
+  const handleSlidingComplete = useHandler(() => {
+    const filteredMigrateWalletList = migrateWalletList.filter(asset => typeof feeState[asset.key] === 'string')
+    navigation.push('migrateWalletCompletion', { migrateWalletList: filteredMigrateWalletList })
+  })
+
+  // Create getMaxSpendable/makeSpend promises for each selected asset. We'll group them by wallet first and then execute all of them while keeping
+  // track of which makeSpends are successful so we can enable the slider. A single failure from any of a wallet's assets will cast them all as failures.
+  useAsyncEffect(async () => {
+    // This bundles the assets by similar walletId with the main asset (ie. ETH) at the end of each array so its makeSpend is called last
+    const bundledWalletAssets: MigrateWalletItem[][] = migrateWalletList.reduce((bundles: MigrateWalletItem[][], asset) => {
```
swansontec, line 124: This one isn't actually O(n^2), because of a subtle difference in the way you have coded it. You could just as easily use a `for of` loop here, but it's not as important as the other places.

src/components/scenes/MigrateWalletCalculateFeeScene.tsx
```diff
@@ -0,0 +1,268 @@
...
+    let successCount = 0
+    const walletPromises = []
+    for (const bundle of bundledWalletAssets) {
+      const wallet = currencyWallets[bundle[bundle.length - 1].key]
+      const {
+        currencyInfo: { pluginId }
+      } = wallet
+
+      let feeTotal = '0'
+      const bundlesFeeTotals: { [key: string]: AssetRowState } = bundle.reduce((map: { [key: string]: AssetRowState }, asset) => {
```
swansontec, line 152: Good candidate for a `Map<string, AssetRowState>`

src/components/scenes/MigrateWalletCompletionScene.tsx
```diff
@@ -0,0 +1,253 @@
...
+      return map
+    }, {})
+  })
+
+  const flatListRef = React.useRef<FlatList<MigrateWalletItem>>(null)
+
+  const handleItemStatus = (item: MigrateWalletItem, status: 'complete' | 'error') => {
+    setItemStatus(currentState => ({ ...currentState, [item.key]: status }))
+    const index = sortedMigrateWalletList.findIndex(asset => asset.key === item.key)
+    flatListRef.current?.scrollToIndex({ animated: true, index, viewPosition: 0.5 })
```
swansontec, line 89: Nice touch.

src/components/scenes/MigrateWalletCompletionScene.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+  const renderStatus = useHandler((item: MigrateWalletItem) => {
+    let icon = <ActivityIndicator style={{ paddingRight: theme.rem(0.3125) }} color={theme.iconTappable} />
+    if (itemStatus[item.key] === 'complete') icon = <IonIcon name="checkmark-circle-outline" size={theme.rem(1.5)} color={theme.iconTappable} />
+    if (itemStatus[item.key] === 'error')
+      icon = <IonIcon name="warning-outline" style={{ paddingRight: theme.rem(0.0625) }} size={theme.rem(1.5)} color={theme.dangerText} />
+    return icon
+  })
+
+  const renderRow = useHandler(({ item }) => {
```
swansontec, line 209: On another scene, you had `useHandler((data: ListRenderItemInfo<MigrateWalletItem>) => {`. Would that work here?

src/components/scenes/MigrateWalletCompletionScene.tsx
```diff
@@ -0,0 +1,253 @@
...
+    const { createWalletIds } = item as MigrateWalletItem
+
+    const walletId = createWalletIds[0]
+    const wallet = currencyWallets[walletId]
```
swansontec, line 213: Maybe we should check `if (wallet == null) return null` for safety here too.

swansontec (changes requested):
We could merge this as-is if we really needed to, since nothing I found would be a show-stopper. However, I did find some things that might be worth fixing.

src/components/scenes/MigrateWalletCalculateFeeScene.tsx
```diff
@@ -171,19 +167,19 @@ const MigrateWalletCalculateFeeComponent = (props: Props) => {
...
+            bundlesFeeTotals.set(asset.key, txFee)
             feeTotal = add(feeTotal, txFee)
 
             // While imperfect, sanity check that the total fee spent so far to send tokens + fee to send mainnet currency is under the total mainnet balance
             if (i === bundle.length - 1 && lt(wallet.balances[asset.currencyCode], feeTotal)) {
               throw new InsufficientFundsError({ currencyCode: asset.currencyCode, networkFee: feeTotal })
             }
           } catch (e: any) {
-            for (const key of Object.keys(bundlesFeeTotals)) {
+            for (const key of [...bundlesFeeTotals.keys()]) {
```
swansontec, line 178: You don't need the `[...keys()]` bit, since `keys()` already returns an iterator that is compatible with `for` loops.

swansontec (approved):
Approved. The one small thing I found doesn't matter too much.

src/components/scenes/MigrateWalletCalculateFeeScene.tsx
peachbits: Ah it's a `for..in` loop that it works. Didn't work for `for..of` and I've conditioned myself away from `for..in` so I just accepted it and did the ugly thing.

PR #4027, opened 2023-03-01
Add migrate wallet warning modal

peachbits:


src/actions/WalletActions.tsx
```diff
@@ -312,3 +316,134 @@ const activateWalletTokens = async (
...
+    // Get synced setting Settings.json with public key map securityCheckedWallets: {walletId: { checked: boolean, modalShown: number }}
+    const settings = await getSyncedSettings(account)
+    let securityCheckedWallets = { ...settings.securityCheckedWallets }
+
+    // Gather list to send to info server
+    const hashedPubKeys = new Map<string, string>()
+    for (const walletId of activeWalletIds) {
+      // Add entries for walletIds that aren't recognized
+      if (securityCheckedWallets[walletId] == null) {
+        securityCheckedWallets = { ...securityCheckedWallets, [walletId]: { checked: false, modalShown: 0 } }
```
swansontec, line 347: Since you have already made a copy, I would just mutate it, like `securityCheckedWallets[walletId] = { checked: false }`. Otherwise it's O(n^2) to keep re-copying the object every time around the loop.

src/components/services/Services.tsx
```diff
@@ -37,6 +39,15 @@ export function Services(props: Props) {
   const account = useSelector(state => (state.core.account !== defaultAccount ? state.core.account : undefined))
   const { navigation } = props
 
+  // Methods to call once all of the currency wallets have been loaded
+  useAsyncEffect(async () => {
+    if (account == null) return
```
swansontec, line 45: On logout, account will be a dummy object, but the dummy object won't have a `waitForAllWallets` method. So maybe check `account?.waitForAllWallets != null` to avoid this case.

swansontec (changes requested):
This is an incredibly solid PR. Really well done.

swansontec (approved):

PR #4029, opened 2023-03-01
Add nested navigators for menus tabs

samholmes:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204091777780323

### DIFF
--- a/src/actions/CryptoExchangeActions.tsx
+++ b/src/actions/CryptoExchangeActions.tsx
@@ -70,7 +70,7 @@ export function getQuoteForTransaction(navigation: NavigationBase, info: SetNati
       })
       dispatch({ type: 'UPDATE_SWAP_QUOTE', data: swapInfo })
     } catch (error: any) {
-      navigation.navigate('exchange', {})
+      navigation.navigate('exchangeTab', { screen: 'exchange' })
 
       const insufficientFunds = asMaybeInsufficientFundsError(error)
       if (insufficientFunds != null && insufficientFunds.currencyCode != null && fromCurrencyCode !== insufficientFunds.currencyCode && fromWalletId != null) {
@@ -106,7 +106,7 @@ export function exchangeTimerExpired(navigation: NavigationBase, swapInfo: GuiSw
       })
       dispatch({ type: 'UPDATE_SWAP_QUOTE', data: swapInfo })
     } catch (error: any) {
-      navigation.navigate('exchange', {})
+      navigation.navigate('exchangeTab', { screen: 'exchange' })
       dispatch(processSwapQuoteError(error))
     }
   }
--- a/src/components/Main.tsx
+++ b/src/components/Main.tsx
@@ -186,6 +186,19 @@ const Drawer = createDrawerNavigator<AppParamList>()
 const Stack = createStackNavigator<AppParamList>()
 const Tab = createBottomTabNavigator<AppParamList>()
 
+const defaultScreenOptions = {
+  title: '',
+  headerTitle: ({ children }: HeaderTitleProps) => <HeaderTitle title={children} />,
+  headerLeft: () => <BackButton />,
+  headerRight: () => <SideMenuButton />,
+  headerShown: true,
+  headerTransparent: true
+}
+const firstSceneScreenOptions = {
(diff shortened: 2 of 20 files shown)


src/components/scenes/GuiPluginListScene.tsx
```diff
@@ -466,18 +466,25 @@ const getStyles = cacheStyles((theme: Theme) => ({
...
     }
   })
 )(withTheme(GuiPluginList))
+
+export function GuiPluginListBuyScene(props: Exclude<OwnProps, 'direction'>) {
+  return <GuiPluginListScene {...props} direction="buy" />
+}
+export function GuiPluginListSellScene(props: Exclude<OwnProps, 'direction'>) {
+  return <GuiPluginListScene {...props} direction="sell" />
+}
```
swansontec, line 490: This is not a great way to do it. Your component gets a `route` prop, and in there is a `name`. You can use that to figure out which direction is happening (route.name === 'pluginListSell'), and simply delete the `direction` prop altogether. This should also clear up the type errors.

src/actions/CryptoExchangeActions.tsx
```diff
@@ -70,7 +70,7 @@ export function getQuoteForTransaction(navigation: NavigationBase, info: SetNati
       })
       dispatch({ type: 'UPDATE_SWAP_QUOTE', data: swapInfo })
     } catch (error: any) {
-      navigation.navigate('exchange', {})
+      navigation.navigate('exchangeTab', {})
```
swansontec, line 73: I am not sure what happens if you navigate to a tab that you are already on. Does it always always always reset the state and go to the first scene? If not, you can force it to do so by passing `{ screen: 'xyz' }` as the route props.

If it does reset to the first scene every time, how does `MenuTabs` avoid resetting the state? Would it make sense to use the `.jump` action there to preserve the state?

src/components/Main.tsx
```diff
@@ -413,7 +408,7 @@ const EdgeAppStack = () => {
       />
       <Stack.Screen
         name="extraTab"
-        component={ExtraTabScene}
+        component={ExtraTabTabScene}
```
swansontec, line 411: This rename is really weird. Why?

src/components/Main.tsx
```diff
@@ -734,56 +729,124 @@ const EdgeAppStack = () => {
...
+const EdgeBuyTabScreen = () => {
+  return (
+    <Stack.Navigator initialRouteName="pluginListBuy" screenOptions={defaultScreenOptions}>
+      <Stack.Screen
         name="pluginListBuy"
         // @ts-expect-error
-        component={GuiPluginListScene}
-        route={{ params: { direction: 'buy' } }}
+        component={GuiPluginListBuyScene}
+        options={firstSceneScreenOptions}
```
swansontec, line 787: Should we put copies of GuiPluginViewScene in here?

swansontec (changes requested):
I have more questions than answers with this one, but at least we should fix the buy / sell direction thing.

src/actions/CryptoExchangeActions.tsx
samholmes: I forgot to add the feature to navigate back to top on the second tap.

src/components/Main.tsx
samholmes: Whoops. Not intentional.

src/components/Main.tsx
swansontec: No, because the tabs don't remember state.

samholmes: Though it may be useful to be able to switch out of the plugin view scene to the wallet scene with one tap vs a bunch of back taps to get out of this scene...

src/components/scenes/GuiPluginListScene.tsx
samholmes: Even though we discussed that we can do this later... I ended up just doing it because upon further inspection, I missed some changes that were necessary anyway. Those changes are to change each `navigate()` call for each tab's default scene for the walletList. Since I'm touch each `navigate()` call in this PR anyway, I might as well implement the final piece

swansontec (approved):

PR #4030, opened 2023-03-01
Sam/enter-amount-back-button

samholmes:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204082095021208


swansontec (approved):
I have no idea why this works, but it all looks fine.

samholmes:
> I have no idea why this works, but it all looks fine.

It works because instead of handing the back button control over to the plugin (which hasn't mounted yet), it is handled with the normal back button component. I also moved the global state change for which plugin is active to the mount event just cause that is the right place for it.

PR #4043, opened 2023-03-06
Fix the log upload regexp

swansontec:
### CHANGELOG

- Only match sensitive words surrounded by quotes (and maybe slashes). So `"loginKey"` or `"loginKey\\\\"` would match, but not `loginKey`.

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204108903827493

### DIFF
--- a/src/components/modals/LogsModal.tsx
+++ b/src/components/modals/LogsModal.tsx
@@ -18,7 +18,7 @@ interface Props {
   logs: MultiLogOutput
 }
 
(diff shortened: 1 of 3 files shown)


paullinator (approved):

swansontec:
There is a slight UI issue here, but since the modal can scroll, and since this sort of thing should never happen, I am OK with leaving it. Otherwise, maybe we could put the yellow card down in the buttons area:

simulator_screenshot_9A4CBA8E-4478-4037-B550-FFAB047AC9B5

PR #4044, opened 2023-03-06
Tweak currency wallet error handling

peachbits:
1) Show errors when creating wallets fails in account creation and when creating a single wallet
2) Show an error row when loading wallet fails

---
  - https://app.asana.com/0/0/1204119791048624

### DIFF
--- a/src/components/themed/WalletListErrorRow.tsx
+++ b/src/components/themed/WalletListErrorRow.tsx
@@ -1,0 +1,40 @@
+import * as React from 'react'
+import { TouchableOpacity } from 'react-native'
+import EntypoIcon from 'react-native-vector-icons/Entypo'
+
+import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
+import { EdgeText } from './EdgeText'
+
+interface Props {
+  onLongPress?: () => void
+  error: Error
+}
+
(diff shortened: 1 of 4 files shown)


peachbits:
Screen Shot 2023-03-06 at 2 22 10 PM

src/components/themed/WalletListErrorRow.tsx
```diff
@@ -0,0 +1,40 @@
...
+function WalletListErrorRowComponent(props: Props) {
+  const { onLongPress, error } = props
+  const theme = useTheme()
+  const styles = getStyles(theme)
+
+  return (
+    <TouchableOpacity style={styles.container} onLongPress={onLongPress}>
+      <EntypoIcon name="warning" size={theme.rem(1)} style={styles.icon} />
+      <EdgeText>{error.message}</EdgeText>
+    </TouchableOpacity>
```
swansontec, line 22: I think we upgrade the visuals later, such as adding the `CryptoIcon` for the wallet, but for now this is sooo much better.

swansontec (changes requested):
Just fix the `useWatch` thing, and maybe Michael can add a task to a future sprint if he cares deeply about the visuals.

swansontec (approved):

PR #4046, opened 2023-03-07
Support expireTime in URIs

paullinator:
### CHANGELOG

none

### Dependencies

none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204124160023937


src/actions/ScanActions.tsx
```diff
@@ -217,7 +217,10 @@ export function handleWalletUris(
         tokenId
       }
 
-      navigation.push('send2', { walletId: wallet.id, spendInfo })
+      // React navigation doesn't like passing non-serializable objects as params. Convert date to string first
+      // https://github.com/react-navigation/react-navigation/issues/7925
+      const isoExpireDate = parsedUri?.expireDate != null ? parsedUri.expireDate.toISOString() : undefined
```
swansontec, line 222: You can do `parsedUri?.expireDate?.toISOString()`. Stringing is fine, per our discussion the other day - it's not a public API, so go ahead and make the data easy to work with.

src/components/scenes/SendScene2.tsx
```diff
@@ -839,6 +860,23 @@ const SendComponent = (props: Props) => {
...
+      const expireS = Math.max(Math.round(timeoutSeconds - nowSeconds), 0)
+      setExpireSeconds(expireS)
+      if (expireS <= 0) {
+        setError(new Error(s.strings.send_address_expired_error_message))
+      }
+      timeoutHandler.current = setTimeout(() => setTick(tick + 1), 500)
+      return () => clearTimeout(timeoutHandler.current)
+    }
+  }, [expireDate, tick])
+
```
swansontec, line 879: I wonder if we can factor out a `CountdownTile` component, that handles the timer internally and presents an `onDone` handler for when we reach the end. That way, we don't re-render the huge SendScene2 component 2x per second. React will only need to diff the tile itself.

The SendScene2 would continue to provide the `onDone` logic, of course, but that only fires once after the 10 minutes are up.

paullinator: good idea. done!

src/actions/ScanActions.tsx
paullinator: done

swansontec (approved):
Oooo, this is niiiiice.

PR #4049, opened 2023-03-09
Fix crash trying to access account.disklet

paullinator:
### CHANGELOG

none

### Dependencies

none

### Requirements

- [ ] Tested on iOS device


swansontec (approved):
Very wac-a-mole, but works for now.

PR #4054, opened 2023-03-13
William/tx details hooks

swansontec:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device


src/components/scenes/TransactionDetailsScene.tsx
```diff
@@ -31,62 +23,42 @@ import { Airship, showError, showToast } from '../services/AirshipInstance'
...
-
-const getAbsoluteAmount = (edgeTransaction: EdgeTransaction): string =>
-  edgeTransaction && edgeTransaction.nativeAmount ? abs(edgeTransaction.nativeAmount) : ''
-
 // Only exported for unit-testing purposes
 class TransactionDetailsComponent extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props)
-    const { amountFiat: defaultAmountFiat = 0, edgeTransaction, thumbnailPath } = props.route.params
+    const { edgeTransaction } = props.route.params
```
samholmes, line 59: Isn't the `defaultAmountFiat` route param used by the send scene? I recall there is some weird logic to how a new trasnsaction's amount metadata propagates from the send scene to the details scene. If you tested with a send and everything looked good, then there is no concern here.

samholmes (approved):
I tried my best to map the code refactors to the original and I see how the new code works, but don't quite understand why the old code was broken. As long as certain edge cases were tested for any regressions (send, sync, resync, etc). Then this looks good to me.

src/components/modals/CategoryModal.tsx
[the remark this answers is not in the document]
swansontec: If you pass the categories in like this, the transaction details scene needs to be connected to this state and re-render when it changes. That doesn't make sense, since the transaction details scene doesn't care about the state at all - it only exists to pass into this modal. We want to limit the connections in top-level scenes as much as possible.

src/components/scenes/TransactionDetailsScene.tsx
swansontec: The transaction list is the only place in the app who passes this param. I am guessing this was done because the transaction list already has to look up the historic rate anyhow, so it was easier to pass it in than duplicate the logic.

Now that we have hooks, we *can* duplicate the logic, and we can make it live-updating instead of relying on the one-shot `componentDidMount`. I don't fully understand why the old code was failing, since all the dots were being connected, but some sort of race condition was preventing the data from getting where it needed to go.

samholmes: Okay, that's a great explanation. I really tried to dig to figure out what the bug was, but if it's one of those obscure issue that's resolved by just simply cleaning up the code, then that's well and good. This PR has solid changes that all look right to me, the only concern is with the fragility around this code and possible regressions there. But this isn't a large concern because I feel like these issues will be resolved during QA and we should be pushing this change out and making things more robust anyway.

All of it is approved with no changes that I can perceive to be needed.

PR #4059, opened 2023-03-14
Fix list usage in create / migrate wallet scenes

swansontec:
FlashList is like an upgraded FlatList, but more aggressive with optimization. Because of this, we need to pass in `extraData` so it will update correctly.

FlashList also doesn't take a CSS `style` prop, so we need to delete those or get yellow warning boxes.

Also, we need to apply `useHandler` correctly, or not at all.

### CHANGELOG

none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204122607311075

### DIFF
--- a/src/components/scenes/CreateWalletSelectFiatScene.tsx
+++ b/src/components/scenes/CreateWalletSelectFiatScene.tsx
@@ -1,4 +1,4 @@
-import { FlashList, ListRenderItemInfo } from '@shopify/flash-list'
+import { FlashList, ListRenderItem } from '@shopify/flash-list'
 import * as React from 'react'
 import { View } from 'react-native'
 import FastImage from 'react-native-fast-image'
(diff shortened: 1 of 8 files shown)


src/components/scenes/CreateWalletSelectFiatScene.tsx
```diff
@@ -211,13 +213,7 @@ const CreateWalletSelectFiatComponent = (props: Props) => {
...
           {s.strings.fragment_create_wallet_instructions}
         </EdgeText>
-        <FlashList
-          style={styles.resultList}
-          automaticallyAdjustContentInsets={false}
-          data={createWalletList}
-          keyExtractor={item => item.key}
-          renderItem={renderCurrencyRow}
-        />
+        <FlashList automaticallyAdjustContentInsets={false} data={createWalletList} keyExtractor={keyExtractor} renderItem={renderCurrencyRow} />
```
peachbits, line 216: I believe for this list it walletNames should be the "extraData"

peachbits (changes requested):

peachbits (approved):

peachbits (approved):

PR #4071, opened 2023-03-17
Add tracking event data to logs

Jon-edge:
### CHANGELOG

added- Add tracking event data to logs

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204207399694359

### DIFF
--- a/src/util/tracking.ts
+++ b/src/util/tracking.ts
@@ -3,6 +3,7 @@ import { getUniqueId, getVersion } from 'react-native-device-info'
 
 import { ENV } from '../env'
 import { fetchReferral } from './network'
+import { consify } from './utils'
 
 export type TrackingEventName =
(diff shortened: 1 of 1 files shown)


src/util/tracking.ts
```diff
@@ -80,15 +82,16 @@ async function logToFirebase(name: TrackingEventName, values: TrackingValues) {
   if (installerId != null) params.aid = installerId
   if (pluginId != null) params.plugin = pluginId
   if (error != null) params.error = error
-  // @ts-expect-error
-  global.firebase.analytics().logEvent(name, params)
```
paullinator, line 84: You are changing logic which is not a part of the task. For now keep both log events.

paullinator (changes requested):

paullinator (approved):

PR #4073, opened 2023-03-20
Replace FIO Actions with makeSpend, signTx, broadcastTx flow

peachbits:
---
  - https://app.asana.com/0/0/1204112329611718
  - https://app.asana.com/0/0/1204211925285852

### DIFF
--- a/src/modules/FioAddress/util.ts
+++ b/src/modules/FioAddress/util.ts
@@ -585,7 +585,7 @@ export const getFioObtData = async (fioWallets: EdgeCurrencyWallet[]): Promise<F
   let obtDataRecords: FioObtRecord[] = []
   for (const fioWallet of fioWallets) {
     try {
-      const { obt_data_records: lastRecords } = await fioWallet.otherMethods.fioAction('getObtData', {})
+      const { obt_data_records: lastRecords } = await fioWallet.otherMethods.getObtData()
       obtDataRecords = [...obtDataRecords, ...lastRecords]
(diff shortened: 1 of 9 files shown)


peachbits:
In draft status since it is blocked by accountbased PR

src/modules/FioAddress/util.ts
```diff
@@ -120,6 +120,28 @@ export const setFioExpiredCheckToDisklet = async (lastChecks: { [fioName: string
   }
 }
 
+export const fioMakeSpend = async (fioWallet: EdgeCurrencyWallet, actionName: string, params: any): Promise<EdgeTransaction> => {
```
swansontec, line 135: I would do `params: unknown` instead of `any`. This is because `unknown` forces you to clean before touching, so it's safer.

src/modules/FioAddress/components/ConnectWallets.tsx
```diff
@@ -197,6 +197,7 @@ class ConnectWallets extends React.Component<Props, LocalState> {
             {walletItems && Object.keys(walletItems).length ? (
               <FlashList
                 data={Object.values(walletItems)}
+                extraData={{ ...this.state.connectWalletsMap, ...this.state.disconnectWalletsMap }}
```
swansontec, line 200: This got fixed in the other PR, so I guess this commit will get dropped in the rebase.

swansontec (approved):

PR #4074, opened 2023-03-20
Visual gap at bottom of GuiPluginViewScene

samholmes:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204196564436860

### DIFF
--- a/src/components/scenes/GuiPluginViewScene.tsx
+++ b/src/components/scenes/GuiPluginViewScene.tsx
@@ -255,7 +255,7 @@ class GuiPluginView extends React.Component<Props, State> {
         : 'Mozilla/5.0 (iPhone; CPU iPhone OS 10_3_1 like Mac OS X) AppleWebKit/603.1.30 (KHTML, like Gecko) Version/10.0 Mobile/14E304 Safari/602.1'
 
     return (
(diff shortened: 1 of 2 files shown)


swansontec (approved):

src/components/scenes/GuiPluginViewScene.tsx
```diff
@@ -255,7 +255,7 @@ class GuiPluginView extends React.Component<Props, State> {
         : 'Mozilla/5.0 (iPhone; CPU iPhone OS 10_3_1 like Mac OS X) AppleWebKit/603.1.30 (KHTML, like Gecko) Version/10.0 Mobile/14E304 Safari/602.1'
 
     return (
-      <SceneWrapper background="body">
+      <SceneWrapper background="body" hasTabs>
```
swansontec, line 258: Actually, it should be `hasTabs={route.name !== 'pluginView'}`, since we have tabs when we get routed as `'pluginViewBuy' | 'pluginViewSell'`, but not when we get routed to as the floating plugin scene.

swansontec (changes requested):

swansontec (approved):

PR #4076, opened 2023-03-21
Fix FlashList in connected wallet scene

peachbits:
Also tweaked the footer style 

- [x] Tested on iOS device

Screen Shot 2023-03-21 at 11 55 41 AM


peachbits:
Screen Shot 2023-03-21 at 11 59 18 AM

src/modules/FioAddress/components/ConnectWallets.tsx
```diff
@@ -197,6 +197,7 @@ class ConnectWallets extends React.Component<Props, LocalState> {
             {walletItems && Object.keys(walletItems).length ? (
               <FlashList
                 data={Object.values(walletItems)}
+                extraData={{ ...this.state.connectWalletsMap, ...this.state.disconnectWalletsMap }}
```
swansontec, line 200: This object gets re-created on every render, which means the list itself will re-render every time as well. I would say `useMemo` to only re-create `extraData` when its dependencies change, but this is a class component, so we are in trouble here.

Since we can't use the easy solution, I suggest `// TODO: Cache this with useMemo for better perf once we switch to hooks`. At least this way we acknowledge the problem.

swansontec (changes requested):

swansontec (approved):

PR #4081, opened 2023-03-24
Always pop SendScene2

paullinator:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204220396495751

### DIFF
--- a/src/components/scenes/SendScene2.tsx
+++ b/src/components/scenes/SendScene2.tsx
@@ -800,6 +800,7 @@ const SendComponent = (props: Props) => {
       playSendSound().catch(error => console.log(error)) // Fail quietly
 
       if (onDone) {
+        navigation.pop()
         onDone(null, broadcastedTx)
       } else {
         navigation.replace('transactionDetails', {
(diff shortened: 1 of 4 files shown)


src/components/scenes/SendScene2.tsx
```diff
@@ -801,6 +801,7 @@ const SendComponent = (props: Props) => {
 
       if (onDone) {
         onDone(null, broadcastedTx)
+        navigation.pop()
```
swansontec, line 804: As mentioned at the meeting yesterday, the `navigation.pop()` should go *before* the `onDone()` call. That gives the `onDone` callback the flexibility to do things like`navigation.replace(next-step-in-the-flow)`, without the send scene's `navigation.pop` messing that up.

swansontec (approved):

src/components/scenes/SendScene2.tsx
paullinator: 👍

PR #4084, opened 2023-03-24
No legacy platform variables

swansontec:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204256160179278


src/components/modals/FlipInputModal.tsx
```diff
@@ -251,9 +249,11 @@ export class FlipInputModalComponent extends React.PureComponent<Props, State> {
   }
 }
 
+const deviceHeight = Dimensions.get('window').height
+
 const getStyles = cacheStyles((theme: Theme) => ({
   hackContainer: {
-    flex: deviceHeight <= MINIMUM_DEVICE_HEIGHT ? 1 : 0
+    flex: deviceHeight <= 580 ? 1 : 0
```
samholmes, line 256: Why was the constant removed? I like that it described the value. Can we at least move it to the top of this file?

samholmes, line 256: Why was the constant removed? I like that it described the value. Can we at least move it to the top of this file?

src/components/modals/HelpModal.tsx
```diff
@@ -162,8 +163,8 @@ const getStyles = cacheStyles((theme: Theme) => ({
     height: theme.rem(2.25)
   },
   footer: {
-    marginTop: PLATFORM.deviceHeight < theme.rem(42) ? 0 : theme.rem(1.5),
-    paddingVertical: PLATFORM.deviceHeight < theme.rem(42) ? theme.rem(0.25) : theme.rem(0.5),
+    marginTop: deviceHeight < theme.rem(42) ? 0 : theme.rem(1.5),
+    paddingVertical: deviceHeight < theme.rem(42) ? theme.rem(0.25) : theme.rem(0.5),
```
samholmes, line 167: Might be an opportunity to explain this logic.

samholmes (approved):
Approved with minor feedback

src/components/modals/FlipInputModal.tsx
swansontec: The constant did not describe anything. If I gave you a constant named "MINIMUM_DEVICE_HEIGHT" and no context, what would you guess it meant? Probably "the app won't run on phones less than 580px tall", or something incorrect like that.

The CSS section is always full of numbers. That's what they are for - to describe how things look. Given it's new home in the CSS section, the number is no longer misleading. It's perfectly obvious that the modal flexes only if the phone is below 580px in size now.

src/components/modals/HelpModal.tsx
swansontec: I have no idea.

src/components/modals/FlipInputModal.tsx
samholmes: Fair enough.

src/components/modals/HelpModal.tsx
samholmes: Ha! Okay

PR #4086, opened 2023-03-24
Move more modules

swansontec:
Move more sources out of the "modules" folder and into the right places.

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device


src/reducers/UiReducer.ts
samholmes, line 1: This file exports a singleton, but not function which instantiates. So we should probably use camelCasing here.

samholmes (approved):
Minor feedback regarding filename

src/reducers/UiReducer.ts
swansontec: Thanks, fixed it.

PR #4087, opened 2023-03-28
William/more lint rules

swansontec:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204275835641700

### DIFF
--- a/src/plugins/borrow-plugins/plugins/aave/AaveNetwork.ts
+++ b/src/plugins/borrow-plugins/plugins/aave/AaveNetwork.ts
@@ -89,6 +89,7 @@ export const makeAaveNetworkFactory = (blueprint: AaveNetworkBlueprint): AaveNet
     //
 
     async getAssetPrice(address): Promise<BigNumber> {
+      // eslint-disable-next-line @typescript-eslint/return-await
       return await IPriceOracle.getAssetPrice(address)
     },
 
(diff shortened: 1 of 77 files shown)


src/plugins/borrow-plugins/plugins/aave/AaveNetwork.ts
```diff
@@ -89,6 +89,7 @@ export const makeAaveNetworkFactory = (blueprint: AaveNetworkBlueprint): AaveNet
     //
 
     async getAssetPrice(address): Promise<BigNumber> {
+      // eslint-disable-next-line @typescript-eslint/return-await
```
paullinator, line 92: why the disable?

src/modules/FioAddress/util.ts
```diff
@@ -788,6 +788,7 @@ export const addBundledTxs = async (fioWallet: EdgeCurrencyWallet | null, fioAdd
   if (fioWallet) {
     try {
       const params = { fioAddress, bundleSets: DEFAULT_BUNDLE_SET_VALUE, maxFee: fee }
+      // eslint-disable-next-line @typescript-eslint/return-await
```
paullinator, line 791: why the disable?

paullinator (approved):
Just one small question

src/plugins/borrow-plugins/plugins/aave/AaveNetwork.ts
swansontec: The `otherMethods` are `any`, so ESLint doesn't see that the return value is a promise. We know the value is a promise, so we disable the rule.

PR #4096, opened 2023-03-31
Add Algorand

peachbits:
This PR includes a couple minor changes in addition to adding Algorand

- Fix spam filter tx dropdown logic
- Fix transaction row show as sentd/receive
- Move token activation text to special info
- Fix wallet row tap doing nothing if token needs to activate and parent wallet has insufficient funds

---
  - https://app.asana.com/0/0/1203860841541940


peachbits:
Ready for review but blocked for merging by https://github.com/EdgeApp/edge-currency-accountbased/pull/537

src/actions/TransactionListActions.ts
```diff
@@ -178,7 +178,10 @@ export function newTransactionsRequest(navigation: NavigationBase, walletId: str
     if (isTransactionForSelectedWallet) dispatch(fetchTransactions(walletId, selectedCurrencyCode, options))
     if (receivedTxs.length) dispatch(checkFioObtData(walletId, receivedTxs))
     if (!isReceivedTransaction(edgeTransaction)) return
-    if (spamFilterOn && !zeroString(exchangeRate) && gte(edgeTransaction.nativeAmount, calculateSpamThreshold(exchangeRate, exchangeDenom))) {
+    if (
+      !spamFilterOn ||
+      (spamFilterOn && !zeroString(exchangeRate) && gte(edgeTransaction.nativeAmount, calculateSpamThreshold(exchangeRate, exchangeDenom)))
```
swansontec, line 183: The first `!spamFilterOn ||` bit ensures that `spamFilterOn` will always be true in the second bit. Therefore, you can drop the `(spamFilterOn && ` part, since that's redundant.

src/util/utils.ts
```diff
@@ -209,7 +209,7 @@ export const isReceivedTransaction = (edgeTransaction: EdgeTransaction): boolean
 }
 
 export const isSentTransaction = (edgeTransaction: TransactionListTx | EdgeTransaction): boolean => {
-  return !!edgeTransaction.nativeAmount && edgeTransaction.nativeAmount.charAt(0) === '-'
+  return edgeTransaction.ourReceiveAddresses.length === 0 || (!!edgeTransaction.nativeAmount && edgeTransaction.nativeAmount.charAt(0) === '-')
 }
```
swansontec, line 213: The transaction details scene has its own direction logic, and there may be other places as well. That makes this change incomplete. What does it mean to receive a negative transaction, anyhow? We may need a different solution for this, like adding an `isSend: boolean` flag to `EdgeTransaction`.

src/actions/WalletActions.tsx
```diff
@@ -227,12 +227,27 @@ const activateWalletTokens = async (
...
+        paymentWalletId,
+        paymentTokenId: tokenId
+      })
+    } catch (e: any) {
+      if (e.name === 'InsufficientFundsError' && e.networkFee != null) {
+        // Allow the user to proceed to token activation scene with fee explanation
+        // @ts-expect-error
+        activationQuote = {
+          networkFee: { nativeAmount: e.networkFee, currencyPluginId: wallet.currencyInfo.pluginId, tokenId }
+        }
```
swansontec, line 244: I don't really like this solution, since the scene needs a quote. If we are turning the error into a quote, that says we should have probably just returned a quote from the engine. The engine-provided quote will then fail when the user calls  `approve()`, as we expect.

src/actions/WalletActions.tsx
```diff
@@ -270,10 +270,9 @@ const activateWalletTokens = async (
     const feeString = `${displayFee} ${feeDenom.name} (${fiatFee} ${fiatCurrencyCode.replace('iso:', '')})`
     let bodyText = s.strings.activate_wallet_token_scene_body
 
-    // HACK: XRP causes the reserve requirement to increase per token activated. There's
-    // No good way to parametrize this and abstract away the asset
-    if (wallet.currencyInfo.pluginId === 'ripple') {
-      bodyText += '\n\n' + s.strings.activate_wallet_token_scene_body_xrp_extra
+    const info = getSpecialCurrencyInfo(wallet.currencyInfo.pluginId)
```
swansontec, line 273: Optional: I have been using this structure in new code:

```ts
const { tokenActivationAdditionalReserveText } = SPECIAL_CURRENCY_INFO[pluginId] ?? {}
```

This gives us a chance to assign a default value if we need it. We don't need the `getSpecialCurrencyInfo` helper in this case, since TypeScript understands the `?? {}` bit.

swansontec (changes requested):
We may need to update accountbased too for some of these fixes.

src/util/utils.ts
peachbits: Honestly, I didn't love my solution and am definitely open to starting a conversation. The case here is a valid transaction has a zero amount and is a send but obviously doesn't start with a '-'. `ourReceiveAddresses` is the closest we have to an implied `isSend` flag which is why I opted for that rather than assuming zero amount txs should be sends.

swansontec (commented):
I am approving what I see so far, but we do need to resolve the direction issue before we merge this. If we are going to go with the hack version, we at least need to do it to the transaction details scene too.

src/util/utils.ts
swansontec: Ok, we can merge this as-is, and then fix the `isSend` thing in the next sprint cycle.

swansontec (approved):

PR #4102, opened 2023-04-05
William/fix fiat deep links

swansontec:
### CHANGELOG

- fixed: Restore the original "edge://plugin/pluginId" deep link behavior for non-fiat plugins. For fiat plugins, parse the new "edge://plugin/pluginId/direction/provider/paymentType" format.

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204331801949563

### DIFF
--- a/src/actions/DeepLinkingActions.ts
+++ b/src/actions/DeepLinkingActions.ts
@@ -5,6 +5,7 @@ import { pickWallet } from '../components/modals/WalletListModal'
 import { showError, showToast } from '../components/services/AirshipInstance'
 import { guiPlugins } from '../constants/plugins/GuiPlugins'
 import { lstrings } from '../locales/strings'
+import { executePlugin } from '../plugins/gui/fiatPlugin'
 import { DeepLink } from '../types/DeepLinkTypes'
 import { Dispatch, RootState, ThunkAction } from '../types/reduxTypes'
 import { NavigationBase } from '../types/routerTypes'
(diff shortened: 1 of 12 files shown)


src/actions/DeepLinkingActions.ts
```diff
@@ -89,6 +97,34 @@ export async function handleLink(navigation: NavigationBase, dispatch: Dispatch,
       return true
     }
 
+    case 'fiatPlugin': {
```
paullinator, line 100: why introduce another link type? You already have to handle native plugins in GuiPluginListScene.tsx so you could just route to it with the plugin object like we did before.

paullinator (approved):

paullinator:
Looks like this should work. Approved with one question

src/actions/DeepLinkingActions.ts
swansontec: This is a confusing question, since you seem to be conflating two separate issues.

The first issue is whether we should route to the GuiPluginListScene, since it knows how to launch plugins. The answer to that should be a resounding nooooo! Run away in terrror! Scenes exist to interact with the user. They aren't supposed to be "callable" like functions. If you need to launch a scene to get access to it's innards, that's a sign that the scene contains too much critical logic. So instead, this PR pulled a bunch of the "smarts" out of the plugin list scene and into the right places, so launching plugins becomes easy again.

The second issue is whether we need a separate link type. The options there are:

1) Do all the parsing in one place, including extracting the direction / provider / payment type for fiat plugins. This means having separate types, since WebView plugins a deep path and deep query instead of those things.
2) Use a single link type, but extract the direction / provider / payment type from `deepPath` as part of launching the plugin.

Of those two options, 1 seemed the best, since all parsing happens in one place.

PR #4103, opened 2023-04-05
Fix unresolved getHistoricalRate promises

paullinator:
Resolve null rates to 0

### CHANGELOG

none

### Dependencies

none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204293288180110


swansontec (approved):
Approved, but we need to keep an eye on this to figure out why it's happening.

PR #4105, opened 2023-04-06
Fix the bottom gap on the transaction list scene

swansontec:
Also remove the second copy, so the scene only exists inside the tab navigator.

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204340335884555
  - https://app.asana.com/0/0/1204343630586365

### DIFF
--- a/src/components/Main.tsx
+++ b/src/components/Main.tsx
@@ -703,15 +703,6 @@ const EdgeAppStack = () => {
         }}
       />
       <Stack.Screen
-        name="transactionList"
-        component={TransactionList}
-        listeners={{
-          focus: () => {
-            requestPermission('contacts').catch(showError)
-          }
-        }}
-      />
-      <Stack.Screen
         name="transactionsExport"
         component={TransactionsExportScene}
         options={{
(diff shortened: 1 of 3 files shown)


src/components/Main.tsx
samholmes, line 1: This should be in a separate commit, but that's just a nit that I'm not necessarily going to push for.

src/components/scenes/GuiPluginViewScene.tsx
```diff
@@ -190,7 +190,7 @@ export function GuiPluginViewScene(props: Props): JSX.Element {
       : 'Mozilla/5.0 (iPhone; CPU iPhone OS 10_3_1 like Mac OS X) AppleWebKit/603.1.30 (KHTML, like Gecko) Version/10.0 Mobile/14E304 Safari/602.1'
 
   return (
-    <SceneWrapper background="theme">
+    <SceneWrapper background="theme" hasTabs={route.name !== 'pluginView'}>
```
samholmes, line 193: Didn't I do this for this scene from a previous review? I was confident that I added this as a fixup from review feedback. What happened there in retrospect, I wonder.

samholmes (approved):
Looks good to me.

src/components/scenes/GuiPluginViewScene.tsx
swansontec: Yes, you did do it. However, it got lost during the EdgeProvider re-write.

PR #4120, opened 2023-04-12
Fix New User Free Fio Handle Scene

Jon-edge:
image

image

### CHANGELOG

- fixed: FIO Create Handle: Unable to delete first two characters of the handle when using backspace.
- fixed: FIO Create Handle: Double tapping clear in the handle input box removes the domain.

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204385705490826
  - https://app.asana.com/0/0/1204378156120794
  - https://app.asana.com/0/0/1204366217589433

### DIFF
--- a/src/components/scenes/Fio/FioCreateHandleScene.tsx
+++ b/src/components/scenes/Fio/FioCreateHandleScene.tsx
@@ -147,13 +147,6 @@ export const FioCreateHandleScene = ({ navigation, route }: Props) => {
 
   // Create the new FIO wallet, default the handle to a cleaned version of the username
   useAsyncEffect(async () => {
-    const wallet = await dispatch(createFioWallet())
-
-    if (!mounted.current) return
-    setWallet(wallet)
-
-    handleChangeFioHandle(accountUserName)
-
     const domains = await fioPlugin.otherMethods.getDomains(freeRegRefCode)
     if (domains.length === 1) {
       if (!mounted.current) return
@@ -161,8 +154,15 @@ export const FioCreateHandleScene = ({ navigation, route }: Props) => {
         setDomainStr(`${FIO_ADDRESS_DELIMITER}${asFreeFioDomain(domains[0]).domain}`)
       } catch (e) {
         setErrorText(lstrings.fio_register_handle_error)
+        return
       }
     }
+    handleChangeFioHandle(accountUserName)
+
+    const wallet = await dispatch(createFioWallet())
+
+    if (!mounted.current) return
+    setWallet(wallet)
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 
@@ -187,7 +187,7 @@ export const FioCreateHandleScene = ({ navigation, route }: Props) => {
           <OutlinedTextInput
             ref={inputRef}
             suffix={domainStr}
-            value={fioHandle}
+            value={domainStr === '' ? '' : fioHandle}
             onChangeText={handleChangeFioHandle}
             autoCapitalize="none"
             autoCorrect={false}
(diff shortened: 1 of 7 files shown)


src/components/scenes/Fio/FioCreateHandleScene.tsx
```diff
@@ -69,12 +69,18 @@ export const FioCreateHandleScene = ({ navigation, route }: Props) => {
   const inputRef = React.useRef<TextInput>(null)
   const mounted = React.useRef<boolean>(true)
 
+  // TODO: Give feedback to the user to indicate why the input was rejected
   const handleChangeFioHandle = useHandler((userInput: string) => {
-    // Clean the userInput:
-    userInput = userInput.replace(domainStr, '').trim()
+    // Dash '-' allowed,
+    // but cannot be the first...
+    if (userInput.indexOf('-') === 0) {
```
paullinator, line 76: Unnecessary change. Could have left the original line 77 and reduce regression potential.

src/actions/LoginActions.tsx
```diff
@@ -73,22 +73,28 @@ export function initializeAccount(navigation: NavigationBase, account: EdgeAccou
...
+
+        // New user FIO handle registration flow (if env is properly configured)
+        const { freeRegApiToken = '', freeRegRefCode = '' } = typeof ENV.FIO_INIT === 'object' ? ENV.FIO_INIT : {}
+        const isCreateHandle =
+          freeRegApiToken !== '' && freeRegRefCode !== '' && (await Airship.show<boolean>(bridge => <FioCreateHandleModal bridge={bridge} />))
+
+        if (isCreateHandle) {
+          // We need to start the wallet creation process before starting the
+          // FIO workflow or wallet creation can fail.
+          createCustomWallets(account, fiatCurrencyCode, selectedEdgetokenIds, dispatch)
```
paullinator, line 92: Definitely want to await this before calling updateWalletsRequest()

src/components/themed/OutlinedTextInput.tsx
```diff
@@ -332,47 +324,48 @@ export const OutlinedTextInput = React.forwardRef<OutlinedTextInputRef, Outlined
...
-            secureTextEntry={hidePassword}
-            // Callbacks:
-            onBlur={handleBlur}
-            onChangeText={onChangeText}
-            onFocus={handleFocus}
-            maxLength={maxLength}
-            onSelectionChange={handleSelectionChange}
-          />
-        )}
+        <View style={[styles.suffixContainer, suffixPadding]}>
```
paullinator, line 327: This affects ALL use of OutlineTextInput and I can't confirm this doesn't cause a regression. I would reduce the regression surface by limiting to textinput, and conditionalize to when `suffix` is being used. Even if it means copy and pasting the entire <TextInput> block, I'd be ok with that and prefer it to affecting all uses.

paullinator (changes requested):

src/components/scenes/Fio/FioCreateHandleScene.tsx
Jon-edge: Line 77 was causing the bug where the user was unable to delete the first two characters via backspace

src/actions/LoginActions.tsx
Jon-edge: Commit removed

src/components/scenes/Fio/FioCreateHandleScene.tsx
```diff
@@ -71,6 +71,9 @@ export const FioCreateHandleScene = ({ navigation, route }: Props) => {
 
   // TODO: Give feedback to the user to indicate why the input was rejected
   const handleChangeFioHandle = useHandler((userInput: string) => {
+    // Clean spaces:
+    userInput = userInput.replace(domainStr, '').trim()
+
```
paullinator, line 76: why did this get added back. There should be no domain in the string

paullinator (changes requested):

src/components/scenes/Fio/FioCreateHandleScene.tsx
Jon-edge: You're right, intended to ensure spaces were cleaned too but it's already handled.

paullinator (approved):

PR #4124, opened 2023-04-13
Replace fake bugsnag key with valid hexidecimal

peachbits:


peachbits:
Screen Shot 2023-04-13 at 4 54 44 PM

swansontec (approved):

PR #4125, opened 2023-04-14
Use tx.isSend from edge-core-js 0.20.1

paullinator:
- Update all code that created EdgeTransaction objects to also include `isSend`

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204343038506603

### DIFF
--- a/src/components/themed/TransactionListRow.tsx
+++ b/src/components/themed/TransactionListRow.tsx
@@ -21,7 +21,6 @@ import {
   decimalOrZero,
   DEFAULT_TRUNCATE_PRECISION,
   getDenomFromIsoCode,
-  isSentTransaction,
   maxPrimaryCurrencyConversionDecimals,
   precisionAdjust,
   truncateDecimals
@@ -106,7 +105,6 @@ export function TransactionListRow(props: Props) {
       fiatSymbol={getSymbolFromCurrency(nonIsoFiatCurrencyCode)}
       onPress={handlePress}
       onLongPress={handleLongPress}
-      isSentTransaction={isSentTransaction(transaction)}
       requiredConfirmations={requiredConfirmations}
       selectedCurrencyName={currencyName || currencyCode}
       thumbnailPath={thumbnailPath}
--- a/src/components/themed/TransactionRow.tsx
+++ b/src/components/themed/TransactionRow.tsx
@@ -1,3 +1,4 @@
+import { eq } from 'biggystring'
 import { EdgeCurrencyWallet } from 'edge-core-js'
 import * as React from 'react'
 import { View } from 'react-native'
@@ -21,7 +22,6 @@ interface Props {
   fiatSymbol: string
   onPress: () => void
   onLongPress: () => void
-  isSentTransaction: boolean
   requiredConfirmations: number
   selectedCurrencyName: string
   thumbnailPath?: string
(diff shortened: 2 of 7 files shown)


src/components/themed/TransactionListRow.tsx
```diff
@@ -106,7 +105,7 @@ export function TransactionListRow(props: Props) {
       fiatSymbol={getSymbolFromCurrency(nonIsoFiatCurrencyCode)}
       onPress={handlePress}
       onLongPress={handleLongPress}
-      isSentTransaction={isSentTransaction(transaction)}
+      isSentTransaction={transaction.isSend}
```
swansontec, line 108: Let's do this differently:

1. Remove the `isSentTransaction` flag, since we already pass the full `transaction` itself to the row.
2. Inside the row, continue to use `> 0` for deciding whether to show the `-` sign or not.
3. Inside the row, pass `>= 0 ? 'income' : 'expense'` as the second parameter to `splitCategory`, to pick the right default category.

swansontec (changes requested):

src/components/themed/TransactionRow.tsx
```diff
@@ -50,6 +49,8 @@ const TransactionRowComponent = (props: Props) => {
 
   const { canReplaceByFee = false } = wallet.currencyInfo
 
+  const isSentTransaction = transaction.nativeAmount.startsWith('-') || (eq(transaction.nativeAmount, '0') && transaction.isSend)
+
   const cryptoAmountString = `${isSentTransaction ? '-' : '+'} ${denominationSymbol ? denominationSymbol + ' ' : ''}${cryptoAmount}`
```
swansontec, line 54: Using `isSentTransaction` on this line seems like a bug. We should put a - sign or a + sign based on whether the amount is positive or negative, and nothing else.

src/components/themed/TransactionRow.tsx
```diff
@@ -85,10 +86,11 @@ const TransactionRowComponent = (props: Props) => {
   const pendingStyle = currentConfirmations === 'confirmed' ? styles.completedTime : styles.partialTime
 
   // Transaction Category
+  const defaultCategory = !isSentTransaction ? 'income' : 'expense'
```
swansontec, line 89: This default category logic also exists in the TransactionDetailsScene on line 67. You should copy the fancy `isSentTransaction` logic there so the two scenes agree. Otherwise QA will send us a bug where a transaction with an `undefined` category says "Income" in the transaction list but "Expense" in the transaction details scene.

swansontec (changes requested):

src/components/themed/TransactionRow.tsx
paullinator: The only time this would be borderline incorrect is if the amount is '0' such as for Algorand token activations which show 0 amount spends on the token tx. At worst it would show "-0" sent which is "kind of" correct.

swansontec: Why the heck does number formatting have anything to do with who signed the transaction!? The point of adding `isSend` was to *fix* bugs like this, not just replace the existing logic. I will approve this, but when QA complains about "-0" transactions, you get to fix it.

swansontec (approved):

PR #4127, opened 2023-04-17
Fix placeholder Bugsnag api key

peachbits:
---
  - https://app.asana.com/0/0/1204407648172338


swansontec (approved):
Approved. Could we use something like "11111111111111111111111111111111" to make it more obvious that this is fake? Otherwise, I see how this would trick bugsnag more effectively.

PR #4128, opened 2023-04-17
Fix async method usage

peachbits:
---
  - https://app.asana.com/0/0/1204414162532213

### DIFF
--- a/src/components/scenes/WcConnectScene.tsx
+++ b/src/components/scenes/WcConnectScene.tsx
@@ -77,6 +77,7 @@ export const WcConnectScene = (props: Props) => {
       })
     } catch (e: any) {
       showError('Failed to connect, try again.')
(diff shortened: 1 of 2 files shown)


src/components/scenes/WcConnectScene.tsx
```diff
@@ -77,6 +77,7 @@ export const WcConnectScene = (props: Props) => {
       })
     } catch (e: any) {
       showError('Failed to connect, try again.')
+      console.error('wcInit error:', e.message)
```
swansontec, line 80: Never access `.message` unless you have first cleaned the error. In this case, you can just do `console.error('wcInit error:', e)` and not worry about the cleaning, since that will log the message just the same.

swansontec (approved):

PR #4131, opened 2023-04-18
Fixed uncentered Edge header logo on Android devices

itayplav:
### CHANGELOG
 
Fixed uncentered Edge header logo on Android devices

### Dependencies

None

### Requirements

- [ ] Tested on iOS device

| Wallet 	| Markets 	|
|---	|---	|
| wallet 	| markets 	|

---
  - https://app.asana.com/0/0/1204206920053793

### DIFF
--- a/src/components/navigation/EdgeLogoHeader.tsx
+++ b/src/components/navigation/EdgeLogoHeader.tsx
@@ -2,6 +2,7 @@ import * as React from 'react'
 import { View } from 'react-native'
 import FastImage from 'react-native-fast-image'
 
+import { lstrings } from '../../locales/strings'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 
 export function EdgeLogoHeader() {
@@ -9,7 +10,7 @@ export function EdgeLogoHeader() {
   const styles = getStyles(theme)
   return (
     <View style={styles.container}>
-      <FastImage style={styles.icon} source={theme.headerIcon} resizeMode="contain" accessibilityHint="Edge Logo Header" />
+      <FastImage style={styles.icon} source={theme.headerIcon} resizeMode="contain" accessibilityHint={lstrings.app_logo_hint} />
     </View>
   )
 }
--- a/src/components/navigation/EdgeLogoHeader.tsx
+++ b/src/components/navigation/EdgeLogoHeader.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react'
-import { Image, View } from 'react-native'
+import { View } from 'react-native'
+import FastImage from 'react-native-fast-image'
 
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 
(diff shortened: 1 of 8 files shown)


src/components/navigation/EdgeLogoHeader.tsx
```diff
@@ -1,14 +1,16 @@
...
 
+import edgeMark from '../../assets/images/edgeLogo/Edge_logo_Icon.png'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 
 export function EdgeLogoHeader() {
   const theme = useTheme()
   const styles = getStyles(theme)
   return (
     <View style={styles.container}>
-      <Image style={styles.icon} source={theme.headerIcon} resizeMode="contain" />
```
swansontec, line 11: The reason the header image comes from the theme is because we have white-label versions of our app with different logos. Simply swapping out the theme allows us to re-brand the app as needed. So, please put the source back the way it was.

swansontec (approved):

swansontec (changes requested):
Oops, hit the wrong button.

src/components/navigation/EdgeLogoHeader.tsx
itayplav: Reverted, and now uses a new `ImageProp` declared in `Theme.ts`.

src/components/navigation/EdgeLogoHeader.tsx
```diff
@@ -8,7 +9,7 @@ export function EdgeLogoHeader() {
...
+      <FastImage style={styles.icon} source={theme.headerIcon} resizeMode="contain" accessibilityHint="Edge Logo Header" />
```
swansontec, line 12: The `accessibilityHint` should be a translated string, and it can't say "Edge", because white-labels again. Maybe add `lStrings.logo_hint: "App logo"`

src/types/Theme.ts
```diff
@@ -1,6 +1,8 @@
 import { asNumber, asObject } from 'cleaners'
 import { ImageSourcePropType } from 'react-native'
```
swansontec, line 2: Weren't you going to remove `ImageSourcePropType`, and just use the `ImageProp`?

swansontec (changes requested):
Getting close.

src/components/navigation/EdgeLogoHeader.tsx
itayplav: If we are treating `accessibilityHint`'s as first-class citizen strings then https://github.com/EdgeApp/edge-login-ui-rn/pull/42 needs to be fixed too.

itayplav: Looks like we are, so this string has been moved to it's own 'section' in [`cdb6412` (#4131)](https://github.com/EdgeApp/edge-react-gui/pull/4131/commits/cdb64129d285558c969d0ec97de46546473c72d3)

swansontec (approved):
Per our engineering housekeeping note a few weeks ago, the label should be app_logo_hint, not hint_app_logo. However, I am approving in any case just to avoid the back & forth.

PR #4139, opened 2023-04-19
Add proper bottom padding to fade-out modals

swansontec:
### CHANGELOG

- fixed: Make the last item tappable in the wallet list modal and default swap setting modal.

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204392945688667

### DIFF
--- a/src/components/themed/ModalParts.tsx
+++ b/src/components/themed/ModalParts.tsx
@@ -41,6 +41,13 @@ export function ModalMessage(props: { children: React.ReactNode; paddingRem?: nu
   return <Text style={[styles.messageText, padding, isWarning && styles.warningText]}>{children}</Text>
 }
 
+/**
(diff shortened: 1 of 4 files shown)


src/components/themed/ModalParts.tsx
```diff
@@ -61,6 +68,8 @@ export function ModalFooter(props: ModalFooterProps) {
   )
 }
 
+ModalFooter.bottomRem = 2.5
```
samholmes, line 71: Oh interesting. What's the benefit for exporting this value as a static property? Usually we export some relevant constants from the module.

samholmes (approved):

src/components/themed/ModalParts.tsx
swansontec: It's a little funky, but cute.

PR #4142, opened 2023-04-20
Add "Private View Key" menu option for Monero wallets

samholmes:
### CHANGELOG

- Changed: Monero "Private View Key" wallet menu option replaces "View XPub Address" and includes disclaimer.

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204176035281206

### DIFF
--- a/src/actions/WalletListMenuActions.tsx
+++ b/src/actions/WalletListMenuActions.tsx
@@ -7,6 +7,7 @@ import { ButtonInfo, ButtonsModal } from '../components/modals/ButtonsModal'
 import { RawTextModal } from '../components/modals/RawTextModal'
 import { TextInputModal } from '../components/modals/TextInputModal'
 import { Airship, showError, showToast } from '../components/services/AirshipInstance'
+import { Alert } from '../components/themed/Alert'
 import { ModalMessage } from '../components/themed/ModalParts'
 import { deleteLoanAccount } from '../controllers/loan-manager/redux/actions'
 import { lstrings } from '../locales/strings'
@@ -139,7 +140,7 @@ export function walletListMenuAction(
         dispatch(showSplitWalletModal(walletId, option.replace('split', '')))
       }
     }
-
+    case 'viewPrivateViewKey':
     case 'viewXPub': {
       return (dispatch, getState) => {
         const state = getState()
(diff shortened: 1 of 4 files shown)


src/locales/en_US.ts
```diff
@@ -208,6 +208,8 @@ const strings = {
   fragment_wallets_copy_seed: 'Copy Seed',
   fragment_wallets_copied_seed: 'Copied Seed',
   fragment_wallets_get_seed_wallet: 'Get Seed',
+  fragment_wallets_view_private_view_key: 'Private View Key',
+  fragment_wallets_view_private_view_key_warning: `Do not to share your Private View Key if not necessary (tax purposes, POS systems, accounting...) because the receiver might see the balance of your Monero wallet.`,
```
swansontec, line 212: "Do not to share" is bad English. How about: "The private view key allows the receiver to see the balance in your Monero wallet. Do not share this key unless necessary, such as for tax purposes, accounting, or similar reasons."

I asked ChatGPT for ideas, but it's responses were much more long-winded. I think simply re-organizing the existing instructions gets us where we need to go.

src/actions/WalletListMenuActions.tsx
```diff
@@ -139,7 +140,56 @@ export function walletListMenuAction(
...
+            message={displayPublicSeed ?? ''}
+            title={lstrings.fragment_wallets_view_private_view_key}
+          >
+            <Alert
+              type="warning"
+              title={lstrings.string_warning}
+              marginRem={0.5}
+              message={lstrings.fragment_wallets_view_private_view_key_warning}
+              numberOfLines={0}
+            />
```
swansontec, line 174: Aside from this one addition, and changing `lstrings.fragment_wallets_view_xpub` to `lstrings.fragment_wallets_view_private_view_key`, this whole switch case is identical to the following one. Can we combine them somehow?

```ts
case 'viewPrivateViewKey':
case 'viewXPub': {
  // ... the same stuff ...
  switchString === 'viewPrivateViewKey' ? lstrings.fragment_wallets_view_private_view_key : lstrings.fragment_wallets_view_xpub
  // ... more similar stuff...
}
```

swansontec (changes requested):

src/actions/WalletListMenuActions.tsx
samholmes: I can combine them, but in my mind these are different cases altogether and the code duplication is correct. If there are separate cases, then we should handle the cases separately. But, I'll try the change you suggest.

swansontec (approved):

PR #4143, opened 2023-04-21
In-app review comment support for Android

Jon-edge:
iOS does not support this functionality

### CHANGELOG

- added: In-app review comment support for android.

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204129806235163

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -439,6 +439,8 @@ PODS:
     - React-Core
   - react-native-image-picker (4.0.6):
     - React-Core
+  - react-native-in-app-review (4.3.3):
+    - React-Core
   - react-native-mail (6.1.1):
     - React-Core
   - react-native-mymonero-core (0.2.7):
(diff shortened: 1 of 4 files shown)


android/app/build.gradle
```diff
@@ -187,6 +187,8 @@ dependencies {
 
     implementation "androidx.swiperefreshlayout:swiperefreshlayout:1.0.0"
 
+    implementation project(':react-native-in-app-review')
+
```
swansontec, line 191: This should not be needed thanks to autolinking. Can you delete it and check that the feature still works?

android/settings.gradle
```diff
@@ -1,3 +1,5 @@
 rootProject.name = 'co.edgesecure.app'
+include ':react-native-in-app-review'
+project(':react-native-in-app-review').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-in-app-review/android')
```
swansontec, line 3: This should not be needed thanks to autolinking. Can you delete it and check that the feature still works?

ios/Podfile
```diff
@@ -21,6 +21,8 @@ target 'edge' do
   # you should disable the next line.
   use_flipper!()
 
+  pod 'react-native-in-app-review', :path => '../node_modules/react-native-in-app-review'
+
```
swansontec, line 25: This should not be needed thanks to autolinking. Can you delete it and check that the feature still works?

swansontec (changes requested):
The package was added in a weird way, but the feature seems pretty simple,

swansontec (approved):

PR #4147, opened 2023-04-22
Add Algorand wallet connect v1 support

peachbits:
---
  - https://app.asana.com/0/0/1204335086711417

### DIFF
--- a/src/components/modals/WcSmartContractModal.tsx
+++ b/src/components/modals/WcSmartContractModal.tsx
@@ -134,7 +134,7 @@ const getStyles = cacheStyles((theme: Theme) => ({
   }
 }))
 
-async function wcRequestResponse(wallet: EdgeCurrencyWallet, uri: string, approve: boolean, payload: WcRpcPayload): Promise<void> {
+async function wcRequestResponse(wallet: EdgeCurrencyWallet, uri: string, approve: boolean, payload: Payload): Promise<void> {
   if (!approve) {
     await wallet.otherMethods.wcRejectRequest(uri, payload)
     return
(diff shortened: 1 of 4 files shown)


src/components/scenes/WcConnectScene.tsx
```diff
@@ -70,23 +68,26 @@ export const WcConnectScene = (props: Props) => {
...
+        showError('Failed to connect, try again.')
+        console.error('wcInit error:', e)
+        navigation.navigate('wcConnections', {})
+      }
+    },
+    [currencyWallets, navigation, styles.currencyLogo, uri]
+  )
 
-  const showWalletListModal = () => {
+  const showWalletListModal = React.useCallback(() => {
```
swansontec, line 90: This function should be wrapped in our `useHandler` hook, *not* `useCallback`. That will keep its identity stable. It should also be named `handleWalletListModal`, since it's being passed to an `onPress` prop.

src/components/modals/WcSmartContractModal.tsx
```diff
@@ -202,3 +194,37 @@ async function wcRequestResponse(wallet: EdgeCurrencyWallet, uri: string, approv
     throw e
   }
 }
+
+const asWcRpcPayload = asObject({
+  id: asEither(asString, asNumber),
+  method: asValue('personal_sign', 'eth_sign', 'eth_signTypedData', 'eth_sendTransaction', 'eth_signTransaction', 'eth_sendRawTransaction'),
```
swansontec: TypeScript is probably going to infer `string` for this parameter. If you want it to infer the proper enum type, put `const asWcRpcPayloadMethod = asValue(...)` on its own line, and then use the named `asWcRpcPayloadMethod` here inside the object.

For some reason, putting `asValue` on its own line helps TypeScript do better inference.

src/components/services/AccountCallbackManager.tsx
```diff
@@ -104,9 +104,14 @@ export function AccountCallbackManager(props: Props) {
...
-        const { dApp, payload, uri, walletId } = obj
-        if (walletId == null) return
-        Airship.show(bridge => <WcSmartContractModal bridge={bridge} wallet={wallet} dApp={dApp} payload={payload} uri={uri} />)
+        try {
+          const clean = asWcSmartContractModalProps(obj)
+          const { dApp, payload, uri } = clean
+
+          Airship.show(bridge => <WcSmartContractModal bridge={bridge} wallet={wallet} dApp={dApp} payload={payload} uri={uri} />)
+        } catch (e: any) {
+          console.warn('Invalid wcNewContractCall params', e.message)
```
swansontec, line 124: Don't do `e.message`. You can do `console.warn(..., e)` or even `console.warn('... ${String(e)}')`

src/components/modals/WcSmartContractModal.tsx
```diff
@@ -31,12 +31,11 @@ interface Props extends WcSmartContractModalProps {
 }
 
 export const WcSmartContractModal = (props: Props) => {
-  const { bridge, wallet, dApp, payload, tokenId, uri } = props
+  const { bridge, wallet, dApp, payload, tokenId, nativeAmount, networkFee, uri } = props
```
swansontec, line 34: Optional: Alphabetize these?

src/components/modals/WcSmartContractModal.tsx
```diff
@@ -114,13 +104,8 @@ export const WcSmartContractModal = (props: Props) => {
...
           <EdgeText>{dAppName}</EdgeText>
         </IconTile>
-        {!zeroString(networkFeeCrypto) && (
-          <CryptoFiatAmountTile
-            title={lstrings.wc_smartcontract_network_fee}
-            nativeCryptoAmount={networkFeeCrypto}
-            denomination={feeDenom}
-            walletId={wallet.id}
-          />
+        {!zeroString(networkFee) && (
```
swansontec, line 107: We are avoiding the `&&` for rendering, since we want to enable strict boolean linting. So `zeroString(networkFee) ? null : <CryptoFiatAmountTile ... />`

src/components/services/AccountCallbackManager.tsx
```diff
@@ -106,9 +106,20 @@ export function AccountCallbackManager(props: Props) {
...
+          Airship.show(bridge => (
+            <WcSmartContractModal
+              bridge={bridge}
+              wallet={wallet}
+              dApp={dApp}
+              nativeAmount={nativeAmount}
+              networkFee={networkFee}
+              payload={payload}
+              tokenId={tokenId}
+              uri={uri}
```
swansontec, line 120: Optional: Alphabetize these props.

swansontec (changes requested):

swansontec (approved):

src/components/modals/WcSmartContractModal.tsx
```diff
@@ -58,7 +58,9 @@ export const WcSmartContractModal = (props: Props) => {
 
   const handleSubmit = () => {
     wcRequestResponse(wallet, uri, true, payload)
-      .then(async () => await Airship.show(bridge => <FlashNotification bridge={bridge} message={lstrings.wc_smartcontract_confirmed} />))
+      .then(() => {
+        Airship.show(bridge => <FlashNotification bridge={bridge} message={lstrings.wc_smartcontract_confirmed} />)
```
swansontec, line 62: This will be an ESLint error once we enable stricter promise handling. Better add `.catch(()=>{})` on the `Airship.show`, since we don't care about the Airship result.

swansontec (changes requested):

swansontec (approved):

PR #4151, opened 2023-04-24
Show FIO popup to existing users who don't have FIO handles

samholmes:
### CHANGELOG

- Add: Show FIO handle creation modal to existing accounts without any registered FIO names

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204252350267523

### DIFF
--- a/src/components/modals/FioCreateHandleModal.tsx
+++ b/src/components/modals/FioCreateHandleModal.tsx
@@ -15,7 +15,7 @@ import { ThemedModal } from '../themed/ThemedModal'
 
 interface Props {
   bridge: AirshipBridge<boolean>
-  createWalletsPromise: Promise<void>
+  createWalletsPromise?: Promise<void>
 }
 
 export const FioCreateHandleModal = (props: Props) => {
@@ -23,7 +23,6 @@ export const FioCreateHandleModal = (props: Props) => {
 
   const theme = useTheme()
   const styles = getStyles(theme)
-  const isCreateWalletsInProgress = typeof createWalletsPromise === 'object' && createWalletsPromise instanceof Promise
   const [showPleaseWait, setShowPleaseWait] = React.useState(false)
 
   const handleCancel = useHandler(() => {
@@ -31,7 +30,7 @@ export const FioCreateHandleModal = (props: Props) => {
   })
 
   const handleConfirm = useHandler(async () => {
-    if (isCreateWalletsInProgress) {
+    if (createWalletsPromise != null) {
       setShowPleaseWait(true)
       await createWalletsPromise
     }
--- a/src/components/services/Services.tsx
+++ b/src/components/services/Services.tsx
@@ -1,3 +1,5 @@
+import { asDate, asJSON, asObject, uncleaner } from 'cleaners'
+import { EdgeAccount } from 'edge-core-js'
 import * as React from 'react'
 
 import { updateExchangeInfo } from '../../actions/ExchangeInfoActions'
@@ -5,6 +7,7 @@ import { registerNotificationsV2 } from '../../actions/NotificationActions'
 import { checkCompromisedKeys } from '../../actions/WalletActions'
 import { ENV } from '../../env'
 import { useAsyncEffect } from '../../hooks/useAsyncEffect'
+import { useHandler } from '../../hooks/useHandler'
 import { useRefresher } from '../../hooks/useRefresher'
 import { makeStakePlugins } from '../../plugins/stake-plugins/stakePlugins'
 import { defaultAccount } from '../../reducers/CoreReducer'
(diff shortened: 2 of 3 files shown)


src/components/modals/FioCreateHandleModal.tsx
```diff
@@ -15,15 +15,15 @@ import { ThemedModal } from '../themed/ThemedModal'
...
+  createWalletsPromise?: Promise<void>
 }
 
 export const FioCreateHandleModal = (props: Props) => {
   const { bridge, createWalletsPromise } = props
 
   const theme = useTheme()
   const styles = getStyles(theme)
-  const isCreateWalletsInProgress = typeof createWalletsPromise === 'object' && createWalletsPromise instanceof Promise
+  const isCreateWalletsInProgress = typeof createWalletsPromise === 'object' && createWalletsPromise != null
```
swansontec, line 26: It seems like `createWalletsPromise != null` should be adequate, since the type system won't let us pass anything else.

I'm not sure how `isCreateWalletsInProgress` could have ever been false before, though, so maybe I'm wrong about this. Maybe just leave it if it's too tricky.

src/components/services/Services.tsx
```diff
@@ -55,6 +60,37 @@ export function Services(props: Props) {
...
+              })
+            )
+          }
+          return promises
+        }, [])
+      )
+    ).filter(wallet => wallet != null)
+
+    if (fioWalletsWithAddresses.length === 0 && freeRegApiToken !== '' && freeRegRefCode !== '') {
+      const itemIds = await account.dataStore.listItemIds('')
```
swansontec, line 84: Do not do this. It's extremely expensive. Just do `account.dataStore.getItem('', FIO_CREATE_HANDLE_IGNORED_AT).catch(() => undefined)`, and then check if you get a string (it's there) or undefined (it's not there). You could also use a cleaner for the file contents.

src/components/services/Services.tsx
```diff
@@ -55,6 +60,37 @@ export function Services(props: Props) {
     dispatch(checkCompromisedKeys(navigation)).catch(e => {
       console.warn('checkCompromisedKeys error:', e)
     })
+
+    // FIO handle registration:
+    const { freeRegApiToken = '', freeRegRefCode = '' } = typeof ENV.FIO_INIT === 'object' ? ENV.FIO_INIT : {}
+    const fioWalletsWithAddresses = (
+      await Promise.all(
+        Object.values(account.currencyWallets).reduce<Array<Promise<EdgeCurrencyWallet>>>((promises, wallet) => {
```
swansontec, line 68: Never iterate the `account.currencyWallets` object. Always use `account.currencyWalletIds` to drive iteration on this object, and then do `account.currencyWallets[walletId]` to read the actual wallet, which might be `undefined` if it hasn't loaded yet. This has to do with the way the core bridges this particular object.

src/components/services/Services.tsx
```diff
@@ -55,6 +60,37 @@ export function Services(props: Props) {
...
+          }
+          return promises
+        }, [])
+      )
+    ).filter(wallet => wallet != null)
+
+    if (fioWalletsWithAddresses.length === 0 && freeRegApiToken !== '' && freeRegRefCode !== '') {
+      const itemIds = await account.dataStore.listItemIds('')
+      if (!itemIds.includes(FIO_CREATE_HANDLE_IGNORED_AT)) {
+        const isCreateHandle = await Airship.show<boolean>(bridge => <FioCreateHandleModal bridge={bridge} />)
```
swansontec, line 86: Maybe call this "shouldCreate". We agreed not to start booleans with "is", but instead to give them sensible names.

src/components/services/Services.tsx
```diff
@@ -55,6 +60,37 @@ export function Services(props: Props) {
...
+        if (isCreateHandle) {
+          navigation.navigate('fioCreateHandle', { freeRegApiToken, freeRegRefCode })
+        } else {
+          account.dataStore.setItem('', FIO_CREATE_HANDLE_IGNORED_AT, new Date().toISOString())
```
swansontec, line 90: For the contents of this file, maybe define `asFioCreateHandleIgnored = asJson(asObject({ ignored: asDate }))`, and then do `uncleaner(asFioCreateHandleIgnored)({ ignored: new Date() })` to JSON-ize the file.

swansontec (changes requested):

src/components/services/Services.tsx
samholmes: If we use a cleaner, I might as well use `src/util/CleanStore.ts`.

swansontec (approved):
Yep, this makes sense now.

PR #4157, opened 2023-04-26
Remove problematic isDone logic from WalletSyncCircle

samholmes:
This logic appears only to be present for two reasons:

1. To prevent the the sync circle from showing briefly while fading out
on a wallet with a fully synced state.
2. To hard-reset (without timing) the syncRatio when re-syncing.

In order to replace this extra state in order to simplify and fix
inconsistency issues:

1. Instead, we default to 0 opacity for the circle and immediately
invoke the `syncRatio` handler in the subscriber useEffect.
2. We derive the done state using `syncRatio.value` for the logic where
its needed.

### CHANGELOG

- Fixed: WalletSyncCircle inconsistency issues with wallets that are stuck syncing.

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204316259619599

### DIFF
--- a/src/components/progress-indicators/WalletSyncCircle.tsx
+++ b/src/components/progress-indicators/WalletSyncCircle.tsx
@@ -26,29 +26,29 @@ export const WalletSyncCircle = (props: Props) => {
   const { size = theme.rem(2), wallet } = props
   // Animation shared state
   const syncRatio = useSharedValue(wallet.syncRatio < RESYNC_THRESHOLD ? MIN_RATIO : wallet.syncRatio)
-  const isDone = useSharedValue(false)
 
   // Subscribe to the sync ratio:
-  React.useEffect(
-    () =>
-      wallet.watch('syncRatio', (ratio: number) => {
-        // If already done but needs to resync reset the flags and animations
-        if (isDone.value && ratio < RESYNC_THRESHOLD) {
-          isDone.value = false
+  React.useEffect(() => {
+    const handler = (ratio: number) => {
+      if (ratio < RESYNC_THRESHOLD) {
+        // Do not animate backwards if a resync happens after the sync is done:
+        if (syncRatio.value > DONE_THRESHOLD) {
           syncRatio.value = RESYNC_THRESHOLD
+        } else {
+          syncRatio.value = withTiming(RESYNC_THRESHOLD, { duration: 1000 })
         }
-        // If the wallet hasn't fully synced show progress animation
-        if (!isDone.value) {
-          // Sync is done
-          if (ratio > DONE_THRESHOLD) isDone.value = true
+      } else if (ratio > MAX_RATIO && ratio < DONE_THRESHOLD) {
+        syncRatio.value = withTiming(MAX_RATIO, { duration: 1000 })
+      } else {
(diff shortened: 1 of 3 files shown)


src/components/progress-indicators/WalletSyncCircle.tsx
```diff
@@ -60,11 +57,13 @@ export const WalletSyncCircle = (props: Props) => {
   const viewBox = `${-strokeWidth * 2} ${-strokeWidth * 2} ${svgSize} ${svgSize}`
 
   // Animated Params
-  const animatedProps = useAnimatedProps(() => ({
-    strokeDashoffset: circumference * (1 - syncRatio.value),
-    opacity: withTiming(syncRatio.value === 1 ? 0 : 1, { duration: 500 }),
-    stroke: withTiming(isDone.value ? theme.walletProgressIconDone : theme.walletProgressIconFill)
-  }))
+  const animatedProps = useAnimatedProps(() => {
+    return {
```
swansontec, line 61: Why add the `return` statement? The `=> ({})` syntax is nice because it says "there is no logic in here," but I can also see the `return` syntax being nice for letting you easily add debug stuff to the lines above.

src/components/progress-indicators/WalletSyncCircle.tsx
```diff
@@ -26,29 +26,26 @@ export const WalletSyncCircle = (props: Props) => {
...
-          else if (ratio > MAX_RATIO && ratio < DONE_THRESHOLD) syncRatio.value = withTiming(MAX_RATIO, { duration: 1000 })
-          else syncRatio.value = withTiming(ratio, { duration: 1000 })
-        }
-      }),
-    [isDone, syncRatio, wallet]
-  )
+      if (ratio < RESYNC_THRESHOLD) syncRatio.value = withTiming(RESYNC_THRESHOLD, { duration: 1000 })
+      else if (ratio > MAX_RATIO && ratio < DONE_THRESHOLD) syncRatio.value = withTiming(MAX_RATIO, { duration: 1000 })
+      else syncRatio.value = withTiming(ratio, { duration: 1000 })
+    }
```
swansontec, line 42: You have the `ratio < RESYNC_RATIO` logic twice, which makes this tricky to read. I would refactor this entire block as follows:

```ts
if (ratio < RESYNC_THRESHOLD) {
  // Do not animate backwards if a resync happens after the sync is done:
  if (syncRatio.value > DONE_THRESHOLD) {
    syncRatio.value = RESYNC_THRESHOLD
  } else {
    syncRatio.value = withTiming(RESYNC_THRESHOLD, { duration: 1000 })
  }
} else if (ratio > MAX_RATIO && ratio < DONE_THRESHOLD) {
  syncRatio.value = withTiming(MAX_RATIO, { duration: 1000 })
} else {
  syncRatio.value = withTiming(ratio, { duration: 1000 })
}
```

This consolidates the two cases involving RESYNC_THRESHOLD, and uses `else` for control flow instead of the early return. If you want to use returns for control flow, that would work too, but it should be consistent for all the cases.

swansontec (approved):
I have a question and a suggestion here, but I am approving the PR since they don't affect the logic in any way.

PR #4158, opened 2023-04-26
Remove WalletListModal allowed/exclude currencyCode props

Jon-edge:
### CHANGELOG

- changed: Remove deprecated props from WalletListModal

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1202827562867788


src/components/modals/WalletListModal.tsx
```diff
@@ -123,22 +112,11 @@ export function WalletListModal(props: Props) {
 
   // #region Init
 
-  // Upgrade deprecated props
-  const [legacyAllowedAssets, legacyExcludeAssets] = React.useMemo(() => {
-    if (allowedCurrencyCodes == null && excludeCurrencyCodes == null) return []
-
-    const lookup = makeCurrencyCodeTable(account.currencyConfig)
-    const allowedAssets = upgradeCurrencyCodes(lookup, allowedCurrencyCodes)
```
swansontec, line 131: Since `WalletListModal` no longer uses `upgradeCurrencyCodes`, we should move this helper into its own utility file. Please create "src/utils/tokenIdTools.ts", an put `upgradeCurrencyCodes` in there, along with the `makeCurrencyCodeTable` function that goes along with it.

That will complete the cleanup for this feature.

swansontec (changes requested):

swansontec (approved):

PR #4161, opened 2023-05-02
Add Tron Stake v2

peachbits:
---
  - https://app.asana.com/0/0/1204436465323062

### DIFF
--- a/src/plugins/stake-plugins/currency/tronStakePlugin.ts
+++ b/src/plugins/stake-plugins/currency/tronStakePlugin.ts
@@ -34,6 +34,7 @@ const policyDefault = {
 const policies: StakePolicy[] = [
   {
     ...policyDefault,
+    deprecated: true,
     stakePolicyId: 'currency:tron:BANDWIDTH',
     rewardAssets: [
       {
@@ -52,6 +53,7 @@ const policies: StakePolicy[] = [
   },
   {
     ...policyDefault,
+    deprecated: true,
     stakePolicyId: 'currency:tron:ENERGY',
     rewardAssets: [
       {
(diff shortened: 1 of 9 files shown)


src/plugins/stake-plugins/types.ts
```diff
@@ -160,8 +160,28 @@ export interface StakePosition {
 // Stake Plugin
 // -----------------------------------------------------------------------------
 
+export interface StakePolicyFilter {
+  wallet?: EdgeCurrencyWallet
+  currencyCode?: string
```
swansontec, line 168: I don't like seeing `currencyCode` used for any money stuff. It needs to become a purely a display thing going forward, not something we can compare based on.

However, I see that currencyCode exists in a bunch of places here, so maybe it's too hard to fix. If it really is a hard fix, I guess we can leave it.

src/plugins/stake-plugins/currency/tronStakePlugin.ts
```diff
@@ -106,15 +146,15 @@ export const makeTronStakePlugin = async (): Promise<StakePlugin> => {
       const policy = getPolicyFromId(stakePolicyId)
       const resource = policy.rewardAssets[0].currencyCode
       const isStake = action === 'stake'
-
+      console.log('156. policy.deprecated', policy.deprecated)
```
swansontec, line 156: Leftover?

swansontec (approved):
Go ahead and delete the extra comment. If you decide to attack the currencyCode thing, then OK, I will do another review.

src/plugins/stake-plugins/types.ts
peachbits: That's a good point. I looked at the potential work involved and currencyCode is pretty pervasive throughout the staking plugin so I'd rather punt

PR #4174, opened 2023-05-08
Fix login lag

paullinator:
Remove await from registerNotificationsV2 as we should NEVER wait on network ops to complete logins.

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204547794852799


swansontec (changes requested):
Do console.log(error). It is never safe to access error properties, like message, without some kind of cleaning or validation first.

swansontec (approved):

PR #4178, opened 2023-05-09
Add functionality to navigate after swiping past the USPs

Jon-edge:
### CHANGELOG

Add navigation to login or signup flow after swiping past the final USP

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204190904815166

### DIFF
--- a/src/components/scenes/GettingStartedScene.tsx
+++ b/src/components/scenes/GettingStartedScene.tsx
@@ -2,7 +2,17 @@ import * as React from 'react'
 import { useEffect } from 'react'
 import { Image, Pressable, Text, View } from 'react-native'
 import { ScrollView, TouchableOpacity } from 'react-native-gesture-handler'
-import Animated, { Extrapolation, interpolate, interpolateColor, SharedValue, useAnimatedStyle, useSharedValue, withTiming } from 'react-native-reanimated'
+import Animated, {
+  Extrapolation,
+  interpolate,
+  interpolateColor,
+  runOnJS,
+  SharedValue,
+  useAnimatedReaction,
+  useAnimatedStyle,
+  useSharedValue,
+  withTiming
+} from 'react-native-reanimated'
 import { useSafeAreaFrame } from 'react-native-safe-area-context'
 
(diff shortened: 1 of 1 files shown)


src/components/scenes/GettingStartedScene.tsx
```diff
@@ -97,7 +126,7 @@ export const GettingStartedScene = (props: Props) => {
           </TouchableOpacity>
         </Space>
       </SkipButton>
-      <SwipeOffsetDetector swipeOffset={swipeOffset} minOffset={0} maxOffset={sections.length}>
+      <SwipeOffsetDetector swipeOffset={swipeOffset} minOffset={0} maxOffset={sections.length + 1}>
```
samholmes, line 129: I realize the `+1` here lacks documentation. Maybe we should factor out the expression to make the code self-documenting. I always default to self-documenting code over adding comments where I can. So, `maxOffset={totalSectionCount}` and `const totalSectionCount = sections.length + 1 // Add one for the final "skip section"` to allow for a clean comment.

src/components/scenes/GettingStartedScene.tsx
```diff
@@ -88,6 +101,22 @@ export const GettingStartedScene = (props: Props) => {
     navigation.navigate('login', {})
   })
 
+  useAnimatedReaction(
+    () => swipeOffset.value,
+    value => {
+      if (value === 5) {
```
samholmes, line 107: Now that I think about it. We could get a nicer animation effect if we hook into the "release" event of the gesture in SwipeOffsetDetector. The event handler would have the new offset as a parameter which could be checked instead of checking the `swipeOffset.value`. This would be better.

samholmes (changes requested):
This is really good. I had some final comments to make it over 9000

src/components/scenes/GettingStartedScene.tsx
samholmes: Forget this one. Scope creep

src/components/scenes/GettingStartedScene.tsx
samholmes: Or, comment however you like

PR #4183, opened 2023-05-11
Fix duplicated Share modal popup glitch on Android

itayplav:
### CHANGELOG

- Fix duplicated share modals popping up when rapidly tapping Share button (Android only).

### Dependencies

none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204353783542603

### DIFF
--- a/src/components/scenes/RequestScene.tsx
+++ b/src/components/scenes/RequestScene.tsx
@@ -3,7 +3,7 @@ import { lt } from 'biggystring'
 import { EdgeAccount, EdgeCurrencyWallet, EdgeEncodeUri } from 'edge-core-js'
 import * as React from 'react'
 import { ActivityIndicator, Linking, Platform, Text, TouchableOpacity, View } from 'react-native'
-import Share from 'react-native-share'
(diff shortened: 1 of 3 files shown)


src/components/scenes/RequestScene.tsx
```diff
@@ -439,20 +439,17 @@ export class RequestSceneComponent extends React.Component<Props, State> {
...
+    const shareOptions: ShareOptions = {
       subject,
       message: Platform.OS === 'ios' ? message : message + edgePayUri,
-      url: Platform.OS === 'ios' ? edgePayUri : ''
+      url: Platform.OS === 'ios' ? edgePayUri : '',
+      failOnCancel: false
     }
 
-    Share.open(shareOptions).catch(e => console.log(e))
+    await Share.open(shareOptions).catch((e: unknown) => showError(e))
```
swansontec, line 449: You can do `.catch(showError)` directly, since the function is designed to be used that way.

swansontec (approved):
Approved with an optional cleanup. Good work with the `usePendingPress` hook - this is definitely the right solution.

PR #4187, opened 2023-05-15
Fix categories with no ':'

paullinator:
Assume the entire category is a primary category and append a ':' to the end. If it doesn't match the GUI will default to income or expense accordingly

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204352698460546

### DIFF
--- a/src/util/categories.ts
+++ b/src/util/categories.ts
@@ -13,6 +13,9 @@ export interface SplitCategory {
  * The subcategory can be localized and freely edited.
  */
 export function splitCategory(fullCategory: string = '', defaultCategory: Category = 'income'): SplitCategory {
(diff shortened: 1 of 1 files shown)


src/util/categories.ts
```diff
@@ -13,6 +13,9 @@ export interface SplitCategory {
  * The subcategory can be localized and freely edited.
  */
 export function splitCategory(fullCategory: string = '', defaultCategory: Category = 'income'): SplitCategory {
+  if (fullCategory.length > 0 && !fullCategory.includes(':')) {
```
swansontec, line 16: This is probably the best solution, unless we want to refactor the whole function to use `string.split(':')`.

PR #4190, opened 2023-05-17
Fix personal_sign wallet connect method

peachbits:
---
  - https://app.asana.com/0/0/1204615731108224

### DIFF
--- a/src/components/modals/WcSmartContractModal.tsx
+++ b/src/components/modals/WcSmartContractModal.tsx
@@ -144,18 +144,23 @@ async function wcRequestResponse(wallet: EdgeCurrencyWallet, uri: string, approv
 
   try {
     switch (payload.method) {
-      case 'personal_sign':
+      case 'personal_sign': {
+        const cleanPayload = asEvmSignPayload(payload)
+        const result = await wallet.signMessage(cleanPayload.params[0])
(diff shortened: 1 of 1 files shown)


src/components/modals/WcSmartContractModal.tsx
```diff
@@ -185,32 +190,31 @@ async function wcRequestResponse(wallet: EdgeCurrencyWallet, uri: string, approv
...
-        gas: asOptional(asString),
-        gasPrice: asOptional(asString),
-        value: asOptional(asString),
-        nonce: asOptional(asString)
-      }),
-      asOptional(asString, '')
-    ),
-    asArray(asString)
+  method: asEvmSignMethod,
+  params: asEither(asTuple(asString, asString))
```
swansontec, line 200: This `asEither` isn't needed anymore, since there is only one option.

Also, it looks like personal_sign only uses params[0]. I'm not sure why the cleaner needs two parameters if we only care about the first one, but maybe there is a reason. Otherwise, `asTuple(asString, asOptional(asString))` or `asArray(asString)` might be needed, followed by some extra checking inside the methods themselves.

swansontec (changes requested):

swansontec (approved):

PR #4197, opened 2023-05-20
Visa fixes

paullinator:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204640751122989


src/plugins/gui/RewardsCardPlugin.tsx
```diff
@@ -285,13 +285,25 @@ export const makeRewardsCardPlugin: FiatPluginFactory = async params => {
...
+          return await getRewardCards().catch(e => {
+            throw new Error(lstrings.rewards_card_error_retrieving_cards)
+          })
+        }
+        return rewardCards
       }
 
+      rewardCards = await showUi.showToastSpinner(
+        lstrings.loading,
+        runWithTimeout(startPluginInner(), 8000, new Error(lstrings.rewards_card_error_timeout_loading))
```
samholmes, line 304: Because we know that the API is low, can we increase from 8 seconds to 30, or 15 seconds? Ideally, the API gets faster, but I'd rather give it a bit more room to succeed. What do you think?

src/components/services/ContactsLoader.ts
```diff
@@ -255,6 +255,18 @@ const merchantPartners = [
     familyName: '',
     recordID: ''
   },
+  {
+    givenName: 'Visa Prepaid Card',
```
samholmes, line 259: Missing ®

src/components/services/ContactsLoader.ts
```diff
@@ -255,6 +255,18 @@ const merchantPartners = [
...
+    hasThumbnail: true,
+    thumbnailPath: 'visa.png',
```
samholmes, line 261: Where is this thumbnail image?

samholmes (approved):

PR #4201, opened 2023-05-22
Make scamWarning explicitly shown for Payment Protocol deep-linking

samholmes:
### CHANGELOG

- Fixed: Bug with scam warning modal shown in unwanted instances; only show once for first manual navigation to send scene.

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204622252075405

### DIFF
--- a/src/actions/ScanActions.tsx
+++ b/src/actions/ScanActions.tsx
@@ -221,7 +221,7 @@ export function handleWalletUris(
       // React navigation doesn't like passing non-serializable objects as params. Convert date to string first
       // https://github.com/react-navigation/react-navigation/issues/7925
       const isoExpireDate = parsedUri?.expireDate?.toISOString()
-      navigation.push('send2', { walletId: wallet.id, minNativeAmount, spendInfo, isoExpireDate })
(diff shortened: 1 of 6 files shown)


src/actions/ScanActions.tsx
```diff
@@ -221,7 +221,7 @@ export function handleWalletUris(
       // React navigation doesn't like passing non-serializable objects as params. Convert date to string first
       // https://github.com/react-navigation/react-navigation/issues/7925
       const isoExpireDate = parsedUri?.expireDate?.toISOString()
-      navigation.push('send2', { walletId: wallet.id, minNativeAmount, spendInfo, isoExpireDate })
```
paullinator, line 224: How was this needed? The SendScene checks for `scamWarning !== true` so `undefined` should be as good as `false`

paullinator (changes requested):

src/actions/ScanActions.tsx
samholmes: Because of the last commit

paullinator (approved):

PR #4204, opened 2023-05-23
Supress push server errors on login

paullinator:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204637819439235

### DIFF
--- a/src/components/scenes/LoginScene.tsx
+++ b/src/components/scenes/LoginScene.tsx
@@ -162,11 +162,15 @@ class LoginSceneComponent extends React.PureComponent<Props, State> {
 
     const { notificationPermissionsInfo } = this.state
     if (notificationPermissionsInfo) {
-      const newSettings = await this.props.dispatch(setDeviceSettings(notificationPermissionsInfo.notificationOptIns))
-      this.props.dispatch({
-        type: 'NOTIFICATION_SETTINGS_UPDATE',
-        data: serverSettingsToNotificationSettings(newSettings)
-      })
+      try {
+        const newSettings = await this.props.dispatch(setDeviceSettings(notificationPermissionsInfo.notificationOptIns))
+        this.props.dispatch({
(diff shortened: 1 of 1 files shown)


peachbits (changes requested):
This doesn't fix anything it just suppresses an error message. Do we know why the response object isn't getting past the cleaner? I couldn't repo this so I haven't seen the bad return value.

src/components/scenes/LoginScene.tsx
```diff
@@ -162,11 +162,15 @@ class LoginSceneComponent extends React.PureComponent<Props, State> {
...
-        data: serverSettingsToNotificationSettings(newSettings)
-      })
+      try {
+        const newSettings = await this.props.dispatch(setDeviceSettings(notificationPermissionsInfo.notificationOptIns))
+        this.props.dispatch({
+          type: 'NOTIFICATION_SETTINGS_UPDATE',
+          data: serverSettingsToNotificationSettings(newSettings)
+        })
+      } catch (e) {
+        console.error(String(e))
```
samholmes, line 172: No need to cast here because console.error should take any. Unlesss console.error was mutated as a global and the API is unexpected in the GUI?

samholmes (approved):
Approve with optional change. If you change it, you should get the stack trace without casting to string.

src/components/scenes/LoginScene.tsx
paullinator: The String cast was recommended by @swansontec in our huddle last week. I'll leave as is for now and discuss further. The main goal is to prevent the red error dropdown. @peachbits I didn't repo this either, but based on the text in the red error, the cleaner `asDevicePayload` called in `setDeviceSettings` is the only cleaner that could generate this error and it's always handled except in the call to `setDeviceSettings` from the code changed above.

PR #4208, opened 2023-05-23
More visa fixes

paullinator:
Disable Dash due to network stall
Remove safe area offset on dashboard

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204654031017140

### DIFF
--- a/src/plugins/gui/scenes/RewardsCardDashboardScene.tsx
+++ b/src/plugins/gui/scenes/RewardsCardDashboardScene.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react'
 import { ActivityIndicator, Image, Text, TouchableOpacity, View } from 'react-native'
-import { useSafeAreaInsets } from 'react-native-safe-area-context'
 import Ionicon from 'react-native-vector-icons/Ionicons'
 
 import visaBrandImage from '../../../assets/images/guiPlugins/visaBrand.png'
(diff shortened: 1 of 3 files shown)


src/plugins/gui/scenes/RewardsCardDashboardScene.tsx
```diff
@@ -116,9 +114,9 @@ const CardListItemContainer = styled(View)(props => ({
   marginRight: props.theme.rem(1)
 }))
 
-const BottomFloat = styled(View)<{ bottomSpace: number }>(props => ({
+const BottomFloat = styled(View)(props => ({
   alignSelf: 'center',
-  bottom: props.bottomSpace,
+  bottom: 0,
```
peachbits, line 119: Optional: Can be removed completely now that it's zero?

samholmes (approved):

src/plugins/gui/scenes/RewardsCardDashboardScene.tsx
samholmes: Nah, cause that’s the way to anchor it to the bottom using absolute positioning

paullinator: @peachbits Exactly what @samholmes said. Otherwise the button ends up at the very top

PR #4211, opened 2023-05-24
Bity: Fix Incorrect Buy Amount Error Messages

Jon-edge:
**_This fix only addresses the buy side._** A separate task has been made to capture updating our error handling to be side-specific: https://app.asana.com/0/1200382638405084/1204654536856719/f

Bity provider was not giving a typed error for amount limits exceeded. 

Since provider error handling assumes that all providers give typed errors, it was incorrectly showing the errors from the other providers due to their higher error priority.

### CHANGELOG

- fix: Incorrect Bity Buy Amount Error Messages

### Dependencies

https://app.asana.com/0/1200382638405084/1204654536856719/f

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204449269628095


src/app.ts
```diff
@@ -2,15 +2,81 @@
...
+import { LogBox, Text, TextInput } from 'react-native'
 import RNFS from 'react-native-fs'
 
+import ENV from '../env.json'
 import { changeTheme, getTheme } from './components/services/ThemeContext'
-import { ENV } from './env'
 import { NumberMap } from './types/types'
 import { log, logToServer } from './util/logger'
 
+console.log = () => {}
```
paullinator, line 13: This file has nothing to do with the commit. Revert out or make another commit

src/plugins/gui/amountQuotePlugin.ts
```diff
@@ -200,6 +200,7 @@ export const amountQuoteFiatPlugin: FiatPluginFactory = async (params: FiatPlugi
           let errors: unknown[] = []
           const quotes = await fuzzyTimeout(quotePromises, 5000).catch(e => {
             errors = e
+            console.error(JSON.stringify({ errors }, null, 2))
```
paullinator, line 203: Why are you stringifying? There's no guarantee the error is JSON

src/locales/en_US.ts
```diff
@@ -1241,14 +1241,14 @@ const strings = {
   stake_resource_energy: 'Energy',
 
   fiat_plugin_select_asset_to_purchase: 'Select Asset to Purchase',
-  fiat_plugin_buy_currencycode: 'Buy %s', // TODO: string key convention
-  fiat_plugin_sell_currencycode_s: 'Sell %s',
-  fiat_plugin_amount_currencycode: 'Amount %s',
-  fiat_plugin_buy_amount_over_limit: 'Max purchase amount is %s',
-  fiat_plugin_buy_amount_under_limit: 'Minimum purchase amount is %s',
+  fiat_plugin_buy_currencycode_s: 'Buy %1$s',
```
paullinator, line 1244: Revert this. No reason to cause a retranslation as none of the strings changed.

src/plugins/gui/pluginUtils.ts
```diff
@@ -15,22 +15,21 @@ export const createStore = (storeId: string, store: EdgeDataStore): FiatProvider
...
 
-// @ts-expect-error
-const ERROR_PRIORITIES: { [errorType: FiatProviderQuoteErrorTypes]: number } = {
-  underLimit: 1,
-  overLimit: 2,
-  paymentUnsupported: 3,
-  regionRestricted: 4,
-  assetUnsupported: 5
-}
+const ERROR_PRIORITIES = new Map<FiatProviderQuoteErrorTypes, number>([
```
paullinator, line 18: I don't understand how this commit is a cleanup. Logic just became more complicated. Fixing the `ts-expect-errors` could have been one-liners each.

src/__tests__/amountQuotePlugin.test.ts
```diff
@@ -28,17 +28,17 @@ describe('creditCardPlugin', function () {
     test('overLimit', function () {
       const errors: FiatProviderError[] = [new FiatProviderError({ providerId: '', errorType: 'overLimit', errorAmount: 50 })]
```
paullinator, line 29: This should have been committed in the previous commit. But I still suggest reverting the string renames anyway

src/plugins/gui/fiatProviderTypes.ts
```diff
@@ -25,13 +25,24 @@ export interface FiatProviderQuote {
   closeQuote: () => Promise<void>
 }
 
-export type FiatProviderQuoteErrorTypes = 'overLimit' | 'underLimit' | 'regionRestricted' | 'assetUnsupported' | 'paymentUnsupported'
+export type FiatProviderQuoteErrorTypes =
+  | 'overLimit'
+  | 'overUnknownLimit'
```
paullinator, line 30: This is overkill to add new error types. Just make the `errorAmount` optional and have the plugin show a different error if the amount isn't provided.

paullinator (changes requested):

src/app.ts
Jon-edge: My mistake, this got in while rebasing

src/plugins/gui/fiatProviderTypes.ts
Jon-edge: The intent was to avoid special casing

src/plugins/gui/pluginUtils.ts
```diff
@@ -48,32 +39,53 @@ export const getBestError = (errorQuotes: FiatProviderError[], currencyCode: str
...
-    // @ts-expect-error
     if (ERROR_PRIORITIES[errorQuote.errorType] < ERROR_PRIORITIES[bestError.errorType]) {
       bestError = errorQuote
       continue
     }
-    // @ts-expect-error
     if (ERROR_PRIORITIES[errorQuote.errorType] === ERROR_PRIORITIES[bestError.errorType]) {
       if (errorQuote.errorType === 'overLimit' && bestError.errorType === 'overLimit') {
-        if (errorQuote.errorAmount > bestError.errorAmount) {
+        if (bestError.errorAmount != null && errorQuote.errorAmount != null && errorQuote.errorAmount > bestError.errorAmount) {
```
paullinator, line 48: Should also handle the case where `bestError.errorAmount == null && errorQuote.errorAmount != null'

Change the `if` to

```
if ((errorQuote.errorAmount ?? 0) > (bestError.errorAmount ?? 0)) {
```

src/plugins/gui/pluginUtils.ts
```diff
@@ -48,32 +39,53 @@ export const getBestError = (errorQuotes: FiatProviderError[], currencyCode: str
...
           bestError = errorQuote
         }
       } else if (errorQuote.errorType === 'underLimit' && bestError.errorType === 'underLimit') {
-        if (errorQuote.errorAmount < bestError.errorAmount) {
+        if (bestError.errorAmount != null && errorQuote.errorAmount != null && errorQuote.errorAmount < bestError.errorAmount) {
```
paullinator, line 52: Change to

`if ((errorQuote.errorAmount ?? Infinity) < (bestError.errorAmount ?? Infinity)) {`

src/plugins/gui/providers/bityProvider.ts
```diff
@@ -191,7 +193,10 @@ const fetchBityQuote = async (bodyData: BityQuoteRequest) => {
     const newData = await result.json()
     return newData
   } else {
-    throw new Error('Bity: Unable to fetch quote: ' + (await result.text()))
+    const bityErrorRes = asBityErrorResponse(await result.json())
```
paullinator, line 196: The `result.json()` might throw if the result is not json. Wrap this in try/catch and add back the 

`throw new Error('Bity: Unable to fetch quote: ' + (await result.text()))`

src/plugins/gui/pluginUtils.ts
```diff
@@ -48,32 +39,53 @@ export const getBestError = (errorQuotes: FiatProviderError[], currencyCode: str
...
-    errorText = sprintf(errorText, localeAmount + ' ' + currencyCode)
-  } else if (bestError.errorType === 'regionRestricted') {
-    errorText = sprintf(errorText, bestError.displayCurrencyCode)
+  return getErrorText(bestError, currencyCode)
+}
+
+const getErrorText = (error: FiatProviderQuoteError, currencyCode: string): string => {
+  let errorText = ''
+  const errorAmount =
+    (error.errorType === 'overLimit' || error.errorType === 'underLimit') && error.errorAmount != null
```
paullinator, line 65: Do not need to check the type. Only `overLimit` and `underLimit` can have `errorAmount`

paullinator (changes requested):

src/plugins/gui/pluginUtils.ts
Jon-edge: The way we have the types set up we need to check the type. Lint prevents usage of error.hasOwnProperty('errorAmount').

Jon-edge: While that's true, the way the types are currently set up, we can't check errorAmount here without first directly
checking errorType against both string literals. I may have addressed this issue
with my original error type changes that were just reverted but will note this
in https://app.asana.com/0/1200382638405084/1204654536856719/f

Duplicated some code in the switch cases instead for now.

src/plugins/gui/providers/bityProvider.ts
Jon-edge: Are we guaranteed result.json() if result.ok? I was relying on
[fuzzyTimeout](https://github.com/EdgeApp/edge-react-gui/blob/a73f679ba2ba7603396641edc1a9052f92cf736b/src/plugins/gui/amountQuotePlugin.ts#L201)
to deal with such errors.

Jon-edge: Are we guaranteed `result.json()` if `result.ok`? As a fallback, I was relying on [`fuzzyTimeout`](https://github.com/EdgeApp/edge-react-gui/blob/a73f679ba2ba7603396641edc1a9052f92cf736b/src/plugins/gui/amountQuotePlugin.ts#L201), which would catch any of the throws from here.

Jon-edge: Oh nevermind, I see I removed that bity-worded error message

Jon-edge: Also added this as a note to handle better in a future task

src/plugins/gui/providers/bityProvider.ts
```diff
@@ -189,14 +189,20 @@ const fetchBityQuote = async (bodyData: BityQuoteRequest) => {
...
-    if (bityErrorRes.errors.some((bityError: { code: string; message: string }) => bityError.code === 'amount_too_large')) {
-      throw new FiatProviderError({ providerId, errorType: 'overLimit' })
+  try {
+    if (result.ok) {
+      const newData = await result.json()
+      return newData
+    } else {
+      const bityErrorRes = asBityErrorResponse(await result.json())
+      if (bityErrorRes.errors.some((bityError: { code: string; message: string }) => bityError.code === 'amount_too_large')) {
+        throw new FiatProviderError({ providerId, errorType: 'overLimit' })
```
paullinator, line 199: This throw will get caught by the outer catch and re-packaged as a generic error. Only wrap line 197 with a try/catch that would run

`throw new Error('Bity: Unable to fetch quote: ' + (await result.text()))`

paullinator (changes requested):
Regression in the fixup commit. Pls fix

paullinator (approved):

PR #4226, opened 2023-05-27
Implement persistent floating card

Jon-edge:
Persistent floating card, that shows app-wide while logged in with an empty username AND ControlPanel isn't open

Simulator Screen Shot - iPhone 14 Pro - 2023-05-26 at 18 13 32

### CHANGELOG

Add persistent card notification for light accounts

### Dependencies

https://app.asana.com/0/1200382638405084/1204451072447705/f

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204680458642169

### DIFF
--- a/src/locales/en_US.ts
+++ b/src/locales/en_US.ts
@@ -1314,6 +1314,13 @@ const strings = {
 
   // #endregion GuiPlugins
 
+  // #region Light Account
+
+  light_account_card_backup_title: 'Back up your account',
+  light_account_card_backup_message: 'Without a backup, you risk losing your funds!',
+
+  // #endregion Light Account
+
   // Currency Labels
   currency_label_AFN: 'Afghani',
   currency_label_ALL: 'Lek',
--- a/src/components/Main.tsx
+++ b/src/components/Main.tsx
@@ -1,5 +1,5 @@
 import { createBottomTabNavigator } from '@react-navigation/bottom-tabs'
-import { createDrawerNavigator } from '@react-navigation/drawer'
+import { createDrawerNavigator, DrawerContentComponentProps, useDrawerStatus } from '@react-navigation/drawer'
 import { HeaderTitleProps } from '@react-navigation/elements'
 import { DefaultTheme, NavigationContainer } from '@react-navigation/native'
 import { createStackNavigator, StackNavigationOptions } from '@react-navigation/stack'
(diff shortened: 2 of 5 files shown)


src/locales/en_US.ts
```diff
@@ -1314,6 +1314,13 @@ const strings = {
 
   // #endregion GuiPlugins
 
+  // #region Light Account
+
+  light_account_card_backup_title: 'Back up your account',
```
paullinator, line 1319: Rename key to `backup_account_title` to make more generic

src/locales/en_US.ts
```diff
@@ -1314,6 +1314,13 @@ const strings = {
...
+  light_account_card_backup_message: 'Without a backup, you risk losing your funds!',
```
paullinator, line 1320: Rename key to `backup_account_message` to make more generic

paullinator (changes requested):

src/locales/en_US.ts
Jon-edge: The suggestion seems more specific rather than generic. 
"Backup Account" describes a subset (modal, reduced signup flow) of the overall "Light Account" feature

src/locales/en_US.ts
Jon-edge: The suggestion seems more specific rather than generic.
"Backup Account" describes a subset (modal, reduced signup flow) of the overall "Light Account" feature

Jon-edge:
I actually think we should make this less generic.

I see a possible future extension of this design as:
**BackupFloatingCard** - Modification to this PR's implementation. Entirely specific to light account
**CompromisedFloatingCard** - Account/wallet(s) compromised. Can push out high priority messages via an Edge-owned server.

**NotificationFloatingCard** - Perhaps can replace airships such as transactions sent/received/confirmed. Could adhere to new settings for notification duration, which to ignore, behave generally similar to native phone notifications.
**PromoFloatingCard** - You were randomly selected to win X coin for providing feedback. Maybe this can just be some informative card. Grabs data directly from 

These can be 4 distinct cards or 2 types of cards.

All of these cards can live under some **FloatingCardArea** component 
- Would be responsible for deciding which cards to show, how to show them, under what conditions, etc.
- Would replace the current FloatingCard component from this PR.

Jon-edge (commented):
I actually think we should make this less generic.

I see a possible future extension of this design as:
`BackupFloatingCard` 
- Modification to this PR's implementation. 
- Entirely specific to light account, 
- Non-dismissible
`DangerFloatingCard` 
- Account/wallet(s) compromised. 
- Can push out high priority messages via an Edge-owned server.

`NotificationFloatingCard` - Perhaps can replace airships such as transactions sent/received/confirmed. Could adhere to new settings for notification duration, which to ignore, behave generally similar to native phone notifications. Each notification could have some category that would allow user and/or system controlled behavior.

All of these cards can live under some `FloatingCardArea` component
- Would be responsible for deciding which cards to show, how to show them, under what conditions, etc.
- Would replace the current FloatingCard component from this PR.

Jon-edge:
implemented in a future light account PR

PR #4234, opened 2023-05-30
Rename tracking event for usp swipe

Jon-edge:
The event `Signup_Account_Review_Done` was incorrectly named implying the end of the signup process, while in reality it happens at the very beginning of the signup flow (after USPs are viewed).

Rename to `Signup_Welcome`

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204664120652186
  - https://app.asana.com/0/0/1204697741640174

### DIFF
--- a/src/components/scenes/GettingStartedScene.tsx
+++ b/src/components/scenes/GettingStartedScene.tsx
@@ -101,12 +101,13 @@ export const GettingStartedScene = (props: Props) => {
         swipeOffset.value = 0
       }, 500)
 
-      if (localUsersLength > 0) navigation.navigate('login', { loginUiInitialRoute: 'login-password' })
-      else {
-        logEvent('Signup_Account_Review_Done', {
-          variantId,
-          variantParams: { method: 'swipe' }
-        })
+      logEvent('Signup_Welcome', {
+        variantId,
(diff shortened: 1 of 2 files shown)


swansontec (approved):

src/components/scenes/GettingStartedScene.tsx
```diff
@@ -101,11 +101,16 @@ export const GettingStartedScene = (props: Props) => {
...
           variantId,
-          variantParams: { method: 'swipe' }
+          variantParams: { doneMethod: 'swipe' }
+        })
+        navigation.navigate('login', { loginUiInitialRoute: 'login-password' })
+      } else {
+        logEvent('Signup_Welcome', {
+          variantId,
+          variantParams: { doneMethod: 'swipe' }
         })
```
swansontec, line 114: I notice that this exact block of code exists in both branches. You could move it above the `if`, so you only have a single copy. Then the only thing inside the `if` would be the `navigate` call.

Jon-edge: Good catch, thanks

PR #4237, opened 2023-05-30
Fix undefined value after wallet creation

itayplav:
### CHANGELOG

none

### Dependencies

none

### Requirements

---
  - https://app.asana.com/0/0/1204541171459583


swansontec (approved):
The `account.swapConfig` is only `undefined` if the user is currently logged out. There may be something weird happening in the login flow, where we haven't fully logged in yet while creating the wallets. In the meantime, this looks like a safe fix.

PR #4242, opened 2023-06-01
Performance improvements for card program

samholmes:
### CHANGELOG

- Changed: Improved performance of purchase of Visa® Card by being hasty and using debouncing techniques

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204616457260607

### DIFF
--- a/src/plugins/gui/providers/ioniaProvider.ts
+++ b/src/plugins/gui/providers/ioniaProvider.ts
@@ -315,7 +315,9 @@ export const makeIoniaProvider: FiatProviderFactory<IoniaMethods> = {
         expiry: Date.now() + ONE_MINUTE,
         rateQueryPromise: ratePromise
       }
-      return await ratePromise
+      const rate = await ratePromise
+      logActivity(`Ionia rates a $${cardAmount} card at ${rate} ${currencyCode}`)
+      return rate
     }
 
     function checkAmountMinMax(fiatAmount: number) {
@@ -414,7 +416,6 @@ export const makeIoniaProvider: FiatProviderFactory<IoniaMethods> = {
 
         const rateAmount = await getCardPurchaseRateAmount(quoteParams.displayCurrencyCode, RATE_QUOTE_CARD_AMOUNT)
         const rateExchangeAmount = await quoteParams.wallet.nativeToDenomination(toBigNumberString(rateAmount), quoteParams.displayCurrencyCode)
-        logActivity(`Ionia rates a $${RATE_QUOTE_CARD_AMOUNT} card at ${rateExchangeAmount} ${quoteParams.displayCurrencyCode}`)
 
         const price = RATE_QUOTE_CARD_AMOUNT / parseFloat(rateExchangeAmount)
         const cryptoAmount =
--- a/src/plugins/gui/providers/ioniaProvider.ts
+++ b/src/plugins/gui/providers/ioniaProvider.ts
@@ -16,6 +16,7 @@ import {
   Cleaner,
   uncleaner
 } from 'cleaners'
+import { EdgeParsedUri } from 'edge-core-js'
 import { sprintf } from 'sprintf-js'
 import URL from 'url-parse'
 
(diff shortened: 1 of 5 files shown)


src/plugins/gui/providers/ioniaProvider.ts
```diff
@@ -229,12 +230,32 @@ export const makeIoniaProvider: FiatProviderFactory<IoniaMethods> = {
...
     const ratesCache: { [currencyCode: string]: { expiry: number; rateQueryPromise: Promise<number> } } = {}
 
     //
     // Private methods:
     //
 
+    async function getPurchaseCard(currencyCode: string, cardAmount: number): Promise<IoniaPurchaseCard> {
+      const newPromise: Promise<IoniaPurchaseCard> = new Promise<IoniaPurchaseCard>(resolve => {
+        // Debounce the request
+        setTimeout(
```
paullinator: This isn't a good debounce implementation as you will still hit the Ionia API on every getQuote. A better implementation is

1. Upon getQuote, save the most recent request amount in local state and set 1s timer
2. If another getQuote comes in, replace the previous request and restart 1s timer
3. Once timer expires, fire off API request
4. If more getQuotes come in, wait until previous API call finishes then go to step 1

src/plugins/gui/providers/ioniaProvider.ts
```diff
@@ -346,34 +367,75 @@ export const makeIoniaProvider: FiatProviderFactory<IoniaMethods> = {
         }
         return fiatProviderAssetMap
       },
-      async getQuote(params: FiatProviderGetQuoteParams) {
-        if (params.wallet == null) throw new Error('missing wallet')
+      async getQuote(quoteParams: FiatProviderGetQuoteParams) {
```
paullinator, line 373: next time don't do rename in the same commit to minimize irrelevant diffs

paullinator (changes requested):

src/plugins/gui/providers/ioniaProvider.ts
```diff
@@ -392,34 +413,75 @@ export const makeIoniaProvider: FiatProviderFactory<IoniaMethods> = {
...
+
+            // Use the purchase card URI to complete the payment over the Payment Protocol:
+            const parsedUri: EdgeParsedUri & { paymentProtocolUrl?: string } = await wallet.parseUri(purchaseCard.uri, quoteParams.displayCurrencyCode)
+            const { paymentProtocolUrl } = parsedUri
+            if (paymentProtocolUrl == null) {
+              throw new Error(lstrings.missing_provider_payment_address_message)
+            }
+
+            // Check if the URI requires a warning to the user:
+            const approved = await addressWarnings(parsedUri, quoteParams.displayCurrencyCode)
```
paullinator, line 460: This breaks the plugin model. addressWarnings should be wrapped by `showUi`. Maybe in the cleanup PR

src/plugins/gui/providers/ioniaProvider.ts
```diff
@@ -392,34 +413,75 @@ export const makeIoniaProvider: FiatProviderFactory<IoniaMethods> = {
...
+                      reject(
+                        new Error(
+                          `User cancelled quote of ${cryptoAmount} ${quoteParams.displayCurrencyCode} to '${purchaseCard.uri}' to purchase ${fiatAmount} USD card.`
+                        )
+                      )
+                    },
+                    onDone: resolve
+                  }
+                })
+                .catch(reject)
```
paullinator, line 483: don't pass reject directly. Instead, put anonymous function that logs the error object and where you are in the code with some context tagging.  Then call the reject(error). This gives us some context.

src/plugins/gui/providers/ioniaProvider.ts
```diff
@@ -277,30 +277,28 @@ export const makeIoniaProvider: FiatProviderFactory<IoniaMethods> = {
...
-              // IoniaPurchaseCard ready!
-            })
+      return await new Promise<IoniaPurchaseCard>((resolve, reject) => {
+        // Clear the previous task timeout if present
+        if (purchaseCardTimeoutId != null) clearTimeout(purchaseCardTimeoutId)
+
+        // Hastily invoke the task promise with a debounce:
+        const newPurchaseCardTimeoutId = setTimeout(() => {
+          if (purchaseCardTimeoutId === newPurchaseCardTimeoutId) {
+            queryPurchaseCard(currencyCode, cardAmount).then(resolve, reject)
```
paullinator, line 295: the `else` case leaves a dangling promise. I think we're guilty of this elsewhere in the GUI but ideally return undefined and have caller of `getPurchaseCard` throw it away.

paullinator (changes requested):

paullinator (approved):

PR #4254, opened 2023-06-08
Fix race condition for exchange settings switch

samholmes:
### CHANGELOG

- Fixed: Fix race condition for exchange settings switch

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204696624144919

### DIFF
--- a/src/components/scenes/SwapSettingsScene.tsx
+++ b/src/components/scenes/SwapSettingsScene.tsx
@@ -193,7 +193,8 @@ export class SwapSettings extends React.Component<Props, State> {
         label={displayName}
         value={pluginEnabled}
         onPress={async () => {
-          await swapConfigs[pluginId].changeEnabled(!pluginEnabled)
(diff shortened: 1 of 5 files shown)


src/components/scenes/SwapSettingsScene.tsx
```diff
@@ -41,6 +42,7 @@ type Props = StateProps & DispatchProps & ThemeProps
 
 interface State {
   enabled: { [pluginId: string]: boolean }
+  swapConfigUnsubscribeFns: CallbackRemover[]
```
swansontec, line 45: On a class-style component, you can add things as properties of `this`, instead of putting them into `State`. I believe `swapConfigUnsubscribeFns` should be a `this` property, not a state property, since it doesn't affect rendering.

In the hooks world we wouldn't put this into a `useState` - it would be managed internally by the `useEffect` hook's own storage.

swansontec (approved):
One small optional change, but otherwise this looks pretty good.

PR #4264, opened 2023-06-15
Speed up the useWatch hook

swansontec:
When changing the watched object (such as switching wallets), there is a one-tick delay before the hook's return value updates. We need to remove this delay to fix some bugs.

### CHANGELOG

- fixed: Do not accidentally back out of the transaction list scene when selecting a token wallet from the drop-down.

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204827955021786
  - https://app.asana.com/0/0/1204845530898852

### DIFF
--- a/src/hooks/useWatch.ts
+++ b/src/hooks/useWatch.ts
@@ -9,12 +9,13 @@ export function useWatch<
   },
   Name extends keyof T
 >(object: T, name: Name): T[Name] {
-  const [out, setOut] = React.useState<any>(object[name])
+  // We need to re-render the component when the value changes.
+  // Since [] !== [], we can use an empty array to force an update:
+  const [, rerender] = React.useState([])
 
   React.useEffect(() => {
-    setOut(object[name])
-    return object.watch(name, setOut)
+    return object.watch(name, () => rerender([]))
   }, [object, name])
 
-  return out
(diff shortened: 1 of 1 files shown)


src/hooks/useWatch.ts
```diff
@@ -9,12 +9,13 @@ export function useWatch<
   },
   Name extends keyof T
 >(object: T, name: Name): T[Name] {
-  const [out, setOut] = React.useState<any>(object[name])
+  // We need to re-render the component when the value changes.
+  // Since [] !== [], we can use an empty array to force an update:
+  const [, rerender] = React.useState([])
```
samholmes, line 14: Wouldn't `NaN` work here because `NaN !== NaN`?

```
const [, rerender] = React.useState(NaN)
```

This feels more intentional of a hack. If you like the suggestion, then please try it and let me know if it works. Otherwise, what you have is fine.

samholmes (approved):

src/hooks/useWatch.ts
swansontec: It won't work. React internally uses `Object.is` for comparisons, and `Object.is(NaN,NaN)` returns true.

PR #4267, opened 2023-06-15
Paul/rsync builds

paullinator:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204836801484055
  - https://app.asana.com/0/0/1204846049105749

### DIFF
--- a/scripts/deploy.ts
+++ b/scripts/deploy.ts
@@ -7,6 +7,7 @@ const argv = process.argv
 const mylog = console.log
 
 const _rootProjectDir = join(__dirname, '../')
+const githubSshKey = process.env.GITHUB_SSH_KEY ?? join(_rootProjectDir, 'id_github')
 
 let _currentPath = __dirname
 
@@ -41,6 +42,7 @@ interface BuildConfigFile {
   hockeyAppToken: string
   productName: string
   projectName: string
+  rsyncLocation?: string
 }
 
 /**
@@ -164,7 +166,6 @@ function buildIos(buildObj: BuildObj) {
     process.env.MATCH_KEYCHAIN_PASSWORD != null &&
     process.env.MATCH_PASSWORD != null
   ) {
-    const githubSshKey = process.env.GITHUB_SSH_KEY ?? join(_rootProjectDir, 'id_github')
     call(`security unlock-keychain -p '${process.env.KEYCHAIN_PASSWORD ?? ''}' "${process.env.HOME ?? ''}/Library/Keychains/login.keychain"`)
     call(`security set-keychain-settings -l ${process.env.HOME ?? ''}/Library/Keychains/login.keychain`)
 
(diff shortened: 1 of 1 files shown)


scripts/deploy.ts
```diff
@@ -374,6 +376,20 @@ function buildCommonPost(buildObj: BuildObj) {
...
+
+    // const rsyncDir = join(buildObj.rsyncLocation, repoBranch, platformType, String(buildNum))
+
+    const datePrefix = new Date().toISOString().slice(2, 19).replace(/:/gi, '').replace(/-/gi, '')
+    const rsyncFile = escapePath(`${datePrefix}--${productNameClean}--${platformType}--${repoBranch}--${buildNum}--${guiHash.slice(0, 8)}.ipa`)
+
+    // call(`rsync -avz -e "ssh -i ${githubSshKey}" ${buildObj.ipaFile} ${rsyncDir}/`)
+    call(`rsync -avz -e "ssh -i ${githubSshKey}" ${buildObj.ipaFile} ${(join(buildObj.rsyncLocation), rsyncFile)}`)
+    mylog('\n*** Upload to rsyncLocation Complete ***')
+  }
```
samholmes: You should co-locate line 328 within his section of code (`const { buildNum, guiHash, platformType, productNameClean, repoBranch } = buildObj`). This destructuring is for line 387, so maybe put it above that line?

scripts/deploy.ts
```diff
@@ -73,6 +77,17 @@ interface BuildObj extends BuildConfigFile {
   ipaFile: string // Also APK
 }
 
+const asLatestTestFile = asObject({
```
samholmes, line 80: Use `asJSON` and `asCodec`?

scripts/deploy.ts
```diff
@@ -383,16 +398,75 @@ function buildCommonPost(buildObj: BuildObj) {
...
     mylog(`\n\nUploading to rsyncLocation ${buildObj.rsyncLocation}`)
-    mylog('***********************\n')
+    mylog('***********************************************************************\n')
 
     // const rsyncDir = join(buildObj.rsyncLocation, repoBranch, platformType, String(buildNum))
 
     const datePrefix = new Date().toISOString().slice(2, 19).replace(/:/gi, '').replace(/-/gi, '')
-    const rsyncFile = escapePath(`${datePrefix}--${productNameClean}--${platformType}--${repoBranch}--${buildNum}--${guiHash.slice(0, 8)}.ipa`)
+    const fileSuffixArr = buildObj.ipaFile.split('.')
+    const fileSuffix = fileSuffixArr[fileSuffixArr.length - 1]
```
samholmes, line 407: Propose using the word "extension" or "ext" instead of "suffix". You can also shorten this to one line.

```
    const [fileExtension] = buildObj.ipaFile.split('.').reverse()
```

scripts/deploy.ts
samholmes: Hmm, upon further review, the cleaner is never really used. Why not just a static type then?

samholmes (changes requested):

scripts/deploy.ts
paullinator: It was going to be used and for now the data doesn't need to be read but only written. Will change

samholmes (approved):

PR #4269, opened 2023-06-16
Derive all scene props from `EdgeSceneProps`

swansontec:
Advantages:

- The `EdgeSceneProps` name doesn't conflict with built-in react-navigation names.
- We can have `route` and `navigation` available on every scene, and ESLint doesn't complain if they are unused.
- We can easily see all the ways to route to each scene, right in the scene itself.
- TypeScript double-checks the props with Main.tsx, so we know the route names are correct and complete. There was a lot of room for miss-matches in the old way.
- Fewer unexplained @ts-expect-errors.
- Less things to import.
- Simpler unit-testing, with better type-checking.

Since this only changes types & tests, we don't need much QA attention. If TypeScript and Jest are happy, it's good.

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204845024114242
  - https://app.asana.com/0/0/1204845303063745


samholmes (approved):
I added some missed scenes.

I notice a type error in fakeSceneProps.ts during my review:

```
[{
	"resource": "/Users/samholmes/Developer/edge-react-gui/src/util/fake/fakeSceneProps.ts",
	"owner": "typescript",
	"code": "2322",
	"severity": 8,
	"message": "Type '{ name: any; key: string; params: RouteParamList[Name]; }' is not assignable to type 'RouteProp<Name>'.\n  Type '{ name: any; key: string; params: RouteParamList[Name]; }' is not assignable to type 'undefined extends AppParamList[Name] ? Readonly<{ params?: Readonly<AppParamList[Name]> | undefined; }> : Readonly<...>'.",
	"source": "ts",
	"startLineNumber": 48,
	"startColumn": 5,
	"endLineNumber": 48,
	"endColumn": 10,
	"relatedInformation": [
		{
			"startLineNumber": 330,
			"startColumn": 3,
			"endLineNumber": 330,
			"endColumn": 8,
			"message": "The expected type comes from property 'route' which is declared here on type 'EdgeSceneProps<Name>'",
			"resource": "/Users/samholmes/Developer/edge-react-gui/src/types/routerTypes.tsx"
		}
	]
}]
```

PR #4270, opened 2023-06-16
Better tx list

swansontec:
Stream transactions from the core in a hook, instead of chunking them into redux.

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203927798750595
  - https://app.asana.com/0/0/1204845530898842


samholmes (commented):
Still working on my review. I'm at everything up to 16a9cc6bfb39649bdc884c756dc027777ec1fcc1 looks good. I'm still working on commit 16a9cc6bfb39649bdc884c756dc027777ec1fcc1 and onward.

src/hooks/useTransactionList.ts
```diff
@@ -0,0 +1,152 @@
...
+            if (result.done) {
+              atEnd = true
+              txs.splice(offset, Infinity) // Trim stragglers
+              requestRender()
+            } else {
+              const { value } = result
+              txs.splice(offset, value.length, ...value)
+              offset += value.length
+              requestRender()
+              if (offset < txs.length) requestMore.current()
```
samholmes, line 124: Why are we calling the `requestMore` recursively if the offset doesn't match the length of the result set? Shouldn't this always match; when would it not match?

src/hooks/useTransactionList.ts
```diff
@@ -0,0 +1,152 @@
...
+    const txs: EdgeTransaction[] = []
+
+    // Sends the mutable state to React,
+    // merging the two transaction lists together:
+    function requestRender() {
+      setOutput({
+        atEnd,
+        transactions: txs.map(tx => changedTxs.get(tx.txid) ?? tx)
+      })
+    }
```
samholmes, line 50: I like this pattern of using a `requestRender` function because we could always add debouncing in the future if there ever is thrashing of events.

samholmes (changes requested):
Pretty solid PR! Minor feedback if you deem it to make sense.

samholmes (approved):
Approved after OTS!

src/components/scenes/TransactionListScene.tsx
```diff
@@ -74,185 +67,136 @@ function TransactionListComponent(props: Props) {
...
+        onRefresh={() => setSearching(true)}
+      />
+    )
+  }, [theme])
 
   const topArea = React.useMemo(() => {
     return (
       <TransactionListTop
-        isEmpty={isTransactionListUnsupported || transactions.length < 1}
+        isEmpty={listItems.length < 1}
```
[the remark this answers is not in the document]
swansontec: It will be empty once we fully finish loading on an empty wallet with no transactions, or on a wallet that doesn't support the transaction list to begin with.

src/hooks/useTransactionList.ts
swansontec: This is for the case were a new transaction arrives from the engine, and we start re-writing the existing transaction list from the beginning. We need to keep going until we reach the end of the existing list.

PR #4272, opened 2023-06-19
Replace wallet connect v1 with v2

peachbits:
---
  - https://app.asana.com/0/0/1204436463195824


src/hooks/useWalletConnect.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Access Wallet Connect
+ */
+export function useWalletConnect(): WalletConnect {
+  const account = useSelector(state => state.core.account)
+  const currencyWallets = useWatch(account, 'currencyWallets')
+
+  // Utils
+  const publicAddresses = React.useMemo(async () => {
```
swansontec, line 40: The `useMemo` hook does not work with async values. If the wallet list updates quickly, you could get a big pile-up of dangling promises.

Instead, you can either use Sam's `useAsyncValue`, which is almost exactly list `useMemo` but handles async calculations, or you can just move this calculation into `getActiveSessions`, which is already `async`, and skip the caching altogether.

src/components/scenes/WcConnectScene.tsx
```diff
@@ -27,6 +27,10 @@ import { SelectableRow } from '../themed/SelectableRow'
 
 interface Props extends EdgeSceneProps<'wcConnect'> {}
 
+export interface WcConnectSceneParams {
```
swansontec, line 35: The word "Scene" should not be part of these type names. We've been using `XyzScene` / `XyzParams`, not `XyzSceneParams`.

src/components/scenes/WcConnectionsScene.tsx
```diff
@@ -78,7 +76,8 @@ export const WcConnectionsScene = (props: Props) => {
       />
     ))
       .then((result: string | undefined) => {
-        if (result) {
+        console.log(result)
```
swansontec, line 79: Extra log?

src/hooks/useWalletConnect.tsx
swansontec: Yup, just skip the caching and turn this into a method that `getActiveSessions` can call. It's not used anywhere else.

src/components/services/WalletConnectService.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }).then(res => {
+    walletConnectRef = res
+    console.log('WalletConnect initialized')
+    resolve(walletConnectRef)
+  })
+})
+
+export const WalletConnectService = () => {
+  useMount(async () => {
+    await walletConnectPromise
```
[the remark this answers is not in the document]
swansontec: OK, so the component does do something later. However, even in the later iteration, there is no error handling on the `useMount` hook. You could move this stuff into an async helper method, and then do `useMount(() => subscribeToWc().catch(error => showError(error)))`. That way you can still use the `async` keyword, but also get error handling on the whole thing.

swansontec (changes requested):
Pretty nice overall.

src/components/scenes/WcConnectionsScene.tsx
```diff
@@ -41,40 +43,10 @@ export const WcConnectionsScene = (props: Props) => {
...
-          }
-        }
-      }
-    }
-    return dAppConnections
-  }
+  useAsyncEffect(async () => {
+    const connections = await walletConnect.getActiveSessions()
+    setConnections(connections)
+  }, [walletConnect, props])
```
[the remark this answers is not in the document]
peachbits: Yes that is intended. The scene remains (I'm not sure why) when we navigate away to the connect and disconnect scenes. When those scenes call navigate to return to the connections scene it uses new props so we get a render and the list updates.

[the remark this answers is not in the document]
swansontec: That's fine, then. Can you comment this, since it's tricky and unexpected?

swansontec (approved):
Approved, but it will be a tricky rebase:

1. Don't forget to drop the first commit.
2. Merge conflict with the router types file.
3. Add a comment about the `props` being in the deps list, since it's unexpected but intended.

PR #4276, opened 2023-06-21
WalletConnectV2 cleanups

peachbits:
Some fixes for:

Inaccurate walletId identification from session
Error handling
Dapp spec non-compliance safety

---
  - https://app.asana.com/0/0/1204868235343560

### DIFF
--- a/src/components/scenes/WcConnectScene.tsx
+++ b/src/components/scenes/WcConnectScene.tsx
@@ -23,7 +23,7 @@ import { SceneWrapper } from '../common/SceneWrapper'
 import { CryptoIcon } from '../icons/CryptoIcon'
 import { WalletListModal, WalletListResult } from '../modals/WalletListModal'
 import { FlashNotification } from '../navigation/FlashNotification'
-import { Airship } from '../services/AirshipInstance'
+import { Airship, showError } from '../services/AirshipInstance'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 import { EdgeText } from '../themed/EdgeText'
 import { MainButton } from '../themed/MainButton'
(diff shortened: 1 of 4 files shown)


src/components/scenes/WcConnectScene.tsx
```diff
@@ -75,6 +75,7 @@ export const WcConnectScene = (props: Props) => {
       navigation.navigate('wcConnections', {})
     } catch (error: any) {
       console.error(`WalletConnect connection error: ${error.message}`)
```
swansontec, line 77: I missed this in the last review, but `error.message` is forbidden. Do `String(error)`. Or just drop the whole line, since we already `showError` on the next line.

src/hooks/useWalletConnect.tsx
```diff
@@ -245,3 +217,27 @@ const getSupportedNamespaces = (chainId: WalletConnectChainId, addr: string) =>
...
+    if (chainId == null) continue
+
+    const address = await currencyWallets[walletId].getReceiveAddress()
+    const account = `${chainId.namespace}:${chainId.reference}:${address.publicAddress}`
+    map.set(account, walletId)
+  }
+  return map
+}
+
+export const getWalletIdFromSessionNamespace = async (namespaces: SessionTypes.Namespaces, accounts: Map<string, string>): Promise<string | undefined> => {
```
swansontec, line 235: This doesn't need to be `async`, since it doesn't have any `await` statements.

swansontec (changes requested):

swansontec (approved):

src/components/scenes/WcConnectScene.tsx
peachbits: I'll leave it for additional context and print appropriately.

PR #4279, opened 2023-06-21
Upgrade to ESLint 8

swansontec:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204868426100921

### DIFF
--- a/src/components/modals/RadioListModal.tsx
+++ b/src/components/modals/RadioListModal.tsx
@@ -3,15 +3,23 @@ import { Image, Text, TouchableOpacity, View } from 'react-native'
 import { AirshipBridge } from 'react-native-airship'
 import IonIcon from 'react-native-vector-icons/Ionicons'
 
+import { useHandler } from '../../hooks/useHandler'
 import { lstrings } from '../../locales/strings'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 import { EdgeText } from '../themed/EdgeText'
 import { ListModal } from './ListModal'
 
+interface Item {
+  // Icon strings are image uri, numbers are local files:
+  icon: string | number | React.ReactNode
+  name: string
+  text?: string
+}
+
 interface Props {
   bridge: AirshipBridge<string | undefined>
   title: string
-  items: Array<{ icon: string | number | React.ReactNode; name: string; text?: string }> // Icon strings are image uri, numbers are local files
+  items: Item[]
   selected?: string
 }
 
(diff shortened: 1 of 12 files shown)


src/components/scenes/CryptoExchangeScene.tsx
```diff
@@ -285,12 +285,14 @@ export class CryptoExchangeComponent extends React.Component<Props, State> {
...
-    })
-    return null
+    ))
+      .then(async result => {
+        const { walletId, currencyCode } = result
+        if (walletId != null && currencyCode != null) {
+          await this.props.onSelectWallet(walletId, currencyCode, whichWallet)
+        }
+      })
+      .catch(error => showError(error))
```
paullinator, line 295: Will this showError give us the stack to show it was called from CryptoExchangeScene? If not, this would be a good opportunity to create and use the onionError() function

src/components/scenes/RequestScene.tsx
```diff
@@ -234,10 +234,13 @@ export class RequestSceneComponent extends React.Component<Props, State> {
...
-      .then((result?: string) => {
-        return result === 'confirm' && addressExplorer != null ? Linking.openURL(sprintf(addressExplorer, requestAddress)) : null
+      .then(async (result?: string) => {
+        if (result === 'confirm' && addressExplorer != null) {
+          const url = sprintf(addressExplorer, requestAddress)
+          await Linking.openURL(url).catch(error => showError(error))
+        }
       })
-      .catch(error => console.log(error))
+      .catch(error => showError(error))
```
paullinator, line 243: Will this showError give us the stack to show it was called from CryptoExchangeScene? If not, this would be a good opportunity to create and use the onionError() function

paullinator (changes requested):
Just a few questions but pretty much approved.

src/components/scenes/CryptoExchangeScene.tsx
swansontec: Yes, "CryptoExchangeScene.tsx:295" will show up in the stack, since we are doing `.catch(error => showError(error))`, instead of `.catch(showError)`.

We don't need any weird utility functions to handle these basic cases.

src/components/modals/RadioListModal.tsx
```diff
@@ -20,19 +28,27 @@ export function RadioListModal(props: Props) {
...
+    const { name, icon, text } = item
+
     const isSelected = selected === name
     const radio = isSelected ? { icon: 'ios-radio-button-on', color: theme.iconTappable } : { icon: 'ios-radio-button-off', color: theme.iconTappable }
     const accessibilityState = isSelected ? { checked: true } : { checked: false }
     const accessibilityHint = `${isSelected ? lstrings.on_hint : lstrings.off_hint} ${name}`
+
+    const iconElement =
+      typeof icon === 'string' ? (
+        <Image resizeMode="contain" source={{ uri: icon }} style={styles.icon} />
```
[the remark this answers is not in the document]
swansontec: It's the same logic as before. Before, we did the `? { uri: icon } : icon` thing on its own line, and then used that later inside the component. However, TypeScript couldn't track that, so I combined the component and the URL switching thing into one line to solve the type error.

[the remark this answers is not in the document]
paullinator: Ah. I missed the removal of imageIcon. All good

paullinator (approved):

PR #4280, opened 2023-06-21
Fix first 101 floating-promises

samholmes:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204848238190307

### DIFF
--- a/src/actions/AccountReferralActions.ts
+++ b/src/actions/AccountReferralActions.ts
@@ -126,7 +126,7 @@ export function activatePromotion(installerId: string): ThunkAction<Promise<void
       plugins: lockStartDates(clean.plugins, now)
     }
     dispatch({ type: 'PROMOTION_ADDED', data: promotion })
-    saveAccountReferral(getState())
+    await saveAccountReferral(getState())
   }
 }
 
@@ -136,7 +136,7 @@ export function activatePromotion(installerId: string): ThunkAction<Promise<void
 export function removePromotion(installerId: string): ThunkAction<Promise<void>> {
   return async (dispatch, getState) => {
     dispatch({ type: 'PROMOTION_REMOVED', data: installerId })
-    saveAccountReferral(getState())
+    await saveAccountReferral(getState())
   }
 }
 
@@ -146,7 +146,7 @@ export function removePromotion(installerId: string): ThunkAction<Promise<void>>
 export function hideMessageTweak(messageId: string, source: TweakSource): ThunkAction<Promise<void>> {
   return async (dispatch, getState) => {
     dispatch({ type: 'MESSAGE_TWEAK_HIDDEN', data: { messageId, source } })
-    saveAccountReferral(getState())
+    await saveAccountReferral(getState())
   }
 }
 
@@ -156,7 +156,7 @@ export function hideMessageTweak(messageId: string, source: TweakSource): ThunkA
 export function ignoreAccountSwap(ignore: boolean = true): ThunkAction<Promise<void>> {
   return async (dispatch, getState) => {
     dispatch({ type: 'ACCOUNT_SWAP_IGNORED', data: ignore })
-    saveAccountReferral(getState())
+    await saveAccountReferral(getState())
   }
 }
 
@@ -201,7 +201,7 @@ export function refreshAccountReferral(): ThunkAction<Promise<void>> {
 
     if (cache.accountMessages.length <= 0 && cache.accountPlugins.length <= 0) return
     dispatch({ type: 'ACCOUNT_TWEAKS_REFRESHED', data: cache })
-    saveReferralCache(getState())
+    await saveReferralCache(getState())
   }
 }
 
--- a/src/actions/CreateWalletActions.tsx
+++ b/src/actions/CreateWalletActions.tsx
@@ -222,12 +222,11 @@ export function createAccountTransaction(
               category: 'Expense:' + sprintf(lstrings.create_wallet_account_metadata_category, createdWalletCurrencyCode),
               notes: sprintf(lstrings.create_wallet_account_metadata_notes, createdWalletCurrencyCode, createdWalletCurrencyCode, config.supportEmail)
             }
-            paymentWallet.saveTxMetadata(edgeTransaction.txid, currencyCode, edgeMetadata).then(() => {
-              navigation.navigate('walletsTab', { screen: 'walletList' })
-              setTimeout(() => {
-                Alert.alert(lstrings.create_wallet_account_payment_sent_title, lstrings.create_wallet_account_payment_sent_message)
-              }, 750)
-            })
+            paymentWallet.saveTxMetadata(edgeTransaction.txid, currencyCode, edgeMetadata).catch(err => console.warn(err))
+            navigation.navigate('walletsTab', { screen: 'walletList' })
+            setTimeout(() => {
+              Alert.alert(lstrings.create_wallet_account_payment_sent_title, lstrings.create_wallet_account_payment_sent_message)
+            }, 750)
           }
         },
         alternateBroadcast:
@@ -235,7 +234,7 @@ export function createAccountTransaction(
       })
     } else {
       // if handle is now unavailable
-      dispatch(createHandleUnavailableModal(navigation, createdWalletId, accountName))
+      await dispatch(createHandleUnavailableModal(navigation, createdWalletId, accountName))
     }
   }
 }
@@ -244,7 +243,7 @@ export function createHandleUnavailableModal(navigation: NavigationBase, newWall
   return async (dispatch, getState) => {
     const state = getState()
     const { account } = state.core
-    account.changeWalletStates({
+    await account.changeWalletStates({
       [newWalletId]: {
         deleted: true
       }
--- a/src/actions/DeepLinkingActions.ts
+++ b/src/actions/DeepLinkingActions.ts
@@ -131,7 +131,7 @@ export async function handleLink(navigation: NavigationBase, dispatch: Dispatch,
     case 'promotion': {
       if (!state.account.accountReferralLoaded) return false
       const { installerId = '' } = link
-      dispatch(activatePromotion(installerId))
+      await dispatch(activatePromotion(installerId))
       return true
     }
 
--- a/src/actions/DeepLinkingActions.ts
+++ b/src/actions/DeepLinkingActions.ts
@@ -38,15 +38,20 @@ export function launchDeepLink(navigation: NavigationBase, link: DeepLink): Thun
  * Maybe we were in the wrong state before, but now we are able
  * to launch the link.
  */
-export function retryPendingDeepLink(navigation: NavigationBase): ThunkAction<Promise<void>> {
-  return async (dispatch, getState) => {
+export function retryPendingDeepLink(navigation: NavigationBase): ThunkAction<void> {
+  return (dispatch, getState) => {
     const state = getState()
     const { pendingDeepLink } = state
     if (pendingDeepLink == null) return
+    // Clear the link as we try to handle
     dispatch({ type: 'DEEP_LINK_HANDLED' })
 
-    const handled = await handleLink(navigation, dispatch, state, pendingDeepLink)
-    // If we handled the link, clear it:
+    const handled = handleLink(navigation, dispatch, state, pendingDeepLink).catch(err => {
+      console.warn(err)
+      return false
+    })
+
+    // If we didn't handled the link, put it back
     if (!handled) {
       dispatch({ type: 'DEEP_LINK_RECEIVED', data: pendingDeepLink })
     }
@@ -137,7 +142,7 @@ export async function handleLink(navigation: NavigationBase, dispatch: Dispatch,
 
     case 'requestAddress': {
       if (!allWalletsLoaded) return false
-      doRequestAddress(navigation, state.core.account, dispatch, link)
+      await doRequestAddress(navigation, state.core.account, dispatch, link)
       return true
     }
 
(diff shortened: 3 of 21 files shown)


src/actions/DeepLinkingActions.ts
```diff
@@ -131,13 +131,13 @@ export async function handleLink(navigation: NavigationBase, dispatch: Dispatch,
     case 'promotion': {
       if (!state.account.accountReferralLoaded) return false
       const { installerId = '' } = link
-      dispatch(activatePromotion(installerId))
+      dispatch(activatePromotion(installerId)).catch(err => showError(err))
```
swansontec, line 134: The `handleLink` function should simply `await` these lines so the error flows out.

Then, we should put catch statements on the places where `handleLink` is being called:

- `SwapDetailsTile` will probably `showError`
- `retryPendingDeepLink` should log the problem and leave the link un-handled
- `launchDeepLink` should probably become `async`, so it can `await` the `handleLink`. This kicks the can up to the `launchDeepLink` callers.

src/actions/DeepLinkingActions.ts
```diff
@@ -131,13 +131,13 @@ export async function handleLink(navigation: NavigationBase, dispatch: Dispatch,
...
       return true
     }
 
     case 'requestAddress': {
       if (!allWalletsLoaded) return false
-      doRequestAddress(navigation, state.core.account, dispatch, link)
+      doRequestAddress(navigation, state.core.account, dispatch, link).catch(err => showError(err))
```
swansontec, line 140: `await` instead of `catch`

src/actions/DeepLinkingActions.ts
```diff
@@ -178,7 +178,7 @@ export async function handleLink(navigation: NavigationBase, dispatch: Dispatch,
     }
 
     case 'price-change': {
-      dispatch(launchPriceChangeBuySellSwapModal(navigation, link))
+      dispatch(launchPriceChangeBuySellSwapModal(navigation, link)).catch(err => showError(err))
```
swansontec, line 181: `await` instead of `catch`

src/actions/DeepLinkingActions.ts
```diff
@@ -210,7 +210,7 @@ export async function handleLink(navigation: NavigationBase, dispatch: Dispatch,
 
       if (matchingWalletIdsAndUris.length === 1) {
         const { walletId, parsedUri } = matchingWalletIdsAndUris[0]
-        dispatch(handleWalletUris(navigation, currencyWallets[walletId], parsedUri))
+        dispatch(handleWalletUris(navigation, currencyWallets[walletId], parsedUri)).catch(err => showError(err))
```
swansontec, line 213: `await` instead of `catch`

src/actions/DeepLinkingActions.ts
```diff
@@ -238,7 +238,7 @@ export async function handleLink(navigation: NavigationBase, dispatch: Dispatch,
         return true
       }
       const { parsedUri, walletId } = widUri
-      dispatch(handleWalletUris(navigation, currencyWallets[walletId], parsedUri))
+      dispatch(handleWalletUris(navigation, currencyWallets[walletId], parsedUri)).catch(err => showError(err))
```
swansontec, line 241: `await` instead of `catch`

src/components/services/AirshipInstance.tsx
```diff
@@ -24,7 +24,7 @@ export function showError(error: unknown, options: ShowErrorWarningOptions = {})
     Bugsnag.notify(`showError: ${translatedError}`)
   }
   console.log(redText('Showing error drop-down alert: ' + makeErrorLog(error)))
-  Airship.show(bridge => <AlertDropdown bridge={bridge} message={translatedError} />)
+  Airship.show(bridge => <AlertDropdown bridge={bridge} message={translatedError} />).catch(err => console.error(err))
```
swansontec, line 27: Do we want the red box of doom in here?

src/actions/FioActions.tsx
```diff
@@ -62,7 +62,7 @@ export function checkFioObtData(wallet: EdgeCurrencyWallet, transactions: EdgeTr
     if (account == null || account.currencyConfig == null || fioWallets.length === 0) {
       checkFioObtDataCounter++
       return setTimeout(() => {
-        dispatch(checkFioObtData(wallet, transactions))
+        dispatch(checkFioObtData(wallet, transactions)).catch(err => console.error(err))
       }, 400)
     }
```
swansontec, line 67: This should be turned into a `while` loop, using `await snooze(400)`, to delay the rest of the function until the FIO wallets appear.

src/actions/FioActions.tsx
```diff
@@ -75,7 +75,7 @@ export function checkFioObtData(wallet: EdgeCurrencyWallet, transactions: EdgeTr
         try {
           const { name } = edgeMetadata
           if (name && (await fioPlugin.otherMethods.isFioAddressValid(name))) {
-            addToFioAddressCache(state.core.account, [name])
+            addToFioAddressCache(state.core.account, [name]).catch(err => console.error(err))
```
swansontec, line 78: `await` instead of `catch`

src/actions/FioActions.tsx
```diff
@@ -89,7 +89,7 @@ export function checkFioObtData(wallet: EdgeCurrencyWallet, transactions: EdgeTr
         edgeMetadata.notes = `${fioNotes}\n${edgeMetadata.notes || ''}`
         edgeMetadata.name = obtForTx.payer_fio_address
 
-        addToFioAddressCache(state.core.account, [obtForTx.payer_fio_address])
+        addToFioAddressCache(state.core.account, [obtForTx.payer_fio_address]).catch(err => console.error(err))
```
swansontec, line 92: `await` instead of `catch`

src/actions/FioActions.tsx
```diff
@@ -144,7 +144,7 @@ function refreshNamesToCheckExpired(navigation: NavigationBase): ThunkAction<Pro
 
     if (namesToCheck.length !== 0) {
       dispatch({ type: 'FIO/CHECKING_EXPIRED', data: true })
-      dispatch(checkExpiredFioDomains(navigation, namesToCheck, fioWalletsById))
+      dispatch(checkExpiredFioDomains(navigation, namesToCheck, fioWalletsById)).catch(err => console.error(err))
```
swansontec, line 147: `await` instead of `catch`

src/actions/FioActions.tsx
```diff
@@ -172,7 +172,7 @@ export function checkExpiredFioDomains(
       dispatch({ type: 'FIO/SET_LAST_EXPIRED_CHECKS', data: expiredLastChecks })
       // @ts-expect-error
       dispatch({ type: 'FIO/EXPIRED_REMINDER_SHOWN', data: true })
-      setFioExpiredCheckToDisklet(expiredLastChecks, state.core.disklet)
+      setFioExpiredCheckToDisklet(expiredLastChecks, state.core.disklet).catch(err => console.error(err))
```
swansontec, line 175: `await` instead of `catch`

src/actions/CreateWalletActions.tsx
```diff
@@ -222,20 +222,23 @@ export function createAccountTransaction(
...
-            })
+            paymentWallet
+              .saveTxMetadata(edgeTransaction.txid, currencyCode, edgeMetadata)
+              .then(() => {
+                navigation.navigate('walletsTab', { screen: 'walletList' })
+                setTimeout(() => {
+                  Alert.alert(lstrings.create_wallet_account_payment_sent_title, lstrings.create_wallet_account_payment_sent_message)
+                }, 750)
+              })
+              .catch(err => showError(err))
```
swansontec, line 233: Should the `catch` go on `saveTxMetadata`, so we still navigate to the right place, even if we fail to save the tx details?

swansontec (changes requested):

src/actions/AccountReferralActions.ts
```diff
@@ -126,7 +126,7 @@ export function activatePromotion(installerId: string): ThunkAction<Promise<void
       plugins: lockStartDates(clean.plugins, now)
     }
     dispatch({ type: 'PROMOTION_ADDED', data: promotion })
-    saveAccountReferral(getState())
+    saveAccountReferral(getState()).catch(err => console.error(err))
```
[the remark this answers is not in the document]
samholmes: I assumed these `saveAccountReferral` calls were concurrent with the redux action routines. If that's not the case, then these were bugs and I will add `await` instead. I was under the impression that these calls were more like the log calls; they shouldn't block. I will look more closely at each call to determine whether they should be concurrent like log or not concurrent (if they're bugs).

My main concern with a lot of "bug fixes" is that they may crop up with a lot of P0s (bugs that existed but never were apparent). That's a good thing, but could impact our sprint, time-wise.

src/util/logger.ts
```diff
@@ -157,12 +157,12 @@ export async function logWithType(type: LogType, ...info: Array<number | string
   global.clog(logs)
 }
 
-export async function log(...info: Array<number | string | null | {}>): Promise<void> {
-  await logWithType('info', ...info)
+export function log(...info: Array<number | string | null | {}>): void {
+  logWithType('info', ...info).catch(err => console.error(err))
```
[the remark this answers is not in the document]
samholmes: use `console.warn`

src/actions/LoginActions.tsx
```diff
@@ -321,13 +321,13 @@ async function safeCreateWallet(account: EdgeAccount, walletType: string, wallet
         data: { currencyCode: wallet.currencyInfo.currencyCode, walletId: wallet.id }
       })
     }
-    dispatch(trackAccountEvent('Signup_Wallets_Created_Success'))
+    dispatch(trackAccountEvent('Signup_Wallets_Created_Success')).catch(err => showError(err))
```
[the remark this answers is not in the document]
samholmes: I'll look into this one more. Question is whether it should be non-async and log it's own errors like `log`, or is it different because it's revenue related and we want to block on this async call.

src/actions/DeepLinkingActions.ts
samholmes: How is `handleLink` being called by `SwapDetailsTile`; I couldn't make that connection?

I implemented `retryPendingDeepLink` such that it catches internally; though correct me if I'm wrong, do we need a fail rate counter so we can avoid thrashing this retry routine?

src/components/services/AirshipInstance.tsx
samholmes: We really shouldn't be erroring here, so I'd argue this should be in devs faces if it were to ever fail.

src/actions/FioActions.tsx
samholmes: No while loop is needed since this function is recursive.

src/actions/TrackingActions.ts
```diff
@@ -37,13 +37,13 @@ export function trackConversion(
  * Tracks an event tied to a particular account's affiliate information,
  * such as creating the initial wallets.
  */
-export function trackAccountEvent(event: TrackingEventName, trackingValues: TrackingValues = {}): ThunkAction<Promise<unknown>> {
-  return async (dispatch, getState) => {
+export function trackAccountEvent(event: TrackingEventName, trackingValues: TrackingValues = {}): ThunkAction<void> {
+  return (dispatch, getState) => {
```
swansontec, line 41: Awesome. However, I wonder if this should have been in the previous commit.

src/components/scenes/SwapSettingsScene.tsx
```diff
@@ -307,8 +309,8 @@ export const SwapSettingsScene = connect<StateProps, DispatchProps, OwnProps>(
     changePreferredSwapPluginType(swapPluginType) {
       dispatch(setPreferredSwapPluginType(swapPluginType))
     },
-    ignoreAccountSwap() {
-      dispatch(ignoreAccountSwap())
+    async ignoreAccountSwap() {
+      await dispatch(ignoreAccountSwap())
```
swansontec, line 313: Awesome. It's always a good sign when we do `await dispatch`. It means we have correctly piped the promise up to the top of the call stack. 👍

src/actions/TrackingActions.ts
```diff
@@ -14,13 +14,13 @@ export function trackConversion(
     pluginId: string
     orderId?: string
   }
-): ThunkAction<Promise<unknown>> {
-  return async (dispatch, getState) => {
+): ThunkAction<unknown> {
```
swansontec, line 17: While we are in here, should this be `ThunkAction<void>`? Seems weird to return `unknown`.

src/components/services/AirshipInstance.tsx
swansontec: Fair enough.

src/actions/FioActions.tsx
```diff
@@ -61,9 +62,8 @@ export function checkFioObtData(wallet: EdgeCurrencyWallet, transactions: EdgeTr
     const fioWallets = state.ui.wallets.fioWallets
     if (account == null || account.currencyConfig == null || fioWallets.length === 0) {
       checkFioObtDataCounter++
-      return setTimeout(() => {
-        dispatch(checkFioObtData(wallet, transactions))
-      }, 400)
+      await snooze(400)
+      await dispatch(checkFioObtData(wallet, transactions))
```
swansontec, line 66: This will will be an infinite recursion. In the old version, they had a `return` statement to break the loop, but that's missing here.

Instead, try:

```ts
let account: EdgeAccount
let FioWallets: string[]

// Loop until the account and wallets appear (the app may not have loaded them yet):
let loopCount = 0
while (true) {
  account = state.core.account
  fioWallets = state.ui.wallets.fioWallets
  if (account != null && account.currencyConfig != null && fioWallets.length >0) break
  if (loopCount++ > MAX_OBT_DATA_CHECKS) return
  await snooze(400)
}
```

swansontec (changes requested):
This looks much, much better! I am really happy with this approach.

Unfortunately, I did notice a missing return statement in the OBT check, so we need to get that fixed before merging.

swansontec (approved):

PR #4284, opened 2023-06-22
Fix transaction list resync issue

samholmes:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204884077340437

### DIFF
--- a/src/hooks/useTransactionList.ts
+++ b/src/hooks/useTransactionList.ts
@@ -43,9 +43,13 @@ export function useTransactionList(wallet: EdgeCurrencyWallet, tokenId: string |
     // Sends the mutable state to React,
     // merging the two transaction lists together:
     function requestRender() {
+      const mergedTxs = txs.filter(tx => !changedTxs.has(tx.txid))
+      mergedTxs.push(...changedTxs.values())
+      mergedTxs.sort((a, b) => b.date - a.date)
+
       setOutput({
         atEnd,
-        transactions: txs.map(tx => changedTxs.get(tx.txid) ?? tx)
+        transactions: mergedTxs
       })
     }
     requestRender()


src/hooks/useTransactionList.ts
```diff
@@ -43,9 +43,38 @@ export function useTransactionList(wallet: EdgeCurrencyWallet, tokenId: string |
...
+            replaceCount = 1
+          }
+          // Keep updating the insert location as we search
+          insertIndex = index
+        }
+        // Insert changed transaction
+        txs.splice(insertIndex, replaceCount, changedTx)
+        // Remove it from the changedTxs map because we've committed to txs
+        changedTxs.delete(txid)
+      }
```
swansontec: Can we factor this into a standalone utility at the end of the file, outside the hook? Something like `function binarySearch(txs: EdgeTransaction[], tx: EdgeTransaction): number { ... }`.

Then the logic would be:

```ts
const location = binarySearch(txs, changedTx)
const replaceCount = txs[location].txid === changedTx.txid ? 1 : 0
txs.splice(insertIndex, replaceCount, changedTx)
```

I think this would be more readable.

Alternatively, we could just leverage the set operations, like `new Set([...txs, ...changeTxs])` to deduplicate everything, then do an `Array.sort` based on the date to put it all back in order. The whole algorithm would be like 4 lines long, but definitely not as slick or as fast as what you have:

```ts
const noDupes = txs.filter(tx => changedTxs.has(tx.txid))
noDupes.push(...changedTxs)
return { 
  transactions: noDupes.sort((a, b) => a.date - b.date)
} 
```

swansontec (approved):
Approved, since it will definitely work, but I have some ideas about how to make it cleaner.

src/hooks/useTransactionList.ts
samholmes: I woke up with this thought to make it it's own utility.

PR #4288, opened 2023-06-24
Add more promo filters

paullinator:
### CHANGELOG

Add app version and country exclusion to promo card filters

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204739577599913

### DIFF
--- a/src/actions/AccountReferralActions.ts
+++ b/src/actions/AccountReferralActions.ts
@@ -2,7 +2,7 @@ import { asArray, asBoolean, asDate, asObject, asOptional, asString } from 'clea
 import { EdgeDataStore } from 'edge-core-js'
 import { EdgeAccount } from 'edge-core-js/types'
 import { Platform } from 'react-native'
-import { getBuildNumber } from 'react-native-device-info'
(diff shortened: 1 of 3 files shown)


src/actions/AccountReferralActions.ts
```diff
@@ -258,13 +259,25 @@ export async function validatePromoCardsInner(dataStore: EdgeDataStore, cards: M
       if (card.maxBuildNum < buildNum) continue
     }
 
+    const version = funcs.getVersion()
+    if (typeof card.version === 'string') {
+      if (card.version !== version) continue
+    }
```
swansontec, line 265: Shouldn't `fungs.getVersion` be called inside the `if` statement? Otherwise, we don't actually care what the version is.

swansontec (approved):
Approved, with one optional change.

src/actions/AccountReferralActions.ts
paullinator: yep. that's fair. will change

PR #4297, opened 2023-06-27
Fix more floating-promises

samholmes:
### CHANGELOG

- Fixed: Enabled no-floating-promises ESLint rule

### Dependencies

- https://github.com/EdgeApp/edge-react-gui/pull/4296

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204887633075462


src/util/utils.ts
```diff
@@ -297,40 +297,23 @@ export const getYesterdayDateRoundDownHour = () => {
 type AsyncFunction = () => Promise<any>
 
 export async function asyncWaterfall(asyncFuncs: AsyncFunction[], timeoutMs: number = 5000): Promise<any> {
```
swansontec, line 299: Change of plans: Let's just ESLint ignore this function for now. It's tricky.

---

This `asyncWaterfall` refactor seems completely wrong. Since each function is (presumably) a network fetch, we don't want to invoke them unless necessary. This means a simple `Promise.race` situation won't work - we need some way to know if the first promise has succeeded, so we can avoid triggering the subsequent fetches in the first place.

There may still be a way to incorporate `Promise.race` as the core of the algorithm, if we use timeouts and flags to coordinate the different "threads" of execution, so we shut down leftover timers when we are done, and avoid triggering fetches we don't need.

I would like to see `asyncWaterfall` implemented in its own PR, along with tests, in one of the currency repos. Then we can copy-paste it over to the GUI.

src/actions/WalletActions.tsx
```diff
@@ -146,7 +146,7 @@ function selectEOSWallet(navigation: NavigationBase, walletId: string, currencyC
           message={sprintf(lstrings.create_wallet_account_unfinished_activation_message, currencyCode)}
           buttons={{ ok: { label: lstrings.string_ok } }}
         />
-      ))
+      )).catch(err => console.warn(err))
```
swansontec, line 149: For the `Airship.show` cases, I think we should default to `showError`, because we really don't want this guy to crash. Otherwise the user will be scratching their head about why the UI didn't appear.

src/actions/WalletActions.tsx
```diff
@@ -277,7 +277,7 @@ const activateWalletTokens = async (
               message={sprintf(msg, feeString)}
               buttons={{ ok: { label: lstrings.string_ok } }}
             />
-          ))
+          )).catch(err => console.warn(err))
```
swansontec, line 280: showError

src/components/Main.tsx
```diff
@@ -263,9 +262,11 @@ const EdgeApp = () => {
...
     }
     backPressedOnce.current = true
-    Airship.show(bridge => <AirshipToast bridge={bridge} message={lstrings.back_button_tap_again_to_exit} />).then(() => {
-      backPressedOnce.current = false
-    })
+    Airship.show(bridge => <AirshipToast bridge={bridge} message={lstrings.back_button_tap_again_to_exit} />)
+      .then(() => {
+        backPressedOnce.current = false
+      })
+      .catch(err => console.error(err))
```
swansontec, line 269: showError

src/components/Main.tsx
```diff
@@ -711,7 +712,9 @@ const EdgeAppStack = () => {
           title: lstrings.title_settings
         }}
         listeners={{
-          focus: () => dispatch(showReEnableOtpModal())
+          focus: () => {
+            dispatch(showReEnableOtpModal()).catch(err => console.warn(err))
```
swansontec, line 716: showError

swansontec (changes requested):
These are just the first few commits. We have some recurring changes that need to be addressed, and then I will review the whole thing again.

samholmes:
I believed I've resolved all the `console.warn` instances and reverted the `asyncWaterfall` changes.

src/components/scenes/TransactionDetailsScene.tsx
```diff
@@ -93,25 +93,25 @@ class TransactionDetailsComponent extends React.Component<Props, State> {
...
+  openPersonInput = async () => {
     const personLabel = this.state.direction === 'receive' ? lstrings.transaction_details_payer : lstrings.transaction_details_payee
-    Airship.show<ContactModalResult | undefined>(bridge => <ContactListModal bridge={bridge} contactType={personLabel} contactName={this.state.name} />).then(
-      person => {
-        if (person != null) this.onSaveTxDetails({ name: person.contactName })
-      }
-    )
+    await Airship.show<ContactModalResult | undefined>(bridge => (
+      <ContactListModal bridge={bridge} contactType={personLabel} contactName={this.state.name} />
+    )).then(person => {
```
swansontec, line 100: `const person = await Airship.show(...)`

src/components/scenes/TransactionDetailsScene.tsx
```diff
@@ -93,25 +93,25 @@ class TransactionDetailsComponent extends React.Component<Props, State> {
...
   }
 
-  openCategoryInput = () => {
+  openCategoryInput = async () => {
     const { category } = this.state
-    Airship.show<string | undefined>(bridge => <CategoryModal bridge={bridge} initialCategory={category} />).then(async category => {
+    await Airship.show<string | undefined>(bridge => <CategoryModal bridge={bridge} initialCategory={category} />).then(async category => {
       if (category == null) return
       this.onSaveTxDetails({ category })
     })
```
swansontec, line 110: `const category = await Airship.show()`

src/components/scenes/TransactionDetailsScene.tsx
```diff
@@ -93,25 +93,25 @@ class TransactionDetailsComponent extends React.Component<Props, State> {
...
   }
 
-  openNotesInput = () => {
-    Airship.show<string | undefined>(bridge => (
+  openNotesInput = async () => {
+    await Airship.show<string | undefined>(bridge => (
```
swansontec, line 114: `const notes = await Airship.show(...)`

src/components/scenes/TransactionDetailsScene.tsx
```diff
@@ -165,7 +165,7 @@ class TransactionDetailsComponent extends React.Component<Props, State> {
 
     Airship.show(bridge => (
       <AdvancedDetailsModal bridge={bridge} transaction={edgeTransaction} url={sprintf(wallet.currencyInfo.transactionExplorer, edgeTransaction.txid)} />
-    ))
+    )).catch(err => showError(err))
```
swansontec, line 168: Drop the `catch`, and just `await` the `Airship.show`

src/components/scenes/TransactionsExportScene.tsx
```diff
@@ -127,16 +127,16 @@ class TransactionsExportSceneComponent extends React.PureComponent<Props, State>
     )
   }
 
-  handleStartDate = () => {
+  handleStartDate = async () => {
     const { startDate } = this.state
-    Airship.show<Date>(bridge => <DateModal bridge={bridge} initialValue={startDate} />).then(date => {
+    await Airship.show<Date>(bridge => <DateModal bridge={bridge} initialValue={startDate} />).then(date => {
       this.setState({ startDate: date })
     })
```
swansontec, line 134: `const date = await Airship.show(...)`

src/components/scenes/TransactionsExportScene.tsx
```diff
@@ -127,16 +127,16 @@ class TransactionsExportSceneComponent extends React.PureComponent<Props, State>
...
   }
 
-  handleEndDate = () => {
+  handleEndDate = async () => {
     const { endDate } = this.state
-    Airship.show<Date>(bridge => <DateModal bridge={bridge} initialValue={endDate} />).then(date => {
+    await Airship.show<Date>(bridge => <DateModal bridge={bridge} initialValue={endDate} />).then(date => {
```
swansontec, line 139: `const date = await Airship.show(...)`

src/components/scenes/WcConnectScene.tsx
```diff
@@ -61,8 +62,8 @@ export const WcConnectScene = (props: Props) => {
     return { subTitleText, bodyTitleText, dAppImage }
   }, [proposal])
 
-  React.useEffect(() => {
-    wallet.getReceiveAddress().then(r => setWalletAddress(r.publicAddress))
+  useAsyncEffect(async () => {
+    await wallet.getReceiveAddress().then(r => setWalletAddress(r.publicAddress))
```
swansontec, line 66:
```ts
const r = await wallet.getReceiveAddress()
setWalletAddress(r.publicAddress)
```

Or just use `useAsyncValue`

src/components/scenes/WcConnectScene.tsx
```diff
@@ -71,29 +72,29 @@ export const WcConnectScene = (props: Props) => {
...
     }
   }
 
-  const handleWalletListModal = useHandler(() => {
-    Airship.show<WalletListResult>(bridge => (
+  const handleWalletListModal = useHandler(async () => {
+    await Airship.show<WalletListResult>(bridge => (
       <WalletListModal bridge={bridge} headerTitle={lstrings.select_wallet} allowedAssets={edgeTokenIds} navigation={navigation} />
-    )).then(({ walletId, currencyCode }: WalletListResult) => {
+    )).then(async ({ walletId, currencyCode }: WalletListResult) => {
```
swansontec, line 86:
```ts
const result = await Airship.show(...)
const { walletId, currencyCode } = result
...
```

src/components/scenes/WcConnectionsScene.tsx
```diff
@@ -67,25 +67,21 @@ export const WcConnectionsScene = (props: Props) => {
...
-        if (result != null) {
-          onScanSuccess(result)
-        } else {
-          showError('No scan result')
-        }
-      })
-      .catch(error => {
-        showError(error)
-      })
+    )).then(async (result: string | undefined) => {
```
swansontec, line 78: `const result = await Airship.show(...)`

src/components/services/EdgeContextCallbackManager.tsx
```diff
@@ -39,9 +39,11 @@ class EdgeContextCallbackManagerComponent extends React.Component<Props> {
...
         if (!errorShown) {
           errorShown = true
-          this.showError(error).then(() => {
-            errorShown = false
-          })
+          this.showError(error)
+            .then(() => {
+              errorShown = false
+            })
+            .catch(err => showError(err))
```
swansontec, line 46: The `this.showError` method should return `void` instead of `Promise<void>`, and should do it's own catching. That way, it works more like our other logging code.

src/components/themed/SelectFioAddress.tsx
```diff
@@ -107,26 +107,26 @@ export class SelectFioAddressComponent extends React.PureComponent<Props, LocalS
...
     }
   }
 
-  selectAddress = () => {
+  selectAddress = async () => {
     const { currencyCode, selectedWalletId } = this.props
-    Airship.show<string | undefined>(bridge => (
+    await Airship.show<string | undefined>(bridge => (
       <AddressModal bridge={bridge} title={lstrings.fio_select_address} currencyCode={currencyCode} walletId={selectedWalletId} useUserFioAddressesOnly />
-    )).then(response => {
```
swansontec, line 121: `const response = await Airship.show(...)`

src/components/themed/SelectFioAddress.tsx
```diff
@@ -107,26 +107,26 @@ export class SelectFioAddressComponent extends React.PureComponent<Props, LocalS
...
+        await this.setFioAddress(response)
       }
     })
   }
 
-  openMessageInput = () => {
-    Airship.show<string | undefined>(bridge => (
+  openMessageInput = async () => {
+    await Airship.show<string | undefined>(bridge => (
       <TextInputModal
```
swansontec, line 130: `const memo = await Airship.show(...)`

src/components/themed/SelectFioAddress2.tsx
```diff
@@ -106,26 +106,26 @@ export class SelectFioAddressComponent extends React.PureComponent<Props, LocalS
...
   }
 
-  selectAddress = () => {
+  selectAddress = async () => {
     const { currencyCode, coreWallet } = this.props
-    Airship.show<string | undefined>(bridge => (
+    await Airship.show<string | undefined>(bridge => (
       <AddressModal bridge={bridge} title={lstrings.fio_select_address} currencyCode={currencyCode} walletId={coreWallet.id} useUserFioAddressesOnly />
-    )).then(response => {
+    )).then(async response => {
```
swansontec, line 120: `const response = await Airship.show(...)`

src/components/themed/SelectFioAddress2.tsx
```diff
@@ -106,26 +106,26 @@ export class SelectFioAddressComponent extends React.PureComponent<Props, LocalS
...
-        this.setFioAddress(response)
+        await this.setFioAddress(response)
       }
     })
   }
 
-  openMessageInput = () => {
-    Airship.show<string | undefined>(bridge => (
+  openMessageInput = async () => {
+    await Airship.show<string | undefined>(bridge => (
```
swansontec, line 128: `const memo = await Airship.show(...)`

swansontec (changes requested):
This looks pretty good now. Still a bunch of things to look at, though.

src/components/scenes/Staking/StakeModifyScene.tsx
```diff
@@ -110,10 +110,16 @@ const StakeModifySceneComponent = (props: Props) => {
...
-              })
+              wallet
+                .nativeToDenomination(nativeMin, currencyCode)
+                .then(minExchangeAmount => {
+                  errMessage += `: ${minExchangeAmount} ${currencyCode}`
+                  setErrorMessage(errMessage)
+                })
+                .catch(err => {
+                  showError(err)
+                  setErrorMessage(errMessage)
```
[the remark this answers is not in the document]
samholmes: If `nativeToDenomination` fails, we see the extra `showError` which shouldn't realistically happen unless there's a bug in function. I'd say we should leave it in to be noisy in this case to catch regressions at least by QA.

src/components/scenes/TransactionDetailsScene.tsx
samholmes: `openAdvancedDetails` is used by a non-async callback event. I removed the async keyword instead.

src/components/services/EdgeContextCallbackManager.tsx
samholmes: But then we'd remove the logic to only show one error every 5 seconds (duration of the dropdown).

src/components/scenes/Staking/StakeModifyScene.tsx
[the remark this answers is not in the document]
swansontec: Ok

src/components/services/EdgeContextCallbackManager.tsx
swansontec: In that case, maybe the singleton logic should live inside of `this.showError` too.

swansontec (approved):
Close enough. Let's merge it.

PR #4300, opened 2023-06-29
Fix showError to submit Error object to Bugsnag

samholmes:
Instead of sending the string to Bugsnag, we send the entire error
object so we retain the stack trace.

We will fallback to using `translateError` for unknown data types.

### CHANGELOG

- Fixed: Bugsnag error reporting for showError includes error stack trace

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1204848238190311

### DIFF
--- a/src/components/services/AirshipInstance.tsx
+++ b/src/components/services/AirshipInstance.tsx
@@ -19,12 +19,20 @@ export interface ShowErrorWarningOptions {
  */
 export function showError(error: unknown, options: ShowErrorWarningOptions = {}): void {
   const { trackError = true, tag } = options
-  const translatedError = tag ? `Tag: ${tag}. ` + translateError(error) : translateError(error)
+  const tagMessage = tag == null ? '' : `Tag: ${tag}. `
+  const translatedMessage = tagMessage + translateError(error)
   if (trackError) {
-    Bugsnag.notify(`showError: ${translatedError}`)
+    if (error instanceof Error) {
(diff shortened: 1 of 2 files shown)


src/components/services/AirshipInstance.tsx
```diff
@@ -19,9 +19,20 @@ export interface ShowErrorWarningOptions {
...
 export function showError(error: unknown, options: ShowErrorWarningOptions = {}): void {
   const { trackError = true, tag } = options
-  const translatedError = tag ? `Tag: ${tag}. ` + translateError(error) : translateError(error)
+  const translatedError = translateError(error)
   if (trackError) {
-    Bugsnag.notify(`showError: ${translatedError}`)
+    const tagMessage = tag == null ? '' : `Tag: ${tag}. `
+    let bugsnagError: Error | string
+    if (error instanceof Error) {
+      bugsnagError = error
```
paullinator, line 27: This case should also attempt a translateError and tack on the result to the tagMessage.

paullinator (changes requested):

src/components/services/AirshipInstance.tsx
samholmes: I tack on the tag to the `message` property of the error below. I don't want to run `translateError` on error objects because it returns a string then we'd lose the stack property of the error.

paullinator (approved):

PR #4302, opened 2023-06-30
Paul/fix fio crash

paullinator:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204909366169318


swansontec (approved):
Nice PR. The `withWallet` is the right solution, but I'm happy to see the extra cleanups being taken care of as well.

PR #4310, opened 2023-07-05
Wallet connect cleanups

peachbits:
---
  - https://app.asana.com/0/0/1204965986448038


src/components/services/WalletConnectService.tsx
```diff
@@ -7,47 +7,13 @@ import { EdgeAccount } from 'edge-core-js'
...
-import { useMount } from '../../hooks/useMount'
+import { useAsyncEffect } from '../../hooks/useAsyncEffect'
 import { getAccounts, getWalletIdFromSessionNamespace } from '../../hooks/useWalletConnect'
 import { useWatch } from '../../hooks/useWatch'
 import { WcSmartContractModal } from '../modals/WcSmartContractModal'
 import { Airship } from '../services/AirshipInstance'
 
-let walletConnectRef: Web3Wallet | undefined
-
-export const walletConnectPromise: Promise<Web3Wallet> = new Promise((resolve, reject) => {
```
swansontec, line 18: I think the `getClient` method should live in this file, and should be async.

Then we can implement `getClient` as follows:

```ts
const getClient = async () => {
  if (walletConnectClient != null) return walletConnectClient
  return new Promise(resolve => waitingClients.push(resolve))
}

const waitingClients: Array<(client: Web3Wallet) => void> = []
```

Basically, if we have a client, return it. If we don't have a client, add a callback to the list and wait around for it to show up.

Then in your `useAsyncEffect`, you can do `waitingClients.forEach(f => f(walletConnectClient))` to un-block all the waiting code.

src/hooks/useWalletConnect.tsx
```diff
@@ -144,14 +147,26 @@ export function useWalletConnect(): WalletConnect {
...
+          client
+            .disconnectSession({ topic, reason: getSdkError('USER_DISCONNECTED') })
+            .then(() => {
+              Airship.show(bridge => <FlashNotification bridge={bridge} message={sprintf(lstrings.wc_dapp_disconnected, dAppName)} onPress={() => {}} />)
+              resolve()
+            })
+            .catch(e => reject(e))
+        }),
+        10000
+      )
```
swansontec, line 174: You don't need the `new Promise` bit in here:

```ts
await runWithTimeout(
  client.disconnectSession({ topic, reason: getSdkError('USER_DISCONNECTED') }),
  10000
)
await Airship.show(bridge => <FlashNotification bridge={bridge} message={sprintf(lstrings.wc_dapp_disconnected, dAppName)} onPress={() => {}} />)
```

src/hooks/useWalletConnect.tsx
```diff
@@ -126,13 +126,24 @@ export function useWalletConnect(): WalletConnect {
...
+          })
+
+          .then(() => {
+            resolve()
+          })
+          .catch(e => reject(e))
+      }),
+      10000
+    )
   })
```
swansontec, line 147: Same as before.

You don't need the `new Promise` bit in here:

```ts
await runWithTimeout(
  client.approveSession({
            id: proposal.id,
            namespaces: buildApprovedNamespaces({
              proposal: proposal.params,
              supportedNamespaces
            })
   })
)
```

swansontec (changes requested):

swansontec (approved):

PR #4315, opened 2023-07-05
Add key to wallet sync circle in the cryptoIcon component

peachbits:
This fixes a visual issue in flashlists

---
  - https://app.asana.com/0/0/1204622252075359

### DIFF
--- a/src/components/icons/CryptoIcon.tsx
+++ b/src/components/icons/CryptoIcon.tsx
@@ -101,7 +101,14 @@ const CryptoIconComponent = (props: Props) => {
 
   return (
     <View style={spacingStyle}>
-      {wallet == null ? null : <WalletSyncCircle size={size} wallet={wallet} />}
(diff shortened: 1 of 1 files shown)


src/components/icons/CryptoIcon.tsx
```diff
@@ -101,7 +101,7 @@ const CryptoIconComponent = (props: Props) => {
 
   return (
     <View style={spacingStyle}>
-      {wallet == null ? null : <WalletSyncCircle size={size} wallet={wallet} />}
+      {wallet == null ? null : <WalletSyncCircle key={`${wallet.id}${String(tokenId)}`} size={size} wallet={wallet} />}
```
swansontec, line 104: I would add a comment to this key, so we remember why it's here:

`// Re-start the animation if the row gets recycled`

swansontec (approved):

PR #4316, opened 2023-07-06
Wallet Connect tweaks

peachbits:
Fix fail to connect if UI is behind on slow/large accounts
Always show disconnect drop down when user initiates it

---
  - https://app.asana.com/0/0/1204975182129323


src/hooks/useWalletConnect.tsx
```diff
@@ -147,14 +144,8 @@ export function useWalletConnect(): WalletConnect {
...
-    try {
-      await runWithTimeout(client.disconnectSession({ topic, reason: getSdkError('USER_DISCONNECTED') }), 10000)
-    } catch (e) {
-      // In testing, this method is pretty unreliable so we can at least remove it locally.
-      console.log('walletConnect disconnectSession error', String(e))
-      client.core.relayer.subscriber.topicMap.delete(topic)
-    }
-    userDeletedSessions.add(topic)
+    // In testing, this method is pretty unreliable. May be worth replacing with something more manual.
+    await runWithTimeout(client.disconnectSession({ topic, reason: getSdkError('USER_DISCONNECTED') }), 10000)
```
swansontec, line 148: Maybe increase the timeout to 20s as in the previous commit.

swansontec (approved):
Approved with optional timeout increase for disconnection.

src/hooks/useWalletConnect.tsx
peachbits: In my testing, the dapp responds pretty quickly but the sdk hangs for an unknown reason. I'd rather it fail sooner so the user can initiate it again if they want.

PR #4325, opened 2023-07-11
Correctly track changes to the account status

swansontec:
We weren't re-rendering correctly when the account logged in, so our previous login-scene state got left behind.

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205003154684908


samholmes (approved):

samholmes:
Approved. Curious as to how this was a 3.13 regression though, if this was only a issue with that line.. That ternary operator must be new to 3.13?

If it’s not a regression to 3.13, then does this warrant an entry in the CHANGELOG?

PR #4333, opened 2023-07-14
Jon/add-token-done-button

Jon-edge:
This PR adds a ButtonsContainer component that aims to enforce consistency across our app for button layouts. While the following shows both row and column button layouts, **the row layout (right) was ultimately used for this change.**

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204787223052337

### DIFF
--- a/src/components/buttons/ButtonsContainer.tsx
+++ b/src/components/buttons/ButtonsContainer.tsx
@@ -1,0 +1,46 @@
+import * as React from 'react'
+import { View } from 'react-native'
+
+import { styled } from '../hoc/styled'
+import { MainButton, MainButtonType } from '../themed/MainButton'
+
+export interface ButtonInfo {
+  label: string
+  onPress: () => void | Promise<void>
+  disabled?: boolean
+}
+
+interface Props {
+  primary?: ButtonInfo
+  secondary?: ButtonInfo
+  escape?: ButtonInfo
+  layout: 'row' | 'column'
+}
+
+/**
+ * A consistently styled view for displaying button layouts.
+ */
+export const ButtonsContainer = React.memo(({ primary, secondary, escape, layout = 'column' }: Props) => {
+  const renderButton = (type: MainButtonType, buttonProps?: ButtonInfo) => {
+    if (buttonProps == null) return null
+    const { label, onPress, disabled } = buttonProps
+    return <MainButton label={label} onPress={onPress} type={type} marginRem={0.5} disabled={disabled} />
+  }
+
+  return (
+    <StyledButtonContainer layout={layout}>
+      {renderButton('primary', primary)}
+      {renderButton('secondary', secondary)}
+      {renderButton('escape', escape)}
+    </StyledButtonContainer>
+  )
+})
+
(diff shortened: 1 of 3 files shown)


src/components/buttons/ButtonsContainer.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { View } from 'react-native'
+
+import { styled } from '../hoc/styled'
+import { MainButton, MainButtonType } from '../themed/MainButton'
+
+export interface ButtonContainerProps {
+  label: string
+  onPress: () => void | Promise<void>
+  disabled?: boolean
+}
```
swansontec, line 11: Call this `ButtonInfo` or something. These aren't the top-level parameters a component expects, so we can't call them `Props`.

src/components/buttons/ButtonsContainer.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+interface Props {
+  buttons: Map<MainButtonType, ButtonContainerProps>
+  layout: 'row' | 'column'
```
swansontec, line 15: Get rid of the map, and instead do `primary?: ButtonInfo; secondary?: ButtonInfo; escape?: ButtonInfo`.

Since we know the keys this component expects, we should not use a map, but a regular static object. Maps are for arbitrary key / value setups like loginId -> last login or something.

src/components/buttons/ButtonsContainer.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/**
+ * A consistently styled view for displaying button layouts.
+ */
+export const ButtonsContainer = React.memo(({ buttons, layout = 'column' }: Props) => {
```
swansontec, line 21: The big question: Can we use this for ButtonsModal, or not? If so, then we should do that. If not, then what do we need to do to get there? Your comment indicates that we want this to be the goal.

swansontec (changes requested):

src/components/buttons/ButtonsContainer.tsx
Jon-edge: Yeah that was the motivation - ButtonsModal and a case-by-case basis where we update the parent and ButtonsContainer gradually

src/components/buttons/ButtonsContainer.tsx
```diff
@@ -0,0 +1,56 @@
...
+})
+
+const StyledButtonContainer = styled(View)<{ layout: 'row' | 'column' }>(props => {
+  const isRow = props.layout === 'row'
+
+  return {
+    flexDirection: isRow ? 'row' : 'column',
+    justifyContent: 'space-between',
+    padding: props.theme.rem(0.5),
+    paddingHorizontal: isRow ? undefined : props.theme.rem(1)
```
swansontec, line 56: There is a layout called "row-reverse". You *could* use this instead of flipping the button order in the render function.

swansontec (approved):
Approved, but there is one change you could make.

PR #4340, opened 2023-07-18
Fix PasswordReminderModal Button Overlap

Jon-edge:
Punted task to modify OutlinedTextInput such that this bug can no longer happen: https://app.asana.com/0/0/1205078580542521/f

image

### CHANGELOG

- fixed: Password Reminder button overlap with error text on wrong password

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205013500013787

### DIFF
--- a/src/components/buttons/ButtonsContainer.tsx
+++ b/src/components/buttons/ButtonsContainer.tsx
@@ -4,6 +4,8 @@ import { View } from 'react-native'
 import { styled } from '../hoc/styled'
 import { MainButton, MainButtonType } from '../themed/MainButton'
 
+const BUTTON_MARGINS = [0.5, 0]
+
 export interface ButtonInfo {
   label: string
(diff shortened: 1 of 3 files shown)


src/components/buttons/ButtonsContainer.tsx
```diff
@@ -4,6 +4,8 @@ import { View } from 'react-native'
 import { styled } from '../hoc/styled'
 import { MainButton, MainButtonType } from '../themed/MainButton'
 
+const BUTTON_MARGINS = [0.5, 0, 0.5, 0]
```
swansontec, line 7: This works like the web, so I think `[0.5, 0]` would do the same thing. On the web, a 2-item array means vertical / horizontal, as opposed to giving each side separately.

swansontec (approved):

PR #4342, opened 2023-07-18
Fix text input

swansontec:
The "suffix" feature on the FIO registration scene works by making the text input tiny. As the user enters text, the input gets wider, and the suffix moves to the right.

This means that the text input itself is tiny to start, so there is no tappable area for a user to long-press on. This is unacceptable, so the FIO effect needs to go. We can still keep the suffix, just not attached to the text:

simulator_screenshot_39BC1F00-A76E-41DC-96DC-595DFDE78E0B

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205064445212746

### DIFF
--- a/src/components/themed/OutlinedTextInput.tsx
+++ b/src/components/themed/OutlinedTextInput.tsx
@@ -212,10 +212,6 @@ export const OutlinedTextInput = React.forwardRef<OutlinedTextInputRef, Outlined
     paddingRight: clearIcon ? theme.rem(2.875) : theme.rem(1)
   }
 
-  const suffixPadding = {
-    paddingRight: clearIcon ? theme.rem(2.875) : theme.rem(1)
-  }
-
   const containerStyle = {
     ...containerPadding,
     ...sidesToMargin(mapSides(fixSides(marginRem, 0.5), theme.rem)),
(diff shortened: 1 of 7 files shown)


src/components/themed/OutlinedTextInput.tsx
```diff
@@ -326,12 +319,14 @@ export const OutlinedTextInput = React.forwardRef<OutlinedTextInputRef, Outlined
   const charLimitLabel = maxLength === undefined ? '' : `${maxLength - value.length}`
 
   const numpad = props.keyboardType === 'decimal-pad'
-  const textStyle = numpad ? styles.numberInput : styles.textInput
-  const suffixStyle = React.useMemo(() => [styles.suffixText, Platform.OS === 'android' ? styles.suffixAndroidAdjust : null], [styles])
```
Jon-edge, line 333: Does this work properly on android?

Jon-edge (approved):
Approved if it looks right without the androidAdjust style

src/components/themed/OutlinedTextInput.tsx
swansontec: After extensive testing, I got it to work pretty well on both platforms.

Jon-edge:
Did a little digging hoping there might be some magic way to use a TouchableOpacity long tap that covers the suffix and everything else to the right, but seems like it would be too involved for now. Maybe sometime after we consolidate the OTI

PR #4348, opened 2023-07-19
Fix missing version suffix

paullinator:
Have updateVersion.ts read the version file written by jenkins. gitVersionFile.ts is responsible for writing the file.

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device


swansontec (approved):
I see that all the logic from this file has moved into `gitVersionFile.ts`, so the change solves the immediate problem.

Unfortunately, this new architecture means there is no longer any *documented* way for third parties to build release versions of the app. We spent a lot of time and energ a year or two ago making sure there was a way to build the app *without* all our custom Jenkins infrastructure. Part of this design was the `updateVersions.ts`, which would auto-update the version info, independently of Jenkins. This is still documented in our readme, even though it no longer works correctly (especially after this change).

So, we either need to update this script with some sort of fallback mode that third parties can use, or we need to update our README with the steps someone needs to get `release-version.json` up-to-date.

Feel free to create a task for this work, based on your chosen path & time estimate.

paullinator:
> I see that all the logic from this file has moved into `gitVersionFile.ts`, so the change solves the immediate problem.
> 
> Unfortunately, this new architecture means there is no longer any _documented_ way for third parties to build release versions of the app. We spent a lot of time and energ a year or two ago making sure there was a way to build the app _without_ all our custom Jenkins infrastructure. Part of this design was the `updateVersions.ts`, which would auto-update the version info, independently of Jenkins. This is still documented in our README, even though it no longer works correctly (especially after this change).
> 
> So, we either need to update this script with some sort of fallback mode that third parties can use, or we need to update our README with the steps someone needs to get `release-version.json` up-to-date.
> 
> Feel free to create a task for this work, based on your chosen path & time estimate.

Actually this is a pretty easy change for external developers. gitVersionFile.ts just has to be called right before updateVersion.ts. I’ll update the docs.

README.md
```diff
@@ -77,8 +77,6 @@ Otherwise, to get an APK, do:
 
 ### Build release version of app
 
-First, run `yarn update-version` to copy the `package.json` version into the native project files, and to assign a unique build number.
-
```
swansontec, line 81: We should delete the section header, now that it's empty.

swansontec (approved):
Approved with one teeny change.

README.md
paullinator: I think it's still relevant as it differentiates building a release version vs a debug version like the header above. Otherwise we'd have iOS, Android, iOS, Android all under the 

### Run the app in debug mode

header.

PR #4358, opened 2023-07-25
Fix GettingStartedScene from routing after forgetting accounts

samholmes:
If the user forgets accounts (deletes), then the `localUsers.length`
change would trigger the effect which handles the routing to the login
scene.

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1205126719735957

### DIFF
--- a/src/components/scenes/GettingStartedScene.tsx
+++ b/src/components/scenes/GettingStartedScene.tsx
@@ -23,6 +23,7 @@ import slide3HeroImage from '../../assets/images/gettingStarted/slide3HeroImage.
 import slide4HeroImage from '../../assets/images/gettingStarted/slide4HeroImage.png'
 import { useAsyncEffect } from '../../hooks/useAsyncEffect'
 import { useHandler } from '../../hooks/useHandler'
+import { useWatch } from '../../hooks/useWatch'
 import { lstrings } from '../../locales/strings'
 import { useSelector } from '../../types/reactRedux'
 import { EdgeSceneProps } from '../../types/routerTypes'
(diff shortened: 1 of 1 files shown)


src/components/scenes/GettingStartedScene.tsx
```diff
@@ -80,7 +80,7 @@ const REMOTE_CONFIG_DEFAULT = {
 
 export const GettingStartedScene = (props: Props) => {
   const { navigation } = props
-  const localUsersLength = useSelector(state => state.core.context.localUsers.length)
+  const hasLocalUsers = useSelector(state => state.core.context.localUsers.length > 0)
```
swansontec, line 83: A "more correct" solution would be:

```ts
const context= useSelector(state => state.core.context)
const localUsers = useWatch(context, 'localUsers')
const const hasLocalUsers = localUsers.length >0
```

Doing it all in the selector means it won't necessarily update correctly, especially since this is deletion-related.

swansontec (changes requested):

src/components/scenes/GettingStartedScene.tsx
samholmes: If `fn` in `useSelector(fn)` is invoked on changes to the redux state tree and `===` compared to the previous selector's return value before triggering a re-render for `GettingStartedScene`, then making the change you proposed will cause more unnecessary re-renders unless I am misunderstanding redux. Please explain if I am misunderstanding.

swansontec: The first line will never cause a re-render, because the context will remain `===` the same for the whole lifetime of the app. The only exception is if the WebView reloads, in which case we should probably render anyhow.

The second line will render any time users are added or removed, which we do want. The `useSelector` function has *no idea* when the core is going to experience these changes, so we need to subscribe to the core itself. Otherwise, if a user gets deleted without a corresponding `dispatch` (which may happen as we rely on Redux less and less), the component will never see the change and never update.

Finally, the third line implements your fix. By creating an intermediate variable, you can precisely control when the `useEffect` gets triggered based on `===` changes.

samholmes: Ah, I see now that you used `useWatch` and I missed that part. Got it. Thanks for the explanation.

swansontec (approved):

PR #4359, opened 2023-07-27
Jon/ab-light-account

Jon-edge:
### CHANGELOG

- changed: Support A/B testing for light account

### Dependencies

https://github.com/EdgeApp/edge-login-ui-rn/pull/110

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205126828151695


src/envConfig.ts
```diff
@@ -265,5 +265,6 @@ export const asEnvConfig = asObject({
       port: asOptional(asString, '8008')
     }),
     { host: 'localhost', port: '8008' }
-  )
+  ),
+  REMOTE_CONFIG_FETCH_INTERVAL: asOptional(asNumber, 12 * 60 * 60 * 1000) // 12 hours
```
paullinator, line 269: This seems VERY different from the prior default. Why 12 hours vs 1s?

fb-remote.json
```diff
@@ -0,0 +1,4 @@
+{
```
paullinator, line 1: Why is this a json file? Since it is committed, might as well be in the TS file that includes it

paullinator: If anything, this can go into env.json or appConfig.ts so we can configure it per whitelabel app

src/components/scenes/GettingStartedScene.tsx
```diff
@@ -98,27 +115,46 @@ export const GettingStartedScene = (props: Props) => {
...
+      })
+      .catch(() => {}) // handled in getRemoteConfigValue
   })
+
   const handlePressSignUp = useHandler(() => {
-    logEvent('Signup_Welcome', { variant: { doneMethod: 'click' } })
-    navigation.navigate('login', { loginUiInitialRoute: localUsersLength > 0 ? 'new-account' : 'new-light-account' })
+    getRemoteConfigValue('create_account_type')
+      .then(createAccountTypeVal => {
+        const createAccountType = String(createAccountTypeVal)
```
paullinator, line 151: This be of type CreateAccountType. See comment in login-ui PR

src/components/scenes/GettingStartedScene.tsx
```diff
@@ -90,6 +87,26 @@ export const GettingStartedScene = (props: Props) => {
   const paginationCount = sections.length + (isFinalSwipeEnabled ? 1 : 0)
   const swipeOffset = useSharedValue(0)
 
+  // Route helpers
+  const getNewAccountRoute = (createAccountType: string): InitialRouteName => {
```
paullinator, line 91: Should accept a CreateAccountType TS type, not string

paullinator (changes requested):

src/envConfig.ts
Jon-edge: The prior wasn't the default. I had it at 1s during development, I will remove the 1s altogether from the commit

src/fbRemoteConfig.ts
```diff
@@ -0,0 +1,66 @@
...
+    const stickyRemoteConfigJson = await stickyRemoteConfigDisklet.getText(REMOTE_CONFIG_STICKY)
+    stickyRemoteConfig = asFbRemoteConfig(JSON.parse(stickyRemoteConfigJson))
+  } catch (err) {
+    // Not found or mismatched. Re-generate with random values according to the
+    // defined distribution.
+    const generatedRemoteConfigSticky: FbRemoteConfig = {
+      swipeLastUsp: Math.random() < stickyDistribution.swipeLastUsp,
+      createAccountType: Math.random() < stickyDistribution.createAccountType ? 'light' : 'full'
+    }
+    const generatedJsonData = JSON.stringify(uncleaner(asFbRemoteConfig)(generatedRemoteConfigSticky))
```
paullinator, line 39: uncleaner not necessary. The type is derived from the cleaner so that ensures type compatibility

src/fbRemoteConfig.ts
```diff
@@ -0,0 +1,66 @@
...
+    await stickyRemoteConfigDisklet.setText(REMOTE_CONFIG_STICKY, generatedJsonData)
+    stickyRemoteConfig = asFbRemoteConfig(generatedRemoteConfigSticky)
```
paullinator, line 41: No need to clean the data you just created.

src/fbRemoteConfig.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+/**
+ * Returns sticky config, waiting for initialization first.
+ */
+export const getStickyRemoteConfig = async (): Promise<FbRemoteConfig> => {
+  while (true) {
+    if (stickyRemoteConfig != null) return stickyRemoteConfig
+    await setTimeout(() => {}, 10)
```
paullinator, line 51: This needs to be snooze.

src/fbRemoteConfig.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+/**
+ * Returns the sticky remote config value
+ */
+export const getStickyRemoteConfigValue = (key: keyof FbRemoteConfig): any => {
```
paullinator, line 58: This can't be used as it's not async

paullinator: All determination of defaults needs to be from the init function

src/components/scenes/GettingStartedScene.tsx
```diff
@@ -134,20 +143,6 @@ export const GettingStartedScene = (props: Props) => {
     }
   )
 
-  // Fetch and activate remote config (A/B testing).
-  useAsyncEffect(async () => {
```
paullinator, line 141: Need to return back to using useAsyncEffect as getting the config MUST be async

paullinator (changes requested):

src/fbRemoteConfig.ts
Jon-edge: We can't use snooze at this level, the app will crash

Jon-edge: Copied snooze instead

paullinator (approved):

samholmes (approved):

PR #4360, opened 2023-07-27
Remove PIN and biometric settings for light accounts

Jon-edge:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205100255666907

### DIFF
--- a/src/components/scenes/SettingsScene.tsx
+++ b/src/components/scenes/SettingsScene.tsx
@@ -282,6 +282,7 @@ export class SettingsSceneComponent extends React.Component<Props, State> {
       days: lstrings.settings_days
     }
     const autoLogoutRightText = autoLogout.value === 0 ? lstrings.string_disable : `${autoLogout.value} ${timeStrings[autoLogout.measurement]}`
+    const isLightAccount = username == null
 
     return (
       <SceneWrapper background="theme" hasTabs={false}>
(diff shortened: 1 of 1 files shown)


src/components/scenes/SettingsScene.tsx
```diff
@@ -313,8 +314,10 @@ export class SettingsSceneComponent extends React.Component<Props, State> {
           <SettingsLabelRow right={autoLogoutRightText} label={lstrings.settings_title_auto_logoff} onPress={this.handleAutoLogout} />
           <SettingsLabelRow right={this.props.defaultFiat.replace('iso:', '')} label={lstrings.settings_title_currency} onPress={this.handleDefaultFiat} />
 
-          <SettingsSwitchRow key="pinRelogin" label={lstrings.settings_title_pin_login} value={this.props.pinLoginEnabled} onPress={this.handlePinToggle} />
-          {this.props.supportsTouchId && (
+          {!isLightAccount ? (
```
peachbits, line 317: Convention is to switch this so the null case is first

peachbits (changes requested):

src/components/scenes/SettingsScene.tsx
Jon-edge: Where do we have such a convention documented? Weird having to think of a flipped condition to visualize what would appear...

Jon-edge: Having it how I originally had it with null case last also aligns with the logical conditions of nullish coalescing syntax (seen elsewhere in the file).

peachbits: May be documented in past meeting notes. It's fresh in my mind since I've been recently asked to make the same change.

peachbits (approved):

PR #4366, opened 2023-07-28
Implement 'Contacts Access' permission settings row

Jon-edge:
image

### CHANGELOG

- changed: Add a 'Contacts Access' permission settings row

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203556417756042


src/components/scenes/SettingsScene.tsx
```diff
@@ -155,6 +165,33 @@ export class SettingsSceneComponent extends React.Component<Props, State> {
     this.props.isLocked ? this.handleUnlock() : navigation.navigate('passwordRecovery', {})
   }
 
+  handleContactsAccess = (): void => {
+    if (this.state.contactsPermission) {
+      // App will restart after denying the permission
+      openSettings().catch(e => console.error(e))
```
swansontec, line 171: As mentioned in the engineering meetings, `console` error handling is banned. Only `showError` is acceptable, unless you have a documented reason to do it otherwise.

src/components/scenes/SettingsScene.tsx
```diff
@@ -155,6 +165,33 @@ export class SettingsSceneComponent extends React.Component<Props, State> {
...
+        buttonNegative: lstrings.string_deny
+      })
+        .then((permissionStatus: PermissionStatus) => {
+          // Can't request permission from within the app if blocked
+          if (permissionStatus === 'blocked') openSettings().catch(e => console.error(e))
+          else this.setState({ contactsPermission: permissionStatus === 'granted' })
+        })
+        // Handle any other potential failure in enabling the permission by
+        // redirecting to the system settings
+        .catch(async e => await openSettings().catch(e => console.error(e)))
```
swansontec, line 188: `showError`

src/components/scenes/SettingsScene.tsx
```diff
@@ -155,6 +165,33 @@ export class SettingsSceneComponent extends React.Component<Props, State> {
...
+    }
+
+    // Refresh this settings row
+    this.loadContactsAccess().catch(e => console.error(e))
```
swansontec, line 192: `showError`

swansontec (changes requested):
If the user turns this switch off we need to stop accessing the contacts on our own. So basically, the `useContactThumbnail` hook should check whether this row is turned on, and then always return `undefined` if so.

If the user tries to turn the switch on, and we have permissions, then simply update the setting that `useContactThumbnail` reads, so it starts returning images again. If the permission is denied, then of course we need to visit the phone settings. If we haven't asked yet, then we need to ask for the first time.

Jon-edge:
The only new verbiage is "Contacts Access" since the original task you said "use contacts for auto-complete," which didn't make sense to multiple people including myself. Also there was a real-estate issue.

src/actions/SettingsActions.tsx
```diff
@@ -298,3 +303,41 @@ export function setSpamFilterOn(spamFilterOn: boolean): ThunkAction<void> {
...
 }
+
+/**
+ * Toggle the 'Contacts Access' Edge setting. Will request permissions if
+ * toggled on/enabled AND system-level contacts permissions are not granted.
+ * Does NOT modify system-level contacts permissions if toggling the 'Contacts
+ * Access' setting OFF
+ */
+export function setContactsPermissionOn(contactsPermissionOn: boolean): ThunkAction<void> {
+  return (dispatch, getState) => {
```
swansontec, line 314: If you make this function `async (dispatch, getState)`, then you could clean up a huge amount of error handling. You would need to add a single `.catch(showError)` on the settings scene to handle the whole mess.

src/components/modals/ContactListModal.tsx
```diff
@@ -58,8 +59,12 @@ export function ContactListModal({ bridge, contactType, contactName }: Props) {
   const handleSubmitEditing = (contactName: string) => bridge.resolve({ contactName, thumbnailPath: null })
 
   React.useEffect(() => {
-    edgeRequestPermission('contacts').catch(showError)
-  }, [])
+    maybeShowContactsPermissionModal(dispatch, contactsPermissionOn)
+
+    // Avoid popping up the modal when component is mounted and the user changes
+    // contactsPermissionOn.
```
swansontec, line 65: Good comment! It clearly explains the need for the lint exception.

swansontec (changes requested):
The last two commits in this PR contain some old stuff, that then gets reverted. Please squash the last two commits into one.

src/actions/SettingsActions.tsx
```diff
@@ -310,34 +310,33 @@ export function setSpamFilterOn(spamFilterOn: boolean): ThunkAction<void> {
  * Does NOT modify system-level contacts permissions if toggling the 'Contacts
  * Access' setting OFF
  */
-export function setContactsPermissionOn(contactsPermissionOn: boolean): ThunkAction<void> {
-  return (dispatch, getState) => {
+export function setContactsPermissionOn(contactsPermissionOn: boolean): ThunkAction<Promise<void>> {
+  return async (dispatch, getState) => {
```
swansontec, line 314: Ahh, vastly more readable!

swansontec (approved):

PR #4367, opened 2023-07-28
Improve`styled` HOC

samholmes:
This enables basic styled components which do not need the props nor
theme.

### CHANGELOG

- Fixed: Updated `styled` HOC to handle caching and optionality of `theme`/`props` parameters.

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1205078606742009

### DIFF
--- a/src/components/hoc/styled.tsx
+++ b/src/components/hoc/styled.tsx
@@ -1,26 +1,72 @@
 import * as React from 'react'
-import { ImageStyle, StyleProp, TextStyle, ViewStyle } from 'react-native'
+import { ImageStyle, StyleProp, StyleSheet, TextStyle, ViewStyle } from 'react-native'
 
-import { Theme, useTheme } from '../services/ThemeContext'
+import { cacheStyles, getTheme, Theme, useTheme } from '../services/ThemeContext'
 
 interface StyleProps {
   style?: StyleProp<any>
 }
-type ValidStyles = ImageStyle | TextStyle | ViewStyle | ValidStyles[]
 
+type ValidStyles = ImageStyle | TextStyle | ViewStyle
+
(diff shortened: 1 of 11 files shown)


swansontec (changes requested):
Squashing the various commits into one makes this much, much easier to understand. Please do that.

The actual caching feature isn't implemented here. To implement the caching feature with this design, you need examine the styler *outside* the render function, to learn what it does.

If the styler is an object, pass that to the oldschool `StyleSheet.create`, and then use that on each render.

If the styler is a function, invoke it once outside the render to learn about the return type. If the styler returns a function, this is a dynamic component, so we'll have to re-call the returned function on every render. If the styler returns an object, then we need to pass the styler to `cacheStyles` to add the theme-based caching. Then we use the cached function on each render.

I imagine your `StyledComponent` inner function will have three variants, one to deal with each of these cases. The logic for adding `displayName` would s

src/components/hoc/styled.tsx
```diff
@@ -1,26 +1,72 @@
...
 
-import { Theme, useTheme } from '../services/ThemeContext'
+import { cacheStyles, getTheme, Theme, useTheme } from '../services/ThemeContext'
 
 interface StyleProps {
   style?: StyleProp<any>
 }
-type ValidStyles = ImageStyle | TextStyle | ViewStyle | ValidStyles[]
 
+export type ValidStyles = ImageStyle | TextStyle | ViewStyle
```
swansontec, line 10: Optional: We put in this export while we were debugging, but no external files use it anymore. This type could be private again.

swansontec (approved):
Nice! You got it clean without any TypeScript hacks!

PR #4373, opened 2023-08-03
Jon/side-menu-cleanup

Jon-edge:
- Remove TOS, Rate display from side menu
- Change termsOfService scene to a generic webView scene
- Convert HelpModal.tsx to function components, split file
- Change icon. The "clipboard" icon is not an appropriate representation of TOS.
- Consistently show a WebViewScene instead of a WebViewModal, as to avoid having multiple modals open at once.

image  image

### CHANGELOG

- Remove TOS, rate display from side menu. Add TOS to help modal

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203963276097142


src/types/routerTypes.tsx
```diff
@@ -283,6 +284,7 @@ export interface RouteParamList {
     currencyCode: string
   }
   walletList: {}
+  webView: WebViewSceneParams
```
swansontec, line 287: Delete the `termsOfService` key a few lines up.

src/__tests__/modals/HelpModal.test.tsx
```diff
@@ -2,15 +2,15 @@ import { describe, expect, it } from '@jest/globals'
...
+import { HelpModal } from '../../components/modals/HelpModal'
 import { fakeAirshipBridge } from '../../util/fake/fakeAirshipBridge'
+import { fakeNavigation } from '../../util/fake/fakeSceneProps'
 
 describe('HelpModal', () => {
   it('should render with loading props', () => {
     const renderer = createRenderer()
 
-    const actual = renderer.render(<HelpModalComponent bridge={fakeAirshipBridge} theme={getTheme()} />)
+    const actual = renderer.render(<HelpModal bridge={fakeAirshipBridge} navigation={fakeNavigation} />)
```
swansontec, line 13: We are using the shallow renderer on this scene. We need to upgrade to the deep renderer, so the snapshot diffs will actually contain something useful:

```ts
import TestRenderer from 'react-test-renderer'

    const renderer = TestRenderer.create(
      <FakeProviders>
        <HelpModal ... />
      </FakeProviders>
    )

    expect(renderer.toJSON()).toMatchSnapshot()
    renderer.unmount()
```

swansontec (changes requested):
Only one tiny change is needed. This is a pretty clean PR. I would personally squash the last two commits together into one, but this is optional.

swansontec (approved):

PR #4384, opened 2023-08-09
Remove Header Buttons for New Account Wallet Creation

Jon-edge:
Split the scene into newAccountFlow/non-newAccountFlow usage to allow independent configuration of header button visibility in Main.tsx

### CHANGELOG

- removed: Header buttons for new accounts' initial wallet creation

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205214041882472

### DIFF
--- a/src/components/scenes/CreateWalletSelectCryptoScene.tsx
+++ b/src/components/scenes/CreateWalletSelectCryptoScene.tsx
@@ -1,6 +1,6 @@
 import { FlashList, ListRenderItem } from '@shopify/flash-list'
 import * as React from 'react'
-import { Switch, View } from 'react-native'
+import { Keyboard, Switch, View } from 'react-native'
 import { sprintf } from 'sprintf-js'
 
 import { enableTokensAcrossWallets, MainWalletCreateItem, PLACEHOLDER_WALLET_ID, splitCreateWalletItems } from '../../actions/CreateWalletActions'
@@ -198,7 +198,7 @@ const CreateWalletSelectCryptoComponent = (props: Props) => {
   })
 
   const handleSubmitEditing = useHandler(() => {
-    handleNextPress().catch(err => showError(err))
+    Keyboard.dismiss()
   })
 
   const renderCreateWalletRow: ListRenderItem<WalletCreateItem> = useHandler(item => {
@@ -255,7 +255,6 @@ const CreateWalletSelectCryptoComponent = (props: Props) => {
             onChangeText={setSearchTerm}
             value={searchTerm}
             label={lstrings.wallet_list_wallet_search}
-            returnKeyType="next"
             marginRem={[0.5, 1]}
             searchIcon
             clearIcon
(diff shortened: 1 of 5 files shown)


src/components/scenes/CreateWalletSelectCryptoScene.tsx
```diff
@@ -237,7 +232,14 @@ const CreateWalletSelectCryptoComponent = (props: Props) => {
...
       <Fade noFadeIn={defaultSelection.length > 0} visible={numSelected > 0} duration={300}>
         <View style={styles.bottomButton}>
-          <MainButton label={lstrings.string_next_capitalized} type="primary" marginRem={[0, -0.5]} onPress={handleNextPress} alignSelf="center" />
+          <MainButton
+            label={lstrings.string_next_capitalized}
+            type="primary"
+            marginRem={[0, -0.5]}
+            onPress={handleNextPress}
+            alignSelf="center"
+            disabled={numSelected === 0}
```
peachbits, line 241: The next button is disabled by the Fade component so this doesn't do anything

src/components/scenes/CreateWalletSelectCryptoScene.tsx
```diff
@@ -93,11 +93,6 @@ const CreateWalletSelectCryptoComponent = (props: Props) => {
   })
 
   const handleNextPress = useHandler(async () => {
-    if (numSelected === 0) {
```
peachbits, line 96: As far as I can tell the only way to get this error is if the user has typed into the search bar and they press the done button on the keyboard without first selecting anything. The keyboard doesn't drop in that case (at least on android) so some kind of feedback for the user is preferable compared to nothing happening at all.

peachbits (changes requested):
I couldn't reproduce the behavior that the second commit aims to fix. You'll need to show me the issue and I'll take another look

src/components/scenes/CreateWalletSelectCryptoScene.tsx
Jon-edge: The fade animation isn't enough, you can still click it while fading. This change was to remove the disabled check in handleNextPress

src/components/scenes/CreateWalletSelectCryptoScene.tsx
Jon-edge: The fade animation isn't enough, you can still click it while fading. 

As mentioned in the commit, we don't want to show errors to the user when they haven't done anything unexpected. Errors look really bad and should be reserved for unexpected things.

src/components/scenes/CreateWalletSelectCryptoScene.tsx
Jon-edge: You can replicate it easily by having the last wallet to disable be an entry near the next button

Jon-edge: Checked this again, can't reproduce. I can just drop this commit

peachbits: Do you have a video of the behavior? I still can't replicate and from code inspection it doesn't seem possible. The Fade component disables pointer events when the component is hidden and in this case that's when selected is 0. You may still be able to see it but pointer events have turned off.

peachbits (approved):

peachbits (approved):

PR #4386, opened 2023-08-10
Paul/xrp dex

paullinator:
### CHANGELOG

added: XRP DEX support
fixed: Exchange quote layout spacing

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205228472265668


src/components/scenes/CryptoExchangeQuoteScene.tsx
```diff
@@ -97,6 +98,22 @@ export class CryptoExchangeQuoteScreenComponent extends React.Component<Props, S
     ))
   }
 
+  showExplanationForCanBePartial = async () => {
```
swansontec, line 101: This method name needs to start with `handle` to satisfy a lint rule we plan to enable one day. Maybe `handleCanBePartialExplanation`

src/locales/en_US.ts
```diff
@@ -785,6 +785,12 @@ const strings = {
   edge_login_fail_message: 'Could not connect to network. Please check your network connection and try again.',
   edge_login_fetching: 'Fetching Edge Login info...',
   modal_addressexplorer_message: 'Show Address in Block Explorer?',
+  can_be_partial_quote: 'Maybe Partial Quote',
```
swansontec, line 788: The string name should have `_title` on the end, per a previous sidebar. I also think we should drop the word "Maybe", since that's included in the explanation below.

swansontec (changes requested):
Only minor changes here.

src/components/scenes/CryptoExchangeQuoteScene.tsx
paullinator: This was to be consistent with other handler `showExplanationForEstimate`

I'll fix with a cleanup to the other handler as well

src/locales/en_US.ts
paullinator: This was also to be consistent with `estimated_exchange_rate` but that's the a fair catch. I'll change to adhere to new standards

swansontec (approved):

PR #4389, opened 2023-08-11
Monero settings

swansontec:
### CHANGELOG

- added: A Monero settings scene.

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204583307528860


src/components/scenes/PromotionSettingsScene.tsx
```diff
@@ -80,19 +81,6 @@ export class PromotionSettingsComponent extends React.Component<Props> {
...
-    backgroundColor: theme.settingsRowBackground,
-    padding: theme.rem(0.5)
-  },
-  textRow: {
-    color: theme.primaryText,
-    fontFamily: theme.fontFaceDefault,
-    fontSize: theme.rem(1),
-    margin: theme.rem(0.5)
-  }
-}))
```
samholmes, line 94: Margin/padding wont match the same presentation as in `SettingsSubHeader` from what I can tell. Is this intentional?

src/components/themed/MaybeMoneroUserSettings.tsx
```diff
@@ -0,0 +1,67 @@
...
+  moneroLwsServer: asString
+})
+type MoneroUserSettings = ReturnType<typeof asMoneroUserSettings>
+
+type Props = CurrencySettingProps<MoneroUserSettings, undefined>
+
+function MoneroUserSettingsComponent(props: Props) {
+  const { defaultSetting, onUpdate, setting } = props
+  const { enableCustomServers, moneroLwsServer } = setting
+  const empty = moneroLwsServer === '' || moneroLwsServer === defaultSetting.moneroLwsServer
```
samholmes, line 25: `isEmpty` or invert the boolean to be `hasCustomServer`

samholmes (changes requested):
A nit and a question

src/components/scenes/PromotionSettingsScene.tsx
swansontec: Yes, the three lines are more spaced out after this change. I think it looks better this way, especially because it's more consistent with the other settings scenes. I think this is part of what Jon was talking about - removing special cases by creating common components.

samholmes (approved):

PR #4397, opened 2023-08-16
Paul/bitwave csv

paullinator:
### CHANGELOG

added: Export Bitwave format CSV

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205274123045129


src/components/scenes/TransactionsExportScene.tsx
```diff
@@ -142,7 +148,8 @@ class TransactionsExportSceneComponent extends React.PureComponent<Props, State>
   handleAndroidToggle = () => {
     this.setState(state => ({
       isExportCsv: !state.isExportCsv,
-      isExportQbo: !state.isExportQbo
+      isExportQbo: !state.isExportQbo,
+      isExportBitwave: !state.isExportBitwave
```
swansontec, line 188: The Android toggles are messed up. We need to do the following:

```ts
  handleQboToggle = () => {
    if (android) {
      this.setState({ isExportQbo: true, isExportCsv: false, isExportBitwave: false })
   } else { 
     this.setState(state => ({ isExportQbo: !state.isExportQbo }))
   }
  }
```

And likewise for the other 2 handlers. Then in the `renderAndroidSwitches`, we use these 3 handlers, instead of the generic `handleAndroidToggle`.

This is because Android can only choose one option at a time. With two options, it doesn't really matter which is which, but with 3 options, it suddenly matters which switch you choose, so we need separate handlers.

swansontec (changes requested):
The export logic is simple enough, but there is a problem with the toggle switches on Android.

src/components/scenes/TransactionsExportScene.tsx
paullinator: good catch. will do.

src/components/scenes/TransactionsExportScene.tsx
```diff
@@ -145,23 +145,25 @@ class TransactionsExportSceneComponent extends React.PureComponent<Props, State>
...
-      isExportBitwave: !state.isExportBitwave
-    }))
-  }
-
   handleQboToggle = () => {
+    if (Platform.OS === 'android' && !this.state.isExportQbo) {
+      this.setState({ isExportCsv: false, isExportBitwave: false })
+    }
+
     this.setState(state => ({ isExportQbo: !state.isExportQbo }))
```
swansontec, line 189: If you press the radio button once on Android, it will select the option. If the press the ratio button again, it will de-select the option and leave nothing chosen. Then, when you hit submit, you will get a crash, because `files[0]` is undefined.

The way I wrote the handler in my original review comment would have avoided this possibility, by ensuring exactly one item can be selected on Android.

swansontec (changes requested):
It's closer, and we could merge it if we really had to, but I see a way the user could crash it.

src/components/scenes/TransactionsExportScene.tsx
paullinator: You can't submit if nothing is selected. The Next button is hidden. This is also true on iOS. I did test this case. But it's fair to say the user shouldn't be allowed to select nothing.

swansontec (approved):

PR #4400, opened 2023-08-18
Custom Token Conflict Warning

Jon-edge:
Matching contract address, different token name
image

Matching built-in token code, different contract address:
image

Old vs new ManageTokensScene:

### CHANGELOG

- added: Warning message when adding a custom token that conflicts with a built-in token
- fixed: Component styling in token management scenes

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205214596907253

### DIFF
--- a/src/components/scenes/EditTokenScene.tsx
+++ b/src/components/scenes/EditTokenScene.tsx
@@ -11,13 +11,13 @@ import { useSelector } from '../../types/reactRedux'
 import { EdgeSceneProps } from '../../types/routerTypes'
 import { getWalletName } from '../../util/CurrencyWalletHelpers'
 import { logActivity } from '../../util/logger'
+import { ButtonsContainer } from '../buttons/ButtonsContainer'
 import { SceneWrapper } from '../common/SceneWrapper'
 import { withWallet } from '../hoc/withWallet'
 import { ButtonsModal } from '../modals/ButtonsModal'
 import { ConfirmContinueModal } from '../modals/ConfirmContinueModal'
 import { Airship } from '../services/AirshipInstance'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
-import { MainButton } from '../themed/MainButton'
 import { OutlinedTextInput } from '../themed/OutlinedTextInput'
 import { SceneHeader } from '../themed/SceneHeader'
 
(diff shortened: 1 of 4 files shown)


src/components/scenes/EditTokenScene.tsx
```diff
@@ -98,13 +98,37 @@ function EditTokenSceneComponent(props: Props) {
...
     } else {
-      const tokenId = await wallet.currencyConfig.addCustomToken(token)
-      await wallet.changeEnabledTokenIds([...wallet.enabledTokenIds, tokenId])
-      logActivity(`Add Custom Token: ${account.username} -- ${getWalletName(wallet)} -- ${wallet.type} -- ${tokenId} -- ${currencyCode} -- ${decimals}`)
+      const { currencyConfig } = wallet
+      const builtinTokens = Object.values(currencyConfig.builtinTokens)
+
+      // Check if custom token input conflicts with built-in tokens
+      const matchingContractToken = builtinTokens.find(
+        builtInToken => builtInToken.networkLocation?.contractAddress === customTokenInput.networkLocation?.contractAddress
```
swansontec, line 109: Do not access `networkLocation`, since chains like EOS, Tron, or Ripple have wildly different stuff in there. I do see that we hard-code this field name a few lines above, which is unfortunate. I will create a task to make this scene dynamic, based on the `EdgeCurrencyInfo` data structures, so we can correctly handle custom tokens on these weird chains.

The only correct way to compare two coins is by `tokenId`, and the only way to get the `tokenId` is to call `addCustomToken`. I would love to expose an `validateCustomToken` method that calculates tokenId without adding the token, but this is may be impossible without a breaking change to the currency plugins.

So, given these constraints, it might make sense to add the token first and then issue the warning. If the user denies the warning, then delete the token. Otherwise, we need to put a big fat TODO comment in this PR, so we know the networkLocation logic is just a placeholder, pending a `validateCustomToken` method getting added.

Instead, we rely on the currency plugins to convert `networkLocation => tokenId`, and then we can use the tokenId to compare identity. Unfortunately, there is no easy way to access this conversion, aside from calling `addCustomToken`.

src/components/scenes/EditTokenScene.tsx
```diff
@@ -98,13 +98,37 @@ function EditTokenSceneComponent(props: Props) {
...
+      )
+
+      const warningMessage = builtinTokens.find(builtInToken => builtInToken.currencyCode === currencyCode)
```
swansontec, line 112: This line will not pass the eslint strict boolean rule. Please add `!= null`

swansontec (changes requested):

src/components/scenes/EditTokenScene.tsx
Jon-edge: I didn't get any violations on yarn lint

swansontec (approved):

PR #4406, opened 2023-08-25
Reorganize Modal UI Heirarchy

Jon-edge:
1. A bottom fade only applies to scrollable areas. It doesn't make sense to be attached to the ModalFooter (close button).
2. The ModalFooter should be attached to all ThemedModals unless otherwise specified. There's no reason for the modal callers to manage it.

Restructure so that the bottom fade is attached to ModalScrollArea, the footer is attached to ThemedModal by default, and adjust stylings to make sense with this new organization. 

Examples of styling adjustments made:

The close button correctly is tappable for the entire horizontal space per correct UX conventions, but the close button's tappable area is completely overlapping the button in the modal contents, making it prone to error:
image

The gradient has also been moved up so it's more apparent. Again note the removal of the overlap between the close button and other tappable areas on the modal:
image

### CHANGELOG

- changed: Modal UI adjustments

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1204800075529342

### DIFF
--- a/src/components/themed/ModalParts.tsx
+++ b/src/components/themed/ModalParts.tsx
@@ -7,6 +7,8 @@ import { lstrings } from '../../locales/strings'
 import { fixSides, mapSides, sidesToPadding } from '../../util/sides'
 import { GradientFadeOut } from '../modals/GradientFadeout'
 import { Theme, useTheme } from '../services/ThemeContext'
+// TODO:
+// KeyboardAwareScrollView (login) instead of ScrollView (here)
 
 interface ModalTitleProps {
   children: React.ReactNode
@@ -17,8 +19,8 @@ interface ModalTitleProps {
 
 interface ModalFooterProps {
   onPress: () => void
-  fadeOut?: boolean | undefined
 }
+
 export function ModalTitle(props: ModalTitleProps) {
   const { center, children, icon = null, paddingRem } = props
   const theme = useTheme()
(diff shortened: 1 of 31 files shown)


src/components/themed/ModalParts.tsx
```diff
@@ -43,51 +45,52 @@ export function ModalMessage(props: { children: React.ReactNode; paddingRem?: nu
...
-    <View>
+    <View style={styles.scrollContainer}>
       <ScrollView contentContainerStyle={styles.scrollPadding}>{children}</ScrollView>
-      <ModalFooter onPress={onCancel} fadeOut />
+      <ModalFooterFade />
+    </View>
+  )
+}
+
+export const ModalFooterFade = () => {
```
swansontec, line 76: It's obvious what this does, but I would like to see a Javadoc comment up here.

src/components/themed/ModalParts.tsx
```diff
@@ -43,51 +45,52 @@ export function ModalMessage(props: { children: React.ReactNode; paddingRem?: nu
...
+    </TouchableOpacity>
   )
 }
 
-ModalFooter.bottomRem = 2.5
+ModalFooter.bottomRem = 3
 
-export function ModalScrollArea(props: { children: React.ReactNode; onCancel: () => void }) {
-  const { children, onCancel } = props
+export function ModalScrollArea(props: { children: React.ReactNode }) {
```
swansontec, line 63: It's obvious what this does, but I would like to see a Javadoc comment up here.

swansontec (approved):
Approved with some very small changes.

PR #4409, opened 2023-08-28
Paul/edge provider extra tab

paullinator:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205305016027593

### DIFF
--- a/src/components/themed/EdgeProviderComponent.tsx
+++ b/src/components/themed/EdgeProviderComponent.tsx
@@ -1,0 +1,205 @@
+import { uncleaner } from 'cleaners'
+import * as React from 'react'
+import { NativeSyntheticEvent, Platform } from 'react-native'
+import { WebView, WebViewMessageEvent } from 'react-native-webview'
+
+import { EdgeProviderServer } from '../../controllers/edgeProvider/EdgeProviderServer'
+import { javascript } from '../../controllers/edgeProvider/injectThisInWebView'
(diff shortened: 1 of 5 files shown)


src/types/types.ts
```diff
@@ -308,6 +308,7 @@ export interface AppConfig {
   website: string
   extraTab?: {
     webviewUrl: string
+    useEdgeProvider: boolean
```
samholmes, line 311: This naming conflicts with hooks naming convention `useXyz`. Perhaps to avoid confusion, could we name it `tabType: 'provider' | 'webview'`?

src/components/themed/EdgeProviderComponent.tsx
samholmes, line 1: Can I trust you that this is just a copy of functionality? If not, what specifically was changed structurally that should be noted?

samholmes (changes requested):

src/components/themed/EdgeProviderComponent.tsx
paullinator: No functionality changed. Just slight refactor to make the logic a component instead of a scene

src/types/types.ts
paullinator: will use `tabType: 'edgeProvider' | 'webview'`

samholmes (approved):

PR #4412, opened 2023-08-29
Paul/add pokt

paullinator:
### CHANGELOG

added: Pokt network for RPC servers

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205371018874233

### DIFF
--- a/src/envConfig.ts
+++ b/src/envConfig.ts
@@ -21,6 +21,7 @@ const asEvmApiKeys = asObject({
   evmScanApiKey: asOptional(asArray(asString), () => []),
   gasStationApiKey: asOptional(asString, ''),
   infuraProjectId: asOptional(asString, ''),
+  poktPortalApiKey: asOptional(asString, ''),
   quiknodeApiKey: asOptional(asString, '')
 }).withRest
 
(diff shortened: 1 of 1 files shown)


src/envConfig.ts
```diff
@@ -14,6 +14,16 @@ function asCorePluginInit<T>(cleaner: Cleaner<T>): Cleaner<T | false> {
...
 }
 
+const asEvmApiKeys = asObject({
+  alethioApiKey: asOptional(asString, ''),
+  amberdataApiKey: asOptional(asString, ''),
+  blockchairApiKey: asOptional(asString, ''),
+  evmScanApiKey: asOptional(asArray(asString), () => []),
+  gasStationApiKey: asOptional(asString, ''),
+  infuraProjectId: asOptional(asString, ''),
+  quiknodeApiKey: asOptional(asString, '')
```
samholmes: Where is `bscscanApiKey`, `ftmscanApiKey`, and `polygonscanApiKey`?

samholmes (changes requested):

src/envConfig.ts
paullinator: Removed as they are deprecated and we have been consistently using `evmScanApiKey` which replaced all of those

samholmes (approved):

PR #4425, opened 2023-08-31
Empty modules folder

swansontec:
We try to put components in src/components, Redux actions in src/actions, etc., but there are a few stragglers in src/modules that need to go to the right place.

While we are doing this, the settings are incredibly messy, and we have needed to clean them up for a while. While the complete solution will take more work, let's at least put each on-disk file into its own source file. So, the code that writes to "Categories.json" should live in "CategoriesActions.ts", and likewise for the other files. 

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205389800586450


Jon-edge (approved):

src/__tests__/components/SelectFioAddress.test.tsx
```diff
@@ -30,7 +30,7 @@ describe('SelectFioAddress', () => {
           payee_fio_address: 'fio@fio',
           payer_fio_address: 'fio2@fio',
           payer_fio_public_key: '123123',
-          status: 'pending',
```
Jon-edge, line 33: Why did this change?

Jon-edge (approved):
Approved with a question to verify

src/__tests__/components/SelectFioAddress.test.tsx
swansontec: Since we typed this as `string`, there was nothing to prevent us from making mistakes. According to the docs, "pending" is not a valid value for this field. So, I picked the actually valid value that had the same meaning.

PR #4427, opened 2023-09-01
Handle the case of entering a custom token that exists

Jon-edge:
### CHANGELOG

- added: Warning message when entering a custom token with a matching built-in currency code and address

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205367674582029


swansontec (approved):
This looks good, if we want this case to be "gentler".

I wonder if it really should be "gentler", though? There is less danger, but then again, the user probably added this custom token because they didn't realize it already existed. In that case, denying the modal should still remove the token, since that's convenient for the user to undo their mistake.

Up to you. It's approved either way.

Jon-edge:
I don't think it's realistic that while the user is trying to manually add a token, they immediately want to remove it

PR #4429, opened 2023-09-02
Retheme CreateWalletAccountSelectScene

Jon-edge:
image image

### CHANGELOG

- changed: Retheme wallet activation scene

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205308890818471


swansontec (approved):
This looks much nicer, both in terms of code and in terms of visuals.

We eventually need to clear out the redux state involved in this scene, but that can be a project for another day. For now this is much closer to the way it should be.

PR #4450, opened 2023-09-08
Paul/cleaner settings

paullinator:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205443364901991


src/actions/LocalSettingsActions.ts
```diff
@@ -149,15 +149,15 @@ export const writeSpendingLimits = async (account: EdgeAccount, spendingLimits:
...
-    .then(JSON.parse)
-    .catch(async () => {
-      // If Settings.json doesn't exist yet, create it, and return it
-      const defaults = asLocalAccountSettings({})
-      return await writeLocalSettings(account, defaults).then(() => defaults)
-    })
-    .then(settings => asLocalAccountSettings(settings))
+  try {
+    const text = await account.localDisklet.getText(LOCAL_SETTINGS_FILENAME)
+    const json = await JSON.parse(text)
```
samholmes, line 154: `JSON.parse` isn't async, so no need to await. 

Just FYI, `asLocalAccountSettings` could be a codec cleaner so we could `asLocalAccountSettings(text)` here if it were to use `asJSON`. Then on the write back to disk `wasLocalAccountSettings(data)`. But that's optional.

src/actions/LocalSettingsActions.ts
```diff
@@ -160,7 +160,7 @@ export const readLocalSettings = async (account: EdgeAccount): Promise<LocalAcco
   }
 }
 
-export const writeLocalSettings = async (account: EdgeAccount, settings: object) => {
+export const writeLocalSettings = async (account: EdgeAccount, settings: LocalAccountSettings) => {
   const text = JSON.stringify(settings)
```
samholmes, line 164: This is were you'd do the uncleaner to get the text: `wasLocalAccountSettings(settings)`. Again, optional.

samholmes (approved):
Approved with one minor change needed (an unnecessary await). An optional change to leverage `asJSON` as an additional commit.

src/actions/SettingsActions.tsx
```diff
@@ -296,7 +296,7 @@ export const asSyncedAccountSettings = asObject({
   autoLogoutTimeInSeconds: asOptional(asNumber, 3600),
   defaultFiat: asOptional(asString, 'USD'),
   defaultIsoFiat: asOptional(asString, 'iso:USD'),
-  preferredSwapPluginId: asOptional(asString, ''),
+  preferredSwapPluginId: asOptional(asString),
```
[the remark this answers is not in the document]
paullinator: The value could always be undefined by the time it is used. We were converting empty string to Undefined before assigning it to the Settings object

src/actions/LocalSettingsActions.ts
paullinator: Good catch on the async. Will fix. Do we really need an uncleaner if the data is always type checked before write

samholmes: Uncleaner is optional. Not necessary. It just moves the JSON serialization into the cleaner.

src/actions/SettingsActions.tsx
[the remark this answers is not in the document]
samholmes: Alright

PR #4453, opened 2023-09-11
Add Claim support to Tron Stake v2

peachbits:
---
  - https://app.asana.com/0/0/1205058026708442

### DIFF
--- a/src/components/scenes/Staking/StakeOverviewScene.tsx
+++ b/src/components/scenes/Staking/StakeOverviewScene.tsx
@@ -148,20 +148,24 @@ const StakeOverviewSceneComponent = (props: Props) => {
           marginRem={[0.25, 0.5, 0.25, 0.5]}
         />
       )}
-      <MainButton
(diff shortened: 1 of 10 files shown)


src/components/scenes/Staking/StakeOverviewScene.tsx
```diff
@@ -149,7 +147,14 @@ const StakeOverviewSceneComponent = (props: Props) => {
         />
       )}
       <MainButton
-        label={unstakeText}
+        label={lstrings.stake_unstake_claim}
+        disabled={!stakePosition?.canUnstakeAndClaim}
```
swansontec: This would violate the strict-boolean rule if we ever enable it. Better do `const {canUnstakeAndClaim = true} = stakePosition`, and then the `canUnstakeAndClaim` variable will work with `!` safely.

Or maybe the default is `false`. It's hard to reason about the way you have it.

swansontec (approved):
One small change, but not worth a re-review.

PR #4459, opened 2023-09-15
Jon/walletlistfooter-update

Jon-edge:
Remove the main wallet list Add Wallet/Add Token and replace with Deposit/Send
image

Deposit
Launch a modal with options:
image

From Bank/Cash
- Launches into the Buy scene
From Another Wallet/Exchange
- launches a wallet picker (with Add Wallet options) and once picked, routes to the QR code scene of that wallet
- Updated wallet picker title for clarity about what's happening:
image

Send 
- Launch a modal with options:
image

To Bank/Cash
- launches into the Sell scene
To Edge From Bank/Cash
- Launches into the Buy scene
To Another Wallet/Exchange
- launches a wallet picker (without Add Wallet options) and once picked, routes to the SendScene of that wallet

### CHANGELOG

- changed: "Add Wallet/Add Token" replaced with "Deposit/Send"

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1203970962685634


src/components/buttons/ButtonsContainer.tsx
```diff
@@ -35,7 +36,7 @@ interface Props {
 /**
  * A consistently styled view for displaying button layouts.
  */
-export const ButtonsContainer = React.memo(({ absolute = false, fade, primary, secondary, escape, layout = 'column' }: Props) => {
+export const ButtonsContainer = React.memo(({ absolute = false, fade, primary, secondary, secondaryAsPrimary, escape, layout = 'column' }: Props) => {
```
paullinator, line 39: Seems like `secondaryAsPrimary` is an odd name. Really this is the "second" secondary button. Something like `secondary2` would be more appropriate.

src/components/modals/TransferModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const handleSell = useHandler(() => {
+    navigation.navigate('sellTab', { screen: 'pluginListSell' })
+    Airship.clear()
+  })
+  const handleBuy = useHandler(() => {
+    navigation.navigate('buyTab', { screen: 'pluginListBuy' })
+    Airship.clear()
+  })
+  const handleSend = useHandler(() => {
+    Airship.show<WalletListResult>(bridge => <WalletListModal bridge={bridge} headerTitle={lstrings.select_wallet} navigation={navigation} />)
```
paullinator, line 54: Use a more precise string than `select_wallet`. Should say "Select wallet to send from"

src/components/modals/TransferModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+      .then(async ({ walletId, currencyCode }: WalletListResult) => {
```
paullinator, line 55: optional: use async/await syntax for readability

paullinator: also use

`.then(async ({ walletId, tokenId }:`

 instead of currencyCode. Then you don't need to call getTokenId

src/components/modals/TransferModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+        Airship.clear()
+      })
+      .catch(err => showError(err))
+  })
+  const handleReceive = useHandler(() => {
+    Airship.clear()
+    Airship.show<WalletListResult>(bridge => (
+      <WalletListModal bridge={bridge} headerTitle={lstrings.select_receive_asset} navigation={navigation} showCreateWallet />
+    ))
+      .then(async ({ walletId, currencyCode }: WalletListResult) => {
```
paullinator, line 71: get the tokenId instead of currencyCode

src/components/modals/TransferModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+        const { title, icon, onPress } = option
+        return (
+          <SelectableRow
+            key={title}
+            title={title}
+            onPress={onPress}
+            icon={<View style={style.iconContainer}>{icon}</View>}
+            arrowTappable
+            autoHeight
+            underline={options.indexOf(option) !== options.length - 1}
```
paullinator, line 158: Get the index by using
```
options.map((option, index) => {
```

paullinator (changes requested):

src/components/buttons/ButtonsContainer.tsx
Jon-edge: I wanted to capture that it is serving as primary in the visual heirarchy, but
skinned visually as a secondary.

src/components/buttons/ButtonsContainer.tsx
```diff
@@ -57,7 +57,7 @@ export const ButtonsContainer = React.memo(({ absolute = false, fade, primary, s
     <Animated.View style={fadeStyle}>
       <StyledButtonContainer absolute={absolute} layout={layout}>
         {renderButton('primary', primary)}
-        {renderButton('secondary', secondaryAsPrimary)}
+        {renderButton('secondary', secondaryRight)}
```
paullinator, line 60: Technically not always to the "right". If the layout is vertical it will be below so hence my recommendation of using `secondary2`

paullinator (changes requested):

src/components/modals/TransferModal.tsx
```diff
@@ -50,34 +49,25 @@ export const TransferModal = ({ account, bridge, depositOrSend, navigation }: Pr
...
-          const wallet = account.currencyWallets[walletId]
-          const tokenId = getTokenId(account, wallet.currencyInfo.pluginId, currencyCode)
 
-          await dispatch(selectWalletToken({ navigation, walletId, tokenId }))
-          navigation.navigate('request', {})
-        }
-      })
-      .catch(err => showError(err))
+    if (walletId != null && tokenId != null) {
+      await dispatch(selectWalletToken({ navigation, walletId, tokenId }))
```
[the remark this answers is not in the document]
Jon-edge: Error gets caught in a showError dropdown without a catch here

PR #4469, opened 2023-09-20
GUI: Unify Tracking

Jon-edge:
### CHANGELOG

- changed: Login-UI Analytics implementation consolidated into GUI

### Dependencies

https://github.com/EdgeApp/edge-login-ui-rn/pull/131

### Requirements

If you have made **any** visual changes to the GUI. Make sure you have:

- [x] Tested on iOS device
- [ ] Tested on Android device
- [ ] Tested on small-screen device (iPod Touch)
- [ ] Tested on large-screen device (tablet)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205535454470917

PR #4473, opened 2023-09-20
Jon/ab/create-login-buttons

Jon-edge:
### CHANGELOG

- added: Experiment config support between gui and login-ui
- added: Experiment config for "Create Account" button labels

### Dependencies

https://github.com/EdgeApp/edge-react-gui/pull/4469

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205532906159293


src/experimentConfig.ts
```diff
@@ -10,19 +10,32 @@ export type ExperimentConfig = ReturnType<typeof asExperimentConfig>
...
 
 /**
- * Generate a random boolean value according to the experiment distribution
+ * Generate a random index value according to the experiment distribution to
+ * determine which variant gets used.
  */
-const generateExperimentConfigVal = (key: keyof typeof experimentDistribution): boolean => {
-  return Math.random() < experimentDistribution[key]
+const generateExperimentConfigVal = <T>(key: keyof typeof experimentDistribution, configVals: T[]): T => {
+  const variantProbability = experimentDistribution[key]
```
paullinator, line 27: Should assert that `variantProbability.length === configVals.length - 1`

paullinator (changes requested):

src/experimentConfig.ts
```diff
@@ -23,45 +26,51 @@ const experimentDistribution = {
...
-
-  // Replaces the "Create Account" button label
-  createAccountText: asOptional<'signUp' | 'getStarted' | 'createAccount'>(
-    asValue('signUp', 'getStarted', 'createAccount'),
-    generateExperimentConfigVal('createAccountText', ['signUp', 'getStarted', 'createAccount'])
-  )
-})
+// behavior/appearance.
+// If no generateCfgValFn given, returns defined defaults
+const asExperimentConfig: any = (generateCfgValFn?: GenerateExperimentConfigValFn) =>
```
paullinator, line 55: A cleaner shouldn't be typed `any`. This would break the return type

src/experimentConfig.ts
```diff
@@ -74,7 +83,7 @@ const experimentConfigPromise: Promise<ExperimentConfig> = (async (): Promise<Ex
   } catch (err) {
     // Not found or incompatible. Re-generate with random values according to
     // the defined distribution.
-    const generatedExperimentConfig = asExperimentConfig({})
+    const generatedExperimentConfig = asExperimentConfig(generateExperimentConfigVal)({})
```
paullinator, line 86: Why all the complexity of creating a function that gets passed in when all you really care about is a boolean like `useDefaults`. Pass that as the first arg to `asExperimentConfig` and you don't need to check for a null function.

paullinator (changes requested):

paullinator (approved):

PR #4475, opened 2023-09-21
Downgrade OS support

swansontec:
### CHANGELOG

- changed: Restore support for Android 7 and iOS v12.4.

### Dependencies

With this change, we can run the app again on these OS's. However, we are now getting a crash on iOS 12.4. Removing RNZcash fixes the crash, so now we need to debug RNZcash and implement a fix over there.

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205551001606090

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -3,7 +3,7 @@
 buildscript {
     ext {
         buildToolsVersion = "33.0.0"
-        minSdkVersion = 27 // Edge modified from 21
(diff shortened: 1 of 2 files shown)


swansontec:
After a few hours of effort, it doesn't look like we can get Zcash working on these older OS versions without crashing. Yes, it's supposed to be possible to conditionalize these things, but doing so will require deep refactoring to third-party code.

PR #4479, opened 2023-09-21
Dynamic TC Savers pools. Was: Stable savers

paullinator:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1201400342342420


src/plugins/stake-plugins/thorchainSavers/tcSaversPlugin.ts
```diff
@@ -201,7 +198,42 @@ let thornodeServers: string[] = THORNODE_SERVERS_DEFAULT
...
+            ...policyDefault,
+            stakePolicyId: `tcsavers/${pluginId}:${lowerCc}=${pluginId}:${lowerCc}`,
+            rewardAssets: [{ pluginId, currencyCode }],
+            stakeAssets: [{ pluginId, currencyCode }]
+          })
+        }
+      })
+      policiesInitialized = true
+    } catch (e) {
+      console.log('Error initializing Thorchain Savers policies', String(e))
```
Jon-edge, line 233: Would this frequently happen? If not, seems useful to promote this to a console warn or error

Jon-edge (approved):
Approved with minor comment

paullinator:
Do not merge. found a bug that will need a fixup

src/plugins/stake-plugins/thorchainSavers/tcSaversPlugin.ts
paullinator: console.warn is annoying in that it shows as a yellow box in the debugger. It's not critical as it can happen if there is a networking failure

paullinator:
This implementation will not work as I've been notified by TC devs that we need to use a smart contract for token deposits. This requires a major overhaul to make the code similar to the swap code that uses memos. Closing this task as a new one will be estimated and planned out.

src/components/scenes/EditTokenScene.tsx
```diff
@@ -197,7 +197,7 @@ function EditTokenSceneComponent(props: Props) {
  * In the future this scene may need to handle other weird networks
  * where the networkLocation has other contents.
  */
-const asMaybeContractLocation = asMaybe(asObject({ contractAddress: asString }))
+export const asMaybeContractLocation = asMaybe(asObject({ contractAddress: asString }))
```
Jon-edge, line 200: Might be a good time to move /aave/cleaners/asEthTokenContractAddress.ts to a
common location and add asMaybeContractLocation to it

paullinator: Will do for the full Thorchain Savers with token task.

PR #4481, opened 2023-09-21
Receive: Fix Balance Text Overflow

Jon-edge:
image

### CHANGELOG

- fixed: Possible balance text overflow in Receive scene

### Dependencies
 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205078841293962


swansontec (approved):
Makes sense. Any time we put text in a horizontal container, it needs to be flexible.

PR #4504, opened 2023-09-30
Jon/fix/retention-analytics

Jon-edge:
Background
Changing svar_legacyLanding's type from boolean to string might have broken retention report views in Firebase, since it was already a tracked boolean dimension. The numbers don't make sense in the retention reports.

It was also discovered that it was misguided to continue to try to augment reported Firebase data to try to more accurately capture first open, because first open is exactly the time where we know it's unlikely Firebase will be working correctly.

Task
Rename legacyLanding to start capturing a fresh dimension.
This is a quick trial fix in hopes of fixing the retention report in firebase
Add new tracking params: firstOpenEpoch, deviceId
firstOpenEpochis a fallback solution in case 1. doesn't work, and will also be useful in moving towards our own reporting solution in general. We need this regardless due to analytics reliability issues for new installs.
Will allow us to create our own retention report with data on our util/referral server.

### CHANGELOG

- added: New analytics for first open epoch and deviceId

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1205619258271472

### DIFF
--- a/src/actions/FirstOpenActions.tsx
+++ b/src/actions/FirstOpenActions.tsx
@@ -1,25 +1,46 @@
+import { asNumber, asObject, asString, asValue } from 'cleaners'
 import { makeReactNativeDisklet } from 'disklet'
 
 import { FIRST_OPEN } from '../constants/constantSettings'
+import { makeUuid } from '../util/utils'
 
 const firstOpenDisklet = makeReactNativeDisklet()
 
-let isFirstOpen: 'true' | 'false'
+const asFirstOpenInfo = asObject({
+  isFirstOpen: asValue('true', 'false'),
+  deviceId: asString,
+  firstOpenEpoch: asNumber
+})
+type FirstOpenInfo = ReturnType<typeof asFirstOpenInfo>
+
+let firstOpenInfo: FirstOpenInfo
 
 /**
  * Returns whether this session was the first time the user opened the app.
- * Repeated calls will return the same result.
+ * Repeated calls will return the same result. Also sets the deviceId &
+ * firstOpenEpoch
  */
-export const getIsFirstOpen = async () => {
-  if (isFirstOpen != null) return isFirstOpen
-  else {
+export const getFirstOpenInfo = async (): Promise<FirstOpenInfo> => {
+  if (firstOpenInfo == null) {
+    let firstOpenText
     try {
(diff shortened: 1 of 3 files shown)


src/actions/FirstOpenActions.tsx
```diff
@@ -1,25 +1,46 @@
...
+    // Not the first time they opened the app, but with an older version that
+    // didn't set a deviceId and firstOpen date
+    let deviceId
+    let firstOpenEpoch
+    if (firstOpenText == null || firstOpenText === '') {
+      deviceId = makeUuid()
+      const firstOpenEpoch = Math.floor(new Date().getTime())
+      await firstOpenDisklet.setText(FIRST_OPEN, JSON.stringify({ deviceId, firstOpenEpoch }))
+    } else {
+      const firstOpenInfo = JSON.parse(firstOpenText)
```
paullinator, line 39: Add cleaner so we don't trust data from disk which may be corrupt. This also gives matching types for your return value

src/actions/FirstOpenActions.tsx
```diff
@@ -1,25 +1,46 @@
...
+let firstOpenEpoch: number | undefined
 
 /**
  * Returns whether this session was the first time the user opened the app.
- * Repeated calls will return the same result.
+ * Repeated calls will return the same result. Also sets the deviceId
  */
-export const getIsFirstOpen = async () => {
-  if (isFirstOpen != null) return isFirstOpen
+export const getFirstOpenInfo = async () => {
```
paullinator, line 16: Add types for function

src/util/tracking.ts
```diff
@@ -122,6 +122,9 @@ export function logEvent(event: TrackingEventName, values: TrackingValues = {})
       // Add all 'sticky' remote config variant values:
       for (const key of Object.keys(experimentConfig)) params[`svar_${key}`] = experimentConfig[key as keyof ExperimentConfig]
 
+      // TEMP HACK: Add renamed var for legacyLanding
+      params.svar_newLegacyLanding = experimentConfig.legacyLanding
```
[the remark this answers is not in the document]
Jon-edge: I limited this change intentionally so that we don't regenerate variants for everyone

PR #4507, opened 2023-10-02
Don't start booting wallets until settings are loaded

swansontec:
Otherwise we might boot wallets that should stay paused.

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205605002662319

### DIFF
--- a/src/reducers/scenes/SettingsReducer.ts
+++ b/src/reducers/scenes/SettingsReducer.ts
@@ -13,7 +13,8 @@ export const initialState: SettingsState = {
   isTouchEnabled: false,
   isTouchSupported: false,
   pinLoginEnabled: false,
-  settingsLoaded: null
+  settingsLoaded: null,
+  userPausedWalletsSet: null
 }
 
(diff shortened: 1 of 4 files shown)


src/reducers/scenes/SettingsReducer.ts
```diff
@@ -94,6 +99,7 @@ export const settingsLegacy = (state: SettingsState = initialState, action: Acti
         mostRecentWallets,
         passwordRecoveryRemindersShown,
         userPausedWallets,
+        userPausedWalletSet: new Set(userPausedWallets),
```
Jon-edge, line 102: userPausedWalletsSet to make it more obvious it's the same as `userPausedWallets`

Jon-edge (approved):

src/reducers/scenes/SettingsReducer.ts
swansontec: Done

PR #4508, opened 2023-10-02
GUI: Fix Back Routing for Legacy Landing Variant

Jon-edge:
Bug steps:
1. Press Sign Up
2. Press Get Started
3. Go Back 2x (new username, password login)
4. Observe the wrong scene: GettingStarted USPs instead of legacy landing

legacyLanding being undefined at the end of the steps to produce the bug was causing incorrect maybeHandleComplete routing to be returned

### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205586035325311

### DIFF
--- a/src/envConfig.ts
+++ b/src/envConfig.ts
@@ -196,7 +196,7 @@ export const asEnvConfig = asObject({
   KEYS_ONLY_PLUGINS: asOptional(asObject(asBoolean), {}),
   USE_FAKE_CORE: asOptional(asBoolean, false),
   USE_FIREBASE: asOptional(asBoolean, true),
-  USE_WELCOME_SCREENS: asOptional(asBoolean, true),
+  USE_WELCOME_SCREENS: asOptional(asBoolean, true), // Used by whitelabels
 
   YOLO_DEEP_LINK: asNullable(asString),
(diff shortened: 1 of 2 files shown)


src/envConfig.ts
```diff
@@ -196,7 +196,6 @@ export const asEnvConfig = asObject({
   KEYS_ONLY_PLUGINS: asOptional(asObject(asBoolean), {}),
   USE_FAKE_CORE: asOptional(asBoolean, false),
   USE_FIREBASE: asOptional(asBoolean, true),
-  USE_WELCOME_SCREENS: asOptional(asBoolean, true),
```
swansontec, line 199: Please check with @paullinator about removing this. It might have something to do with whitelabel apps, where we we don't want the Edge-branded welcome scenes to show at all, regardless of our A/B testing.

swansontec (approved):
I am approving this, pending clarification with @paullinator about the legacy landing pages. If he's good, I'm good. Otherwise, we'd need to drop the first commit.

src/envConfig.ts
Jon-edge: Commit dropped

swansontec (approved):

PR #4509, opened 2023-10-03
Paul/banxa sell

paullinator:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205579278471371


src/actions/TrackingActions.ts
```diff
@@ -33,6 +34,31 @@ export function trackConversion(
   }
 }
 
+export async function trackConversionWithReferral(
+  event: TrackingEventName,
+  opts: {
+    currencyCode: string
+    exchangeAmount: number
```
Jon-edge, line 41: While we're at it, can we add these to the tracking event instead of
just the dollarValue? 

At least for currencyCode, I can easily think of a hypothetical experiment where we are trying to
improve token UI/UX, and using measured token transactions as one of the metrics
of success.

src/locales/en_US.ts
```diff
@@ -1307,6 +1307,9 @@ const strings = {
   fiat_plugin_sell_region_restricted: 'Region restricted from selling %s',
   fiat_plugin_sell_no_quote: 'Unable to get sell quote',
   fiat_plugin_fetching_assets: 'Fetching supported assets',
+  fiat_plugin_sell_cancelled: 'Sell order cancelled',
+  fiat_plugin_sell_failed_try_again: 'Sell order failed. Please try again.',
+  fiat_plugin_sell_failed_to_send_try_again: 'Failed to send funds for sell transaction. Please try again.',
```
Jon-edge, line 1312: Wording suggestion:
'Funds transfer failed, unable to complete sell order. Your funds have not
been lost, please try again.'
'Crypto transfer failed, unable to complete sell order. Your funds have not
been lost. please try again.'

Trying to make it clear that:
1. This is talking about the same thing as the other two messages: `Sell order`
2. The funds simply failed to leave the wallet, they did not get sent and
   disappear. Describing crypto failing to leave the wallet, not that fiat
   failed to arrive.

src/plugins/gui/providers/banxaProvider.ts
```diff
@@ -273,12 +353,10 @@ export const banxaProvider: FiatProviderFactory = {
         return allowedCurrencyCodes[direction]
       },
       getQuote: async (params: FiatProviderGetQuoteParams): Promise<FiatProviderQuote> => {
-        const { pluginId, regionCode, exchangeAmount, amountType, paymentTypes, fiatCurrencyCode, displayCurrencyCode, direction } = params
+        const { pluginId, regionCode, exchangeAmount, amountType, paymentTypes, fiatCurrencyCode, displayCurrencyCode, direction, tokenId } = params
         if (!paymentTypes.some(paymentType => allowedPaymentTypes[direction][paymentType] === true))
           throw new FiatProviderError({ providerId, errorType: 'paymentUnsupported' })
```
Jon-edge, line 358: Now that we have buy + sell, sounds like a good time to add `direction` to these errors.

paullinator: We could but the parent plugin already knows the direction so it should be able to imply that

src/locales/en_US.ts
paullinator: Let me pass this through Michael for feedback

Jon-edge: Main takeaways with Michael:

1. Don't tell the user to try again if the error message is describing a
   situation in which crypto has been sent already
2. Any other situation: `fiat_plugin_sell_failed_try_again: 'Sell order failed. Please try again.'` is sufficient to
   display, additional details on the specifics of the error can be hidden from the user

src/plugins/gui/providers/banxaProvider.ts
Jon-edge: Yeah that's true as it's designed now, and the design reads like there's some possibility that the error info wouldn't match the direction param in getBestError. 

The error would be more versatile if that more info just came attached to the error. Then the consumers of the error (analytics, bugsnag) wouldn't need to know about the implementation of the feature to fully report the full context of the error.

Jon-edge (approved):

PR #4514, opened 2023-10-06
Paul/paybis

paullinator:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1205041005078292

### DIFF
--- a/src/plugins/gui/pluginUtils.ts
+++ b/src/plugins/gui/pluginUtils.ts
@@ -1,5 +1,5 @@
 import { div, toFixed } from 'biggystring'
-import { EdgeDataStore } from 'edge-core-js'
+import { EdgeCurrencyWallet, EdgeDataStore } from 'edge-core-js'
 import { sprintf } from 'sprintf-js'
 
 import { formatNumber } from '../../locales/intl'
@@ -131,4 +131,13 @@ export const debugSpewStore = async (store: EdgeDataStore): Promise<void> => {
 
 export const assert = (condition: boolean, message: string) => {
   if (!condition) throw new Error(message)
+}
+
+const TESNET_KEYWORDS = ['test', 'goerli', 'kovan', 'rinkby']
+// Rough approximation of whether this wallet is a testnet
+export const isWalletTestnet = (wallet: EdgeCurrencyWallet): boolean => {
+  const { pluginId } = wallet.currencyConfig.currencyInfo
+
+  const match = TESNET_KEYWORDS.some(word => pluginId.includes(word))
+  return match
 }
--- a/src/plugins/gui/pluginUtils.ts
+++ b/src/plugins/gui/pluginUtils.ts
@@ -128,3 +128,7 @@ export const debugSpewStore = async (store: EdgeDataStore): Promise<void> => {
   console.log('*** theStore ***')
   console.log(JSON.stringify(theStore, null, 2))
 }
+
+export const assert = (condition: boolean, message: string) => {
+  if (!condition) throw new Error(message)
+}
--- a/src/plugins/gui/providers/paybisProvider.ts
+++ b/src/plugins/gui/providers/paybisProvider.ts
@@ -1,13 +1,15 @@
 import { eq, round } from 'biggystring'
 import { asArray, asDate, asMaybe, asObject, asOptional, asString, asValue } from 'cleaners'
-import { EdgeFetchOptions } from 'edge-core-js'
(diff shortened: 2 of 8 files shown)


src/plugins/gui/providers/paybisProvider.ts
```diff
@@ -0,0 +1,470 @@
...
+          displayCurrencyCode,
+          direction,
+          tokenId
+        } = params
+        const paymentType = paymentTypes.find(paymentType => allowedPaymentTypes[direction][paymentType] === true)
+        if (paymentType == null) throw new FiatProviderError({ providerId, errorType: 'paymentUnsupported' })
+
+        const pairs = paybisPairs?.data
+        if (pairs == null) {
+          throw new FiatProviderError({ providerId, errorType: 'assetUnsupported' })
```
Jon-edge, line 302: Seems a odd to be throwing this without comparing to the getQuote params

src/plugins/gui/providers/paybisProvider.ts
```diff
@@ -0,0 +1,470 @@
...
+              throw new FiatProviderError({ providerId, errorType: 'underLimit', errorAmount: Number(minMatch[1]), displayCurrencyCode: minMatch[2] })
+            }
+          }
+          throw new Error(lastError?.error.message ?? 'Paybis Unknown paymentMethodError')
+        }
+
+        // Should only have one payment method
+        if (paymentMethods.length !== 1) {
+          throw new Error('Invalid number of quoted payment methods')
+        }
```
Jon-edge, line 368:
```suggestion
        assert(paymentMethods.length !== 1, 'Invalid number of quoted payment methods')
```

src/plugins/gui/providers/paybisProvider.ts
```diff
@@ -0,0 +1,470 @@
...
+            const { coreWallet, showUi } = approveParams
+            const receiveAddress = await coreWallet.getReceiveAddress()
+
+            const bodyParams = {
+              cryptoWalletAddress: {
+                currencyCode: paybisCc,
+                address: receiveAddress.segwitAddress ?? receiveAddress.publicAddress
+              },
+              partnerUserId,
+              locale: locale.localeIdentifier.slice(0, 2),
```
Jon-edge, line 412: We should start referencing the same locale logic i.e. pickLanguage() in
preparation for locale UI settings

src/plugins/gui/pluginUtils.ts
```diff
@@ -128,3 +128,16 @@ export const debugSpewStore = async (store: EdgeDataStore): Promise<void> => {
   console.log('*** theStore ***')
   console.log(JSON.stringify(theStore, null, 2))
 }
+
+export const assert = (condition: boolean, message: string) => {
+  if (!condition) throw new Error(message)
+}
```
Jon-edge, line 134: Might not be worth it, but we can also have a generic flavor:

```suggestion
export const assert = (
  condition: boolean,
  ErrorType: new (...args: any[]) => Error = Error,
  props: Record<string, any> = { message: 'An error occurred' }
) => {
  if (!condition) {
    const error = new ErrorType()
    Object.assign(error, props)
    throw error
  }
}

// usage:
assert(true, undefined, { message: "Something went wrong" })
assert(minMatch != null, FiatProviderError, { providerId, errorType: 'underLimit', errorAmount: Number(minMatch[1]), displayCurrencyCode: minMatch[2] })

```

src/plugins/gui/providers/paybisProvider.ts
paullinator: This error implies we were not able to get the supported assets in the first place.

src/plugins/gui/providers/paybisProvider.ts
paullinator: Nice. I'll use it.

src/plugins/gui/pluginUtils.ts
paullinator: I like it, but I don't want to add the code now as this PR won't be using it. I'm open to it being added once it once it's used.

src/plugins/gui/providers/paybisProvider.ts
Jon-edge: Ah right this was the error type that was used as more like a result

Jon-edge (approved):

PR #4517, opened 2023-10-07
Upgrade react-native-linear-gradient 2.8.3

paullinator:
### CHANGELOG

 none

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1205663510577187

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -3,8 +3,8 @@ PODS:
   - boost (1.76.0)
   - BugsnagReactNative (7.20.2):
     - React-Core
-  - BVLinearGradient (2.5.6):
-    - React
+  - BVLinearGradient (2.8.3):
+    - React-Core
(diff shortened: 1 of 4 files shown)


ios/Podfile.lock
```diff
@@ -1343,4 +1343,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 05dc3e3e330cf28ff2c3e6b3517d100d067246b6
 
-COCOAPODS: 1.13.0
+COCOAPODS: 1.12.1
```
swansontec, line 1346: Please upgrade your CocoaPods. Looks like you are on an older version.

swansontec (approved):

ios/Podfile.lock
paullinator: good catch. done

PR #4526, opened 2023-10-11
Show EdgeTxAction Info in Transaction List

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device


src/locales/en_US.ts
```diff
@@ -675,6 +675,28 @@ const strings = {
...
+  transaction_details_swap_order_fill_subcat_2s: '%1$s to %2$s Order Filled',
+  transaction_details_swap_order_cancel_subcat_2s: '%1$s to %2$s Order Canceled',
+  transaction_details_stake_subcat_1s: 'Stake %1$s',
+  transaction_details_stake_subcat_2s: 'Stake %1$s and %2$s',
+  transaction_details_stake_order_subcat_1s: 'Create Stake %1$s Order',
+  transaction_details_stake_order_subcat_2s: 'Create Stake %1$s and %2$s Order',
+  transaction_details_unstake_subcat_1s: 'Unstake %1$s',
+  transaction_details_unstake_subcat_2s: 'Unstake %1$s and %2$s',
+  transaction_details_unstake_order_subcat_1s: 'Create Unstake %1$s Order',
+  transaction_details_unstake_order_subcat_2s: 'Create Unstake %1$s and %2$s Order',
```
swansontec, line 698: I feel like these aren't really "categories", but more like notes. From an accounting point of view, I would expect something simple like "Expense:Staking" or "Exchange:Swap" for the category, and then details like "Trade XYZ token for ZZZ token" as a memo or note or something, but maybe I am wrong. I would ask @paullinator for a second opinion on this design.

src/components/scenes/TransactionDetailsScene.tsx
```diff
@@ -94,6 +102,87 @@ class TransactionDetailsComponent extends React.Component<Props, State> {
     return await wallet.accelerate(transaction)
   }
 
+  /**
+   * Given an EdgeTxAction, returns the display value for pre-filling the 'Category' tile
+   */
+  getTxActionSplitCategory = (action: EdgeTxAction | undefined): SplitCategory | undefined => {
```
swansontec, line 108: Can we fold this logic into the `splitCategory` function, or at least move it into the same file so it's accessible as a utility function?

Maybe we could call it `getTransactionCategory(walllet: EdgeCurrencyWallet, tx: EdgeTransaction): SplitCategory`, and then this function could look at both `tx.action` and `tx.metadata.category`, which would both be available in that case.

That way, we can utilize this function in the CSV export / QBO export functions, or any other places where we need the category.

swansontec (changes requested):

src/locales/en_US.ts
Jon-edge: Yeah agreed, defaults for notes sound better. I'll sync up with him on this

src/actions/CategoriesActions.ts
```diff
@@ -255,3 +257,99 @@ export const defaultCategories = [
...
+/**
+ * Given an EdgeTxAction, returns the display value for pre-filling the
+ * 'Category' and 'Notes' tiles, if they are not already user-modified.
+ */
+export const getTxActionDisplayInfo = (
+  tx: EdgeTransaction,
+  wallet: EdgeCurrencyWallet,
+  tokenId?: string
+): { splitCategory: SplitCategory; notes?: string } | undefined => {
+  const action = tx.action
```
swansontec, line 270: Could be `const {action} = tx`

src/actions/CategoriesActions.ts
```diff
@@ -255,3 +257,99 @@ export const defaultCategories = [
...
+          category: 'expense',
+          subcategory: sprintf(lstrings.transaction_details_swap_order_cancel)
+        }
+      }
+    case 'stake': {
+      let subcategory
+      if (action.stakeAssets.length === 1) subcategory = sprintf(lstrings.transaction_details_stake_subcat_1s, ...getCurrencyCodes(action.stakeAssets))
+      else if (action.stakeAssets.length === 2) subcategory = sprintf(lstrings.transaction_details_stake_subcat_2s, ...getCurrencyCodes(action.stakeAssets))
+      else {
+        console.error(`Unsupported number of assets for '${type}' EdgeTxActionSwapType`)
```
swansontec, line 311: I would use warn instead of error, since as a dev I don't want a big red screen of death if my wallet gets into this state.

swansontec (approved):

PR #4529, opened 2023-10-17
Use EdgeMemos on SendScene2

peachbits:
---
  - https://app.asana.com/0/0/1205724850151850


swansontec (approved):
This is the better fix.

PR #4534, opened 2023-10-19
Exchange Quote Picker

Jon-edge:
image
image

Before final cleanup commit:
image

After final (WIP) cleanup commit:
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205453208664797


src/actions/CryptoExchangeActions.tsx
```diff
@@ -285,14 +298,14 @@ function processSwapQuoteError(error: unknown): ThunkAction<void> {
...
-export function shiftCryptoCurrency(navigation: NavigationBase, swapInfo: GuiSwapInfo, onApprove: () => void): ThunkAction<Promise<void>> {
+export function shiftCryptoCurrency(navigation: NavigationBase, quote: EdgeSwapQuote, onApprove: () => void): ThunkAction<Promise<void>> {
   return async (dispatch, getState) => {
     const state = getState()
     const { account } = state.core
     dispatch({ type: 'START_SHIFT_TRANSACTION' })
 
-    const { fromDisplayAmount, quote, request, fee, fromFiat, fromTotalFiat, toDisplayAmount, toFiat } = swapInfo
-    const { isEstimate, fromNativeAmount, toNativeAmount, networkFee, pluginId, expirationDate } = quote
+    const { fromDisplayAmount, fee, fromFiat, fromTotalFiat, toDisplayAmount, toFiat } = getSwapInfo(state, quote)
```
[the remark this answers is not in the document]
paullinator: This entire commit was unnecessary work

src/components/scenes/CryptoExchangeQuoteScene.tsx
```diff
@@ -154,12 +152,7 @@ export const CryptoExchangeQuoteScene = (props: Props) => {
               walletId={request.toWallet.id}
               walletName={getWalletName(request.toWallet)}
             />
-            <TouchableOpacity style={styles.pluginRowPoweredByRow} onPress={handlePoweredByTap}>
-              <EdgeText style={styles.footerText}>{lstrings.plugin_powered_by_space + ' '}</EdgeText>
```
paullinator: footerText not used.

src/components/scenes/CryptoExchangeQuoteScene.tsx
```diff
@@ -69,27 +107,69 @@ export const CryptoExchangeQuoteScene = (props: Props) => {
...
-    return <CircleTimer timeExpired={async () => await dispatch(exchangeTimerExpired(navigation, quote, onApprove))} expiration={expirationDate} />
+    return <CircleTimer timeExpired={async () => await dispatch(exchangeTimerExpired(navigation, selectedQuote, onApprove))} expiration={expirationDate} />
   }
 
+  const renderRow = useHandler((item: { item: EdgeSwapQuote; section: Section; index: number }) => {
+    const quote = item.item
+    return (
+      <TouchableOpacity
+        onPress={() => {
+          navigation.replace('exchangeQuote', {
```
paullinator, line 123: Instead of a full `navigation.replace`, would have been simpler to make selectedQuote a `useState` which you then change with a `setSelectedQuote`

Line 48 could be

`const { initialSelectedQuote, quotes, onApprove } = route.params`

Then 

`const [ selectedQuote, setSelectedQuote ] = useState(initialSelectedQuote)`

paullinator, line 123: Instead of a full `navigation.replace`, would have been simpler to make selectedQuote a `useState` which you then change with a `setSelectedQuote`

Line 48 could be

`const { selectedQuote: initialSelectedQuote, quotes, onApprove } = route.params`

Then 

`const [ selectedQuote, setSelectedQuote ] = useState(initialSelectedQuote)`

src/components/data/row/SwapProviderRow.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const { fromWallet, fromTokenId, toWallet, toTokenId } = request
+  const theme = useTheme()
+  const styles = getStyles(theme)
+
+  const isToQuote = quoteFor === 'to'
+  const wallet = isToQuote ? fromWallet : toWallet
+  const tokenId = isToQuote ? fromTokenId : toTokenId
+  const nativeAmount = isToQuote ? fromNativeAmount : toNativeAmount
+
+  const toAmount = useCryptoText({ wallet, tokenId, nativeAmount })
```
paullinator, line 29: Misnomer. Might be a 'from' or 'to'. Should be `exchangeAmount`

paullinator (changes requested):

src/actions/CryptoExchangeActions.tsx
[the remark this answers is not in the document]
Jon-edge: There were a lot of motivations.

1. Work towards removing all these old guiXXX types, which are often tied into unnecessary redux and relied on a lot of duplicated handling for crypto/fiat numbers that weren't necessarily consistent.
2. At least take a step towards unifying the handling for crypto/fiat amount displays for this feature. I implemented a lot of common display/conversion handling functionality with the intention of using towards this very purpose. Going to take a bit more work to get rid of _all_ the crypto/fiat amount display handling in favor of common utils, hence leaving a lot of this here in getSwapInfo. 
3. This way, at least the scenes and components can use the crypto/fiat hooks + utilities as intended by that design, and the api for the components becomes simple and makes sense for what the components are displaying. In the future we can unify crypto/fiat calcs for actions/reducers as well. 
4. This is crazy overly complicated, prone to making mistakes (evidenced in this PR) and missing mistakes in reviews if we keep needing these giant blocks of conversion code every time we need to deal with simple fiat and crypto amounts.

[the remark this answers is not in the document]
Jon-edge: If the only concern was fiat exchange rates, I'd argue that the new implementation is better.

In the old implementation, worse case scenario, the user is saving minute-old prices + the time it takes to execute. Sure, they're guaranteed to save what they see, but what they see isn't really the true price at that time.

In the new implementation, worse-case scenario for saved fiat price metadata is only the price delta in the time it takes the quote to execute.

src/actions/CryptoExchangeActions.tsx
```diff
@@ -174,11 +181,20 @@ async function fetchSwapQuote(state: RootState, request: EdgeSwapRequest): Promi
   const feeDisplayAmount = toFixed(feeTempAmount, 0, 6)
 
   // Format fiat fee:
-  const feeDenominatedAmount = await fromWallet.nativeToDenomination(feeNativeAmount, request.fromWallet.currencyInfo.currencyCode)
-  const feeFiatAmountRaw = parseFloat(convertCurrency(state, request.fromWallet.currencyInfo.currencyCode, fromWallet.fiatCurrencyCode, feeDenominatedAmount))
-  const feeFiatAmount = formatNumber(feeFiatAmountRaw || 0, { toFixed: 2 })
+  const feeCryptoAmount = div(feeNativeAmount, feeDenomination.multiplier, DECIMAL_PRECISION)
```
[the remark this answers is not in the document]
Jon-edge: Mainly changed it just so my earlier commit would build - where I use this method in the scene selector. Needed this to become synchronous for that commit to work. 

The intention is to later clean all this up to use common utility methods.

src/components/data/row/SwapProviderRow.tsx
Jon-edge: Made it a bit more literal, 'exchange' already has multiple meanings in our codebase

Jon-edge:
Fixed unit tests on the last commit

src/components/data/row/SwapProviderRow.tsx
paullinator: I'll accept the update, but `exchangeAmount` has a very specific meaning. I might agree that it's not the best term but we do use it consistently.

paullinator (approved):

src/components/themed/ExchangeQuoteComponent.tsx
```diff
@@ -22,39 +26,54 @@ interface Props {
...
 
+  // Fees are assumed to be denominated in the native currency
   const feeNativeAmount = networkFee.nativeAmount
-  const feeCryptoText = useCryptoText({ wallet, tokenId, nativeAmount: feeNativeAmount, withSymbol: true })
-  const { currencyCode, denomination, isoFiatCurrencyCode } = useTokenDisplayData({
-    tokenId,
-    wallet
+  const feeCryptoText = useCryptoText({ wallet: fromWallet, nativeAmount: feeNativeAmount, withSymbol: true })
+  const {
+    currencyCode: nativeCurrencyCode,
```
paullinator, line 41: use `parent`

src/components/themed/ExchangeQuoteComponent.tsx
```diff
@@ -22,39 +26,54 @@ interface Props {
...
+    wallet: fromWallet,
+    tokenId: fromTokenId
+  })
+
   const feeFiatText = useFiatText({
     autoPrecision: true,
-    cryptoCurrencyCode: currencyCode,
-    cryptoExchangeMultiplier: denomination.multiplier,
-    isoFiatCurrencyCode,
+    cryptoCurrencyCode: nativeCurrencyCode,
```
paullinator, line 59: use parent

src/components/themed/ExchangeQuoteComponent.tsx
```diff
@@ -22,39 +26,54 @@ interface Props {
...
   const feeNativeAmount = networkFee.nativeAmount
-  const feeCryptoText = useCryptoText({ wallet, tokenId, nativeAmount: feeNativeAmount, withSymbol: true })
-  const { currencyCode, denomination, isoFiatCurrencyCode } = useTokenDisplayData({
-    tokenId,
-    wallet
+  const feeCryptoText = useCryptoText({ wallet: fromWallet, nativeAmount: feeNativeAmount, withSymbol: true })
+  const {
+    currencyCode: nativeCurrencyCode,
+    denomination: nativeDenomination,
+    isoFiatCurrencyCode: nativeIsoFiatCurrencyCode
```
paullinator, line 43: `nativeIsoFiatCurrencyCode` will always be the same as `fromIsoFiatCurrencyCode` since the wallet is the same. Simplify to `isoFiatCurrencyCode`

src/components/themed/ExchangeQuoteComponent.tsx
```diff
@@ -22,39 +26,54 @@ interface Props {
...
+    const cryptoAmount = div(feeNativeAmount, nativeDenomination.multiplier, DECIMAL_PRECISION)
+    return convertCurrency(state, nativeCurrencyCode, nativeIsoFiatCurrencyCode, cryptoAmount)
+  })
+
+  const fromFiatAmount = useSelector(state => {
+    const cryptoAmount = div(fromNativeAmount, fromDenomination.multiplier, DECIMAL_PRECISION)
+    return convertCurrency(state, fromCurrencyCode, fromIsoFiatCurrencyCode, cryptoAmount)
   })
 
+  const totalFiatText = `${getSymbolFromCurrency(fromIsoFiatCurrencyCode)}${formatFiatString({ fiatAmount: add(feeFiatAmount, fromFiatAmount) })}`
```
paullinator, line 75: revert to using no symbol

paullinator (approved):

PR #4539, opened 2023-10-19
Overlay styling for "Wallet Paused"

Jon-edge:
Simulator Screenshot - iPhone 14 - 2023-10-19 at 16 23 44

Simulator Screenshot - iPhone 14 - 2023-10-19 at 16 23 59

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205654034220682

### DIFF
--- a/src/components/themed/WalletListCurrencyRow.tsx
+++ b/src/components/themed/WalletListCurrencyRow.tsx
@@ -1,12 +1,15 @@
 import { EdgeCurrencyWallet, EdgeToken } from 'edge-core-js'
 import * as React from 'react'
-import { TouchableOpacity } from 'react-native'
+import { StyleSheet, TouchableOpacity, View } from 'react-native'
 
 import { useHandler } from '../../hooks/useHandler'
+import { lstrings } from '../../locales/strings'
+import { useSelector } from '../../types/reactRedux'
 import { triggerHaptic } from '../../util/haptic'
 import { CurrencyRow } from '../data/row/CurrencyRow'
 import { CustomAsset, CustomAssetRow } from '../data/row/CustomAssetRow'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
+import { EdgeText } from './EdgeText'
 
 interface Props {
   customAsset?: CustomAsset
@@ -34,6 +37,8 @@ const WalletListCurrencyRowComponent = (props: Props) => {
   } = props
   const theme = useTheme()
   const styles = getStyles(theme)
+  const pausedWallets = useSelector(state => state.ui.settings.userPausedWalletsSet)
+  const isPaused = pausedWallets != null && pausedWallets.has(wallet.id)
 
   // Currency code and wallet name for display:
   const currencyCode = customAsset?.currencyCode ?? token?.currencyCode ?? wallet.currencyInfo.currencyCode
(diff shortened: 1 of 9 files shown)


src/components/themed/WalletListCurrencyRow.tsx
```diff
@@ -50,13 +55,40 @@ const WalletListCurrencyRowComponent = (props: Props) => {
...
+        <CurrencyRow showRate={showRate && !isPaused} token={token} tokenId={tokenId} wallet={wallet} />
+      )}
     </TouchableOpacity>
   )
 }
 
 const getStyles = cacheStyles((theme: Theme) => ({
-  // Layout:
+  overlayLabel: {
+    color: theme.overlayDisabledTextColor
```
Jon-edge, line 77: Though default text color in edgeDark is white, I want to force this to be white no
matter what so it contrasts against the background which will always be on the
darker side

src/components/themed/WalletListCurrencyRow.tsx
```diff
@@ -50,13 +55,40 @@ const WalletListCurrencyRowComponent = (props: Props) => {
...
+  },
+  overlayLabelContainer: {
+    ...StyleSheet.absoluteFillObject,
+    alignItems: 'center',
+    display: 'flex',
+    justifyContent: 'center',
+    zIndex: 2
```
swansontec, line 84: Instead of using `zIndex`, just move the `isPaused ? <>..</> : null` block *after* the `<CurrencyRow>`rendering block. Then it will naturally go on top of the row component.

src/components/themed/WalletListCurrencyRow.tsx
```diff
@@ -50,13 +55,40 @@ const WalletListCurrencyRowComponent = (props: Props) => {
...
 
 const getStyles = cacheStyles((theme: Theme) => ({
-  // Layout:
+  overlayLabel: {
+    color: theme.overlayDisabledTextColor
+  },
+  overlayLabelContainer: {
+    ...StyleSheet.absoluteFillObject,
+    alignItems: 'center',
+    display: 'flex',
```
swansontec, line 82: Remove `display: flex`. It's the default.

src/components/themed/WalletListCurrencyRow.tsx
```diff
@@ -50,13 +55,40 @@ const WalletListCurrencyRowComponent = (props: Props) => {
...
+    justifyContent: 'center',
+    zIndex: 2
+  },
+  overlayContainer: {
+    ...StyleSheet.absoluteFillObject,
+    backgroundColor: theme.overlayDisabledColor,
+    opacity: theme.overlayDisabledOpacity,
+    zIndex: 1
+  },
```
swansontec, line 91: Just put the background color on `overlayLabelContainer`, but use either `RGBA(0, 0, 0, 0.5)` or `#00000080` to put the opacity into the color itself, removing the need for `theme.overlayDisabledOpacity` as well.

swansontec (approved):
This looks really good, but I have some ideas about how to simplify the CSS. If you don't want to do this, then we could merge as-is.

src/components/themed/WalletListCurrencyRow.tsx
Jon-edge: Just realized we have that in the palette

swansontec (approved):

src/components/themed/WalletListCurrencyRow.tsx
```diff
@@ -50,13 +55,33 @@ const WalletListCurrencyRowComponent = (props: Props) => {
...
 }
 
 const getStyles = cacheStyles((theme: Theme) => ({
-  // Layout:
+  overlayLabel: {
+    color: theme.overlayDisabledTextColor
+  },
+  overlayLabelContainer: {
+    ...StyleSheet.absoluteFillObject
+  },
```
swansontec, line 78: This appears to be unused now.

swansontec (changes requested):

swansontec (approved):

PR #4543, opened 2023-10-23
Jon/fix/experiment-config-init

Jon-edge:
- Fix experiment config disk copy management
- Fix generateExperimentConfigVal: The logic was incorrect for checking each bucket of probabilities for 3+ variants

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205764219371829

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - changed: Change FlipInput styling to make the edit functionality more obvious
 - changed: Enable max spend for Filecoin
 - changed: Move asset-specific settings into their own settings page
+- changed: Experiment config probability distribution support percentage based values
 - removed: Moonpay sell via ACH
 - removed: Banxa buy via Pix
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - changed: Enable max spend for Filecoin
 - changed: Move asset-specific settings into their own settings page
 - changed: Experiment config probability distribution support percentage based values
+- fixed: Write updated experiment configs to disk
 - removed: Moonpay sell via ACH
 - removed: Banxa buy via Pix
 
--- a/src/experimentConfig.ts
+++ b/src/experimentConfig.ts
@@ -92,17 +92,19 @@ const asExperimentConfig: Cleaner<ExperimentConfig> = asObject({
  * This config value is available through the module's getter functions.
  */
 const experimentConfigPromise: Promise<ExperimentConfig> = (async (): Promise<ExperimentConfig> => {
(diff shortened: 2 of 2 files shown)


samholmes (changes requested):
Change the probability distribution configuration to use percentages, and derive the sum as the denominator for you calculations (incase the sum doesn't add up to 100%).

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - added: Paybis sell to debit card
 - changed: Change FlipInput styling to make the edit functionality more obvious
 - changed: Move asset-specific settings into their own settings page
+- fixed: Experiment config generation for more than 2 variants
```
samholmes, line 10: `fixed: Write updated experiment configs to disk` and `changed: Experiment config probability distribution support percentage based values`

src/experimentConfig.ts
```diff
@@ -40,20 +39,27 @@ const experimentDistribution = {
 const generateExperimentConfigVal = <T>(key: keyof typeof experimentDistribution, configVals: T[]): T => {
   const variantProbability = experimentDistribution[key]
 
-  if (variantProbability.length !== configVals.length - 1) {
+  if (variantProbability.length !== configVals.length) {
     console.error(`Misconfigured experimentDistribution for: '${key}'`)
   } else {
+    // Distribute the probability of each config value
+    const totalProbability = variantProbability.reduce((sum, probability) => sum + probability, 0)
```
samholmes, line 46: I don't think this value is a "probability" but instead a `denomination`, so that might be a better name because that's what it is for each probability we calculate.

src/experimentConfig.ts
```diff
@@ -40,20 +39,27 @@ const experimentDistribution = {
...
+    if (totalProbability > 101 || totalProbability < 99) {
+      console.warn(`Config values for ${key} do not add up to 100% +/- 1%`)
+    }
+    const distributedProbabilities = variantProbability.map(probability => probability / totalProbability)
```
samholmes, line 50: Division by `zero` isn't handled when it should.

samholmes: We should probably check if the distribution array is empty too?

samholmes (changes requested):

src/experimentConfig.ts
Jon-edge: I assume you are referring only to `variantDenomination` because total denomination makes no sense

samholmes (approved):
Approved with the added fixup

PR #4549, opened 2023-10-27
Fix Android Auto-Logoff Picker Color

Jon-edge:
For Android only:
- Dynamically color the picker in contrast to the OS theme
- Add an overlay with a constant style to hide the dynamically colored native picker.

image
Screenshot 2023-10-26 at 7 29 12 PM

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205773982652046


src/components/modals/AutoLogoutModal.tsx
```diff
@@ -39,52 +57,103 @@ export const AutoLogoutModal = (props: Props) => {
...
-  }
+  // In Android, the Picker will be hidden with the text overlay.
+  // Determine correct contrasting text color for the native Android Picker
+  // menu. The selected option will show the normal theme color in a text
+  // overlay, but the Picker's color itself must be set to a color that
+  // contrasts with the Android OS theme.
+  React.useEffect(() => {
+    if (isAndroid) {
+      const { UIManagerModule } = NativeModules
+      const uiMode = UIManagerModule && UIManagerModule.getConstants().UI_MODE_NIGHT_YES
```
swansontec, line 68: This query isn't async, so we don't need the `useState` / `useEffect` pair. You could instead use `androidNativeTextColor = React.useMemo(..., [theme, isAndroid])`, which would still query the OS only once, but also save a render.

swansontec (approved):
Approved, but with two optional cleanups.

PR #4561, opened 2023-11-02
Add Posthog Analytics

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205865322432409


src/app.ts
```diff
@@ -34,6 +103,23 @@ console.log('***********************')
...
+  'Warning: isMounted(...) is deprecated'
+]
+// $FlowExpectedError
+// @ts-expect-error
+console.ignoredYellowBox = IGNORED_WARNINGS
+
+// Ignore errors and warnings(used for device testing)
+if (ENV.DISABLE_WARNINGS) {
+  LogBox.ignoreAllLogs() //  (IGNORED_WARNINGS)
+}
```
samholmes, line 121: We shouldn't be filtering warnings because they motivate us to handle the tech-debt rather than bury it. I'm okay with a DISABLE_WARNINGS ENV flag if developers want to use this, but it still discourages actually handling the warnings, so it's not great.

src/app.ts
```diff
@@ -2,19 +2,88 @@
...
+  const fmgMsg = typeof message === 'string' ? `'${message}${color != null ? `:${color}` : ''}'` : `${JSON.stringify({ message }, null, 2)}`
+
+  realConsDbg(`${deco}${fmgMsg}${reset}`)
+}
+
+console.debug('TEST   TEST')
+
 Bugsnag.start({
+  // @ts-expect-error
+  apiKey: ENV.BUGSNAG_API_KEY,
```
samholmes, line 81: Why wasn't this in place before and why the `@ts-expect-error`?

src/app.ts
```diff
@@ -2,19 +2,88 @@
...
 
+import ENV from '../env.json'
 import { changeTheme, getTheme } from './components/services/ThemeContext'
-import { ENV } from './env'
 import { NumberMap } from './types/types'
 import { log, logToServer } from './util/logger'
 
+console.log = () => {}
+console.info = () => {}
+console.warn = () => {}
```
samholmes, line 15: This doesn't seem like a good idea.

src/app.ts
```diff
@@ -2,19 +2,88 @@
...
+
+const realConsDbg = console.debug
+
+type FgColor = 'b' | 'r' | 'g' | 'c' | 'm' | 'y'
+type BgColor = 'B' | 'R' | 'G' | 'C' | 'M' | 'Y'
+
+console.debug = (message?: any, color?: FgColor | BgColor | [FgColor, BgColor] | string) => {
```
samholmes, line 22: I've never been a fan of overloading console methods. This certainly is a lot of mucking about with our debug method. These log utils should be reliable and without unexpected behavior, and so all of this seems risky for the win of getting some colors? I'm not in favor of this. Make this overloading an option for developers if they want to add colors to their console.debug method and either with an ENV flag or a new utility like `prettyDebug` or something that they can import from `app.ts` while debugging. This also seems more suited for a scratch file.

samholmes (changes requested):

src/app.ts
Jon-edge: Sorry this entire file wasn't supposed to end up in this review

src/app.ts
Jon-edge: Sorry this entire file wasn't supposed to end up in this review

src/app.ts
Jon-edge: Sorry this entire file wasn't supposed to end up in this review

src/app.ts
Jon-edge: Sorry this entire file wasn't supposed to end up in this review

src/util/tracking.ts
samholmes, line 1: This fixup is for the commit before the one it's fixing up

samholmes (approved):
Approved. Just be sure to apply the fixup correctly.

PR #4574, opened 2023-11-09
Fix ENS Name Resolution

Jon-edge:
image
image

ENS name resolution only works on Ethereum network. 
`ethers.getDefaultProvider` does not reliably succeed in resolving ENS names.

- Port over `validText` from loginUI's implementation of `OutlinedTextInput` and fix styling on the modal. This change caused differences in unit test colors. **The original implementation rendered red in tests for some reason while the new one renders mint, as expected.**
- Rely on accountbased to multicast `ethers` `JsonRpcProviders` instead to reliably resolve ENS names.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-currency-accountbased/pull/666

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205732732237945

### DIFF
--- a/src/components/modals/AddressModal.tsx
+++ b/src/components/modals/AddressModal.tsx
@@ -1,6 +1,5 @@
 import { FlashList } from '@shopify/flash-list'
 import { EdgeAccount, EdgeCurrencyConfig, EdgeCurrencyWallet } from 'edge-core-js'
-import { ethers } from 'ethers'
 import * as React from 'react'
 import { ActivityIndicator, Image, TouchableWithoutFeedback, View } from 'react-native'
 import { AirshipBridge } from 'react-native-airship'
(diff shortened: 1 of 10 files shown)


src/components/modals/AddressModal.tsx
```diff
@@ -96,8 +98,16 @@ export class AddressModalComponent extends React.Component<Props, State> {
...
-    this.setState({ statusLabel: status })
+  setShowSpinner = (showSpinner: boolean) => {
+    this.setState({ showSpinner })
+  }
+
+  setValidLabel = (validLabel: string | undefined) => {
+    this.setState({ validLabel })
+  }
+
+  setErrorLabel = (errorLabel: string | undefined) => {
```
paullinator, line 108: Get rid of these wrappers. Just call this.setState directly. one-liner wrappers not too useful here.

paullinator (changes requested):

paullinator (approved):
Small optional change requested

PR #4577, opened 2023-11-11
Prevent crash when tx has no spendTargets

paullinator:
A tx created by makeTx that also has swapData  will crash since it is assume to be created by makeSpend where core always adds the spendTargets.

In the case of a makeTx there are no spendTargets so use tx.nativeAmount instead.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205930491280477

### DIFF
--- a/src/components/tiles/SwapDetailsTiles.tsx
+++ b/src/components/tiles/SwapDetailsTiles.tsx
@@ -1,3 +1,4 @@
+import { abs, sub } from 'biggystring'
 import { EdgeCurrencyWallet, EdgeTransaction, EdgeTxSwap } from 'edge-core-js'
 import * as React from 'react'
 import { Linking, Platform, View } from 'react-native'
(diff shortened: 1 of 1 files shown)


src/components/tiles/SwapDetailsTiles.tsx
```diff
@@ -89,7 +90,8 @@ export function SwapDetailsTiles(props: Props) {
   )
   if (destinationDenomination == null) return null
 
-  const sourceAmount = convertNativeToDisplay(walletDefaultDenom.multiplier)(spendTargets[0].nativeAmount)
+  const sourceNativeAmount = sub(abs(transaction.nativeAmount), transaction.networkFee)
```
peachbits, line 93: This math won't work for tokens. You'll need to check if the currencyCode matches the mainnet before subtracting the fee. Also noting that we need to add tokenId to EdgeTxSwap

peachbits (changes requested):

peachbits (approved):

PR #4584, opened 2023-11-15
Use memos for Thorchain Savers & add token support

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205564968041578


src/components/scenes/Staking/StakeOverviewScene.tsx
```diff
@@ -141,6 +142,11 @@ const StakeOverviewSceneComponent = (props: Props) => {
           stakeProviderInfo={stakePolicy.stakeProviderInfo}
         />
       </View>
+      {stakePosition == null ? (
+        <View style={styles.shimmer}>
+          <Shimmer isShown />
+        </View>
+      ) : null}
```
peachbits, line 149: This would look better than the current behavior but the hardcoded height is not ideal since there could be one or two items in the list, maybe even 3? Since the list has an unknown number I'd prefer if we moved the fetchStakePosition into the previous scene and pass the allocations into this scene as props.

src/plugins/stake-plugins/thorchainSavers/tcSaversPlugin.ts
```diff
@@ -555,7 +557,7 @@ const stakeRequest = async (opts: EdgeGuiPluginOptions, request: ChangeQuoteRequ
     // 2. Only use UTXOs from the primary address (index 0)
     // 3. Force change to go to the primary address
     otherParams: { outputSort: 'targets', utxoSourceAddress, forceChangeAddress },
-    metadata: { name: 'Thorchain Savers', category: 'Transfer:Staking' }
+    metadata: { name: 'Thorchain Savers', category: `Transfer:Thorchain Savers ${sprintf(lstrings.transaction_details_stake_subcat_1s, currencyCode)}` }
```
peachbits, line 560: Is this string supposed to be in notes?

peachbits (changes requested):

src/plugins/stake-plugins/thorchainSavers/tcSaversPlugin.ts
paullinator: No. This is a subcat string. Transfer categories should have subcategories that specify the unique "entity" or "account" that is receiving the funds. "Thorchain Savers Stake BTC" is a specific account that funds are being transferred to and from. Although admittedly the name would be better described as "Thorchain Savers Staked BTC"

peachbits (approved):

PR #4585, opened 2023-11-16
Add GLIF Infinity Pool Staking

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1205644799202600


src/plugins/stake-plugins/generic/GenericStakePlugin.ts
```diff
@@ -0,0 +1,101 @@
+import { ChangeQuote, ChangeQuoteRequest, StakePlugin, StakePluginFactory, StakePolicy, StakePolicyFilter, StakePosition, StakePositionRequest } from '../types'
+import { makeGlifInfinityPoolAdapter } from './policyAdapters/GlifInfinityPoolAdapter'
+import { StakeAdapterConfig, StakePolicyAdapter } from './policyAdapters/types'
+import { StakePluginInfo, StakePolicyConfig } from './types'
+
+export const makeGenericStakePlugin =
```
swansontec, line 6: OK, so you are converting an "adapter" into a "plugin".

If writing a "plugin" is difficult because of all the boilerplate, then maybe we should simplify things so the "adapter" is all you really need to write, and *that* becomes the official plugin type.

I don't know how the existing plugins look. If there is an easy way to port them to this new "adapter" architecture, then we should just do it, and then just delete the whole stake plugin concept.

If we cannot port the existing stake plugins to "adapter", then why not? Are they more complicated? Do they need special logic? If there is something that makes the generic ones different, then maybe we should rename "Generic" to "EVM" or "Simple" or whatever that difference is.

src/plugins/stake-plugins/generic/policyAdapters/GlifInfinityPoolAdapter.ts
```diff
@@ -0,0 +1,270 @@
+import '@ethersproject/shims'
+
+import { gt } from 'biggystring'
+import { asObject, asString } from 'cleaners'
+import { EdgeCurrencyWallet } from 'edge-core-js'
+import { BigNumber, ethers } from 'ethers'
+
+import { GlifInfinityPool__factory, GlifPoolToken__factory, /* GlifRouter__factory, */ GlifSimpleRamp__factory } from '../../../contracts'
```
swansontec, line 8: Leftover comment?

src/plugins/stake-plugins/generic/policyAdapters/GlifInfinityPoolAdapter.ts
```diff
@@ -0,0 +1,270 @@
...
+
+    async fetchStakeQuote(wallet, requestAssetId, requestNativeAmount): Promise<ChangeQuote> {
+      const requestAssetCurrencyCode = requestAssetId.currencyCode
+      const isRequestAssetNative = requestAssetCurrencyCode === policyConfig.parentCurrencyCode
+
+      if (!isRequestAssetNative) throw new Error('Token staking not supported')
+
+      const { maxFeePerGas, maxPriorityFeePerGas, nextNonce, txs, walletAddress, walletSigner } = await workflowUtils(wallet)
+
+      // Deposik into pool contract:
```
swansontec, line 106: Is "deposik" a teknikal term?

swansontec (approved):
This looks pretty good, but I have an architecture question.

src/plugins/stake-plugins/generic/GenericStakePlugin.ts
samholmes: I originally attempted porting plugins, but found a lot of my time was going into this effort and not the actual task. I decided to design the new plugin with clean slate and start porting other plugins as a break away task to clean up tech debt. The reason is mainly a resource constraint; if I had more time, I would port the other plugins over to this adapter pattern and then flatten the abstractions.

babel.config.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
   presets: ['module:metro-react-native-babel-preset'],
-  plugins: ['react-native-reanimated/plugin']
+  plugins: [require('@babel/plugin-transform-export-namespace-from'), 'react-native-reanimated/plugin']
```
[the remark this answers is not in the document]
samholmes: Upgrading these packages didn't fix the issue:

```
    "@babel/core": "^7.23.3",
    "@babel/preset-env": "^7.23.3",
    "@babel/preset-typescript": "^7.23.3",
    "@babel/runtime": "^7.23.4",
```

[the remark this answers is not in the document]
swansontec: The `"@babel/preset-env": "^7.23.3"` package does pull this in as a dependency, but something about the Metro / React Native setup must be preventing it from working. So I guess this is the best way.

src/plugins/stake-plugins/generic/GenericStakePlugin.ts
swansontec: As long as there is an Asana task, this is fine.

swansontec (approved):

PR #4586, opened 2023-11-17
UI4: Transaction Details Scene

Jon-edge:
Implement UI4 changes for Transaction Details Scene

image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-react-gui/pull/4593

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205826921045607


Jon-edge:
@CodiumAI-Agent /describe

src/actions/ReceiveDropdown.tsx
```diff
@@ -71,7 +71,7 @@ export function showReceiveDropdown(navigation: NavigationBase, transaction: Edg
 
           navigation.navigate('transactionDetails', {
             edgeTransaction: transaction,
-            walletId
+            wallet
```
swansontec, line 74: We are not allowed to pass the wallet itself to a scene, but only the walletId. However, we have a functional middleware, `withWallet`, that will turn the `walletId` route param back into a `wallet` prop.

We do it this way so the app won't crash on auto-logout. The `withWallet` middlware handles all the crash cases.

src/components/scenes/TransactionDetailsScene.tsx
```diff
@@ -14,142 +14,122 @@ import { lstrings } from '../../locales/strings'
...
-  }
-
-  async makeAcceleratedTx(transaction: EdgeTransaction): Promise<EdgeTransaction | null> {
-    const { wallet } = this.props
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
 
+  const makeAcceleratedTx = async (transaction: EdgeTransaction): Promise<EdgeTransaction | null> => {
     return await wallet.accelerate(transaction)
   }
```
swansontec, line 164: Can this function body just move into the `useEffect` above? It seems kinda pointless now.

src/components/scenes/TransactionDetailsScene.tsx
```diff
@@ -197,36 +171,29 @@ class TransactionDetailsComponent extends React.Component<Props, State> {
...
-        const newLine = i + 1 < spendTargets.length ? '\n' : ''
-        recipientsAddresses = `${recipientsAddresses}${spendTargets[i].publicAddress}${newLine}`
-      }
+  const personLabel = direction === 'receive' ? lstrings.transaction_details_sender : lstrings.transaction_details_recipient
+  const personName = action != null ? TX_ACTION_LABEL_MAP[action.type] : name !== '' ? name : personLabel
+  const personHeader = sprintf(lstrings.transaction_details_person_name, personLabel)
+
+  // spendTargets recipient addresses format
+  let recipientsAddresses = ''
+  if (transaction.spendTargets) {
```
swansontec, line 246: Add `!= null`

src/components/scenes/TransactionDetailsScene.tsx
```diff
@@ -81,6 +90,64 @@ export const TransactionDetailsScene = (props: Props) => {
 
   const { name = '' } = localMetadata
 
+  // #region Crypto Fiat Rows
```
swansontec, line 94: This is a good cleanup! The only reason `TransactionFiatTiles` existed was so we could use hooks, and now that the parent is finally converted, all this stuff can finally move over.

src/locales/en_US.ts
```diff
@@ -1405,7 +1405,7 @@ const strings = {
   bank_info_title: 'Bank Info',
   home_address_title: 'Home Address',
   input_output_currency: 'Currency',
-  n_a: 'n/a',
+  n_a: 'N/A',
```
swansontec, line 1408: Instead of "n_a", could it be "empty_notes_placeholder" or something like that? Other languages might want to express "empty notes" using other idioms.

swansontec (changes requested):
I found a few small things, but overall I am really happy with how this cleaned up!

If the UI4 work is going to side on the sidelines for a few weeks, would it make sense to cherry-pick the function-style component upgrade into its own mini-PR? That way we can get the work in right away.

src/locales/en_US.ts
Jon-edge: I'll add another placeholder for notes, but I want to change 'N/A' regardless because it's being used elsewhere

src/components/scenes/TransactionDetailsScene.tsx
Jon-edge: Yeah we still need to make a slightly different call in the accelerate modal so I'll just write out the `accelerate` call directly

src/components/scenes/TransactionDetailsScene.tsx
swansontec: Still not fixed, but not a big deal.

Jon-edge: Oops, I'll get this in before merging

PR #4592, opened 2023-11-22
Add use of Pokt RPCs for Fantom

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205997135183169

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- added: Add Pokt RPCs as option for Fantom
 - added: Support for Thorchain Savers EVM token deposits
 - changed: Use memos for Thorchain Savers withdrawals
 - fixed: Various margin styling alignments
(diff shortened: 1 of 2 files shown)


src/app.ts
```diff
@@ -2,14 +2,16 @@
...
-import { Text, TextInput } from 'react-native'
+import { LogBox, Text, TextInput } from 'react-native'
 import RNFS from 'react-native-fs'
 
 import { changeTheme, getTheme } from './components/services/ThemeContext'
 import { ENV } from './env'
 import { NumberMap } from './types/types'
 import { log, logToServer } from './util/logger'
 
+LogBox.ignoreAllLogs()
```
samholmes, line 13: Nope. We should show annoying log boxes. I think this was a mistake?

samholmes (changes requested):

src/app.ts
paullinator: Oops. that snuck in there.

samholmes (approved):

PR #4593, opened 2023-11-22
UI4: Card & Row Components

Jon-edge:
- CardUi4
- SectionView
- RowUi4

See branches future'd with this branch for additional visuals.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205995877853214


src/components/ui4/RowUi4.tsx
```diff
@@ -0,0 +1,155 @@
...
+  ) : (
+    <TouchableHighlight
+      accessible={false}
+      onPress={handlePress}
+      onLongPress={handleLongPress}
+      disabled={handlePress == null && handleLongPress == null}
+      underlayColor={theme.touchHighlightUi4}
+    >
+      <View style={styles.container}>
+        {icon && <View style={styles.iconContainer}>{icon}</View>}
```
swansontec, line 86: The `&&` will be a lint error eventually. You mean `icon == null ? null : <View ... />`

src/components/ui4/RowUi4.tsx
```diff
@@ -0,0 +1,155 @@
...
+        <View style={styles.content}>
+          {title == null ? null : (
+            <EdgeText disableFontScaling ellipsizeMode="tail" style={error ? styles.textHeaderError : styles.textHeader}>
+              {title}
+            </EdgeText>
+          )}
+          {children == null
+            ? typeof body === 'string' && (
```
swansontec, line 94: `body == null ? null : <EdgeText .../>`. We don't need to check for string-ness, since TypeScript does that already.

src/components/ui4/RowUi4.tsx
```diff
@@ -0,0 +1,155 @@
...
+          {type === 'delete' && <FontAwesomeIcon name="times" style={styles.tappableIcon} size={theme.rem(1)} />}
+          {type === 'questionable' && <SimpleLineIcons name="question" style={styles.tappableIcon} size={theme.rem(1)} />}
+        </View>
+      </View>
+    </TouchableHighlight>
+  )
+}
+
+// TODO: Adjust margin/padding so everything combines with correct layout no
+// matter the combination of UI4 components.
```
swansontec, line 114: Ambitious. I assume this will happen later as more stuff gets built out?

src/theme/variables/edgeDark.ts
```diff
@@ -394,5 +397,8 @@ export const edgeDark: Theme = {
   guiPluginLogoMoonpay: guiPluginLogoMoonpay,
 
   // UI 4.0:
+  cardBackgroundUi4: palette.teal,
+  cardRadiusUi4: 1,
```
swansontec, line 401: We usually avoid putting dimensions in the theme, and instead rely on `rem` to scale things.

If we are committed to having the radius in the theme for some reason, then it's weird to do `theme.rem(theme.cardRadius)`. Maybe just return the radius in dp's, not rem's, since the theme controls the scaling already.

swansontec (changes requested):
Small changes needed.

src/theme/variables/edgeDark.ts
swansontec: Just checked, and the other theme dimensions are already in dp's, such as `thinLineWidth` or `secondaryButtonOutlineWidth`.

src/components/ui4/RowUi4.tsx
Jon-edge: Yes, I tried to start doing it and then found I had to make more changes upon tackling subsequent tasks. Figured it would be better to just handle it at the end to make sure everything works and avoid all the rebasing, it's mostly fine right now.

src/theme/variables/edgeDark.ts
Jon-edge: This is a rem, I can rename the var

swansontec: I don't think it should be rem in the first place, but the rename is also an acceptable solution.

PR #4594, opened 2023-11-23
Properly return tokenId from WalleListModal

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205981561051523

### DIFF
--- a/src/components/modals/WalletListModal.tsx
+++ b/src/components/modals/WalletListModal.tsx
@@ -128,11 +128,11 @@ export function WalletListModal(props: Props) {
   const handlePaymentMethodPress = useHandler((fiatAccountId: string) => () => {
     bridge.resolve({ fiatAccountId })
   })
-  const handleWalletListPress = useHandler((walletId: string, currencyCode: string, _tokenId?: string, customAsset?: CustomAsset) => {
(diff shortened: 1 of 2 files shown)


src/components/modals/WalletListModal.tsx
```diff
@@ -128,11 +128,11 @@ export function WalletListModal(props: Props) {
   const handlePaymentMethodPress = useHandler((fiatAccountId: string) => () => {
     bridge.resolve({ fiatAccountId })
   })
-  const handleWalletListPress = useHandler((walletId: string, currencyCode: string, _tokenId?: string, customAsset?: CustomAsset) => {
+  const handleWalletListPress = useHandler((walletId: string, currencyCode: string, tokenId?: string, customAsset?: CustomAsset) => {
```
samholmes, line 131: Interesting this wouldn't have been implemented. The only call-site where tokenId isn't passed is line 61 in WalletLIstCreateRow.tsx. If users can create token wallets from within the modal and the tokenId isn't passed, then this may mean we fail in this case. If only native wallets can can be created from the modal, then this isn't an issue.

samholmes (approved):
Approved with optional potential fix.

PR #4597, opened 2023-11-24
UI4: Transaction List Rows

Jon-edge:
image
image

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205826921045605


src/components/themed/TransactionListRow.tsx
```diff
@@ -163,36 +163,42 @@ export function TransactionListRow(props: Props) {
...
-          <View style={styles.categoryAndTimeContainer}>
-            {categoryText && <EdgeText style={styles.category}>{categoryText}</EdgeText>}
-            <EdgeText style={pendingStyle}>{pendingText}</EdgeText>
+      <FastImage style={styles.icon} source={{ uri: thumbnailPath }} />
+    </View>
+  )
+
+  return (
+    <CardUi4 icon={icon} onPress={handlePress} onLongPress={handleLongPress}>
+      <SectionView>
```
swansontec, line 185: Doesn't the card already include the `SectionView` functionality inside itself? This seems redundant.

src/components/themed/TransactionListRow.tsx
```diff
@@ -104,30 +101,43 @@ export function TransactionListRow(props: Props) {
...
+    action != null
+      ? TX_ACTION_LABEL_MAP[action.type]
+      : transaction.metadata && transaction.metadata.name
+      ? transaction.metadata.name
+      : (isSentTransaction ? lstrings.fragment_transaction_list_sent_prefix : lstrings.fragment_transaction_list_receive_prefix) + selectedCurrencyName
+
+  // Icon & Thumbnail
+  const thumbnailPath = useContactThumbnail(name)
+
+  const isSwapIcon = action != null && (action.type === 'swap' || action.type === 'swapOrderFill')
```
swansontec, line 115: Not all currency plugins use the new action system. We should also check `EdgeTransaction.swapData`, since we set that on *all* outgoing swap transactions.

swansontec (changes requested):
Only small changes needed here.

src/components/themed/TransactionListRow.tsx
Jon-edge: You're right, I think I previously had some special handling that tweaked things but since removed it.

swansontec (approved):

PR #4598, opened 2023-11-25
UI4: Wallet List Rows

Jon-edge:
image

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

https://github.com/EdgeApp/edge-react-gui/pull/4593

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205826921045599


src/components/ui4/CurrencyViewUi4.tsx
```diff
@@ -0,0 +1,80 @@
...
+ * A view representing the data from a wallet, used for rows, cards, etc.
+ */
+const CurrencyViewUi4Component = (props: Props) => {
+  const { nativeAmount, showRate = false, token, tokenId, wallet } = props
+  const { pluginId } = wallet.currencyInfo
+  const { showTokenNames = false } = SPECIAL_CURRENCY_INFO[pluginId] ?? {}
+  const theme = useTheme()
+
+  // Currency code for display:
+  const allTokens = wallet.currencyConfig.allTokens
```
swansontec, line 38: `const { allTokens } = wallet.currencyConfig`

src/components/ui4/CurrencyViewUi4.tsx
```diff
@@ -0,0 +1,80 @@
...
+    )
+  }
+
+  // Balance stuff:
+  const showBalance = useSelector(state => state.ui.settings.isAccountBalanceVisible)
+  const balance = useWalletBalance(wallet, tokenId)
+  const icon = <CryptoIcon sizeRem={2} tokenId={tokenId} walletId={wallet.id} />
+  const tickerText = showRate && wallet != null ? <TickerText wallet={wallet} tokenId={tokenId} /> : null
+  const cryptoText = showBalance ? <CryptoText wallet={wallet} tokenId={tokenId} nativeAmount={nativeAmount ?? balance} withSymbol /> : null
+  const fiatText = showBalance ? <FiatText nativeCryptoAmount={nativeAmount ?? balance} tokenId={tokenId} wallet={wallet} /> : null
```
swansontec: 👍 This is really clean & nice. It looks like all your hard work is really starting to pay off, now that we have these nice re-usable components for showing amounts in various ways. I am really impressed, Jon!

src/components/ui4/CurrencyViewUi4.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  let displayCurrencyCode = currencyCode
+  if (showTokenNames && tokenFromId != null) {
+    displayCurrencyCode = `${tokenFromId.displayName}`
```
swansontec, line 71: The `'${}'` bit seems useless?

src/locales/intl.ts
```diff
@@ -274,7 +274,17 @@ export const trimEnd = (val: string): string => {
...
+  percentVal,
+  intlOpts,
+  maxPrecision = 1,
+  minPrecision = 0
+}: {
+  percentVal: string | number
+  maxPrecision?: number
+  minPrecision?: number
+  intlOpts?: IntlNumberFormatOptionsType
+}): string => {
```
swansontec, line 287: I think the value to convert should remain as the first parameter. It's the only non-optional "option", and would never be omitted.

I also dislike destructuring in the argument list, especially with default values, since we are mixing interface & implementation. Maybe do:

```ts
export const toPercentString = (percentVal: string | number, opts: {...}): string {
  const {
    intlOpts,
    maxPrecision = 1,
    minPrecision = 0
  } = opts
```

I think this would be easier to use & understand, but I am open to keeping it as-is if you have strong opinions otherwise.

swansontec (changes requested):
It would be nice to see this PR organized differently, with new components introduced at the beginning, and then integrated into the wallet row at the end. I think that would flow better than tweaking the row step-by-step.

Jon-edge:
> It would be nice to see this PR organized differently, with new components introduced at the beginning, and then integrated into the wallet row at the end. I think that would flow better than tweaking the row step-by-step.

Thought there would be value in seeing the first diff for related components to highlight changes, but if not I'll just squash em. I'll see what I can do to rearrange.

src/components/ui4/AssetChangeTextUi4.tsx
```diff
@@ -1,69 +1,68 @@
...
-  if (gt(yesterdayDeltaPct, '0')) return { percentString: `+${percentString}`, deltaColorStyle: theme.positiveText }
-  return { percentString: `-${percentString}`, deltaColorStyle: theme.negativeText }
+  if (gt(yesterdayDeltaPct, '0')) {
+    return { percentString: `+${percentString}`, deltaColorStyle: theme.positiveText }
+  } else if (lt(yesterdayDeltaPct, '0')) {
+    return { percentString: `-${percentString}`, deltaColorStyle: theme.negativeTextMutedUi4 }
+  } else {
+    return { percentString, deltaColorStyle: theme.negativeTextMutedUi4 }
+  }
 }
```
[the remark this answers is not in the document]
swansontec: Thanks for rebasing the commits - it makes more sense to read now!

However, we still need to move the utility function at the bottom of the file.

swansontec (approved):

PR #4599, opened 2023-11-29
Auto Detect Tokens (GUI)

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-currency-accountbased/pull/678
https://github.com/EdgeApp/edge-core-js/pull/572

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205654034220620


src/reducers/CoreReducer.ts
```diff
@@ -52,5 +54,32 @@ export const core = combineReducers<CoreState, Action>({
...
     }
+  },
+
+  enabledDetectedTokens(state: { [walletId: string]: string[] } = {}, action: Action): { [walletId: string]: string[] } {
+    switch (action.type) {
+      case 'CORE/NEW_TOKENS': {
+        const newState = { ...state }
+        const { walletId, enablingTokenIds } = action.data
+
+        if (enablingTokenIds.length === 0) showError(`enabledDetectedTokens: Received empty tokens array for ${walletId}`)
```
swansontec, line 65: I don't see why this test is needed. Was it for debugging purposes?

src/__tests__/scenes/__snapshots__/CryptoExchangeQuoteScene.test.tsx.snap
```diff
@@ -1393,5 +1393,17 @@ exports[`CryptoExchangeQuoteScreenComponent should render with loading props 1`]
       </View>
     </RCTScrollView>
   </View>
+  <View
```
swansontec, line 1396: Don't these test changes go with the previous commit? This one didn't change any JSX.

swansontec (approved):
I am surprised by this PR. I thought we were going to use the new-transaction drop-down,  not the bottom notification area. The bottom area is more for persistent notifications that sit around a long time, not for one-shot events like this.

However, I do see that it works, so I presume you have a good reason for this.

Besides this, I would like you to delete the `showError` in the reducer - reducers shouldn't have side-effects like this.

src/reducers/CoreReducer.ts
Jon-edge: No, it's an assertion

Jon-edge:
> I am surprised by this PR. I thought we were going to use the new-transaction drop-down, not the bottom notification area. The bottom area is more for persistent notifications that sit around a long time, not for one-shot events like this.
> 
> However, I do see that it works, so I presume you have a good reason for this.
> 
> Besides this, I would like you to delete the `showError` in the reducer - reducers shouldn't have side-effects like this.

Fair point, we should probably add some persistence duration options later. Honestly there wasn't any spec on the appearance of the notification beyond the term "notification" so I took some liberties. 

The problem with the airship was that receiving a new token would overlap the auto-detect message with the tx received airship making it kind of messy. I also wanted to just get this all laid out so we have the option to eventually consolidate notifications neatly into a nicer looking component that can handle displaying multiple items. We could even expand the notification feature later with a dedicated scene and settings for the user to enable/disable notifications of interest.

src/__tests__/scenes/__snapshots__/CryptoExchangeQuoteScene.test.tsx.snap
Jon-edge: I went back and updated each commit with `yarn test -u`

PR #4604, opened 2023-11-29
Add SimpleTextInput

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205875781941080


src/components/icons/ThemedIcons.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return <IconRenderer>{fontSize => <ThemedAntDesignIcon name="search1" adjustsFontSizeToFit fontSize={fontSize} {...props} />}</IconRenderer>
+}
+
+const IconRenderer = (props: { children: (fontSize: SharedValue<number>) => React.ReactNode }) => {
+  const fontSize = useSharedValue(1)
+  const handleLayout = (event: LayoutChangeEvent) => {
+    fontSize.value = event.nativeEvent.layout.height
+  }
+  return <IconContainer onLayout={handleLayout}>{props.children(fontSize)}</IconContainer>
+}
```
swansontec, line 32: It seems like you are trying to fit the icon into the available space, which is a really bad design. If you get it wrong, you will end up with an infinite rendering loop, as the text size causes the layout to change which causes the text size to change, etc...

Also, using the text size prop is a bad idea in general, since it affects layout, and layout updates are expensive. It's much better to animate sizes using CSS transforms, not layout props like text size / margin / padding. That way we don't have to run the flexbox code on every frame.

So, I would much rather let the text input component drive the animation directly, and remove animation at the icon level - just handle the scaling using CSS transforms. Or, if that doesn't work for some reason, leave the text input in charge of animations, but allow the text height to be animated like you have it.

If those fail, then I would expect a very clear comment explaining what's going on here and what the dangers are - you don't want to just grab `SearchIcon` and use it wherever. It's a ticking time-bomb, designed just for use in a text input that knows what it's doing.

swansontec (changes requested):

swansontec (approved):

PR #4610, opened 2023-12-01
UI4: Promo Cards

Jon-edge:
image

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205826921045601

### DIFF
--- a/src/components/ui4/ButtonUi4.tsx
+++ b/src/components/ui4/ButtonUi4.tsx
@@ -1,0 +1,162 @@
+import * as React from 'react'
+import { ActivityIndicator, Text, TouchableOpacity, ViewStyle } from 'react-native'
+import LinearGradient from 'react-native-linear-gradient'
+import { cacheStyles } from 'react-native-patina'
+
+import { usePendingPress } from '../../hooks/usePendingPress'
+import { fixSides, mapSides, sidesToMargin } from '../../util/sides'
+import { Theme, useTheme } from '../services/ThemeContext'
+
+export type ButtonTypeUi4 = 'primary' | 'secondary' | 'tertiary'
+
+interface Props {
+  children?: React.ReactNode
+
+  // Called when the user presses the button.
+  // If the callback returns a promise, the button will disable itself
+  // and show a spinner until the promise resolves.
+  onPress?: () => void | Promise<void>
+
+  // Whether to center the button or stretch to fill the screen.
+  // Defaults to 'auto', letting the parent component be in charge:
+  alignSelf?: 'auto' | 'stretch' | 'center'
+
+  // True to dim the button & prevent interactions:
+  disabled?: boolean
+
+  // If this is set, the component will insert a text node after its children:
+  label?: string
+
+  // The gap around the button. Takes 0-4 numbers (top, right, bottom, left),
+  // using the same logic as the web `margin` property. Defaults to 0.
+  marginRem?: number[] | number
+
+  // True to show a spinner after the contents:
+  spinner?: boolean
+
+  // Which visual style to use. Defaults to primary (solid):
+  type?: ButtonTypeUi4
+}
+
+/**
+ * A stand-alone button to perform the primary action in a modal or scene.
+ */
+export function ButtonUi4(props: Props) {
+  const { alignSelf = 'center', children, disabled = false, label, marginRem, onPress, type = 'primary', spinner = false } = props
+
+  // `onPress` promise logic:
+  const [pending, handlePress] = usePendingPress(onPress)
+
+  // Styles:
+  const theme = useTheme()
+  const styles = getStyles(theme)
+
+  let buttonShadow, spinnerColor, textStyle, gradientProps
+  if (type === 'primary') {
+    textStyle = styles.primaryText
+    buttonShadow = styles.primaryShadow
+
+    spinnerColor = theme.buttonPrimaryUi4.spinnerColor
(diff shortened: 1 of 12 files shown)


src/components/ui4/ButtonUi4.tsx
```diff
@@ -0,0 +1,154 @@
+import * as React from 'react'
+import { ActivityIndicator, Text, TouchableOpacity, ViewStyle } from 'react-native'
+import LinearGradient from 'react-native-linear-gradient'
+import { cacheStyles } from 'react-native-patina'
+
+import { usePendingPress } from '../../hooks/usePendingPress'
+import { fixSides, mapSides, sidesToMargin } from '../../util/sides'
+import { Theme, useTheme } from '../services/ThemeContext'
+
+export type ButtonTypeUi4 = 'primary' | 'secondary' | 'tertiary'
```
swansontec, line 10: Please don't do "tertiary". We chose the word "escape" because this button style must only be used to leave a scene or modal. We want the styles to have meaning, not just "oooh I randomly like the way this looks." This "escape" makes it clear that this is the button style you use for cancel / exit / etc., and nothing else.

If "tertiary" is something that was decided on logically with the rest of the team, then I would love to hear the logic for when/how these styles should be used.

src/components/ui4/ButtonUi4.tsx
```diff
@@ -0,0 +1,154 @@
...
+  return (
+    <TouchableOpacity disabled={disabled || pending} style={[buttonShadow, styles.borderRadiusCommon]} onPress={handlePress}>
+      <LinearGradient {...gradientProps} style={[styles.gradientLayoutCommon, dynamicGradientStyles, styles.borderRadiusCommon]}>
+        {pending ? null : children}
+        {pending || label == null ? null : (
+          <Text adjustsFontSizeToFit minimumFontScale={0.75} numberOfLines={1} style={textStyle}>
+            {label}
+          </Text>
+        )}
+        {!pending && !spinner ? null : <ActivityIndicator color={spinnerColor} style={styles.spinnerCommon} />}
```
swansontec, line 90: Can we use this opportunity to fix an old `MainButton` mistake? Can we make `spinner` work the same as `pending`? So it's always `(pending || spinner) ...` for each of these three conditions. That way, passing `spinner={true}` as a prop would achieve the same result as having `onPress` return a promise.

src/types/Theme.ts
```diff
@@ -23,6 +23,20 @@ interface TextShadowParams {
   textShadowRadius: number
 }
 
+interface ButtomParamsUi4 {
+  textStyle: {
+    fontFamily: string
+    color: string
+  }
+  gradientProps: LinearGradientProps
+  shadowParams: ThemeShadowParams // TODO: Address android/ios inconsistencies with lib
```
swansontec, line 32: We need to specify exactly what's allowed in the gradient options, so please define `ThemeGradientProps` like you have defined `ThemeShadowParams` or even the loose `containerStyle` below. That way we only have colors in the theme, not arbitrary code.

swansontec (changes requested):
This is pretty much in-line with the other PR's I have reviewed. I have some questions about the "tertiary" idea, and want to tighten up the style definitions used for gradients.

src/components/ui4/ButtonUi4.tsx
Jon-edge: I don't want us to make up new terms (i.e. tile) making it a bit confusing to communicate between devs and everyone else. Primary, secondary, and tertiary are the standard UI/UX terms

https://www.codecademy.com/resources/docs/uiux/button

src/types/Theme.ts
Jon-edge: Actually on second thought - for this one, I intentionally specified
LinearGradientProps because I wanted the flexibility of controlling any aspect
of the gradient, not just a subset. In the implementation, I splat out the theme props.

src/components/ui4/ButtonUi4.tsx
```diff
@@ -78,16 +78,24 @@ export function ButtonUi4(props: Props) {
...
       <LinearGradient {...gradientProps} style={[styles.gradientLayoutCommon, dynamicGradientStyles, styles.borderRadiusCommon]}>
-        {pending ? null : children}
-        {pending || label == null ? null : (
-          <Text adjustsFontSizeToFit minimumFontScale={0.75} numberOfLines={1} style={textStyle}>
-            {label}
-          </Text>
-        )}
-        {!pending && !spinner ? null : <ActivityIndicator color={spinnerColor} style={styles.spinnerCommon} />}
+        {maybeSpinner}
+        {children}
```
swansontec, line 97: The current design hides the content when we show the spinner, but you are putting the spinner to the left of the content. This might be fine if we somehow dim the content, but we don't, so I think it will be confusing.

Can we just go back to hiding the content?

```ts
const hideContent = pending || spinning

<LinearGradient ...>
  { hideContent ? null : children }
  { hideContent ? null : maybeText }
  { !hideContent ? null : <ActivityIndicator ...> }
</LinearGradient>
```

swansontec (changes requested):

src/components/ui4/ButtonUi4.tsx
Jon-edge: Wasn't intentional. I was trying to keep it the same but misread the original implementation.

PR #4627, opened 2023-12-12
Sam/flip-input

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205826921045611


src/components/themed/FlipInput2.tsx
```diff
@@ -193,86 +194,151 @@ export const FlipInput2 = React.forwardRef<FlipInputRef, Props>((props: Props, r
...
+    marginTop: isIos ? 0 : -theme.rem(0.75),
+    marginBottom: isIos ? 0 : -theme.rem(1)
   }
 })
+
+function useAnimatedColorInterpolateFn(fromColor: string, toColor: string) {
+  const interpolateFn = useMemo(() => {
+    return (focusValue: SharedValue<number>) => {
+      'worklet'
+      return interpolateColor(focusValue.value, [0, 1], [fromColor, toColor])
```
swansontec, line 339: Please pass the "RGB" color space parameter: https://docs.swmansion.com/react-native-reanimated/docs/utilities/interpolateColor/#arguments

The default is "HSV" interpolation, which looks awful in almost every circumstance. This is probably why you had to change the transparent outline color a few commits ago - there should be no real difference between `#00000000` and `#00f1a200`, but with "HSV" interpolation, getting it wrong produces flashes or random colors.

swansontec (approved):
Only a few small things in here - not worth a re-review if you fix them.

src/components/icons/ThemedIcons.tsx
```diff
@@ -60,9 +61,13 @@ function makeFontIcon(IconComponent: React.ComponentType<VectorIconProps>, name:
 //
 
 const AnimatedAntDesignIcon = Animated.createAnimatedComponent(AntDesignIcon)
+const AnimatedFontello = Animated.createAnimatedComponent(Fontello)
 
 export const CloseIcon = makeFontIcon(AntDesignIcon, 'close')
 export const CloseIconAnimated = makeAnimatedFontIcon(AnimatedAntDesignIcon, 'close')
 
+export const FlipIcon = makeFontIcon(Fontello, 'exchange')
+export const FlipIconAnimated = makeAnimatedFontIcon(AnimatedFontello, 'exchange')
```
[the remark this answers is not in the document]
samholmes: We don't, but I think we should have the both animated and non-animated component available always for scene which need either.

src/components/themed/FlipInput2.tsx
samholmes: Looks like switching from HSV to RGB didn't have the effect I'm going for. Instead, what's required to get closer to the same effect would be to use HSV and the `useCorrectedHSVInterpolation` option. I think being explicit in the theme is the right move because it gives all the control up to the theme and is less verbose.

PR #4631, opened 2023-12-15
App Tabs and Search

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-react-gui/pull/4683

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1205826921045596

### DIFF
--- a/src/state/SceneDrawerState.tsx
+++ b/src/state/SceneDrawerState.tsx
@@ -1,0 +1,145 @@
+import { useMemo } from 'react'
+import { LayoutChangeEvent, Platform } from 'react-native'
+import { runOnJS, useAnimatedReaction, useSharedValue, withTiming } from 'react-native-reanimated'
+
+import { useHandler } from '../hooks/useHandler'
+import { useSharedEvent } from '../hooks/useSharedEvent'
+import { useState } from '../types/reactHooks'
+import { createStateProvider } from './createStateProvider'
+import { useSceneScrollContext } from './SceneScrollState'
+
+export const [SceneDrawerProvider, useSceneDrawerState] = createStateProvider(() => {
+  const [isRatioDisabled, setIsRatioDisabled] = useState(false)
+  const drawerHeight = useSharedValue(0)
+  const drawerOpenRatio = useSharedValue(1)
+  const drawerOpenRatioStart = useSharedValue(1)
+
+  return useMemo(
+    () => ({
+      drawerHeight,
+      drawerOpenRatio,
+      drawerOpenRatioStart,
+      isRatioDisabled,
+      setIsRatioDisabled
+    }),
+    [drawerHeight, drawerOpenRatio, drawerOpenRatioStart, isRatioDisabled]
+  )
+})
+
+export const useDrawerOpenRatio = () => {
+  const { scrollBeginEvent, scrollEndEvent, scrollMomentumBeginEvent, scrollMomentumEndEvent, scrollY } = useSceneScrollContext()
+
+  const scrollYStart = useSharedValue<number | undefined>(undefined)
+  const snapTo = useSharedValue<number | undefined>(undefined)
+  const { drawerHeight, drawerOpenRatio, drawerOpenRatioStart, isRatioDisabled, setIsRatioDisabled } = useSceneDrawerState()
+
+  function resetDrawerRatio() {
+    snapTo.value = 1
+  }
+
+  function snapWorklet() {
+    'worklet'
+    scrollYStart.value = undefined
+    snapTo.value = Math.round(drawerOpenRatio.value)
+  }
+  function snap() {
+    scrollYStart.value = undefined
+    snapTo.value = Math.round(drawerOpenRatio.value)
+  }
+  function delayedSnap() {
+    'worklet'
+    runOnJS(setTimeout)(snap, 300)
+  }
+
+  // Scroll event subscriptions:
+  useSharedEvent(scrollBeginEvent, nativeEvent => {
+    'worklet'
+    scrollYStart.value = nativeEvent?.contentOffset.y
+    drawerOpenRatioStart.value = drawerOpenRatio.value
+  })
+  useSharedEvent(scrollEndEvent, () => {
+    'worklet'
+    if (Platform.OS === 'android') {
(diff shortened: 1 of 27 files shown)


src/components/hoc/withContextProvider.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+export function withContextProvider<T>(getValue: () => T): [React.FunctionComponent<{ children: React.ReactNode }>, React.Context<T | undefined>] {
```
swansontec, line 3: This isn't really a HOC, since it takes a callback function, not a component. It needs another name and another home.

I am also worried about updating the context value on every render - I think this will force a re-render of each linked child. According to the docs, React uses `===` on the context value to decide whether or not an update has occurred. Using a raw `return {...}` will never be `===` equal, so the update will always take place.

So, if we want to implement these contexts correctly, we need to do something like this:

```ts
export const SceneDrawerContext = React.createContext(undefined)

export function SceneDrawerProvider(props: { children: React.ReactNode }) {
  const [isRatioDisabled, setIsRatioDisabled] = React.useState(false)
  const drawerHeight = useSharedValue(0)
  const drawerOpenRatio = useSharedValue(1)
  const drawerOpenRatioStart = useSharedValue(1)

  const value = React.useMemo(
    () => ({
      drawerHeight,
      drawerOpenRatio,
      drawerOpenRatioStart,
      isRatioDisabled,
      setIsRatioDisabled
    }),
    [drawerHeight, drawerOpenRatio, drawerOpenRatioStart, isRatioDisabled]
  )

  return <SceneDrawerContext.Provider value={value}>{props.children}</SceneDrawerContext.Provider>
}
```

I'm not sure if your `withContextProvider` helper function adds much, compared to just doing things directly as above. Maybe you can refactor your helper to somehow help with the memorization, but I don't have a clear idea about what this would look like.

Another idea is to combine `makeUseContextValue` and `withContextProvider` into a single step - `createStateProvider`. This would reflect the `StateProvier` nomenclature used later, and will create both the hook and the provider in one step.

src/state/index.ts
swansontec, line 1: We don't allow files named `index.ts`. I would put `const stateProviders = [SceneScrollProvider, SceneDrawerProvider]` into `services/Providers.tsx`, and just delete this file.

src/state/StateProvider.tsx
swansontec, line 1: I would also delete this file as well, and move the logic into `services/Providers.tsx`. It can be vastly simplified using `reduce`:

```ts
const stateProviders = [
  SceneScrollProvider,
  SceneDrawerProvider
]

export const renderStateProviders = (children: JSX.Element) => stateProviders.reduce(
  (element, Provider) => <Provider>{element}</Provider>,
  children
)

export function Providers(props: Props) {
  ... Redux setup code ...

  return (
    <ReduxProvider state={}>
      <LoginUiProvider themeOverride={theme}>
        {renderStateProviders(
          <MenuProvider><Airship><Main/></Airship></MenuProvider>
        )}
      </LoginUiProvider>
    </ReduxProvider>
  )
```

Then you can also use the `renderStateProviders(...)` function in `FakeProviders.tsx`.

Since this is a normal function that assembles JSX, instead of a React component, it's not only simpler, but also more efficient.

src/state/SceneDrawerState.tsx
```diff
@@ -8,14 +9,18 @@ import { useState } from '../types/reactHooks'
 import { makeUseContextValue } from '../util/makeUseContextValue'
 import { useSceneScrollContext } from './SceneScrollState'
 
+const SCROLL_DISTANCE = Dimensions.get('window').height / 10
```
swansontec, line 12: The window size can change when a tablet rotates. I guess this is fine, since this is just a fudge factor. We could even consider using `theme.rem`, or even a constant `dp` value.

swansontec (changes requested):
The state provider system needs a little clean-up, but the animation stuff seems OK. I would need a sidebar to understand it a little better, but I don't see any red flags aside from the window dimensions issue.

src/state/SceneScrollState.tsx
```diff
@@ -0,0 +1,65 @@
...
+export type SceneScrollHandler = (event: NativeSyntheticEvent<NativeScrollEvent>) => void
+
+export const useSceneScrollHandler = (): SceneScrollHandler => {
+  const sceneScrollContext = React.useContext(SceneScrollContext)
+  const dragStartY = useSharedValue(0)
+
+  const handler = useAnimatedScrollHandler({
+    onScroll: (nativeEvent: NativeScrollEvent) => {
+      'worklet'
+      if (sceneScrollContext == null) return
```
samholmes, line 31: Not needed check if we use the useSceneScrollContext hook

src/state/StateProvider.tsx
samholmes: Test this at runtime to make sure this approach works, and if so let's go with it. Even if the reducer doesn't work, let's still get rid of index.ts (leverage the helper function).

src/state/SceneDrawerState.tsx
samholmes: Maybe go with a constant (e.g. `theme.rem(3)`) or try leveraging the menu tab height again now that I have those constants in place.

src/state/createStateProvider.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * This creates a "state provider".
+ *
+ * @param hookFn the hook function to return the context value
+ * @returns The context provider component and the useContextHook
+ */
+export function createStateProvider<Value>(hookFn: () => Value): [React.FunctionComponent<{ children: React.ReactNode }>, () => Value] {
+  const Context = React.createContext<Value | undefined>(undefined)
+  function WithContext({ children }: { children: React.ReactNode }) {
```
swansontec: You could put `interface Props { children: React.ReactNode }` at the top, and then you could do `React.FunctionComponent<Props>` and `({children}: Props)` instead of duplicating the props type.

swansontec (approved):
Yeah, the state provider stuff looks really nice now!

PR #4639, opened 2023-12-19
UI4: Home Scene

Jon-edge:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

https://github.com/EdgeApp/edge-react-gui/pull/4632

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206061725717187


src/components/ui4/MarketsCardUi4.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const styles = getStyles(theme)
+  const defaultIsoFiat = useSelector(state => `iso:${getDefaultFiat(state)}`)
+  const defaultFiat = useSelector(state => getDefaultFiat(state))
+  const fiatSymbol = React.useMemo(() => getSymbolFromCurrency(defaultFiat), [defaultFiat])
+
+  const [coinRankingDatas, setCoinRankingDatas] = React.useState<CoinRankingData[]>([])
+
+  const timeoutHandler = React.useRef<Timeout | undefined>()
+
+  const handleCoinRow = useHandler((coinRow, index) => {
```
swansontec, line 47: Please call this `renderEventRow` instead. We use the word "handle" for things passed to `onWhatever` props.

Also remove the `useHandler` wrapper, for the same reason. Since we call the function immediately as part of the render, instead of passing it to a prop, there is no benefit to memoizing it.

src/components/ui4/CarouselUi4.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { useState } from 'react'
+import { View } from 'react-native'
+import Carousel, { Pagination } from 'react-native-snap-carousel'
+
+import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
+
+interface Props {
+  children: React.ReactNode | React.ReactNode[]
```
swansontec, line 8: The preferred syntax is `children?: React.ReactNode`. The `ReactNode` type already includes the array case.

src/components/ui4/CarouselUi4.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const numChildren = nonNullChildren.length
+
+  const [activeIndex, setActiveIndex] = useState(0)
+
+  const renderItem = ({ item }: { item: React.ReactNode }) => <View style={[styles.childContainer, { width: width * 0.9, height }]}>{item}</View>
+
+  if (children == null || numChildren === 0) return null
+
+  return (
+    <View>
```
swansontec, line 33: Does the `View` do anything? Could you use `<>` instead?

If it's a drop-in replacement, great! If it's not, don't waste any time on this one.

src/components/ui4/SplitRowsView.tsx
```diff
@@ -5,26 +5,33 @@ import { cacheStyles } from 'react-native-patina'
...
+
   return (
     <View style={styles.container}>
-      {children.map((row, index) => (
-        <View key={index} style={styles.row}>
-          <View style={styles.leftColumn}>{row.left}</View>
-          <View style={styles.rightColumn}>{row.right}</View>
-        </View>
-      ))}
+      {Array.isArray(children) ? children.map((row: SplitRow, index: number) => getSplitRow(row, `${index}-${row.left}`)) : getSplitRow(children)}
```
swansontec, line 34: Is it valid to put a React node into `${}`? It seems like the index alone would be a unique key.

src/components/ui4/scenes/HomeSceneUi4.tsx
```diff
@@ -0,0 +1,201 @@
...
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  )
+
+  return (
+    <NotificationSceneWrapper navigation={navigation} scroll hasTabs padding={theme.rem(TEMP_PADDING_REM)}>
+      <SectionView wideSpacing>
+        <>
```
swansontec, line 118: Do these fragments do anything? It seems like `{/* */}` comments or even blank lines could work just as well to divide the sections.

swansontec (changes requested):
Some small changes needed here.

src/components/ui4/scenes/HomeSceneUi4.tsx
Jon-edge: Yes, the fragments are telling the `SectionView` which children should be displayed together as a single section group. In this section, we have 3 children in one section.

swansontec: Oooh! That's really neat.

swansontec (approved):

PR #4653, opened 2023-12-22
Cleanups for new core

swansontec:
We need to separate the two cases where `tokenId: undefined` means "parent currency" vs. when it means "we don't have this information".

For the `GuiCurrencyInfo` case, we just make tokenId mandatory. For the `WalletListResult`, I used a discriminated union to remove tokenId from cases where it's not relevant.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

These are the preparatory fixes for #4641, as part of my PR review.

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206232712799533


src/components/modals/WalletListModal.tsx
samholmes, line 1: The type is called `WalletListResult`, so perhaps a change to the commit message is all that is needed to clarify this.

samholmes (approved):
Minor nit

PR #4655, opened 2023-12-25
UI4: Temp Styling Fudge Factor

Jon-edge:
Expose the margins/padding of some common UI4 components such as SectionView, SectionHeader, Row, and Card. We may not end up needing to use all exposed margins/paddings, but this task exposes them to make the UI4 transition easier.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device


src/components/ui4/CardUi4.tsx
```diff
@@ -24,6 +24,7 @@ interface Props {
 
   // DO NOT USE after a scene is fully UI4! Margins should all align without adjustment.
   marginRem?: number[] | number
+  paddingRem?: number[] | number
```
swansontec, line 27: Suggestion: You can add a comment like `/* @deprecated Only here during the UI4 transition */` to each of these props, to get a strike-through effect whenever somebody uses them. This will make the restriction very obvious, right at the call sites.

swansontec (approved):

PR #4663, opened 2023-12-29
Integrate new input components

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206145085315934


src/plugins/gui/components/GuiFormField.tsx
```diff
@@ -12,8 +12,8 @@ interface Props {
   fieldType?: FormFieldType
   key?: string
   autofocus?: boolean
-  returnKeyType?: OutlinedTextInputReturnKeyType
-  fieldRef?: React.Ref<OutlinedTextInputRef> | undefined
+  returnKeyType?: FilledTextInputReturnKeyType
+  fieldRef?: React.Ref<FilledTextInputRef> | undefined
```
swansontec, line 16: Because you have changed the ref type, you also need to update `src/plugins/gui/scenes/AddressFormScene.tsx`which still uses `OutlinedTextInputRef`.

src/components/modals/TextInputModal.tsx
```diff
@@ -95,24 +96,25 @@ export function TextInputModal(props: Props) {
...
-        multiline={multiline}
-        // Our props:
-        error={errorMessage}
-        marginRem={[1, 0.5, 1.5, 0.5]}
-        onChangeText={handleChangeText}
-        onSubmitEditing={handleSubmit}
-        value={text}
-        maxLength={maxLength}
-      />
+      <Space top={1} right={0.5} bottom={1.5} left={0.5} expand>
```
swansontec, line 99: `horizontal={0.5}`

swansontec (changes requested):
This branch doesn't build, because deleting `OutlinedTextInput` broke two type-level things. Specifically, `WalletAndCurrencyConstants.ts` uses `OutlinedTextInputProps['keyboardType']`, and another file uses `OutlinedTextInputRef`.

I think the correct solution for `WalletAndCurrencyConstants` is to define `inputType: 'default' | 'number-pad'`, so it's just the subset of keyboard types that's relevant to importing keys. That will break the dependency between the two files, and make it clearer what we expect here.

src/constants/WalletAndCurrencyConstants.ts
```diff
@@ -120,7 +120,7 @@ export interface ImportKeyOption {
     knowledgeBaseUri?: string
   }
   required: boolean
-  inputType: OutlinedTextInputProps['keyboardType']
+  inputType: FilledTextInputProps['keyboardType']
```
swansontec, line 123: I still think `inputType: 'default' | 'number-pad'` is the correct solution here. There is no reason to support every possible keyboard type.

swansontec (approved):
One change requested, but you can skip it if you don't like it.

PR #4672, opened 2024-01-02
Disable buy/sell tab buttons for Light Account

Jon-edge:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206262048565622

### DIFF
--- a/src/components/themed/MenuTabs.tsx
+++ b/src/components/themed/MenuTabs.tsx
@@ -2,6 +2,7 @@ import { BottomTabBarProps } from '@react-navigation/bottom-tabs'
 import * as React from 'react'
 import { useMemo } from 'react'
 import { TouchableOpacity, View } from 'react-native'
+import { AirshipBridge } from 'react-native-airship'
 import LinearGradient from 'react-native-linear-gradient'
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
 import Ionicon from 'react-native-vector-icons/Ionicons'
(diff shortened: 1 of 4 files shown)


src/components/themed/MenuTabs.tsx
```diff
@@ -63,9 +69,35 @@ export const MenuTabs = (props: BottomTabBarProps) => {
...
+        if (isLightAccount) {
+          Airship.show((bridge: AirshipBridge<BackupForTransferModalResult | undefined>) => {
+            return <BackupForTransferModal bridge={bridge} />
+          })
+            .then((userSel?: BackupForTransferModalResult) => {
+              if (userSel === 'upgrade') {
+                navigation.navigate('upgradeUsername', {})
+              }
+            })
+            .catch(error => showError(error))
```
swansontec, line 81: I wonder if you could make a helper function, so it's more like:

```ts
 await showUpgradeModal(() => navigation.navigate('upgradeUsername', {}))
```

That way 90% of the code vanishes, and just the `navigation` stuff has to be duplicated.

swansontec (approved):
Optional cleanup I just thought of.

src/components/themed/MenuTabs.tsx
Jon-edge: Ah yeah, I couldn't get it work earlier because I was trying to take a navigation as a prop directly

PR #4676, opened 2024-01-04
UI4: Merge legacy themes

Jon-edge:
image
image
image

- Merge legacy + UI4 themes where possible
- Use new lib for consistent android/ios icon shadows
- Update button styling according to latest spec
- Give full control of inter-button spacing to ButtonsViewUi4
- Deprecate CryptoIconUi4

### CHANGELOG

Does this branch warrant an entry to the CHANGELOG?

- [ ] Yes
- [x] No

### Dependencies

https://github.com/EdgeApp/edge-react-gui/pull/4673

### Requirements

If you have made **any** visual changes to the GUI. Make sure you have:

- [x] Tested on iOS device
- [x] Tested on Android device
- [ ] Tested on small-screen device (iPod Touch)
- [ ] Tested on large-screen device (tablet)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206254827085647

PR #4681, opened 2024-01-05
UI4: Replace Legacy Tile & Card

Jon-edge:
image
image
image
image
image
image
image
image

image

- Replace all legacy Card and Tile components with CardUi4 and RowUi4.
- At least the first tap from the tab or side menu items are restyled in this PR.
- Some additional changes for buttons will be added after rebase on top of latest jon/ui4/hack-legacy-buttons

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

https://github.com/EdgeApp/edge-react-gui/pull/4676
https://github.com/EdgeApp/edge-react-gui/pull/4673

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206210011193785


src/components/scenes/SendScene2.tsx
```diff
@@ -531,6 +532,8 @@ const SendComponent = (props: Props) => {
   const renderSelectFioAddress = () => {
     if (hiddenFeaturesMap.fioAddressSelect) return null
     const fioTarget = spendInfo.spendTargets.some(target => target.otherParams?.fioAddress != null)
+    if (!fioTarget) return null
```
paullinator, line 535: This seems safe but only due to distributed logic that we currently always send a fioTarget if we send a fioRequest although they are not technically both needed. What motivated this change?

src/components/settings/SettingsHeaderRow.tsx
```diff
@@ -13,33 +13,27 @@ interface Props {
...
   row: {
-    // Layout:
-    minHeight: theme.rem(2.75),
-    padding: theme.rem(1),
-
-    // Children:
     alignItems: 'center',
     flexDirection: 'row',
-    justifyContent: 'flex-start'
+    justifyContent: 'flex-end',
```
paullinator, line 35: why `flex-end`? `flex-start` seems more appropriate

paullinator (changes requested):

src/components/scenes/SendScene2.tsx
Jon-edge: Not sure... It may have been needed for some prior version of my changes, it doesn't seem
necessary anymore. Removing.

src/components/settings/SettingsHeaderRow.tsx
Jon-edge: You're right, but in this case it totally didn't matter either way because the
text fills the space while being left `textAligned`. Removed

paullinator (approved):

PR #4683, opened 2024-01-05
Fixes following SceneWrapper refactor

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device


src/components/scenes/CreateWalletCompletionScene.tsx
```diff
@@ -161,12 +161,12 @@ const CreateWalletCompletionComponent = (props: Props) => {
...
     <SceneWrapper background="theme">
-      {({ safeAreaInsets }) => (
-        <View style={[styles.content, { marginBottom: -safeAreaInsets.bottom }]}>
+      {({ insetStyles }) => (
+        <View style={[styles.content, insetStyles]}>
           <SceneHeader title={lstrings.title_create_wallets} withTopMargin />
           <FlashList
             automaticallyAdjustContentInsets={false}
-            contentContainerStyle={{ paddingTop: theme.rem(0.5), paddingBottom: safeAreaInsets.bottom }}
+            contentContainerStyle={{ paddingTop: theme.rem(0.5) }}
```
swansontec, line 169: This seems wrong to me. We should be stretching the scroll view outwards to meet the edge of the physical screen. It looks like you got it right on the other scenes in this commit, so this one is just different.

If fixing this is too hard, I see that this is at least less-broken than what we had before, so maybe we can leave it for later.

src/components/scenes/MigrateWalletCompletionScene.tsx
```diff
@@ -239,12 +239,11 @@ const MigrateWalletCompletionComponent = (props: Props) => {
 
   return (
     <SceneWrapper background="theme">
-      {({ safeAreaInsets }) => (
-        <View style={[styles.content, { marginBottom: -safeAreaInsets.bottom }]}>
+      {({ insetStyles }) => (
+        <View style={[styles.content, insetStyles]}>
```
swansontec, line 243: Same here. We want the list to go underneath the tab bar.

swansontec (approved):
Mostly approved, but there are two scenes where it looks like we aren't extending the scroll view correctly.

src/components/scenes/CreateWalletCompletionScene.tsx
samholmes: https://github.com/EdgeApp/edge-react-gui/assets/517469/22163368-e8b1-4ecc-8624-83bb329a22c6

The reason why it's different is because the bottom padding is in the View container instead of the scroll element. This is because the button is a secondary button so it doesn't look good to have it overlap. Maybe this floating button area should be in a bar with a background in the future.

src/components/scenes/MigrateWalletCompletionScene.tsx
samholmes: Same issue here. We're using a secondary button instead of a primary so we don't have a floating button

PR #4685, opened 2024-01-06
Blurred Dots Background

swansontec:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device


src/theme/variables/edgeDark.ts
```diff
@@ -121,10 +121,16 @@ export const edgeDark: Theme = {
   loadingIcon: palette.edgeMint,
 
   // Background
-  // backgroundGradientColors: [palette.navyAqua, palette.navyAquaDarker], // For vertical gradient
-  backgroundGradientColors: [palette.darkestNavy, palette.darkAqua],
-  backgroundGradientStart: { x: 0, y: 0 },
-  backgroundGradientEnd: { x: 1, y: 0 },
+  background: {
+    blurR: 100,
+    color: 'black',
```
samholmes, line 126: palette?

src/theme/variables/edgeDark.ts
```diff
@@ -121,10 +121,16 @@ export const edgeDark: Theme = {
   loadingIcon: palette.edgeMint,
 
   // Background
-  // backgroundGradientColors: [palette.navyAqua, palette.navyAquaDarker], // For vertical gradient
-  backgroundGradientColors: [palette.darkestNavy, palette.darkAqua],
-  backgroundGradientStart: { x: 0, y: 0 },
-  backgroundGradientEnd: { x: 1, y: 0 },
+  background: {
+    blurR: 100,
```
samholmes, line 125: blurRadius?

src/types/Theme.ts
```diff
@@ -97,9 +104,15 @@ export interface Theme {
...
-  backgroundGradientEnd: { x: number; y: number }
+  background: {
+    blurR: number
+    color: string
+    dotOpacity: number
+    dots: ThemeDot[]
+  }
+  // backgroundGradientColors: string[]
+  // backgroundGradientStart: { x: number; y: number }
+  // backgroundGradientEnd: { x: number; y: number }
```
samholmes, line 115: Cleanup comments

src/types/Theme.ts
```diff
@@ -97,9 +104,15 @@ export interface Theme {
...
   // Background
-  backgroundGradientColors: string[]
-  backgroundGradientStart: { x: number; y: number }
-  backgroundGradientEnd: { x: number; y: number }
+  background: {
+    blurR: number
+    color: string
+    dotOpacity: number
+    dots: ThemeDot[]
+  }
```
samholmes, line 112: I like the extra level for the theme file, should we be doing this for things like inputs and other components?

src/theme/variables/edgeLight.ts
```diff
@@ -116,9 +116,15 @@ export const edgeLight: Theme = {
...
-  backgroundGradientColors: [palette.lightestGray, palette.lightestGray],
-  backgroundGradientStart: { x: 0, y: 0 },
-  backgroundGradientEnd: { x: 0, y: 1 },
+  background: {
+    blurR: 100,
+    color: palette.lightestGray,
+    dotOpacity: 0.3,
+    dots: [
+      { color: '#15cb7f', cx: 100, cy: -200, r: 200 },
+      { color: '#7f15cb', cx: -100, cy: 200, r: 200 }
```
samholmes, line 125: Why isn't there a use of percentages for edgeLight?

src/theme/variables/testDark.ts
```diff
@@ -122,9 +122,16 @@ export const testDark: Theme = {
...
-  backgroundGradientStart: { x: 0, y: 0 },
-  backgroundGradientEnd: { x: 0, y: 1 },
+  background: {
+    blurR: 100,
+    color: 'black',
+    dotOpacity: 0.4,
+    dots: [
+      { color: '#997833', cx: '45%', cy: '-20%', r: 100 }, // Mid-right
+      { color: '#999999', cx: '-40%', cy: '-40%', r: 90 }, // Top-left
+      { color: '#15cb7f', cx: '-65%', cy: '50%', r: 250 } // Bottom left
```
samholmes, line 132: Why not introduce these colors to the theme palette?

src/theme/variables/edgeDark.ts
```diff
@@ -121,10 +121,16 @@ export const edgeDark: Theme = {
...
+    color: 'black',
+    dotOpacity: 0.4,
+    dots: [
+      { color: '#997833', cx: '45%', cy: '-20%', r: 100 }, // Mid-right
+      { color: '#999999', cx: '-40%', cy: '-40%', r: 90 }, // Top-left
+      { color: '#15cb7f', cx: '-65%', cy: '50%', r: 250 } // Bottom left
```
samholmes, line 131: Can we express the dot's radius in percentages?

src/components/ui4/DotsBackground.tsx
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * To do this, we slice a ring off the outside of each dot and
+   * replace it with the gradient. The sliced-off ring has a width of blurR,
+   * and the gradient ring has a width of 2 * blurR.
+   * This means the circle radius will expand by blurR.
+   *
+   * If the circle is smaller than blurR,
+   * then we can't slice off a big enough ring,
+   * so we reduce the opacity towards 0 to compensate.
+   */
```
samholmes, line 41: I have a vague idea on how this works. Looking closely it does look like its the correct implementation for this description, but I have to double check math more to verify.

src/components/ui4/DotsBackground.tsx
```diff
@@ -20,8 +20,9 @@ export function DotsBackground(props: Props): JSX.Element {
     accentDots[0] = { ...accentDots[0], color: accentColor }
   }
 
-  const [{ width, height }, setLayout] = React.useState({ width: 0, height: 0 })
+  const [{ width, height }, setLayout] = React.useState(cachedLayout)
   const handleLayout = useHandler((event: LayoutChangeEvent) => {
+    cachedLayout = event.nativeEvent.layout
```
samholmes, line 25: Could this safely be implemented in `useLayout`? I think so.

src/components/scenes/LoginScene.tsx
```diff
@@ -75,11 +71,6 @@ export function LoginSceneComponent(props: Props) {
   // Effects
   // ---------------------------------------------------------------------
 
-  const [backgroundImage = theme.backgroundImage] = useAsyncValue<ImageProp | undefined>(async () => {
-    const url = pickRandom(theme.backgroundImageServerUrls)
-    return await getBackgroundImage(disklet, url, theme.backgroundImage)
-  }, [disklet, theme])
```
samholmes, line 81: What if we could pass a "background component" to the app as part of the infra for white-labeling etc. The `DotsBackground` component just becomes a parameter to the app defined perhaps in the appConfig or something.

Can a component be a value for a context provider which all the dependents could access to render the appropriate component. This is a bit uncommon of a pattern, but I did consider trying out this pattern for the dynamic search bar to sort of "portal" the scene's drawer component (render function) to the MenuTabs component which is rendered in a disparate part of the component tree.

samholmes (changes requested):
My feedback is a few fixups regarding theming configuration and mainly a concern about configurability for white-labels. I think we have the opportunity to consider infrastructure here to render different background components whether some complex SVG, a simple background image, or a gradient of some kind. All of these requirements can be handled in a custom component which may be "used" throughout the app and configured for each white-label configuration.

src/components/scenes/TransactionListScene.tsx
```diff
@@ -248,10 +246,9 @@ function TransactionListComponent(props: Props) {
   })
 
   return (
-    <SceneWrapper hasNotifications hasTabs>
+    <SceneWrapper accentColor={iconColor} hasNotifications hasTabs>
       {({ insetStyles }) => (
         <>
-          <LinearGradient colors={[backgroundGradientColor, '#00000000']} start={{ x: 0, y: 0 }} end={{ x: 0, y: 1 }} style={StyleSheet.absoluteFill} />
```
samholmes, line 254: The design spec requires this remains:

https://app.asana.com/0/search?q=wallet%20colors&child=1205856947404193

Screenshot 2023-11-29 at 3 48 29 PM

samholmes (changes requested):
Also this shouldn't have been removed.

paullinator (changes requested):
Overall I see that the implementation works but I'm concerned that there was a lot of custom engineering work done for a specific design. A blurview on top of flat dots would have allowed more flexibility in the future should we ever want squares, triangles, or even images under the blurview. As is we'd have to write very specific code once again for other implementations.

src/components/ui4/DotsBackground.tsx
swansontec: Not really. I would totally use the `useLayout` hook, aside from this one problem.

Basically, every time we would transition from one scene to another, the background would glitch black as we waited for the layout to calculate. However, the calculation would always produce the same value, since we are full-screen on the phone every time. To solve the flash-of-black, I could either use `Dimensions.get(`window`)` as the default value for the layout, or I could use the last-seen value. I chose the last-seen value, which I cache in a global variable.

Since these assumptions are so specific to this one component, I don't see any way to incorporate them into the hook in a generic way.

src/components/ui4/DotsBackground.tsx
swansontec: To do a blur mathematically, imagine a circle of radius `blurRadius`, which slides over the scene from left to right. The average color inside the circle becomes the new color for the pixel at the very center of the circle. This imaginary circle is called the "filter kernel" in signal processing lingo.

As it slides, the filter kernel first begins to intersect the background dot at a distance of `blurRadius` away. As it continues to slide through, the ratio of dot color to background color increases, until the filter kernel is fully inside the background dot. This happens when the kernel is `blurRadius` in. So, the total distance from 0% to 100% opacity is `2 * blurRadius`.

The "filter kernel" corresponds to the camera's shutter or iris in physical reality, where a single point on the film receives color from the entire circular lens. If the camera is out-of-focus, this circle will be projected out into space and perform the averaging operation I just described.

src/theme/variables/edgeDark.ts
swansontec: I did it using the `scale` function.

src/components/ui4/DotsBackground.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const accentDots = [...dots]
+  if (accentColor != null) {
+    accentDots[0] = { ...accentDots[0], color: accentColor }
+  }
+
+  const [{ width, height }, setLayout] = React.useState({ width: 0, height: 0 })
+  const handleLayout = useHandler((event: LayoutChangeEvent) => {
+    setLayout(event.nativeEvent.layout)
+  })
```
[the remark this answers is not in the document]
swansontec: Neat! However, I can't use it here because of the caching hack.

src/components/scenes/LoginScene.tsx
swansontec: This is a good idea, but maybe out of scope for this PR.

src/theme/variables/testDark.ts
samholmes: Perhaps use the same or similar names from `edgeDark.ts`? If you have to add colors to those themes palettes, I say it's worth-while to keep things integrated for when design efforts are focused on these files.

src/components/common/SceneWrapper.tsx
```diff
@@ -29,8 +28,8 @@ interface SceneWrapperProps {
   // to changes in the gap.
   children: React.ReactNode | ((info: { safeAreaInsets: EdgeInsets; insets: EdgeInsets; insetStyles: InsetStyles }) => React.ReactNode)
 
-  // Settings for when using ScrollView
-  keyboardShouldPersistTaps?: 'always' | 'never' | 'handled'
+  // Adjusts the blurred dots background:
+  accentColor?: string
```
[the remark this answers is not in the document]
swansontec: I did it by adding a property to the theme, since the app shouldn't care how many dots the theme wants or where they are positioned.

src/types/Theme.ts
```diff
@@ -3,6 +3,9 @@ import { asNumber, asObject } from 'cleaners'
 export type ImageProp = { uri: string } | number
 
 export interface ThemeDot {
+  // We the dot color with the accent color, we want to keep the dot color,
+  // or drop the dot:
+  accent?: 'keep' | 'drop'
```
paullinator, line 8: Hmm. I won't fight back hard on this but this is another bizarre api at is SO specific to our current design. I would have still preferred a simple accent array that allows any of the dots to be changed programmatically after the Theme is applied.

src/components/scenes/TransactionListScene.tsx
```diff
@@ -249,6 +251,7 @@ function TransactionListComponent(props: Props) {
     <SceneWrapper accentColor={iconColor} hasNotifications hasTabs>
       {({ insetStyles }) => (
         <>
+          <LinearGradient colors={[backgroundGradientColor, '#00000000']} start={{ x: 0, y: 0 }} end={{ x: 0, y: 1 }} style={StyleSheet.absoluteFill} />
```
paullinator, line 254: Hard coding black in to the gradient will not allow us to do light mode or whitelabels with different colors. Due to time constraints, I won't push back on this today but I'll create a task to add the gradient color to the theme. Ideally something like

`theme.txListBackgroundGradient` and `theme.txDetailsBackgroundGradient` which would have the full gradient capabilities of arbitrary direction and colors.

paullinator:
@samholmes I added my comments but nothing that needs fixing in the PR. I'll let you give it the final approval

src/components/ui4/DotsBackground.tsx
samholmes: This seems like something that could be better understood by a video. Though, thanks for clarifying as it does help me better understand blur algorithms.

src/theme/variables/edgeDark.ts
```diff
@@ -121,10 +121,16 @@ export const edgeDark: Theme = {
   loadingIcon: palette.edgeMint,
 
   // Background
-  // backgroundGradientColors: [palette.navyAqua, palette.navyAquaDarker], // For vertical gradient
-  backgroundGradientColors: [palette.darkestNavy, palette.darkAqua],
```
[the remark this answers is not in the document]
samholmes: I don't see the backgroundGradient reinstated from the fixup commits.

samholmes (approved):
Approved with some optional feedback

src/theme/variables/testDark.ts
samholmes: I added an extra commit for test theme files

src/components/scenes/TransactionListScene.tsx
swansontec: Oh, good catch! We should use `theme.background.color for this one`.

samholmes (approved):

PR #4687, opened 2024-01-06
Add promo cards to HomeScene

Jon-edge:
Pending close button functionality

Screenshot 2024-01-11 at 3 44 22 PM
Screenshot 2024-01-11 at 3 44 28 PM

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206075308977979

### DIFF
--- a/src/components/ui4/PromoCardsUi4.tsx
+++ b/src/components/ui4/PromoCardsUi4.tsx
@@ -28,8 +28,9 @@ export const PromoCardsUi4 = (props: Props) => {
   // Check for PromoCard2 from info server:
   React.useEffect(() => {
     const osType = Platform.OS.toLowerCase()
-    const osVersion = DeviceInfo.getSystemVersion()
+    const osVersionRaw = DeviceInfo.getSystemVersion()
     const version = getVersion()
+    const osVersion = Array.from({ length: 3 }, (_, i) => osVersionRaw.split('.')[i] || '0').join('.')
 
     fetchInfo(`v1/inforollup/${config.appId ?? 'edge'}?os=${osType}&osVersion=${osVersion}&appVersion=${version}`)
       .then(async res => {
--- a/src/components/ui4/PromoCardsUi4.tsx
+++ b/src/components/ui4/PromoCardsUi4.tsx
@@ -1,0 +1,107 @@
+import { asArray } from 'cleaners'
+import { asInfoRollup, asPromoCard2, PromoCard2 } from 'edge-info-server/types'
+import * as React from 'react'
+import { Platform } from 'react-native'
+import DeviceInfo, { getBuildNumber, getVersion } from 'react-native-device-info'
+
+import { getCountryCodeByIp } from '../../actions/AccountReferralActions'
+import { getLocaleOrDefaultString } from '../../locales/intl'
+import { config } from '../../theme/appConfig'
+import { NavigationBase } from '../../types/routerTypes'
+import { fetchInfo } from '../../util/network'
+import { EdgeAnim } from '../common/EdgeAnim'
+import { useTheme } from '../services/ThemeContext'
+import { CarouselUi4 } from './CarouselUi4'
+import { PromoCardUi4 } from './PromoCardUi4'
+
+interface Props {
+  navigation: NavigationBase
+  screenWidth: number
+}
+
+export const PromoCardsUi4 = (props: Props) => {
+  const { navigation, screenWidth } = props
+  const theme = useTheme()
+
+  const [promos, setPromos] = React.useState<PromoCard2[]>([])
+
+  // Check for PromoCard2 from info server:
+  React.useEffect(() => {
+    const osType = Platform.OS.toLowerCase()
+    const osVersion = DeviceInfo.getSystemVersion()
+    const version = getVersion()
+
+    fetchInfo(`v1/inforollup/${config.appId ?? 'edge'}?os=${osType}&osVersion=${osVersion}&appVersion=${version}`)
+      .then(async res => {
+        if (!res.ok) {
+          console.error(await res.text())
+          return
+        }
+        const infoData = await res.json()
+        const infoPromoCards = asArray(asPromoCard2)(asInfoRollup(infoData).promoCards2)
+
+        const buildNumber = getBuildNumber()
+        const countryCode = await getCountryCodeByIp()
+        const currentIsoDate = new Date().toISOString()
+
+        setPromos(
+          infoPromoCards.filter(infoPromoCard => {
+            const { appVersion, minBuildNum, maxBuildNum, exactBuildNum, countryCodes, excludeCountryCodes, osTypes, osVersions, startIsoDate, endIsoDate } =
+              infoPromoCard
+
+            // Validate app version
+            // Ignore everything else if build version is specified and mismatched.
+            if (exactBuildNum != null && exactBuildNum !== buildNumber) return false
+
+            // Ignore min/max if app version is specified and mismatched.
+            if (appVersion != null && appVersion !== version) return false
+
+            // Look at min/max only if exact build or app version is not specified.
+            if (minBuildNum != null && minBuildNum > buildNumber) return false
+            if (maxBuildNum != null && maxBuildNum < buildNumber) return false
+
+            // Validate country
+            const isCountryInclude =
+              countryCodes == null ||
+              countryCodes.length === 0 ||
+              countryCodes.map(countryCode => countryCode.toLowerCase()).includes(countryCode.toLowerCase())
+            const isCountryExclude =
+              excludeCountryCodes != null &&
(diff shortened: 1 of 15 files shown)


src/components/ui4/scenes/HomeSceneUi4.tsx
```diff
@@ -71,6 +76,61 @@ export const HomeSceneUi4 = (props: Props) => {
       .catch(e => console.log(String(e)))
   }, [])
 
+  // Check for PromoCard2 from info server:
+  React.useEffect(() => {
+    fetchInfo(`v1/promoCards2/${config.appId ?? 'edge'}`)
```
paullinator, line 81: Use the `v2/rollup` endpoint which gives you EVERYTHING on the info server. Ideally this should be queried at app boot time and stored in redux for use by any component, but for now, query it in one useEffect and use it for both blogPosts and promocard2

src/components/ui4/scenes/HomeSceneUi4.tsx
```diff
@@ -71,6 +76,61 @@ export const HomeSceneUi4 = (props: Props) => {
...
+
+        setPromos(
+          infoPromoCards.filter(infoPromoCard => {
+            const { appVersion, minBuildNum, maxBuildNum, exactBuildNum, countryCodes, excludeCountryCodes, osTypes, osVersions, startIsoDate, endIsoDate } =
+              infoPromoCard
+
+            const isExactVersion = (appVersion != null && appVersion === version) || (exactBuildNum != null && exactBuildNum === buildNumber)
+            const isMinVersion = minBuildNum == null || minBuildNum <= buildNumber
+            const isMaxVersion = maxBuildNum == null || maxBuildNum >= buildNumber
+            const isVersionMatch = isExactVersion || (isMinVersion && isMaxVersion)
```
paullinator, line 101: isVersionMatch should also be true if there is no version specified in the card.

src/components/ui4/scenes/HomeSceneUi4.tsx
```diff
@@ -71,6 +76,61 @@ export const HomeSceneUi4 = (props: Props) => {
...
+
+            const isCountryInclude =
```
paullinator, line 103: This logic style is very difficult to follow and creates more code execution than necessary. Change this to use early exits once you know a card rule doesn't pass

ie.

```
if (appVersion != null && version !== appVersion) return false
if (minBuildNum != null && builderNumber < minBuildNum) return false
...
```

src/components/ui4/scenes/HomeSceneUi4.tsx
```diff
@@ -139,11 +199,19 @@ export const HomeSceneUi4 = (props: Props) => {
               />
             </EdgeAnim>
           </>
+          {promos == null || promos.length === 0 ? null : (
+            <EdgeAnim style={{ height: theme.rem(11.5) }} enter={{ type: 'fadeInUp', distance: 30 }}>
+              <CarouselUi4 height={theme.rem(10)} width={screenWidth}>
+                {promos.map(promoInfo => (
+                  <PromoCardUi4 navigation={navigation} promoInfo={promoInfo} key={promoInfo.localeMessages[0]} />
```
paullinator, line 206: Shouldn't this be `localeMessages[language]` or something like that

paullinator (changes requested):

src/components/ui4/scenes/HomeSceneUi4.tsx
Jon-edge: It's just the key and we're guaranteed the first `localeMessage`

Jon-edge:
Close functionality is broken in that the Carousel doesn't properly handle removals from its children array.

Punted and created a spinoff task: https://app.asana.com/0/1200382638405084/1206315466361645/f

src/components/ui4/scenes/HomeSceneUi4.tsx
Jon-edge: There's a bug in the `v1/infoRollup` endpoint. It does not properly clean the `osVersion` because it's using `asSemver`.

It should be using a string for android (`"api33"`) and a number for iOS. This will allow a distinction between the two in case somehow in the future there is an overlap.

Created a task:
 https://app.asana.com/0/1200382638405084/1206315466361651/f

paullinator: We shouldn't be using the Android API version, that's not granular enough as there can be regressions within releases with the same API version. See this chart

https://en.wikipedia.org/wiki/Android_version_history

We should be using the actual android version number which is semver based

src/components/ui4/scenes/HomeSceneUi4.tsx
paullinator: But we shouldn't be using the first message. We should be using the one that matches the language.

paullinator: We should be using the language not the first entry. That's why we even have a localeMessage in the first place.

src/components/ui4/scenes/HomeSceneUi4.tsx
paullinator: This is not a bug, it's a feature. We should be using the true Android version which is semver based, not the API level. There can be regressions in versions within the same API level so the API level isn't granular enough.

```
import DeviceInfo from 'react-native-device-info';

// Get the Android version
const androidVersion = DeviceInfo.getSystemVersion();
```

src/components/ui4/PromoCardsUi4.tsx
```diff
@@ -0,0 +1,109 @@
...
+            // Ignore everything else if build version is specified and mismatched.
+            if (exactBuildNum != null && exactBuildNum !== buildNumber) return false
+
+            // Ignore min/max if app version is specified and mismatched.
+            if (appVersion != null && appVersion !== version) return false
+
+            // Look at min/max only if exact build or app version is not specified.
+            const isMinVersion = minBuildNum == null || minBuildNum <= buildNumber
+            const isMaxVersion = maxBuildNum == null || maxBuildNum >= buildNumber
+            if (!isMinVersion || !isMaxVersion) return false
```
paullinator, line 60: I won't require a fixup but this should have just been
```
if (minBuildNum == null || minBuildNum > buildNumber) return false
if (maxBuildNum == null || maxBuildNum < buildNumber) return false
```

Get rid of those darn `isMinVersion` type variables. They're unneeded.

src/components/ui4/PromoCardsUi4.tsx
```diff
@@ -0,0 +1,109 @@
...
+            if (!isCountryInclude || isCountryExclude) return false
+
+            // Validate OS type
+            if (osTypes != null && osTypes.length > 0 && !osTypes.map(osType => osType.toLowerCase()).includes(osType)) return false
+
+            // Validate OS version
+            if (osVersions != null && osVersions.length > 0 && !osVersions.includes(osVersion.toString())) return false
+
+            // Validate date range
+            const startDate = startIsoDate != null && !isNaN(Date.parse(startIsoDate)) ? new Date(startIsoDate) : null
```
paullinator, line 80: Iso dates can be lexographically compared. No need to convert with `new Date`. Just do

```
const currentIsoDate = new Date().toISOString()
if (startIsoDate > currentIsoDate) return false
if (endIsoDate < currentIsoDate) return false
```

paullinator (changes requested):
Fix android version, use v1/rollup, use proper language

src/components/ui4/scenes/HomeSceneUi4.tsx
paullinator: See this chart https://en.wikipedia.org/wiki/Android_version_history

paullinator (changes requested):
I added a commit to clean the osVersion into semver. After rebasing on develop, visuals look broken. Needs to be fixed before merge.

paullinator (changes requested):
Added commit to normalize osVersion into semver. Some visuals needs fixing though

paullinator:
Broken visuals attached.

1. Learn and Promocard do not align on the left side. Need to fix while keeping the dots centered
2. Markets preview has totally broken spacing
3. Not quite "broken" but the CTA button on promo card is way too big. This was always supposed to have a different button component than ButtonUi4. Something like CardMiniButton which should have its own theme entry since it will likely need a different color.

Screenshot 2024-01-10 at 22 17 40
Screenshot 2024-01-10 at 22 17 32

src/components/ui4/PromoCardsUi4.tsx
Jon-edge: I don't want to assume that the dates given in the info data are correctly formed.

paullinator (approved):

paullinator (approved):

PR #4694, opened 2024-01-09
Use value directly in placeholderTextColor

peachbits:
---
  - https://app.asana.com/0/0/1206280885176690


samholmes (changes requested):
I would expect that passing the shared value to this property would allow for reanimated to animate the InputText property. If we pass the value directly, we get no animation behavior. If we don't want animation behavior, then the correct fixed is to change the variable from a `SharedValue<string>` to `string`

paullinator:
Ack. ok then a different fix is required

samholmes:
This PR commit the issue with placeholder color animation: https://github.com/EdgeApp/edge-react-gui/pull/4699/commits/7c1eab06c1bd64e197d8fd7a4d652e7633733959

PR #4699, opened 2024-01-10
Text Input Fixes

samholmes:
This allows for the color interpolation to actually animate.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206275754188804

### DIFF
--- a/src/components/themed/FilledTextInput.tsx
+++ b/src/components/themed/FilledTextInput.tsx
@@ -14,6 +14,7 @@ import Animated, {
 } from 'react-native-reanimated'
 
 import { useHandler } from '../../hooks/useHandler'
+import { SpaceProps, useSpaceStyle } from '../../hooks/useSpaceStyle'
 import { styled, styledWithRef } from '../hoc/styled'
 import { AnimatedIconComponent, CloseIconAnimated, EyeIconAnimated, EyeOffIconAnimated } from '../icons/ThemedIcons'
 import { useTheme } from '../services/ThemeContext'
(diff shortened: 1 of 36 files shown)


src/components/themed/FilledTextInput.tsx
```diff
@@ -273,8 +273,7 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
         </Container>
       </TouchableWithoutFeedback>
       <MessagesContainer>
-        {valid ? <Message>{valid}</Message> : null}
-        {error ? <Message danger>{error}</Message> : null}
+        <Message danger={error != null}>{valid || error || null}</Message>
```
paullinator: ?? instead of ||

src/components/themed/WalletListHeader.tsx
```diff
@@ -65,7 +66,7 @@ export class WalletListHeaderComponent extends React.PureComponent<Props> {
 
     return (
       <>
-        <View style={styles.searchContainer}>
+        <Space sideways around={0.5}>
```
paullinator, line 69: Why is this not using spaceProps?

paullinator (changes requested):

src/components/themed/WalletListHeader.tsx
samholmes: Technically, the Space component is using SpaceProps (that's all it uses). But, I think you're asking, why are we using a Space component here instead of putting the space prop in the SimpleTextInput. This is because we need a container component to layout the child elements in a row and the `sideways` prop acheives that.

paullinator (approved):

PR #4702, opened 2024-01-11
Paul/asset gradient

paullinator:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206303846046806

### DIFF
--- a/src/components/scenes/RequestScene.tsx
+++ b/src/components/scenes/RequestScene.tsx
@@ -12,17 +12,18 @@ import { toggleAccountBalanceVisibility } from '../../actions/LocalSettingsActio
 import { selectWalletToken } from '../../actions/WalletActions'
 import { Fontello } from '../../assets/vector'
 import { getSpecialCurrencyInfo, SPECIAL_CURRENCY_INFO } from '../../constants/WalletAndCurrencyConstants'
+import { useIconColor } from '../../hooks/useIconColor'
 import { lstrings } from '../../locales/strings'
 import { getDisplayDenomination, getExchangeDenomination } from '../../selectors/DenominationSelectors'
 import { getExchangeRate } from '../../selectors/WalletSelectors'
 import { config } from '../../theme/appConfig'
-import { connect } from '../../types/reactRedux'
+import { connect, useSelector } from '../../types/reactRedux'
 import { EdgeSceneProps, NavigationBase } from '../../types/routerTypes'
 import { GuiCurrencyInfo } from '../../types/types'
-import { getTokenIdForced, isKeysOnlyPlugin } from '../../util/CurrencyInfoHelpers'
+import { getTokenId, getTokenIdForced, isKeysOnlyPlugin } from '../../util/CurrencyInfoHelpers'
 import { getAvailableBalance, getWalletName } from '../../util/CurrencyWalletHelpers'
 import { triggerHaptic } from '../../util/haptic'
-import { convertNativeToDenomination, truncateDecimals, zeroString } from '../../util/utils'
+import { convertNativeToDenomination, darkenHexColor, truncateDecimals, zeroString } from '../../util/utils'
 import { EdgeAnim } from '../common/EdgeAnim'
 import { SceneWrapper } from '../common/SceneWrapper'
 import { AddressModal } from '../modals/AddressModal'
@@ -41,6 +42,7 @@ import { MainButton } from '../themed/MainButton'
 import { SceneHeader } from '../themed/SceneHeader'
 import { ShareButtons } from '../themed/ShareButtons'
 import { CardUi4 } from '../ui4/CardUi4'
+import { AccentColors } from '../ui4/DotsBackground'
 
 interface OwnProps extends EdgeSceneProps<'request'> {}
 
@@ -64,6 +66,10 @@ interface CurrencyMinimumPopupState {
   [pluginId: string]: ModalState
 }
 
+interface HookProps {
+  iconColor?: string
+}
+
 type Props = StateProps & DispatchProps & OwnProps & ThemeProps
 
 interface State {
@@ -82,7 +88,7 @@ interface AddressInfo {
 
 const inputAccessoryViewID: string = 'cancelHeaderId'
 
-export class RequestSceneComponent extends React.Component<Props, State> {
+export class RequestSceneComponent extends React.Component<Props & HookProps, State> {
   flipInputRef: React.RefObject<ExchangedFlipInputRef>
   unsubscribeAddressChanged: (() => void) | undefined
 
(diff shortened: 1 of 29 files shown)


src/types/Theme.ts
```diff
@@ -92,9 +92,11 @@ export interface Theme {
   loadingIcon: string
 
   // Background
+  backgroundGradientColors: string[]
+  backgroundGradientStart: { x: number; y: number }
+  backgroundGradientEnd: { x: number; y: number }
```
samholmes, line 97: Why not keep this in the `background:` field as:

```js
color?: string // maybe not optional
gradient?: {
  colors: string[]
  start: Point
  end: Point
}
```
Now you can use `background.gradient != null` as the condition for rendering the `MaybeLinearGradient` component.

src/util/utils.ts
```diff
@@ -595,3 +595,33 @@ export const base58ToUuid = (base58String: string): string => {
...
+
+  // Parse the hexadecimal values
+  const r = parseInt(hexColor.slice(0, 2), 16)
+  const g = parseInt(hexColor.slice(2, 4), 16)
+  const b = parseInt(hexColor.slice(4, 6), 16)
+
+  // Multiply each color component by the scale factor
+  const scaledR = Math.round(r * scaleFactor)
+  const scaledG = Math.round(g * scaleFactor)
+  const scaledB = Math.round(b * scaleFactor)
```
samholmes, line 621: How do you deal with overflow?

src/components/scenes/TransactionListScene.tsx
```diff
@@ -274,16 +271,26 @@ function TransactionListComponent(props: Props) {
   // Change the second dot color and remove the 3rd dot
   const overrideDots: UpdateDots = [undefined, { color: iconColor }, null]
 
+  const backgroundColors = [...theme.assetBackgroundGradientColors]
+  if (iconColor != null) {
+    const scaledColor = multiplyHexColor(iconColor, theme.assetBackgroundColorScale)
+    backgroundColors[0] = scaledColor
```
samholmes, line 277: This scaling seems like a change rather than a refactor.

src/components/ui4/DotsBackground.tsx
```diff
@@ -9,11 +9,15 @@ import { useTheme } from '../services/ThemeContext'
 
 export type UpdateDots = Array<Partial<ThemeDot> | undefined | null>
 interface Props {
+  // Optional backgroundGradient overrides
+  backgroundGradientColors?: string[]
+  backgroundGradientStart?: { x: number; y: number }
+  backgroundGradientEnd?: { x: number; y: number }
```
samholmes, line 15: If `ThemeGradient` was a type, then it could be one field:

```ts
gradient?: ThemeGradient
```

src/components/scenes/RequestScene.tsx
```diff
@@ -82,7 +88,7 @@ interface AddressInfo {
 
 const inputAccessoryViewID: string = 'cancelHeaderId'
 
-export class RequestSceneComponent extends React.Component<Props, State> {
+export class RequestSceneComponent extends React.Component<Props & HookProps, State> {
```
samholmes, line 91: Conjoin HookProps in the definition of `Props`.

src/hooks/useIconColor.ts
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react'
+import { getColors } from 'react-native-image-colors'
+
+import { useState } from '../types/reactHooks'
+import { EdgeAsset } from '../types/types'
+import { getCurrencyIconUris } from '../util/CdnUris'
+
+export const useIconColor = (edgeAsset: EdgeAsset): string | undefined => {
```
samholmes, line 8: Why aren't we using this this `TransactionListScene` for consistency?

samholmes (changes requested):

src/types/Theme.ts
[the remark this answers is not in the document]
paullinator: can't easily search for it and it is no longer the entire background specification. Only the dots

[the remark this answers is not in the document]
samholmes: `backgroundDots.dots`? This field is the app's background, and it contains dots. If you want to more search-ability, I'm okay with `appBackground` even.

src/util/utils.ts
```diff
@@ -596,7 +596,14 @@ export const base58ToUuid = (base58String: string): string => {
...
 }
 
-export const multiplyHexColor = (hexColor: string, scaleFactor: number): string => {
+/**
+ * Darken a color by a scale factor.
+ * @param hexColor of format '#1f1f1f1f'
+ * @param scaleFactor 0-1 with 0 being black, 1 is unchanged
+ * @returns darkened hex color string
+ */
+export const darkenHexColor = (hexColor: string, scaleFactor: number): string => {
```
samholmes, line 605: Ah this makes a lot more sense. Can we keep the alpha channel for the hexColor?

src/types/Theme.ts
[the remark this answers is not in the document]
paullinator: this is everything in `backgroundDots`

```    
    blurRadius: number
    dotOpacity: number
    dots: ThemeDot[]
```

Looks to me to be very specifically about the dots

src/components/ui4/DotsBackground.tsx
samholmes: `backgroundGradient` is fine too. What I'm golfing for is keeping these values together under one type.

src/components/scenes/RequestScene.tsx
paullinator: Because `RequestSceneConnected` specifically Receives a combo of just `OwnProps` and `HookProps`. Only the `RequestSceneComponent` gets everything. I needed to bifurcate the prop types.

src/components/scenes/TransactionListScene.tsx
samholmes: Ah, I see, so it's a way to achieve the ability to darken the background color from the theme file (because it's currently set to `#00000000`, transparent black). Now we can darken or lighten the value...wait, we can lighten the value if we pass a scale > 1, so clipping is still an issue that should be handled in the "darkenHexColor" function. I'd argue "scaleHexColor" might be a better name because darkening isn't the only use.

src/types/Theme.ts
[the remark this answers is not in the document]
samholmes: Ah, cause we moved background color out of there. That makes sense.

src/components/ui4/DotsBackground.tsx
```diff
@@ -86,12 +87,18 @@ export function DotsBackground(props: Props): JSX.Element {
...
 
   return (
-    <View style={[StyleSheet.absoluteFill, { backgroundColor: color }]} onLayout={handleLayout}>
+    <LinearGradient
+      style={StyleSheet.absoluteFill}
+      colors={theme.backgroundGradientColors}
+      end={theme.backgroundGradientEnd}
+      start={theme.backgroundGradientStart}
+      onLayout={handleLayout}
+    >
```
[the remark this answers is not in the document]
samholmes: We don't have to.

src/hooks/useIconColor.ts
samholmes: My suggestion was to gut the icon grab out of the `CyptoIconUi4` in favor of this hook. But, no worries if not.

samholmes (approved):
I'm approving this as I think it's ready. The only concern I have is with the hex color function which doesn't clip the color values.

PR #4710, opened 2024-01-13
Fix and Tweak Animations

paullinator:
Disable layout animations that are also gesture animated. These are broken on Android

### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206322748919970
  - https://app.asana.com/0/0/1206124087516563

### DIFF
--- a/src/components/common/EdgeAnim.tsx
+++ b/src/components/common/EdgeAnim.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { ViewProps } from 'react-native'
+import { Platform, ViewProps } from 'react-native'
 import Animated, {
   AnimateProps,
   ComplexAnimationBuilder,
@@ -35,6 +35,9 @@ interface Anim {
 interface Props extends AnimateProps<ViewProps> {
   enter?: Anim
   exit?: Anim
+
+  /** only animate on ios */
+  ios?: boolean
   visible?: boolean
 }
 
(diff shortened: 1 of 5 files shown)


src/components/common/EdgeAnim.tsx
```diff
@@ -33,6 +34,12 @@ interface Anim {
 }
 
 interface Props extends AnimateProps<ViewProps> {
+  /**
+   * disable animation
+   * anim => disable animation but still render a container view
+   * view => render the children with no container view
+   * */
+  disable?: 'anim' | 'view'
```
Jon-edge, line 42: Why is this an enumeration? I think it's clear through the API and component name that a boolean true here means just the lack of animation. 

In general I think it's clearer to have named value enums, though the rest of our app has a pattern of `disable` being a boolean flag. One less thing to hover would be nice and usage would have less code. 

If there's future plans to have more disable configuration types then a different prop name seems more appropriate

src/components/scenes/CoinRankingScene.tsx
```diff
@@ -81,8 +81,9 @@ const CoinRankingComponent = (props: Props) => {
     const key = `${index}-${item}-${rank}-${currencyCode}-${lastUsedFiat}`
     debugLog(LOG_COINRANK, `renderItem ${key.toString()}`)
 
+    const disable = index >= MAX_LIST_ITEMS_ANIM ? 'view' : undefined
```
Jon-edge, line 84: If `disable` we're just a boolean this would simply be `const disable = index >= MAX_LIST_ITEMS_ANIM`

Jon-edge (changes requested):

src/components/scenes/CoinRankingScene.tsx
paullinator: As mentioned above. There are necessarily two types of view disables so that we don't render extra Views if not needed.

Jon-edge (approved):

PR #4716, opened 2024-01-16
Tweak FilledTextInput spacing

paullinator:
Also add more test cases to InputTester

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206354619455584

### DIFF
--- a/src/components/themed/FilledTextInput.tsx
+++ b/src/components/themed/FilledTextInput.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import { useMemo } from 'react'
-import { ActivityIndicator, Platform, TextInput, TouchableOpacity, TouchableWithoutFeedback, View } from 'react-native'
+import { ActivityIndicator, Platform, Text, TextInput, TouchableOpacity, TouchableWithoutFeedback, View } from 'react-native'
 import Animated, {
   interpolate,
   interpolateColor,
@@ -18,10 +18,10 @@ import { SpaceProps, useSpaceStyle } from '../../hooks/useSpaceStyle'
 import { styled, styledWithRef } from '../hoc/styled'
 import { AnimatedIconComponent, CloseIconAnimated, EyeIconAnimated } from '../icons/ThemedIcons'
 import { useTheme } from '../services/ThemeContext'
-import { EdgeText } from './EdgeText'
 import { NumericInput } from './NumericInput'
 
 const AnimatedTextInput = Animated.createAnimatedComponent(TextInput)
+const isAndroid = Platform.OS === 'android'
 
 export type FilledTextInputReturnKeyType = 'done' | 'go' | 'next' | 'search' | 'send' // Defaults to 'done'
 
@@ -132,6 +132,8 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
   const hasIcon = LeftIcon != null
   const hasValue = value !== ''
 
+  const spaceStyle = useSpaceStyle(spaceProps)
+
   // Show/Hide password input:
   const [hidePassword, setHidePassword] = React.useState(secureTextEntry ?? false)
   const handleHidePassword = () => setHidePassword(!hidePassword)
@@ -207,9 +209,9 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
   const InputComponent = numeric ? StyledNumericInput : StyledAnimatedTextInput
 
   return (
-    <>
+    <View style={spaceStyle}>
       <TouchableWithoutFeedback accessible={false} testID={testID} onPress={() => focus()}>
-        <Container disableAnimation={disableAnimation} focusAnimation={focusAnimation} multiline={multiline} scale={scale} spaceProps={spaceProps}>
+        <Container disableAnimation={disableAnimation} focusAnimation={focusAnimation} multiline={multiline} scale={scale}>
           <SideContainer scale={leftIconSize}>{LeftIcon == null ? null : <LeftIcon color={iconColor} size={leftIconSize} />}</SideContainer>
 
           <InnerContainer focusValue={focusValue} hasPlaceholder={placeholder != null}>
@@ -280,7 +282,7 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
           <Message>{charactersLeft}</Message>
         </MessagesContainer>
       ) : null}
-    </>
+    </View>
   )
 })
 
@@ -289,8 +291,7 @@ const Container = styled(Animated.View)<{
   focusAnimation: SharedValue<number>
   multiline: boolean
   scale: SharedValue<number>
-  spaceProps: SpaceProps
-}>(theme => ({ disableAnimation, focusAnimation, multiline, scale, spaceProps }) => {
+}>(theme => ({ disableAnimation, focusAnimation, multiline, scale }) => {
   const rem = theme.rem(1)
   const interpolateInputBackgroundColor = useAnimatedColorInterpolateFn(
     theme.textInputBackgroundColor,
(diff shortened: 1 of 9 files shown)


paullinator:
Screenshot 2024-01-16 at 10 19 48

src/components/themed/FilledTextInput.tsx
```diff
@@ -312,14 +313,14 @@ const Container = styled(Animated.View)<{
...
-      paddingHorizontal: theme.rem(1)
+      paddingHorizontal: theme.rem(0.4)
     },
     useAnimatedStyle(() => ({
       backgroundColor: interpolateInputBackgroundColor(focusAnimation, disableAnimation),
       borderColor: interpolateOutlineColor(focusAnimation, disableAnimation),
       opacity: interpolate(scale.value, [1, 0.5], [1, 0]),
       marginHorizontal: interpolate(scale.value, [1, 0], [0, 2 * rem]),
-      paddingVertical: scale.value * 0.8 * rem
+      paddingVertical: scale.value * 0.75 * rem
```
samholmes, line 323: I feel like we would need the extra vertical space to make it feel less crammed.

src/components/themed/FilledTextInput.tsx
```diff
@@ -346,15 +347,18 @@ const InnerContainer = styled(Animated.View)<{
   hasPlaceholder: boolean
 }>(theme => ({ hasPlaceholder, focusValue }) => {
   const rem = theme.rem(1)
+  const androidHShift = isAndroid ? rem / 8 : 0
+
   return [
     {
+      left: androidHShift,
```
samholmes, line 354: Why the Android specific rule here?

src/components/themed/FilledTextInput.tsx
```diff
@@ -346,15 +347,18 @@ const InnerContainer = styled(Animated.View)<{
   hasPlaceholder: boolean
 }>(theme => ({ hasPlaceholder, focusValue }) => {
   const rem = theme.rem(1)
+  const androidHShift = isAndroid ? rem / 8 : 0
```
samholmes, line 350: Why device by 8?

src/components/themed/FilledTextInput.tsx
```diff
@@ -365,7 +369,8 @@ const InnerContainer = styled(Animated.View)<{
 
 const PrefixAnimatedText = styled(Animated.Text)<{ visibility: SharedValue<number> }>(theme => ({ visibility }) => {
   const rem = theme.rem(1)
-  const isAndroid = Platform.OS === 'android'
+  const androidVShift = isAndroid ? rem / 16 : 0
```
samholmes, line 372: Why device by 16?

src/components/themed/FilledTextInput.tsx
```diff
@@ -375,34 +380,49 @@ const PrefixAnimatedText = styled(Animated.Text)<{ visibility: SharedValue<numbe
     useAnimatedStyle(() => {
       return {
         opacity: visibility.value,
-        top: isAndroid ? -1 : 0,
+        top: androidVShift,
```
samholmes, line 383: I never knew why there was this android specific rule here.

src/components/themed/FilledTextInput.tsx
```diff
@@ -375,34 +380,49 @@ const PrefixAnimatedText = styled(Animated.Text)<{ visibility: SharedValue<numbe
...
-}))
+const SuffixText = styled(EdgeText)(theme => {
+  const rem3 = theme.rem(3)
+  const androidVShift = isAndroid ? rem3 / 16 : 0
+
+  return {
+    color: theme.secondaryText,
+    fontFamily: theme.fontFaceDefault,
+    includeFontPadding: false,
+    marginTop: androidVShift,
```
samholmes, line 398: What is the root cause for the special casing between platforms, and can we isolate it to where it originates instead of spread it out throughout the SimpleTextInput component?

src/components/themed/FilledTextInput.tsx
```diff
@@ -429,7 +449,7 @@ const PlaceholderText = styled(Animated.Text)<{
     useAnimatedStyle(() => {
       return {
         color: interpolatePlaceholderTextColor(focusAnimation, disableAnimation),
-        fontSize: interpolate(shift.value, [0, 1], [fontSizeBase, 0.8 * fontSizeBase])
+        fontSize: interpolate(shift.value, [0, 1], [fontSizeBase, 0.75 * fontSizeBase])
```
samholmes, line 452: Ah, could this be the reason for the vertical padding change in the container component? If so, I think keeping the extra padding would give the component more room visually. One of my comments is on the screenshots showing the component to be looking a little crowded when in focus, perhaps we can alleviate this.

samholmes (changes requested):
In addition to my comments, the prefix is not vertically aligned with the input text from the visuals provided in the PR. This should be adjusted so that the prefix aligns with the input text just like the suffix does (should?).

src/components/themed/FilledTextInput.tsx
samholmes: Let's go with .25 rem increments on all our designs. So, change SimpleTextInput to follow this design guideline.

src/components/themed/FilledTextInput.tsx
samholmes: Look into it, but if too cumbersome, it's okay for now.

src/components/themed/FilledTextInput.tsx
samholmes: Add comment `// 2 pixels based on 16 pixel rem`

src/components/themed/FilledTextInput.tsx
samholmes: Add comment `// 1 pixel based on 16 pixel rem`

src/components/themed/FilledTextInput.tsx
paullinator: This originates from the InputField rendering lower than a Text field. It seems the InputField is more "incorrect" so I moved the shift to the input fields instead.

src/components/themed/FilledTextInput.tsx
paullinator: comments added

src/components/themed/FilledTextInput.tsx
paullinator: comments added

src/components/themed/FilledTextInput.tsx
paullinator: comments added

paullinator:
Screenshot 2024-01-17 at 17 38 42

paullinator:
Ignore blue background. That was removed in the fixup

src/components/themed/FilledTextInput.tsx
```diff
@@ -347,6 +345,8 @@ const InnerContainer = styled(Animated.View)<{
   hasPlaceholder: boolean
 }>(theme => ({ hasPlaceholder, focusValue }) => {
   const rem = theme.rem(1)
+
+  // Need 2 pixels of shift given a 16 point rem settings
```
samholmes, line 349: Still it's weird that you need to shift the container two pixels to the right on Android. Why would Android be rendering views differently?

src/components/themed/FilledTextInput.tsx
```diff
@@ -208,9 +209,9 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
   const InputComponent = numeric ? StyledNumericInput : StyledAnimatedTextInput
 
   return (
-    <>
+    <View style={spaceStyle}>
```
samholmes, line 212: Why was this done? Isn't the whole point of using the space props within the component to avoid the extra view? If this is our result, we might as well just use a <Space> component around the input when we want the space props, at least then we'd only be rendering an extra view when we actually want the space.

paullinator: This fixes the bug where extra margin was added by a space prop but inbetween the input box and the bottom right message. it's too late in the sprint to make a change that would break the API given how many places use this component

src/components/themed/FilledTextInput.tsx
paullinator: Cant tell you why but that seems to be the case with text inputs vs text views

samholmes: This is an Animated.View not a TextInput

paullinator: Yes, but it's the view of the TextInput

src/components/themed/FilledTextInput.tsx
samholmes: In case, we can just use a Space component and pass the space props directly to it. Also, override sideways to false.

PR #4721, opened 2024-01-17
Refactor SceneWrapper, header, and footer

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206347017487102


src/components/scenes/CryptoExchangeQuoteScene.tsx
```diff
@@ -184,31 +184,38 @@ export const CryptoExchangeQuoteScene = (props: Props) => {
...
-                type="warning"
-                onPress={handleCanBePartialExplanation}
-              />
-            ) : null}
-
-            <Slider parentStyle={styles.slider} onSlidingComplete={doShift} disabled={pending} showSpinner={pending} />
-            {renderTimer()}
-          </ScrollView>
+          <View style={{ ...undoInsetStyle, marginTop: 0 }}>
+            <ScrollView contentContainerStyle={[{ ...insetStyle, paddingTop: 0 }, styles.container]}>
```
swansontec, line 191: You don't need to add a `View` here. The `ScrollView` itself takes a `style` prop.

As a side note, all the builtin react-native lists take a `style` prop, but the `FlashList` does not. This is another reason to move away from `FlashList`

src/components/scenes/MigrateWalletCompletionScene.tsx
```diff
@@ -227,28 +227,31 @@ const MigrateWalletCompletionComponent = (props: Props) => {
...
-      <MainButton
-        spinner={!done}
-        disabled={!done}
-        label={!done ? undefined : lstrings.string_done_cap}
-        type="secondary"
-        marginRem={[1]}
-        onPress={() => navigation.navigate('walletsTab', { screen: 'walletList' })}
-        alignSelf="center"
-      />
+      <View style={styles.bottomButton}>
```
swansontec, line 230: Rant: This is what Jon has been harping on - we should be using the `ButtonsView` to put a button on a scene in a particular layout, instead of centering it ourselves.

*However*, I see that this is a special case, since we are floating the button at the bottom of the screen. Since we are in a hurry, there is no need to go adding a special mode to the `ButtonsView`! Arguably this could even go in the scene footer area, but again, no need to do that in this PR.

src/components/ui4/BlurBackground.tsx
```diff
@@ -11,5 +13,14 @@ export interface BlurBackgroundProps {
...
+
+  return <BlurView blurType={theme.isDark ? 'dark' : 'light'} style={styles.blurView} overlayColor={overlayColor} />
 }
+
+const getStyles = cacheStyles(() => ({
+  blurView: {
+    ...StyleSheet.absoluteFillObject,
+    backgroundColor: isAndroid ? '#00000055' : undefined
+  }
+}))
```
swansontec: This is all wrong. By default, the `BlurView` looks consistent on both iOS and Android. However, Android lets you manually adjust the tint color using the `overlayColor`, so we ourselves started making Android look different. But now we don't like the inconsistency, so we tweak it back a different way!!!

What about just deleting the `overlayColor` prop and letting things automatically be the same on both platforms? This will also deal with the light-mode, which your current solution doesn't address.

src/components/scenes/CoinRankingScene.tsx
```diff
@@ -242,8 +239,7 @@ const CoinRankingComponent = (props: Props) => {
           </View>
           <DividerLine marginRem={[0, 0, 0, 1]} />
           <View style={{ ...undoInsetStyle, marginTop: 0 }}>
-            <AnimatedFlashList
-              estimatedItemSize={theme.rem(3.75)}
+            <Animated.FlatList
```
swansontec, line 242: Interesting. Does this work with the gesture handler? Not sure what happens if we require both reanimated and gesture-handler support.

swansontec (changes requested):
This looks super-good for such a huge PR. There are only minor issues.

src/components/scenes/MigrateWalletCompletionScene.tsx
samholmes: Ah. Good to know.

I also was inclined to revisit this and other scenes with floating components to instead render these components int he footer.

src/components/ui4/BlurBackground.tsx
samholmes: overlayColor applied on the whole screen on Android. That was the reason for the transparent value.

src/components/scenes/CoinRankingScene.tsx
samholmes: It appears to work. Isn't reanimated and gesture handler both by software mansion? I figured they'd both be using the same gesture handling efficiencies, but maybe that assumption is wrong?

src/components/ui4/BlurBackground.tsx
swansontec: Oh, interesting. So it's a bug-fix to the blur library!? In that case, we should document it as such.

src/components/scenes/CoinRankingScene.tsx
swansontec: If it works, it works.

swansontec (approved):

swansontec (approved):
The blur changes look good. We could mention that the colors are hard-coded to match the iOS colors on Android, so the theme is irrelevant to that goal.

PR #4722, opened 2024-01-17
Fix animated icon glitches

swansontec:
Swapping the eye icon was leading to color glitches, due to some bug deep within react-native-reanimated.

The solution involves rendering a single `Animated.Text` component, and simply changing the text content inside. This avoids re-wiring the Reanimated dependency graph, which is what seems to be causing the problem.

We achieve this with *almost* the same icon API as before by adding an `off` property to the eye icon, to switch between the two variants. We also refactor the animated icon implementation to use a raw text element, instead of the react-native-vector-icon wrapper component.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206268337019885

### DIFF
--- a/src/components/icons/ThemedIcons.tsx
+++ b/src/components/icons/ThemedIcons.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
-import Animated, { AnimatedProps, SharedValue, useAnimatedStyle } from 'react-native-reanimated'
+import Animated, { SharedValue, useAnimatedStyle } from 'react-native-reanimated'
 import AntDesignIcon from 'react-native-vector-icons/AntDesign'
-import { IconProps as VectorIconProps } from 'react-native-vector-icons/Icon'
+import { type Icon } from 'react-native-vector-icons/Icon'
 import IonIcon from 'react-native-vector-icons/Ionicons'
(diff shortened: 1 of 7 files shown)


src/components/icons/ThemedIcons.tsx
```diff
@@ -27,56 +26,79 @@ export interface IconProps {
...
+    fontSize: size?.value ?? defaultSize,
+    fontStyle: 'normal',
+    fontWeight: 'normal'
+  }))
+
+  return (
+    <Animated.Text accessible={accessible} style={style}>
+      {String.fromCodePoint(glyphMap[name])}
+    </Animated.Text>
+  )
```
samholmes, line 58: We should document this approach with comments. Something to mention is the importance of using a single Animated.Text component for animated icons to avoid the buggy behavior in reanimated. Also, mention the limitation that switching between animated icons should be avoided completely in favor of parameterizing the icon name prop instead; this may limit our ability to switch between icons in different typesets.

samholmes (approved):
Awesome!

PR #4726, opened 2024-01-18
Paul/send anim

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206379503084295

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - added: `minerTip` to `feeRateUsed` processing
 - added: Make the alert drop-down swipeable.
+- added: Multi output payments for UTXO currencies
 - changed: Update various scenes with UI4 components
 - changed: Light account re-enabled at 50% distribution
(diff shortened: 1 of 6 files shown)


paullinator:
https://github.com/EdgeApp/edge-react-gui/assets/6079354/4364c01a-e60c-43d7-9098-b57a0be90dd7

samholmes (approved):
Approved. I assume the multiple targets is ready to go (do we have tasks for QA to test this feature and were the testing hours included in QA's scoped?).

My comment is whether you can make the animation not only transition using the Y flip transition but also a fade-in/fade-out? Combining those two effects might make it feel a bit more polished IMO.

paullinator:
The multi target was implemented a long time ago and went through "some" testing with no issues and was only held back due to missing animations.

There's no off-the-shelf way to add a fade transition. That would really be nice. But we can create a custom animation in the future. @samholmes

PR #4730, opened 2024-01-19
Add background accent colors to ChangeMiningFeeScene

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206379497162107

### DIFF
--- a/src/components/scenes/ChangeMiningFeeScene.tsx
+++ b/src/components/scenes/ChangeMiningFeeScene.tsx
@@ -5,9 +5,13 @@ import Evilicons from 'react-native-vector-icons/EvilIcons'
 import MaterialCommunityIcons from 'react-native-vector-icons/MaterialCommunityIcons'
 
 import { FEE_STRINGS } from '../../constants/WalletAndCurrencyConstants'
+import { useIconColor } from '../../hooks/useIconColor'
 import { lstrings } from '../../locales/strings'
+import { useSelector } from '../../types/reactRedux'
 import { EdgeSceneProps } from '../../types/routerTypes'
 import { FeeOption } from '../../types/types'
+import { getTokenId } from '../../util/CurrencyInfoHelpers'
+import { darkenHexColor } from '../../util/utils'
 import { SceneWrapper } from '../common/SceneWrapper'
 import { cacheStyles, Theme, ThemeProps, withTheme } from '../services/ThemeContext'
 import { SettingsRadioRow } from '../settings/SettingsRadioRow'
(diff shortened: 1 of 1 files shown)


src/components/scenes/RequestScene.tsx
```diff
@@ -629,6 +629,6 @@ export const RequestScene = (props: OwnProps) => {
   const { pluginId = '' } = wallet.currencyInfo ?? {}
   const tokenId = getTokenId(account, pluginId, currencyCode)
 
-  const iconColor = useIconColor({ pluginId, tokenId: tokenId !== undefined ? tokenId : '' })
+  const iconColor = useIconColor({ pluginId, tokenId: tokenId !== undefined ? tokenId : null })
```
paullinator, line 632: This will load the icon color for the mainnet asset. Better not to load a color at all otherwise it will flicker between the two

src/components/scenes/ChangeMiningFeeScene.tsx
```diff
@@ -186,4 +213,16 @@ const getStyles = cacheStyles((theme: Theme) => {
...
+
+export const ChangeMiningFeeScene = (props: OwnProps) => {
+  const account = useSelector(state => state.core.account)
+  const currencyCode = useSelector(state => state.ui.wallets.selectedCurrencyCode)
+  const walletId = useSelector(state => state.ui.wallets.selectedWalletId)
+  const wallet = account.currencyWallets[walletId] ?? {}
+  const { pluginId = '' } = wallet.currencyInfo ?? {}
+  const tokenId = getTokenId(account, pluginId, currencyCode)
+
+  const iconColor = useIconColor({ pluginId, tokenId: tokenId !== undefined ? tokenId : null })
```
paullinator, line 226: don't default tokenId to null as that will pull the parent asset icon color

paullinator (changes requested):

src/components/scenes/ChangeMiningFeeScene.tsx
samholmes: Isn’t that the goal? If there is no token id then this would mean the wallet is not a token wallet but a mainnet wallet.

src/components/scenes/RequestScene.tsx
samholmes: Why would it load the icon for the mainnet when there is a tokenId?

paullinator (approved):

PR #4738, opened 2024-01-20
Improve footer implementation via "render-portals"

samholmes:
A render-portal is a technique to render some components in another location in the component tree by padding a render function to a hook which is provided by some global context.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206394910124605


src/state/SceneFooterState.tsx
```diff
@@ -28,6 +38,81 @@ export const [SceneFooterProvider, useSceneFooterState] = createStateProvider(()
...
+/**
+ * Used by scenes to register a render function for the global footer area.
+ *
+ * @param renderFn the render function to be used in the scene/tab-bar footer
+ * @param deps the dependencies for the render function to trigger re-renders
+ */
+export const useSceneFooterRender = (renderFn: FooterRender = defaultFooterRender, deps: DependencyList) => {
+  const { setRenderFooter } = useSceneFooterRenderState()
+
+  // The callback will allow use to trigger a re-render when the deps change
```
swansontec, line 72: "will allow use"? Not sure what this means.

src/state/SceneFooterState.tsx
```diff
@@ -28,6 +38,81 @@ export const [SceneFooterProvider, useSceneFooterState] = createStateProvider(()
...
+  const [renderFooter, setRenderFooter] = useState<FooterRender>(() => defaultFooterRender)
+
+  return useMemo(
+    () => ({
+      renderFooter,
+      setRenderFooter
+    }),
+    [renderFooter, setRenderFooter]
+  )
+})
```
swansontec, line 57: I'm not sure why we need a separate context here. Could this go in the one above?

I don't think we should fix this right now, though, since it's minor enough and we need to focus on getting the release out.

swansontec (approved):
Only one tiny typo.

src/state/SceneFooterState.tsx
samholmes: I don't want to re-render all components which use the other context because it may not be an relevant update for them. I also landed on this design when fixing strange infinite loops, so I don't know if moving this state into the other context will cause this infinite loop re-render or not.

swansontec (approved):

PR #4739, opened 2024-01-20
Paul/fix exchange scenes

paullinator:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206387498116275

### DIFF
--- a/src/components/scenes/CryptoExchangeQuoteProcessingScene.tsx
+++ b/src/components/scenes/CryptoExchangeQuoteProcessingScene.tsx
@@ -5,6 +5,7 @@ import { ActivityIndicator, View } from 'react-native'
 import { useAsyncEffect } from '../../hooks/useAsyncEffect'
 import { lstrings } from '../../locales/strings'
 import { EdgeSceneProps } from '../../types/routerTypes'
+import { EdgeAnim } from '../common/EdgeAnim'
 import { SceneWrapper } from '../common/SceneWrapper'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 import { EdgeText } from '../themed/EdgeText'
@@ -18,6 +19,8 @@ export interface ExchangeQuoteProcessingParams {
 }
 
 interface Props extends EdgeSceneProps<'exchangeQuoteProcessing'> {}
+
+const ANIM_DURATION = 5000
 
 export function CryptoExchangeQuoteProcessingScene(props: Props) {
   const theme = useTheme()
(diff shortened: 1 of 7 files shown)


paullinator:
Screenshot 2024-01-19 at 20 53 17
Screenshot 2024-01-19 at 20 53 57
Screenshot 2024-01-19 at 21 59 34
Screenshot 2024-01-19 at 20 54 40

src/components/scenes/CryptoExchangeQuoteProcessingScene.tsx
```diff
@@ -61,27 +64,39 @@ export function CryptoExchangeQuoteProcessingScene(props: Props) {
...
-            onPress: () => {
-              mounted.current = false
-              onCancel()
-            }
-          }}
-          layout="column"
-          sceneMargin
-        />
+      {({ insetStyle }) => (
+        <View style={{ paddingBottom: insetStyle.paddingBottom, width: '100%', height: '100%' }}>
```
samholmes, line 68: Why the extra view instead of using the one with `styles.container`?

Also, why only use `paddingBottom` from insetStyle? Using all of the rules usually means we're avoiding insets on left/right and even top sides of the device (e.g. for rotated iPad devices)

src/components/scenes/GuiPluginListScene.tsx
```diff
@@ -386,39 +387,49 @@ class GuiPluginList extends React.PureComponent<Props, State> {
...
         ) : (
-          <FlatList
+          <Animated.FlatList
             data={plugins}
+            onScroll={this.props.handleScroll}
+            ListHeaderComponent={renderTop}
             renderItem={this.renderPlugin}
             keyExtractor={(item: GuiPluginRow) => item.pluginId + item.title}
-            contentContainerStyle={{ ...insetStyle, paddingTop: 0 }}
+            contentContainerStyle={{ paddingTop: 0 }}
```
samholmes, line 436: Remove

samholmes (changes requested):
Test on iPad sim to make sure the padding is correct during rotated orientations

src/components/scenes/CryptoExchangeQuoteProcessingScene.tsx
paullinator: The extra View covers all the content including the button at the bottom while the inner View is just the text and spinner.

I didn't actually need the insets at all since all the content is centered. Removed.

samholmes (approved):

PR #4745, opened 2024-01-23
Jon/ui4/fix/getting-started

Jon-edge:
image
image

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206421723683370
  - https://app.asana.com/0/0/1206340123332880
  - https://app.asana.com/0/0/1206215216300264

### DIFF
--- a/src/components/scenes/GettingStartedScene.tsx
+++ b/src/components/scenes/GettingStartedScene.tsx
@@ -1,7 +1,7 @@
 import { CreateAccountType, InitialRouteName } from 'edge-login-ui-rn'
 import * as React from 'react'
 import { useEffect } from 'react'
-import { Image, Platform, Pressable, Text, View } from 'react-native'
+import { Image, Pressable, Text, View } from 'react-native'
 import { ScrollView, TouchableOpacity } from 'react-native-gesture-handler'
 import Animated, {
   Extrapolation,
@@ -37,7 +37,7 @@ import { styled } from '../hoc/styled'
 import { SwipeOffsetDetector } from '../interactions/SwipeOffsetDetector'
 import { Space } from '../layout/Space'
 import { EdgeText } from '../themed/EdgeText'
-import { MainButton } from '../themed/MainButton'
+import { ButtonsViewUi4 } from '../ui4/ButtonsViewUi4'
 
 const ANIM_DURATION = 1000
 
@@ -171,8 +171,6 @@ export const GettingStartedScene = (props: Props) => {
     }
   }, [isLoggedIn, localUsers, navigation])
 
-  const disableType = Platform.OS === 'android' ? 'view' : undefined
-
   return (
     <SceneWrapper hasHeader={false}>
       <SkipButton swipeOffset={swipeOffset}>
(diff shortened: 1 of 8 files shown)


src/components/ui4/ButtonUi4.tsx
```diff
@@ -1,3 +1,7 @@
+/**
+ * IMPORTANT: Changes in this file MUST be synced with edge-react-gui!
```
swansontec, line 2: We are already in edge-react-gui. Do you mean edge-login-ui-rn?

src/components/ui4/ButtonUi4.tsx
```diff
@@ -141,16 +146,18 @@ export function ButtonUi4(props: Props) {
...
-      <LinearGradient
-        {...gradientProps}
-        style={[styles.contentCommon, dynamicGradientStyles, mini ? styles.contentSizeMini : styles.contentSizeDefault, ...finalContainerCommon]}
-      >
-        {hideContent ? null : children}
-        {hideContent ? null : maybeText}
-        {!hideContent ? null : <ActivityIndicator color={spinnerColor} style={styles.spinnerCommon} />}
-      </LinearGradient>
-    </TouchableOpacity>
+    <MaybeEdgeAnim when={animProps != null} {...animProps}>
```
swansontec, line 149: This layering makes no sense. The button user should just do `<EdgeAnim><ButtonUi4/></EdgeAnim>` themselves. There is no reason for the button to be animation-aware.

At the very least, the `ButtonsViewUi4` should be in charge of adding the `<EdgeAnim>` wrapper.

src/components/ui4/ButtonsViewUi4.tsx
```diff
@@ -19,6 +20,8 @@ export interface ButtonInfo {
   disabled?: boolean
   spinner?: boolean
   testID?: string
+
+  animProps?: EdgeAnimProps
```
swansontec, line 24: Optional: I don't especially like this, because it's too low level. We don't let the user tweak other button properties at this level of detail. Could it be something simpler, like an `animate: boolean` prop on the top-level `ButtonsViewUi4` component? Then the buttons view is in charge of the visuals again, not the scene.

I don't think we should fix this now, because we are running out of time, but at least start thinking about other solutions.

src/components/scenes/GettingStartedScene.tsx
```diff
@@ -394,7 +404,7 @@ const SectionCoverAnimated = styled(Animated.View)<{ swipeOffset: SharedValue<nu
       justifyContent: 'space-between',
       backgroundColor: '#0F1D26',
       paddingVertical: theme.rem(1),
-      paddingBottom: insets.bottom,
+      paddingBottom: insets.bottom + themeRem,
```
swansontec, line 407: Please use `theme.rem(1)`, like on the line above. We use the `themeRem` in animated props because we can't call methods across the reanimated bridge, so `theme.rem()` is unusable. Otherwise prefer the normal `theme.rem(1)` syntax.

swansontec (changes requested):

src/components/ui4/ButtonsViewUi4.tsx
Jon-edge: I agree, it would be nice to be simpler. 

It's like this because each scene has its own timing depending on the number of items animating, I couldn't figure out how to get around that beyond giving animation handling to the SceneWrapper, and the SceneWrapper takes chunks of components like SectionView does.

src/components/ui4/ButtonUi4.tsx
Jon-edge: We have this mix of requirements right now. 

In the ideal future, no scene touches `ButtonUi4` (or `MainButton` which is just a `ButtonUi4` wrapper) but we do now because of time restrictions. Since we have a mix of scenes where they sometimes use `ButtonUi4`/`MainButton` directly and other times `ButtonsViewUi4`, I had to add it to both.

src/components/ui4/ButtonsViewUi4.tsx
Jon-edge: Maybe instead we give `ButtonsViewUi4` an `animStartDistance` time?

src/components/ui4/ButtonUi4.tsx
Jon-edge: I should just mention both to avoid copy pasta mistakes.

src/components/ui4/ButtonsViewUi4.tsx
swansontec: Yeah, something like this would be good.

src/components/ui4/ButtonsViewUi4.tsx
```diff
@@ -1,29 +1,35 @@
...
+  spinner?: boolean
+  testID?: string
+
+  animDistanceStart?: number
 }
 
 interface Props {
   // Specifies whether the component should be positioned absolutely.
   // Default value is false.
   absolute?: boolean
```
Jon-edge, line 32: Removed this fade prop because our animations effectively replace it, and it's
not used anyway.

swansontec (approved):

PR #4748, opened 2024-01-23
Fix margin on BackupModal

peachbits:
---
  - https://app.asana.com/0/0/1206072694496683

### DIFF
--- a/src/components/modals/BackupModal.tsx
+++ b/src/components/modals/BackupModal.tsx
@@ -82,7 +82,6 @@ const getStyles = cacheStyles((theme: Theme) => ({
     alignItems: 'center',
     flexDirection: 'column',
     justifyContent: 'space-between',
-    marginTop: theme.rem(2),
     height: '100%'
   },
   image: {


peachbits:
Screenshot_1705968212

src/components/modals/BackupModal.tsx
```diff
@@ -82,7 +82,6 @@ const getStyles = cacheStyles((theme: Theme) => ({
     alignItems: 'center',
     flexDirection: 'column',
     justifyContent: 'space-between',
-    marginTop: theme.rem(2),
     height: '100%'
```
swansontec, line 85: I think the `height: 100%` might be the problem. I think we mean `flexGrow: 1`, so the area expands to fill the modal? Explicit dimensions are a big red flag for layout problems, since they erase the intrinsic content size.

Nevertheless, if it works, we can merge.

swansontec (approved):

src/components/modals/BackupModal.tsx
peachbits: flexGrow causes the image to go away. The best looking version, by just tweaking these styles, is to just remove marginTop

PR #4751, opened 2024-01-23
Bump minimum iOS version to 14

peachbits:
---
  - https://app.asana.com/0/0/1206418560569860


swansontec (approved):
Approved, but did we check with QA / support about this? I guess we can't support a device we can't test on, but nothing in accountbased should prevent the app from running, even if BigInt support is missing. It sounds like we should debug accountbased as well, regardless, since there should be *zero* library code in the root bundle, just currency infos.

peachbits:
As discussed IRL, edge-exchange-plugins isn't chunked as edge-currency-accountbased is so imported sdks with new fancy features are always present. There's a good performance benefit to eventually adding that to exchange plugins, but since iOS 13 phones are upgradable to iOS 14, this change makes sense for now and exchange plugins can be updated later.

PR #4753, opened 2024-01-24
Paul/smal fixes2

paullinator:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206387496803858

### DIFF
--- a/src/components/scenes/TransactionDetailsScene.tsx
+++ b/src/components/scenes/TransactionDetailsScene.tsx
@@ -382,7 +382,7 @@ const TransactionDetailsComponent = (props: Props) => {
       {recipientsAddresses === '' ? null : (
         <EdgeAnim enter={{ type: 'fadeInDown', distance: 60 }}>
           <CardUi4>
-            <RowUi4 rightButtonType="copy" title={lstrings.transaction_details_recipient_addresses} body={recipientsAddresses} />
+            <RowUi4 maximumHeight="large" rightButtonType="copy" title={lstrings.transaction_details_recipient_addresses} body={recipientsAddresses} />
           </CardUi4>
         </EdgeAnim>
       )}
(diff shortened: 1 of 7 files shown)


src/components/scenes/TransactionDetailsScene.tsx
```diff
@@ -382,7 +382,7 @@ const TransactionDetailsComponent = (props: Props) => {
       {transaction.spendTargets == null ? null : (
         <EdgeAnim enter={{ type: 'fadeInDown', distance: 60 }}>
           <CardUi4>
-            <RowUi4 rightButtonType="copy" title={lstrings.transaction_details_recipient_addresses} body={recipientsAddresses} />
+            <RowUi4 maximumHeight="large" rightButtonType="copy" title={lstrings.transaction_details_recipient_addresses} body={recipientsAddresses} />
```
Jon-edge, line 385: This fix doesn't work for a number of addresses. The legacy sm/med/large prop
from Tile I think we should deprecate.

I think a better fix for now is to give a complex child that properly handles
printing out the addresses on individual lines.

Jon-edge, line 385: This fix doesn't work for a number of addresses beyond 3. The legacy sm/med/large prop
from Tile I think we should deprecate, I don't think maximumHeight fits in our new design.

A more complete fix would be supplying a number of EdgeText components for
each address as a child to RowUi4.

Jon-edge (changes requested):

src/components/scenes/TransactionDetailsScene.tsx
paullinator: I don’t see how this doesn’t fit in our design? I tested with 5 rows and it works fine. 

It’s also a waste to create text components for each row as the content is literally just a multi line string. That’s more code and View overhead for no reason I can see. 

Can you justify your thoughts on how this doesn’t fit our design?

Jon-edge: Can't reproduce the issue anymore

PR #4762, opened 2024-01-25
Only show the backup notification card if FIO names are present

peachbits:
Also, only allow it to be interacted with once and then hidden forever.

---
  - https://app.asana.com/0/0/1206426372035219

### DIFF
--- a/src/actions/DeviceSettingsActions.ts
+++ b/src/actions/DeviceSettingsActions.ts
@@ -23,6 +23,18 @@ export const writeDisableAnimations = async (disableAnimations: boolean) => {
   return await writeDeviceSettings(updatedSettings)
 }
 
+export const writeHasInteractedWithBackupModal = async (hasInteractedWithBackupModal: boolean) => {
+  try {
+    const raw = await disklet.getText(DEVICE_SETTINGS_FILENAME)
+    const json = JSON.parse(raw)
+    deviceSettings = asDeviceSettings(json)
+  } catch (e) {
+    console.log(e)
+  }
+  const updatedSettings = { ...deviceSettings, hasInteractedWithBackupModal }
+  return await writeDeviceSettings(updatedSettings)
+}
+
 const readDeviceSettings = async (): Promise<DeviceSettings> => {
   try {
     const text = await disklet.getText(DEVICE_SETTINGS_FILENAME)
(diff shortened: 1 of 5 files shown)


src/components/notification/NotificationView.tsx
```diff
@@ -25,15 +30,31 @@ const NotificationViewComponent = (props: Props) => {
...
 
   const [autoDetectTokenCards, setAutoDetectTokenCards] = React.useState<React.JSX.Element[]>([])
 
   const handlePress = useHandler(async () => {
+    setHasInteractedWithBackupModal(true)
     await showBackupModal({ navigation })
+    await account.disklet.setText(backupNotificationKey, '')
   })
 
+  useAsyncEffect(
```
paullinator, line 45: Move to use the new framework in `DeviceSettingsActions.ts`

src/components/notification/NotificationView.tsx
```diff
@@ -48,22 +46,9 @@ const NotificationViewComponent = (props: Props) => {
   const handlePress = useHandler(async () => {
     setHasInteractedWithBackupModal(true)
     await showBackupModal({ navigation })
-    await account.disklet.setText(backupNotificationKey, '')
+    hasInteractedWithBackupModalLocal = true
```
paullinator, line 49: need to call `writeHasInteractedWithBackupModal`

src/actions/DeviceSettingsActions.ts
```diff
@@ -23,6 +23,18 @@ export const writeDisableAnimations = async (disableAnimations: boolean) => {
   return await writeDeviceSettings(updatedSettings)
 }
 
+export const writeHasInteractedWithBackupModal = async (hasInteractedWithBackupModal: boolean) => {
```
paullinator, line 26: unused?

src/components/notification/NotificationView.tsx
```diff
@@ -110,9 +110,9 @@ const NotificationViewComponent = (props: Props) => {
 
   return (
     <NotificationCardsContainer hasTabs={hasTabs} insetBottom={insetBottom} footerHeight={footerHeight} footerOpenRatio={footerOpenRatio}>
-      <Fade visible={isBackupWarningShown}>
+      <EdgeAnim visible={isBackupWarningShown} enter={{ type: 'fadeIn' }}>
```
paullinator, line 98: and prop `exit={{type: 'fadeOut'}}`

paullinator (changes requested):

src/components/notification/NotificationView.tsx
paullinator: also declare prop object globally in file so it doesn't cause a re-render

paullinator (approved):

PR #4770, opened 2024-01-26
Fix TextInputModal for Android

Jon-edge:
The android adjustPan glitch does not appear to be happening in the new modals, so it was also removed.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device


swansontec (approved):
Approved, but I don't see where the gap is coming from between the text input and the button. Is this a text input feature? I see an `expand` prop, which makes me think so.

PR #4771, opened 2024-01-26
Add setting to disable animations

paullinator:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206447894981394

### DIFF
--- a/src/actions/DeviceSettingsActions.ts
+++ b/src/actions/DeviceSettingsActions.ts
@@ -1,0 +1,41 @@
+import { makeReactNativeDisklet } from 'disklet'
+
+import { asDeviceSettings, DeviceSettings } from '../types/types'
+
+const disklet = makeReactNativeDisklet()
+const DEVICE_SETTINGS_FILENAME = 'DeviceSettings.json'
+let deviceSettings: DeviceSettings = asDeviceSettings({})
+
+export const getDeviceSettings = (): DeviceSettings => deviceSettings
+export const initDeviceSettings = async () => {
+  deviceSettings = await readDeviceSettings()
+}
+
+export const writeDisableAnimations = async (disableAnimations: boolean) => {
+  try {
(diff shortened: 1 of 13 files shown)


src/actions/DeviceSettingsActions.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const readDeviceSettings = async (): Promise<DeviceSettings> => {
+  try {
+    const text = await disklet.getText(DEVICE_SETTINGS_FILENAME)
+    const json = JSON.parse(text)
+    const settings = asDeviceSettings(json)
+    return settings
+  } catch (e) {
+    const defaults = asDeviceSettings({})
+    return await writeDeviceSettings(defaults).then(() => defaults)
```
swansontec, line 34: This isn't needed. Just return `{}` and be done with it. The file is only interesting once the user changes an actual setting.

Having an empty file on disk doesn't change anything or help anything - if it's not there, you return the defaults, and if the file is there, it only contains defaults anyhow, so why bother?

src/app.ts
```diff
@@ -213,3 +214,5 @@ if (ENV.DEBUG_THEME) {
   }
   themeFunc().catch(err => console.error(err))
 }
+
+initDeviceSettings().catch(err => console.error(err))
```
swansontec, line 218: Please avoid `console.error` in general. This is the red box of doom, which should only be used for genuine emergencies. Just use console.log or console.warn or something.

swansontec (approved):
Two optional things to fix, but otherwise this is fine.

swansontec (approved):

swansontec:
I am going to create an Asana task to store `developerPlugin` (used for custom webview plugins) in your new device store. This will allow us to drop RNCAsyncStorage from our native deps.

PR #4780, opened 2024-01-27
GUI: Sync ButtonUi4 Spinner Fixes from LoginUI

Jon-edge:
image

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

https://github.com/EdgeApp/edge-login-ui-rn/pull/157

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206455248718481

### DIFF
--- a/src/components/ui4/ButtonUi4.tsx
+++ b/src/components/ui4/ButtonUi4.tsx
@@ -1,6 +1,5 @@
 /**
- * IMPORTANT: Changes in this file MUST be synced between edge-react-gui and
- * edge-login-ui-rn!
+ * IMPORTANT: Changes in this file MUST be synced with edge-react-gui!
  */
 
 import * as React from 'react'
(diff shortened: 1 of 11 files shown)


src/components/ui4/ButtonUi4.tsx
```diff
@@ -196,9 +210,9 @@ const getStyles = cacheStyles((theme: Theme) => {
       // Reduce the bounds of a tertiary button so it doesn't appear to be too
```
paullinator, line 210: tertiary buttons should not have any difference in spacing. On some styles they may actually be filled with borders. Please create a task to fix in 4.1

src/components/ui4/ButtonUi4.tsx
```diff
@@ -114,41 +113,56 @@ export function ButtonUi4(props: Props) {
...
   return (
-    <TouchableOpacity disabled={disabled || pending} style={finalContainerCommon} onPress={handlePress} testID={testID}>
+    <TouchableOpacity disabled={disabled || pending || spinner} style={finalContainerCommon} onPress={handlePress} testID={testID}>
       <LinearGradient
         {...gradientProps}
         style={[styles.contentCommon, dynamicGradientStyles, mini ? styles.contentSizeMini : styles.contentSizeDefault, ...finalContainerCommon]}
       >
-        {hideContent ? null : children}
-        {hideContent ? null : maybeText}
+        <View style={contentOpacityStyle}>
```
paullinator, line 162: This is a dangerous change! This is the type of high regression potential change I was referring to this week.

You are changing the way children are laid out in the button. This needs a full audit of everywhere ButtonUi4 is used. Note the difference before and after your change in the screenshots
Screenshot 2024-01-27 at 17 27 08

Screenshot 2024-01-27 at 17 27 22

The first image is not ideal but the issue is that it's different which it shouldn't be.

Change the View alignment style that affects children to exactly match what the LinearGradient had. Or better yet, punt this to 4.1 as it's not that urgent.

paullinator (changes requested):

src/components/ui4/ButtonUi4.tsx
Jon-edge: This was a request by Madison

Jon-edge:
Updated screens with Dev Tab additions

paullinator (approved):

PR #4785, opened 2024-01-30
Restyle SepaFormScene

Jon-edge:
image

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206346500344678


src/plugins/gui/scenes/SepaFormScene.tsx
```diff
@@ -64,40 +62,28 @@ export const SepaFormScene = React.memo((props: Props) => {
 
   return (
     <SceneWrapper hasNotifications>
```
swansontec, line 64: Can it be `<SceneWrapper scroll>` so we don't lose the scroll feature? It won't be keyboard aware, of course.

swansontec (approved):
Approved with optional change.

PR #4772, opened 2024-01-26
Improve SceneWrapper performance

samholmes:
### CHANGELOG

Does this branch warrant an entry to the CHANGELOG?

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

If you have made **any** visual changes to the GUI. Make sure you have:

- [x] Tested on iOS device
- [x] Tested on Android device
- [ ] Tested on small-screen device (iPod Touch)
- [ ] Tested on large-screen device (tablet)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206435742488679

PR #4787, opened 2024-01-30
Sam/performance2

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Follow-up to:

https://github.com/EdgeApp/edge-react-gui/pull/4772

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206435742488679

### DIFF
--- a/src/components/common/SceneWrapper.tsx
+++ b/src/components/common/SceneWrapper.tsx
@@ -7,7 +7,7 @@ import Reanimated from 'react-native-reanimated'
 import { EdgeInsets, useSafeAreaFrame, useSafeAreaInsets } from 'react-native-safe-area-context'
 
 import { SCROLL_INDICATOR_INSET_FIX } from '../../constants/constantSettings'
-import { FooterRender, PortalSceneFooter, useSceneFooterState } from '../../state/SceneFooterState'
+import { FooterRender, PortalSceneFooter } from '../../state/SceneFooterState'
 import { useSceneScrollHandler } from '../../state/SceneScrollState'
 import { useSelector } from '../../types/reactRedux'
 import { NavigationBase } from '../../types/routerTypes'
@@ -64,6 +64,9 @@ interface SceneWrapperProps {
   backgroundGradientColors?: string[]
   backgroundGradientStart?: { x: number; y: number }
   backgroundGradientEnd?: { x: number; y: number }
+
+  // The necessary height of the footer to include in the insets
+  footerHeight?: number
 
   // True if this scene has a header (with back button & such):
   hasHeader?: boolean
@@ -162,6 +165,7 @@ function SceneWrapperInnerComponent(props: SceneWrapperInnerProps) {
     backgroundGradientStart,
     backgroundGradientEnd,
     children,
+    footerHeight = 0,
     hasHeader = true,
     hasNotifications = false,
     hasTabs = false,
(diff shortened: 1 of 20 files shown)


src/state/SceneScrollState.tsx
```diff
@@ -43,50 +43,65 @@ export type SceneScrollHandler = (event: NativeSyntheticEvent<NativeScrollEvent>
  * if the scene is focused (react-navigation's useIsFocused).
  */
 export const useSceneScrollHandler = (): SceneScrollHandler => {
-  const setScrollState = useSceneScrollContext(state => state.setScrollState)
+  const scrollState = useSceneScrollContext(state => state.scrollState)
 
-  const localScrollState: ScrollState = useScrollState()
```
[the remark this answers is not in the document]
swansontec: Oh, I see you did this in the final commit! Maybe move that commit earlier in the series, right after "Refactor local scene scroll handlers to use the same global scroll state"

src/components/common/SceneWrapper.tsx
```diff
@@ -206,19 +204,10 @@ function SceneWrapperInnerComponent(props: SceneWrapperInnerProps) {
     () => ({
       top: safeAreaInsets.top + maybeHeaderHeight,
       right: safeAreaInsets.right,
-      bottom: maybeInsetBottom + maybeNotificationHeight + maybeTabBarHeight + footerHeight,
+      bottom: maybeInsetBottom + maybeNotificationHeight + maybeTabBarHeight,
```
swansontec, line 207: This approach will not work. If we convert `insetStyles` and `undoInsetStyles` into animated styles, this will cause big performance problems with the drawer animation.

This is because styles like `bottom` affect layout, which implies a Yoga recalculation. Since we are applying this at the scene level, the yoga update will affect the entire scene, which isn't going to be cheap. Because this is an animation, the calculation needs to happen on every frame.

On the flip side, we definitely need to include the footer height in the insets / undoInsets numbers to get correct rendering.

So, we want the `footerHeight` to be an normal React value, not an animated value.

The problem is that we do *not* want to store the `footerHeight` in the global state. The `footerHeight` is already unique on a scene-by-scene basis. If scene A has a 5 rem footer, and scene B has a 3 rem footer, why does scene A need to re-render when we switch to scene B? This makes no sense.

Instead, the `renderFooter` should be a prop on the `SceneWrapper` component, along with a `footerHeight` property. *IF* the scene wrapper knows that we have a tab bar, then, and only then, do we send the `footerHeight` and `renderFooter` into the tabbar context. The only one subscribed to the tabbar context is the tabbar, so there is very little rendering impact.

Now, you might say that a static `footerHeight` is a bad idea. What if we want to measure our content? It's fine:

```ts
function SomeScene(props) {
  const [footerHeight, setFooterHeight] = useState()
  const handleLayout = useHandler(layoutEvent => setFooterHeight(...))

  const renderFooter = useCallback(() => {
    return <View onLayout={handleLayout}>...</View>
  }, [...])

  return <SceneWrapper
    renderFooter={renderFooter}
    footerHeight={footerHeight}
    >...</SceneWrapper>
}
```

If a scene uses this pattern, then it will render twice - once at mount, and again immediately once the layout updates. However, *only* the affected scene will re-render, and also the tab bar if used.

On the other hand, we already know the height of the search bar in rem's, so there is no need to bother with this on a scene that simply has search.

swansontec (changes requested):

src/components/common/SceneWrapper.tsx
samholmes: I decided to measure content inline especially for WalletListScene where the footer does change based on the context of whether we're sorting wallets or not. So, since I have that infra on the WalletListScene, I decided to use it on the other scenes as well. I think rendering a scene twice that has a footer should be safe enough considering that we have to re-render for other scene state anyway.

The real expense now is only on initial render for a scene to assigned the scroll handler, which we can try to golf for more performance later, or we can accept this performance hit and optimize other areas.

swansontec (approved):

PR #4789, opened 2024-01-30
Add StakeKit insufficientFunds error handling to action endpoints

peachbits:
---
  - https://app.asana.com/0/0/1206398950862158

### DIFF
--- a/src/plugins/stake-plugins/generic/util/stakeKitUtils.ts
+++ b/src/plugins/stake-plugins/generic/util/stakeKitUtils.ts
@@ -9,6 +9,8 @@ import type {
   YieldBalanceRequestDto,
   YieldDto
 } from '@stakekit/api-hooks'
+import { asMaybe, asObject, asValue } from 'cleaners'
+import { InsufficientFundsError } from 'edge-core-js'
 
 import { ENV } from '../../../../env'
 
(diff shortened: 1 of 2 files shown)


src/plugins/stake-plugins/generic/util/stakeKitUtils.ts
```diff
@@ -63,3 +71,10 @@ export const yieldGetSingleYieldBalances = async (integrationId: string, yieldBa
 export const yieldYieldOpportunity = async (integrationId: string): Promise<YieldDto> => {
   return await fetchGet<YieldDto>(`/v1/yields/${integrationId}`)
 }
+
+const checkInsufficientFunds = (res: any): void => {
```
swansontec, line 75: I generally prefer `unknown` over `any`, since it prevents you from messing with the value until you clean it. Much safer.

src/plugins/stake-plugins/generic/util/stakeKitUtils.ts
```diff
@@ -63,3 +71,10 @@ export const yieldGetSingleYieldBalances = async (integrationId: string, yieldBa
...
+  const maybeStakeKitInsufficientFundsError = asMaybe(asObject({ message: asValue('InsufficientFundsError') }))(res)
+  if (maybeStakeKitInsufficientFundsError != null) {
+    throw new InsufficientFundsError({ tokenId: null })
```
swansontec, line 78: What about making the `tokenId` a parameter? Then you would be doing `checkInsufficientFunds(out, null)` on the lines above. Unless you know for sure there there will never be tokens in this subsystem - then the current way is fine.

swansontec (approved):
Two optional changes.

src/plugins/stake-plugins/generic/util/stakeKitUtils.ts
peachbits: The plugin is for Coreum native token staking only. If we start adding other StakeKit plugins we should add this for sure

PR #4793, opened 2024-01-31
UI4: Narrow Device Scaling

Jon-edge:
An Android emulator was set up to approximate the display of the Fold 3 device based on where the `PasswordReminderModal`'s text appeared to wrap in the bug report.

LoginUi 3.2.3 fixes the legacy landing scene in English:
image

The following is an artificially narrow appearance with the `ButtonUi4` changes mirrored to LoginUi:
image

Other screens:
fio
markets
assetList
pwModal
exchangeScene
image

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206468789158141

### DIFF
--- a/src/components/scenes/GuiPluginListScene.tsx
+++ b/src/components/scenes/GuiPluginListScene.tsx
@@ -349,8 +349,10 @@ class GuiPluginList extends React.PureComponent<Props, State> {
         onLongPress={async () => await this.openPlugin(item, true).catch(showError)}
         paddingRem={[1, 0.5, 1, 0.5]}
       >
-        <View style={styles.rowContainer}>
-          <EdgeText style={styles.titleText}>{item.title}</EdgeText>
+        <View style={styles.cardContentContainer}>
+          <EdgeText style={styles.titleText} numberOfLines={1}>
+            {item.title}
+          </EdgeText>
           {item.description === '' ? null : <EdgeText style={styles.subtitleText}>{item.description}</EdgeText>}
           {poweredBy != null && item.partnerIconPath != null ? (
             <>
@@ -450,8 +452,10 @@ const getStyles = cacheStyles((theme: Theme) => ({
     marginLeft: -theme.rem(0.5),
     width: '100%'
   },
-  rowContainer: {
-    flexDirection: 'column'
+  cardContentContainer: {
+    flexDirection: 'column',
+    flexShrink: 1,
+    marginRight: theme.rem(0.5)
   },
   sceneContainer: {
     flex: 1,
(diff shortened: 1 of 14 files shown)


src/components/scenes/GuiPluginListScene.tsx
```diff
@@ -450,8 +452,10 @@ const getStyles = cacheStyles((theme: Theme) => ({
     marginLeft: -theme.rem(0.5),
     width: '100%'
   },
-  rowContainer: {
-    flexDirection: 'column'
+  cardContainer: {
+    flexDirection: 'column',
+    flexShrink: 1,
```
paullinator, line 457: What's the flexShrink for? This view is in a ScrollView and wouldn't ever need to shrink

paullinator (changes requested):

src/components/scenes/GuiPluginListScene.tsx
Jon-edge: The `cardContainer` is the card's contents. I'll rename to clarify further. Without flexShrink it looks like this:
image

It actually needs a right padding as well because on an even narrow device, it looks like this:
image

src/components/ui4/RowUi4.tsx
```diff
@@ -162,8 +162,7 @@ const getStyles = cacheStyles((theme: Theme) => ({
   },
   textHeader: {
     color: theme.secondaryText,
-    fontSize: theme.rem(0.75),
-    paddingRight: theme.rem(1)
```
paullinator, line 166: why did this change?

paullinator (changes requested):

src/components/ui4/RowUi4.tsx
Jon-edge: This was a remnant style from when I used to have the right button laid out differently. Now it's causing too much spacing with long header text that shrinks too early. This was apparent in the FIO screen

paullinator: But now we have to verify nothing else broke. :(

paullinator: Please double, triple, quadruple verify that this still looks good everywhere you can find a RowUi4. And check on iOS and Android. 4.0 is about to launch today as all perf issues have been resolved.

Jon-edge: All of the weird unexpected bugs that came up from changes during our integration were due to various combinations of flex properties.

We're not messing with flex properties, this is just a constant padding that was _wrong._ Other callers that have long titles will also have this incorrect extra padding. 

This style is what's used when RowUi4 manages the title (and in this case the title in relation to the right button), which does not affect any custom child content or any other weird callers that manage the entire contents of the card. The title is separate from the children or body. This padding change is no more risky than adding font scaling back to the title.

Before:
image

After:
image

Very minor change which is completely localized to the title.

paullinator (approved):

PR #4795, opened 2024-01-31
Use threshold-based footer animation for Android

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206481494003972

### DIFF
--- a/src/state/SceneFooterState.tsx
+++ b/src/state/SceneFooterState.tsx
@@ -9,6 +9,8 @@ import { useState } from '../types/reactHooks'
 import { createStateProvider } from './createStateProvider'
 import { useSceneScrollContext } from './SceneScrollState'
 
+const isAndroid = Platform.OS === 'android'
+
 //
 // Providers
 //
@@ -161,7 +163,7 @@ export const FooterAccordionEventService = () => {
 
   useSharedEvent(scrollEndEvent, () => {
     'worklet'
-    if (Platform.OS === 'android') {
+    if (isAndroid) {
       runOnJS(delayedSnap)()
       return
     }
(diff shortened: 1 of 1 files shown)


src/state/SceneFooterState.tsx
```diff
@@ -161,7 +163,7 @@ export const FooterAccordionEventService = () => {
 
   useSharedEvent(scrollEndEvent, () => {
     'worklet'
-    if (Platform.OS === 'android') {
+    if (isAndroid) {
```
paullinator, line 166: Not related to this review, but it is a bit suspect that you have to do a `runOnJs` here for Android. This implies you'll always cross-the-bridge every time the user's finger lifts off the screen on scroll.

paullinator (approved):
Approved, but I'd like a discussion on the two points below.

src/state/SceneFooterState.tsx
```diff
@@ -172,14 +174,14 @@ export const FooterAccordionEventService = () => {
 
   useSharedEvent(scrollMomentumBeginEvent, nativeEvent => {
     'worklet'
-    if (Platform.OS === 'android') return
+    if (isAndroid) return
```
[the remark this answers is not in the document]
samholmes: You're proposing to remove momentum scrolling for iOS?

src/state/SceneFooterState.tsx
samholmes: The runOnJS is because there is no `setTimeout` on the UI worklet thread.

paullinator: I may need a recap on the advantage of the setTimeout. What happens if we remove it? Can you try and see if perf improves

samholmes:
Deprecated for https://github.com/EdgeApp/edge-react-gui/pull/4804

PR #4800, opened 2024-01-31
Remove unimported files

swansontec:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device


src/components/scenes/Fio/FioAddressRegisteredScene.tsx
```diff
@@ -21,10 +20,10 @@ export class FioAddressRegistered extends React.Component<Props> {
       const styles = getStyles(theme)
 
       return (
-        <FormattedText style={styles.text}>
+        <Text style={styles.text}>
```
swansontec, line 23: Note that we should still port to `EdgeText` at some point. This commit gets rid of `FormattedText` in a way that's 1:1 compatible with what we rendered before; it doesn't try to upgrade anything.

Jon-edge (approved):

PR #4811, opened 2024-02-02
Add better string validation to `toPercentString`

swansontec:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206500529318664

### DIFF
--- a/src/util/cleaners.ts
+++ b/src/util/cleaners.ts
@@ -1,0 +1,17 @@
+import { asString } from 'cleaners'
+
+/**
+ * Accepts strings that are valid numbers according to biggystring.
+ *
+ * Valid strings contain some digits,
+ * with an optional decimal point plus some more digits. That is all.
+ *
+ * We reject blank strings, hex strings, exponential strings ("1e10"),
+ * "Infinity", "NaN", and other similar invalid inputs.
+ */
+export function asBiggystring(raw: unknown): string {
+  const clean = asString(raw)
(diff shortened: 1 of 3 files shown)


src/util/cleaners.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Valid strings contain some digits,
+ * with an optional decimal point plus some more digits. That is all.
+ *
+ * We reject blank strings, hex strings, exponential strings ("1e10"),
+ * "Infinity", "NaN", and other similar invalid inputs.
+ */
+export function asBiggystring(raw: unknown): string {
+  const clean = asString(raw)
+  if (/^-?(?:0|[1-9]\d*)(?:\.\d+)$/.test(clean)) return clean
```
paullinator: This is stricter than biggystring in that in it rejects '1.' or '3.' but biggystring allows it. No big deal as I don't think we generate those in the app. 

However this regex rejects whole numbers like '0', '1', '10'.  This causes the snapshot diff above. I submitted a patch to fix. Pls review

paullinator (changes requested):

paullinator (approved):

src/util/cleaners.ts
swansontec: Oops, thanks for finding the missing `?`.

PR #4818, opened 2024-02-03
Improve Keyboard Awareness

samholmes:
PR changes:

- Fix various misses in the calculation for insets in the SceneWrapper
- Removes old KeyboardTracker component in favor of react-native-keyboard-controller and Reanimated components
- Adds the ability to track keyboard on scroll-SceneWrappers and defaults to `avoidKeyboard` when `scroll={true}`

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206333633923757


swansontec (changes requested):
In the "Remove tab height from insets when keyboard is open" commit, you mention that `WalletListScene` and `TransactionListScene` are broken now. This PR doesn't contain a commit that fixes them, though!

swansontec (approved):
OK, we clarified that this *fixed* the scenes in question. They comment was trying to explain that they were *already* broken, not that the commit broke them.

PR #4824, opened 2024-02-07
Jon/analytics/referral

Jon-edge:
Make any analytics event always report account and device referral data.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206501486890533


src/__tests__/scenes/ChangePasswordScene.test.tsx
```diff
@@ -2,18 +2,23 @@ import { describe, expect, it } from '@jest/globals'
 import * as React from 'react'
 import { createRenderer } from 'react-test-renderer/shallow'
```
Jon-edge, line 3: Attempted to use `import TestRenderer from 'react-test-renderer'`, but errors for wrapped LoginUi scenes.

swansontec: Please create a P4 Asana task in 4.2.0 to fix this, then. We are probably importing some native library that isn't mocked properly. As it stands, the snapshots don't include *any* rendered output from the login-ui scenes, so we aren't really testing anything at this point.

Another way would be to put `it.skip`, so we just skip the test entirely, but still use the deep renderer. No sense running a useless test.

src/components/scenes/PasswordRecoveryScene.tsx
```diff
@@ -1,47 +1,34 @@
...
+  return (
+    <SceneWrapper>
+      <PasswordRecoveryScreen
+        branding={{ appName: config.appName }}
+        account={account}
+        context={context}
+        onLogEvent={logEvent}
+        onComplete={handleComplete}
+        // @ts-expect-error
+        onCancel={handleComplete}
```
swansontec, line 34: The reason we have `@ts-expect-error` is that there is no `onCancel` prop. So you could just delete this one.

src/components/scenes/SendScene2.tsx
```diff
@@ -754,7 +754,8 @@ const SendComponent = (props: Props) => {
       if (alternateBroadcast != null) {
         broadcastedTx = await alternateBroadcast(signedTx)
       } else {
-        broadcastedTx = await coreWallet.broadcastTx(signedTx)
+        // broadcastedTx = await coreWallet.broadcastTx(signedTx)
+        console.debug('sent')
```
swansontec, line 758: Please put this back.

swansontec (changes requested):
We definitely need to fix the send functionality before merging this!

src/util/tracking.ts
```diff
@@ -128,37 +131,54 @@ export function trackError(
...
+
+        // Populate referral params:
+        const state = getState()
+        const { deviceReferral, account } = state
+        const { accountReferral } = account
+        params.refDeviceId = deviceReferral.installerId
+        params.refDeviceCcs = deviceReferral.currencyCodes
+
+        const { creationDate, installerId } = accountReferral
+        params.refAccountDate = installerId == null || creationDate == null ? undefined : creationDate.toISOString().replace(/-\d\dT.*/, '')
```
paullinator, line 152: We shouldn't be reporting the device installerId. It's only relevant if we don't have an account installerId. So really we should only have

```
params.installerId = accountReferral.installerId ?? deviceReferral.installerId
params.refCurrencyCodes = accountReferral.currencyCodes ?? deviceReferral. currencyCodes
params.creationDate = accountReferral.creationDate
```

src/util/tracking.ts
```diff
@@ -48,22 +50,23 @@ export type TrackingEventName =
...
   currencyCode?: string // Wallet currency code
   dollarValue?: number // Conversion amount, in USD
   error?: unknown | string // Any error
-  installerId?: string // Account installerId, i.e. referralId
   orderId?: string // Unique order identifier provided by plugin
   pluginId?: string // Plugin that provided the conversion
   numSelectedWallets?: number // Number of wallets to be created
   destCurrencyCode?: string
-  destExchangeAmount?: string
+  destExchangeAmount?: number
```
paullinator, line 63: why did you change this to a number? This just added more work in the PR for everywhere this is passed into logEvent.

paullinator (changes requested):

src/util/tracking.ts
Jon-edge: This was changed because we need to make sure we're always reporting numeric values to PostHog so we can properly filter gt, Lt, etc

src/util/tracking.ts
Jon-edge: We need this to report ID's prior to login

src/util/tracking.ts
Jon-edge: Changed back to a string and just did a check within `logEvent` to ensure the string represents a number. 

This is very important because if we pollute a value field with a string that can't be interpreted as a number, we lose numeric comparison functionality for that field in PostHog.

swansontec: I fully support using numbers as numbers. We should only be using biggystring in cases where we need absolute correctness (like sending customer funds), and we are willing to pay the price in terms of performance and inconvenience. Otherwise switching to plain numbers is a helpful cleanup.

src/components/scenes/PasswordRecoveryScene.tsx
swansontec: Not resolved yet.

src/util/tracking.ts
swansontec: After a lot of back & forth, we did agree to track both values separately (Jon, William, and Peter). I agree with Paul that it seems unnecessary at first glance, but after thinking about it more deeply, separate tracking is in fact the correct solution.

swansontec (approved):

src/components/scenes/PasswordRecoveryScene.tsx
Jon-edge: Accidentally squashed my fixup earlier and missed this in the re-rebase to split it up

src/util/tracking.ts
paullinator: While I mostly agree, the risk of having a conversion that leaves us with an invalid, non-parsable string (scientific notation) is risky. At minimum our nomenclature should be consistent. `nativeAmount` and `exchangeAmount` are used extensively throughout the app and they are always strings. The introduction of nativeAmount in this PR https://github.com/EdgeApp/edge-react-gui/pull/4825 actually introduces a bug due to blindly converting `nativeAmount` to a number and assuming it's usable.

swansontec: That makes sense.

src/util/tracking.ts
paullinator: ok. That's fair.  Can I just request a simple rename as I don't think the current codes are properly represented, especially the term `deviceId` since it is definitely NOT a deviceId, and we can't change them later without breaking the reporting tools. 

I'd ask for

```
refDeviceInstallerId
refDeviceCurrencyCodes
refAccountInstallerId
refAccountCurrencyCodes
refAccountDate
```

paullinator (changes requested):

src/util/tracking.ts
Jon-edge: We do stringify numbers coming from our partners in the plugins sometimes, but I'm opting to keep it consistent with string types only in our internal-facing API for consistency.

paullinator (approved):

PR #4829, opened 2024-02-08
Make FilledTextInput take up constant space

Jon-edge:
Screenshot 2024-02-09 at 3 25 24 PM

- Used negative margins. Not ideal, but a lower impact change to avoid affecting the layouts app-wide.
- Only apply this special styling if there is no counter, since we already styled the app according to visibility of counters.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206471966737303


samholmes (changes requested):
Please provide before and after screenshots

samholmes (changes requested):
Requesting a review-bar to discuss approach

samholmes (approved):

PR #4834, opened 2024-02-09
UI4: Fix Nugget Styling

Jon-edge:
- Remove 'flex: 1' where appropriate.
- Restyle ButtonUi4 and fix doubled custom margins.
- Increase ButtonUi4 tappable areas.
- MainButton: No callers directly called ButtonUi4 with custom margins.
  - Built-in horizontal margins removed. Remnant from when we had full width buttons.
  - Mostly doubled callers' custom margins to account for fix.
  - Removed custom horizontal margins and capped bottom margins for some users.

Red background fill shows increased tappable areas
image

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206471966737297
  - https://app.asana.com/0/0/1206501486890527

### DIFF
--- a/src/components/themed/MainButton.tsx
+++ b/src/components/themed/MainButton.tsx
@@ -40,7 +40,7 @@ interface Props {
  * A stand-alone button to perform the primary action in a modal or scene.
  */
 export function MainButton(props: Props) {
-  const { children, disabled = false, label, marginRem = [0, 1, 0, 1], onPress, type = 'primary', paddingRem, layout, spinner = false } = props
+  const { children, disabled = false, label, marginRem, onPress, type = 'primary', paddingRem, layout, spinner = false } = props
 
   return (
     <ButtonUi4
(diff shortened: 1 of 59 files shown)


src/components/themed/MainButton.tsx
```diff
@@ -40,7 +40,7 @@ interface Props {
  * A stand-alone button to perform the primary action in a modal or scene.
  */
 export function MainButton(props: Props) {
-  const { children, disabled = false, label, marginRem = [0, 1, 0, 1], onPress, type = 'primary', paddingRem, layout, spinner = false } = props
+  const { children, disabled = false, label, marginRem = 0, onPress, type = 'primary', paddingRem, layout, spinner = false } = props
```
swansontec, line 43: Optional: Because you are passing `marginRem = 0` as the default, none of the `MainButton` components will get extra tappable area. If you let this just be `undefined`, then your padding stretch will kick in.

swansontec (changes requested):

PR #4855, opened 2024-02-18
Paul/perf mon

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206533398594286


src/app.ts
```diff
@@ -18,6 +19,7 @@ import { NumberMap } from './types/types'
 import { log, logToServer } from './util/logger'
 import { initInfoServer } from './util/network'
 
+// LogBox.ignoreAllLogs()
```
swansontec, line 22: Leftover comment?

package.json
```diff
@@ -168,6 +169,7 @@
     "react-native-email-link": "^1.14.5",
     "react-native-fast-image": "^8.5.11",
     "react-native-fast-shadow": "^0.1.0",
+    "react-native-file-access": "^3.0.5",
```
swansontec, line 172: This seems unused?

swansontec (changes requested):
I think there is an extra native dependency in here. Can we remove it?

src/app.ts
paullinator: Actually. I should have added a comment. I intentionally added that to easily disable the logbox.

package.json
paullinator: It was asked to be added by bugsnag

https://docs.bugsnag.com/performance/integration-guides/react-native/

It appears to be imported extensively in their code although it's not a peer dependency.

swansontec (approved):

PR #4858, opened 2024-02-20
Sanitize mnemonic imports by standardizing capitalization and removing extra spaces

peachbits:
Also, show errors from import and make import field multiline


peachbits:
image

src/components/scenes/CreateWalletImportScene.tsx
```diff
@@ -46,7 +46,16 @@ const CreateWalletImportComponent = (props: Props) => {
...
   const handleNext = useHandler(async () => {
     textInputRef.current?.blur()
-    const cleanImportText = importText.trim()
+    let cleanImportText = importText.trim()
+
+    // Clean up mnemonic seeds
+    const cleanImportTextArray = cleanImportText.split(' ')
+    if (cleanImportTextArray.length > 1) {
+      cleanImportText = cleanImportTextArray
+        .filter(part => part !== ' ') // remove extra spaces
```
paullinator, line 55: can't possibly have an empty ' ' as `part` since that's your split character. But you can have an empty string if there are two spaces in a row. Change to

`.filter(part => part !== '') // remove empty spaces`

paullinator (changes requested):

src/components/scenes/CreateWalletImportScene.tsx
```diff
@@ -57,6 +66,7 @@ const CreateWalletImportComponent = (props: Props) => {
     const promises = pluginIds.map(
       async pluginId =>
         await currencyConfig[pluginId].importKey(cleanImportText).catch(e => {
+          showError(e)
```
paullinator, line 69: Are there specific error codes thrown here that we can translate. If so show the translation.

paullinator (changes requested):

src/components/scenes/CreateWalletImportScene.tsx
peachbits: Unfortunately, no, the currency plugins all use generic errors

paullinator (approved):
Nice! I like the tests

PR #4864, opened 2024-02-22
Fix currencyCode red error

paullinator:
Check and filter out null token objects

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206662647999769


samholmes (approved):
I wonder why there'd be `enabledTokenIds` that are not defined in `allTokens`. Approved anyway.

PR #4874, opened 2024-02-26
Only call loadAccountReferral on new accounts

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206689987490075

### DIFF
--- a/src/actions/LoginActions.tsx
+++ b/src/actions/LoginActions.tsx
@@ -53,10 +53,11 @@ export function initializeAccount(navigation: NavigationBase, account: EdgeAccou
     const syncedSettings = await readSyncedSettings(account)
     const { walletsSort } = syncedSettings
     dispatch({ type: 'LOGIN', data: { account, walletSort: walletsSort } })
-    await dispatch(loadAccountReferral(account))
-    const newAccount = account.newAccount
+    const { newAccount } = account
+    const referralPromise = dispatch(loadAccountReferral(account))
 
     if (newAccount) {
+      await referralPromise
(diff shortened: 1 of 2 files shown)


src/actions/LoginActions.tsx
```diff
@@ -53,8 +53,10 @@ export function initializeAccount(navigation: NavigationBase, account: EdgeAccou
     const syncedSettings = await readSyncedSettings(account)
     const { walletsSort } = syncedSettings
     dispatch({ type: 'LOGIN', data: { account, walletSort: walletsSort } })
-    await dispatch(loadAccountReferral(account))
-    const newAccount = account.newAccount
+    const { newAccount } = account
+    if (newAccount) {
+      await dispatch(loadAccountReferral(account))
+    }
```
swansontec, line 59: This will break Peter's analytics. We need to load this information on every login, but we only need to *await* the load for new accounts. Instead, I would propose:

```ts
const { newAccount } = account
const referralPromise = dispatch(loadAccountReferral(account))

if (newAccount) {
  await referralPromise
  ... plus the rest of the wallet-creation stuff from the existing if statement right below ...

} else {
  navigation.push('edgeApp', {})
  // Insert this as a new line in the existing `else` statment:
  referralPromise.catch(() => { /* silently fail - we just get worse analytics */ })
}
```

swansontec (changes requested):

PR #4888, opened 2024-03-06
Jon/edge-wrapped-touch

Jon-edge:
New built-in await/debounce features were only implemented but not integrated. The design spec does not cleanly lend itself to be used for the most common `Button` components without further manual work due to the potential manual overrides callers of button components may be doing.

Besides simple search/replace of jsx and fixing imports, manual integration changes include:

1. Ensuring callers' defined `onPress` handlers return `void` or `Promise<void>`. Some callers were passing `onPress: ()-> null | undefined`. 
Unsure how it worked with the original `'react-native'` `Touchable*` but was unable to replicate support for this use case in the new components.

2. Changing all users of `Touchable*` from `'react-native-reanimated'` to the `'react-native'` counterpart used in all new Edge-wrapped Touchables.
On code and git history inspection, it was determined that users of the `'react-native-reanimated'` version of Touchable either:
- Didn't need to (files I personally authored and wasn't aware of the auto-import pitfall at the time)  
- Accidentally changed from the `'react-native'` counterpart somewhere in the commit history (everything else).

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205298523695455

### DIFF
--- a/src/components/buttons/MinimalButton.tsx
+++ b/src/components/buttons/MinimalButton.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react'
-import { ActivityIndicator, Text, TextStyle, TouchableOpacity, ViewStyle } from 'react-native'
+import { ActivityIndicator, Text, TextStyle, ViewStyle } from 'react-native'
 import { cacheStyles } from 'react-native-patina'
 
 import { usePendingPress } from '../../hooks/usePendingPress'
 import { fixSides, mapSides, sidesToMargin } from '../../util/sides'
+import { EdgeTouchableOpacity } from '../common/EdgeTouchableOpacity'
 import { Theme, useTheme } from '../services/ThemeContext'
 
 interface Props {
@@ -49,14 +50,14 @@ const MinimalButtonComponent = (props: Props) => {
   )
 
   return (
-    <TouchableOpacity disabled={disabled || pending} style={memoizedStyle} onPress={handlePress}>
+    <EdgeTouchableOpacity disabled={disabled || pending} style={memoizedStyle} onPress={handlePress}>
       {pending ? null : (
         <Text adjustsFontSizeToFit minimumFontScale={0.25} numberOfLines={1} style={highlighted ? styles.labelSelected : styles.label}>
           {label}
         </Text>
       )}
       {!pending ? null : <ActivityIndicator color={theme.secondaryButtonText} style={styles.spinner} />}
-    </TouchableOpacity>
+    </EdgeTouchableOpacity>
   )
 }
 
--- a/src/components/hoc/withExtendedTouchable.tsx
+++ b/src/components/hoc/withExtendedTouchable.tsx
@@ -1,0 +1,89 @@
+import React, { ComponentType, ReactNode, useEffect, useRef, useState } from 'react'
+import { GestureResponderEvent, TouchableHighlightProps, TouchableOpacityProps, TouchableWithoutFeedbackProps } from 'react-native'
+
+import { useHandler } from '../../hooks/useHandler'
+import { showError } from '../services/AirshipInstance'
+
+type PressFn = (event: GestureResponderEvent) => void | Promise<void>
+
+interface ExtendedProps {
+  onPress?: PressFn
+  onLongPress?: PressFn
+  awaitChild?: ReactNode // What to render during async await, if different from children. This does not get rendered while debounce is in effect, only while awaiting the onPress.
+  debounce?: boolean // Enforce a strict 500ms debounce delay for synchronous press handlers.
+}
+
+type PropsWithOptionalChildren<P = unknown> = P & { children?: ReactNode }
+
+const DEBOUNCE_DELAY = 500
+
+export function withExtendedTouchable<T extends TouchableOpacityProps | TouchableWithoutFeedbackProps | TouchableHighlightProps>(
+  WrappedComponent: ComponentType<PropsWithOptionalChildren<T>>
+): React.FC<PropsWithOptionalChildren<T> & ExtendedProps> {
+  const ExtendedTouchable: React.FC<PropsWithOptionalChildren<T> & ExtendedProps> = ({
+    children,
+    onPress,
+    onLongPress,
+    awaitChild,
+    debounce = false,
+    ...restProps
+  }) => {
+    const [isDebouncing, setIsDebouncing] = useState<boolean>(false)
+    const [isAwaiting, setIsAwaiting] = useState<boolean>(false)
+    const timeoutRef = useRef<NodeJS.Timeout | null>(null)
+
+    useEffect(() => {
+      return () => {
+        if (timeoutRef.current !== null) {
+          clearTimeout(timeoutRef.current)
+        }
+      }
+    }, [])
+
+    const runPress = (event: GestureResponderEvent, pressFn?: PressFn) => {
+      if (pressFn == null) return
+      if (isDebouncing || isAwaiting) return
+
+      // Set debounce state before we even know if onPress is async or not to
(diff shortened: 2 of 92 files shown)


src/components/buttons/MinimalButton.tsx
```diff
@@ -49,14 +50,14 @@ const MinimalButtonComponent = (props: Props) => {
   )
 
   return (
-    <TouchableOpacity disabled={disabled || pending} style={memoizedStyle} onPress={handlePress}>
+    <EdgeTouchableOpacity disabled={disabled || pending} style={memoizedStyle} onPress={handlePress}>
```
swansontec, line 53: This guy has also incorporated the `usePendingPress` hook.

src/components/settings/SettingsRow.tsx
```diff
@@ -42,7 +43,7 @@ const SettingsRowComponent = (props: Props) => {
   const [pending, handlePress] = usePendingPress(onPress)
 
   return (
-    <TouchableHighlight accessible={false} underlayColor={theme.settingsRowPressed} style={styles.row} onPress={handlePress}>
+    <EdgeTouchableHighlight accessible={false} underlayColor={theme.settingsRowPressed} style={styles.row} onPress={handlePress}>
```
swansontec, line 46: This guy has also incorporated the `usePendingPress` hook.

src/components/themed/ManageTokensRow.tsx
```diff
@@ -84,9 +85,9 @@ export const ManageTokensRowComponent = (props: Props) => {
         <EdgeText style={styles.displayName}>{token.displayName}</EdgeText>
       </View>
       {!isCustom ? null : (
-        <TouchableOpacity style={styles.editIcon} onPress={handleEdit}>
+        <EdgeTouchableOpacity style={styles.editIcon} onPress={handleEdit}>
```
swansontec, line 88: This guy has also incorporated the `usePendingPress` hook, and uses it to drive an animation, so no need for the `EdgeTouchableOpacity`.

src/components/themed/MiniButton.tsx
```diff
@@ -47,14 +48,14 @@ export function MiniButton(props: Props) {
   }
 
   return (
-    <TouchableOpacity disabled={disabled || pending} style={[styles.button, dynamicStyles]} onPress={handlePress}>
+    <EdgeTouchableOpacity disabled={disabled || pending} style={[styles.button, dynamicStyles]} onPress={handlePress}>
```
swansontec, line 51: This guy has also incorporated the `usePendingPress` hook.

src/components/themed/ShareButtons.tsx
```diff
@@ -34,10 +35,10 @@ function ShareButton(props: { text: string; onPress: () => Promise<void>; icon:
   const [pending, handlePress] = usePendingPress(onPress)
 
   return (
-    <TouchableOpacity accessible={false} style={styles.button} onPress={handlePress} disabled={pending}>
+    <EdgeTouchableOpacity accessible={false} style={styles.button} onPress={handlePress} disabled={pending}>
```
swansontec, line 38: This guy has also incorporated the `usePendingPress` hook.

swansontec (commented):
I think this is a really good & necessary cleanup, and I like a lot of things about it.

I think we need a discussion around the mixing of `usePendingPress` and `EdgeTouchable*`. The two tools do the same job.

- One option is to just use both. The `EdgeTouchable*` logic will end up being a no-op, because `usePendingPress` returns a synchronous `onPress` callback, and includes its own error handling too.
- We could base the `withExtendedTouchable` HOC on the `usePendingPress` hook, so the logic lives in one place.
- Otherwise, we could try to eliminate the `usePendingPress` hook and just fold all this logic into the `withExtendedTouchable` HOC. However, we would need to think more carefully about how we handle spinners, particularly with fade-in and fade-out animations.

If we go with option 1, then we could pretty much approve this as-is, aside from the missing `showError`

src/components/hoc/withExtendedTouchable.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const memoizedProps = useMemo(
+      () => ({
+        // HACK: Can't get typescript to not complain when defining restProps as T
+        ...(restProps as any),
+        disabled: isDebouncing || isAwaiting,
+        onPress: handlePress,
+        onLongPress: handleLongPress
+      }),
+      [restProps, isDebouncing, isAwaiting, handlePress, handleLongPress]
+    )
```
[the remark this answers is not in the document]
Jon-edge: Yeah I was unsure about the memoization here, thanks for digging deeper.

src/components/hoc/withExtendedTouchable.tsx
```diff
@@ -0,0 +1,98 @@
...
+        setIsAwaiting(false)
+
+        // Pass along the error after resetting busy state. Not this component's
+        // job to handle errors.
+        throw err
+      }
+    }
+
+    const handlePress = useHandler((event: GestureResponderEvent) => {
+      if (onPress == null) return
```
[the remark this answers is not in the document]
Jon-edge: I think TS complained without this

src/components/hoc/withExtendedTouchable.tsx
```diff
@@ -0,0 +1,98 @@
...
+      runPress(onPress, event)
+    })
+
+    const handleLongPress = useHandler((event: GestureResponderEvent) => {
+      if (onLongPress == null) return
```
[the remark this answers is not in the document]
Jon-edge: I think TS complained without this

src/components/ui4/ButtonUi4.tsx
```diff
@@ -152,13 +153,13 @@ export function ButtonUi4(props: Props) {
   }, [disabled, hideContent, layout, mini, paddingRem, styles, theme, type])
 
   return (
-    <TouchableOpacity disabled={disabled || pending || spinner} style={touchContainerStyle} onPress={handlePress} testID={testID}>
+    <EdgeTouchableOpacity disabled={disabled || pending || spinner} style={touchContainerStyle} onPress={handlePress} testID={testID}>
```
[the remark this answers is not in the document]
Jon-edge: Yeah I think we need to audit all the callers and their spinner/pending/disabled use cases. Leaving as-is for now

src/components/buttons/MinimalButton.tsx
Jon-edge: This caller explicitly sets different opacities depending on the type of "busy" state

Jon-edge: We can probably blanket apply this styling to any `EdgeTouchableOpacity` but I'd like to wait to decide on that in case we missed something

src/components/settings/SettingsRow.tsx
Jon-edge: Has animation logic tied to `pending`

src/components/themed/MiniButton.tsx
Jon-edge: Yeah we really ought to decide on one visual transition and stick to it globally...

src/components/themed/ShareButtons.tsx
Jon-edge: This is the only one I can 100% update without changing visual behavior.

Jon-edge:
I think we still want to use our `Edge*` components because we need the debouncing logic. I guess I can add the debouncing logic into the `usePendingPress` hook for now

swansontec (changes requested):
One tiny change is needed.

swansontec (approved):

PR #4890, opened 2024-03-06
Refactor Conversion Tracking

Jon-edge:
Conversion tracking was broken in some cases. Wrong values, including our very ambiguous `pluginId`, were being used to calculate dollar conversion values. As a result, `logEvent` either silently failed without sending out the tracking event, or reported the incorrect amount values.

Refactor tracking in general, introduce an MVP of `CryptoAmount`, and make props more explicit. All of those combined ought to reduce the likelihood of similar future mistakes.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206683289208449

### DIFF
--- a/src/components/scenes/Fio/FioAddressRegisterSelectWalletScene.tsx
+++ b/src/components/scenes/Fio/FioAddressRegisterSelectWalletScene.tsx
@@ -1,4 +1,3 @@
-import { mul, toFixed } from 'biggystring'
 import { EdgeAccount, EdgeCurrencyConfig, EdgeCurrencyWallet, EdgeDenomination, EdgeTransaction } from 'edge-core-js'
 import * as React from 'react'
 import { Alert, Image, View } from 'react-native'
@@ -6,12 +5,12 @@ import { sprintf } from 'sprintf-js'
 
 import { FIO_STR } from '../../../constants/WalletAndCurrencyConstants'
 import { lstrings } from '../../../locales/strings'
-import { getExchangeDenomByCurrencyCode, selectDisplayDenom } from '../../../selectors/DenominationSelectors'
+import { selectDisplayDenom } from '../../../selectors/DenominationSelectors'
 import { config } from '../../../theme/appConfig'
 import { connect } from '../../../types/reactRedux'
 import { EdgeSceneProps } from '../../../types/routerTypes'
 import { EdgeAsset } from '../../../types/types'
-import { getTokenIdForced } from '../../../util/CurrencyInfoHelpers'
+import { CryptoAmount } from '../../../util/CryptoAmount'
 import { getWalletName } from '../../../util/CurrencyWalletHelpers'
 import { getRegInfo } from '../../../util/FioAddressUtils'
 import { logEvent, TrackingEventName, TrackingValues } from '../../../util/tracking'
@@ -30,7 +29,6 @@ interface StateProps {
   fioPlugin?: EdgeCurrencyConfig
   fioWallets: EdgeCurrencyWallet[]
   fioDisplayDenomination: EdgeDenomination
-  pluginId: string
   isConnected: boolean
 }
 
@@ -150,12 +148,16 @@ export class FioAddressRegisterSelectWallet extends React.Component<Props, Local
         this.props.onSelectWallet(walletId, paymentCurrencyCode)
 
         const wallet = account.currencyWallets[walletId]
-        const exchangeDenomination = getExchangeDenomByCurrencyCode(wallet.currencyConfig, paymentCurrencyCode)
-        const exchangeAmount = allPaymentInfo[paymentCurrencyCode].amount
-        let nativeAmount = mul(exchangeAmount, exchangeDenomination.multiplier)
-        nativeAmount = toFixed(nativeAmount, 0, 0)
+        const { amount: exchangeAmount, address: paymentAddress } = allPaymentInfo[paymentCurrencyCode]
 
-        const tokenId = getTokenIdForced(account, wallet.currencyInfo.pluginId, paymentCurrencyCode)
+        const cryptoAmount = new CryptoAmount({
+          exchangeAmount,
+          currencyCode: paymentCurrencyCode,
+          currencyConfig: wallet.currencyConfig
+        })
+
+        const { tokenId, nativeAmount } = cryptoAmount
+
         const sendParams: SendScene2Params = {
           walletId,
           tokenId,
@@ -170,7 +172,7 @@ export class FioAddressRegisterSelectWallet extends React.Component<Props, Local
             spendTargets: [
               {
                 nativeAmount,
-                publicAddress: allPaymentInfo[paymentCurrencyCode].address
+                publicAddress: paymentAddress
               }
             ],
             metadata: {
@@ -190,8 +192,12 @@ export class FioAddressRegisterSelectWallet extends React.Component<Props, Local
                 [{ text: lstrings.string_ok_cap }]
               )
 
-              const { currencyCode, pluginId } = wallet.currencyInfo
-              onLogEvent('Fio_Handle_Register', { nativeAmount, pluginId, currencyCode })
+              onLogEvent('Fio_Handle_Register', {
+                conversionValues: {
+                  conversionType: 'crypto',
+                  cryptoAmount
+                }
+              })
               navigation.navigate('homeTab', { screen: 'home' })
             }
           }
@@ -283,7 +289,6 @@ export const FioAddressRegisterSelectWalletScene = connect<StateProps, DispatchP
     fioPlugin: state.core.account.currencyConfig.fio,
     fioDisplayDenomination: selectDisplayDenom(state, params.selectedWallet.currencyConfig, null),
     defaultFiatCode: state.ui.settings.defaultIsoFiat,
-    pluginId: params.selectedWallet.currencyInfo.pluginId,
     isConnected: state.network.isConnected
   }),
   dispatch => ({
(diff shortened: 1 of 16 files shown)


src/util/CryptoAmount.ts
```diff
@@ -0,0 +1,102 @@
...
+ *   currencyCode: 'USDC',
+ *   nativeAmount: '12345678',
+ * })
+ */
+export class CryptoAmount {
+  public readonly pluginId: string
+  public readonly chainCode: string
+  public readonly currencyCode: string // Token currency code or chain currency code if no token specified
+  public readonly exchangeAmount: string
+  public readonly nativeAmount: string
```
swansontec, line 53: This class makes no separation between *what* value we are talking about, and how we *display* that value. The concepts are jumbled together.

If we want to uniquely identify a value, we need `{ nativeAmount, pluginId, tokenId }`, and that is all. With those three values, we know what coin we are talking about and how much there is. Anything else, we can derive.

Therefore, I would expect the class members to be `tokenId: string, nativeAmountString, currencyConfig: EdgeCurrencyConfig`. The `currencyConfig` object contains the `pluginId`, and is our link to the core.

Then, the class can expose all the existing things using accessor methods, like this:

```ts
get currencyCode(): string {
  const { currencyCode } = this.tokenId == null ? this.currencyConfig.currencyInfo : this.currencyConfig.allTokens[this.tokenId]
  return currencyCode
}

get chainCode(): string {
  return this.currencyConfig.currencyInfo.currencyCode
}

get exchangeAmount(): string {
  const { multiplier } = getExchangeDenom(this.currencyConfig, this.tokenId)
  return div(nativeAmount, multiplier, mulToPrecision(multiplier))
}

get pluginId (): string {
  return this.currencyConfig.currencyInfo.pluginId
}
```

And besides these, we can provide formatting methods like:

```ts
formatDisplayAmount({ redux, precision: ... }): string {
  const {multiplier} = getDisplayDenom(redux, this.currencyConfig, this.tokenId)
  ...
}

formatExchangeAmount({ precision: ... }): string {
  ...
}
```

So you see that the class can be convenient and fully-featured, while still be "lean" with what it stores.

At the end of the day, storing the `currencyCode` as a class member is banned. Absolutely do not store the currency code. We can maybe discuss storing the `exchangeAmount`, but *only* as an optimization.

If you constructor wants to take a `currencyCode`, that's fine. Just mark it `/** @deprecated */` and immediatrely convert it to a `tokenId`, and I have no complaint.

Jon-edge: If you look at the usage, you see that when we interface with external providers, we often have `exchangeAmount` and `currencyCode` as the input, where we then repeatedly jump through the same hoops to convert values from there. DRYing that out was one of the main design considerations for this class. I can make it more explicit though to make it clear that internally we don't want to start from those inputs.

For the purposes of this PR, I'm not really trying to handle display, I'm mostly keeping the implementation to what I need it for for now, which is reporting `exchangeAmount` and `dollarValues`

src/util/CryptoAmount.ts
```diff
@@ -0,0 +1,148 @@
...
+      currencyCode?: never
+    }
+  | {
+      tokenId?: never
+      currencyCode: string
+    }
+  | {
+      tokenId?: never
+      currencyCode?: never
+    }
```
swansontec, line 28: During the 4.0 sprint, Paul did a huge amount of work to make `tokenId: string | null` everywhere, so folks *must* to provide it, even if it's just `null` for the parent currency. We had a number of bugs where people simply forgot to pass it because it was optional, and we got the wrong coin. With that in mind, this should be:

```ts

type TokenOrCurrencyCodeArgs =
  | {
      tokenId: EdgeTokenId
      currencyCode?: never
    }
  | {
      tokenId?: never
      currencyCode: string
    }
```

That will force the user to provide one or the other, but not both.

src/util/CryptoAmount.ts
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * const cryptoAmountWithNative = CryptoAmount.createWithNativeAmount({
+ *   currencyCode: 'USDC',
+ *   nativeAmount: '12345678',
+ * })
+ */
+export class CryptoAmount {
+  public readonly currencyConfig: EdgeCurrencyConfig
+  public readonly nativeAmount: string
+  public readonly tokenId: string | null
```
swansontec, line 62: Perfect!

src/util/CryptoAmount.ts
```diff
@@ -0,0 +1,148 @@
...
+    if (currencyCode != null) {
+      // Ensure currencyCode is recognized, if given as a constructor argument.
+      const foundTokenId = Object.keys(currencyConfig.allTokens).find(edgeToken => currencyConfig.allTokens[edgeToken].currencyCode === currencyCode)
+      if (foundTokenId == null) {
+        throw new Error(`CryptoAmount: Could not find tokenId for currencyCode: ${currencyCode}, pluginId: ${currencyConfig.currencyInfo.pluginId}.`)
+      } else {
+        this.tokenId = foundTokenId
+      }
+    } else {
+      this.tokenId = tokenId ?? null
```
swansontec, line 78: The `?? null` part should not be needed once the types are fixed.

src/util/CryptoAmount.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  getExchangeDenom(): EdgeDenomination {
+    const { allTokens, currencyInfo } = this.currencyConfig
+
+    if (this.tokenId == null) return currencyInfo.denominations[0]
+    for (const tokenId of Object.keys(allTokens)) {
+      const token = allTokens[tokenId]
+      if (token.currencyCode === this.currencyCode) return token.denominations[0]
+    }
```
swansontec, line 144: This search should not be needed, because we already know `this.tokenId`.

src/components/scenes/Fio/FioAddressRegisterSelectWalletScene.tsx
```diff
@@ -151,12 +148,16 @@ export class FioAddressRegisterSelectWallet extends React.Component<Props, Local
...
-        nativeAmount = toFixed(nativeAmount, 0, 0)
+        const { amount: exchangeAmount, address: paymentAddress } = allPaymentInfo[paymentCurrencyCode]
+
+        const cryptoAmount = new CryptoAmount({
+          exchangeAmount,
+          currencyCode: paymentCurrencyCode,
+          currencyConfig: wallet.currencyConfig
+        })
+
+        const { tokenId, nativeAmount } = cryptoAmount
```
swansontec, line 159: Oooh, nice. Your new class may be even more useful than I expected. I like where this is going.

src/plugins/gui/providers/banxaProvider.ts
```diff
@@ -652,7 +652,7 @@ export const banxaProvider: FiatProviderFactory = {
                                 conversionType: 'sell',
                                 destFiatCurrencyCode: fiatCurrencyCode,
                                 destFiatAmount: priceQuote.fiat_amount,
-                                sourceAmount: CryptoAmount.fromExchangeAmount({
+                                sourceAmount: CryptoAmount({
```
swansontec, line 655: Don't you need a `new` here? Also drop the `pluginId`.

src/plugins/gui/providers/kadoProvider.ts
```diff
@@ -767,7 +767,7 @@ export const kadoProvider: FiatProviderFactory = {
                           conversionType: 'sell',
                           destFiatCurrencyCode: 'USD',
                           destFiatAmount: fiatAmount,
-                          sourceAmount: CryptoAmount.fromExchangeAmount({
+                          sourceAmount: CryptoAmount({
```
swansontec, line 770: Don't you need a `new` here? Also drop the `pluginId`.

src/plugins/gui/providers/paybisProvider.ts
```diff
@@ -614,7 +614,7 @@ export const paybisProvider: FiatProviderFactory = {
                           conversionType: 'sell',
                           destFiatCurrencyCode: fiatCurrencyCode,
                           destFiatAmount: fiatAmount,
-                          sourceAmount: CryptoAmount.fromExchangeAmount({
+                          sourceAmount: CryptoAmount({
```
swansontec, line 617: Don't you need a `new` here? Also drop the `pluginId`.

src/util/tracking.ts
```diff
@@ -59,22 +56,44 @@ export type TrackingEventName =
 
 export type OnLogEvent = (event: TrackingEventName, values?: TrackingValues) => void
 
+// Known dollar amount revenue
```
swansontec, line 59: Can you use `/** */` for these comments? Then VSCode will treat them as documentation and show them when you hover the type name.

swansontec (changes requested):
This is looking really good now!

We need to do a few more changes, so go ahead and squash the fixups before you do your final round of changes.

src/util/CryptoAmount.ts
Jon-edge: I actually started with what you suggested but was opting to appease current callers. I can see how this impl would be dangerous in the same way as the issue you mentioned in the future.

src/plugins/gui/providers/banxaProvider.ts
Jon-edge: Global replace fail my bad

src/util/tracking.ts
Jon-edge: Thought that only was for fns! This is obvious in hindsight thanks

swansontec (approved):

PR #4897, opened 2024-03-08
Paul/small fixes

paullinator:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206673263365906

### DIFF
--- a/src/components/cards/PoweredByCard.tsx
+++ b/src/components/cards/PoweredByCard.tsx
@@ -54,6 +54,7 @@ const getStyles = cacheStyles((theme: Theme) => ({
     flexDirection: 'column'
   },
   poweredByContainer: {
+    marginHorizontal: theme.rem(0.25),
     flexDirection: 'row',
     alignItems: 'center'
   },
--- a/src/components/scenes/TransactionDetailsScene.tsx
+++ b/src/components/scenes/TransactionDetailsScene.tsx
@@ -84,6 +84,7 @@ const TransactionDetailsComponent = (props: Props) => {
 
   // Look up wallet stuff:
   const isoFiatCurrencyCode = useWatch(wallet, 'fiatCurrencyCode')
+  const walletName = useWatch(wallet, 'name')
   const fiatCurrencyCode = isoFiatCurrencyCode.replace('iso:', '')
   const fiatSymbol = getSymbolFromCurrency(fiatCurrencyCode)
 
@@ -98,7 +99,9 @@ const TransactionDetailsComponent = (props: Props) => {
   )
 
   // Look up the historical price:
-  const isoDate = new Date(date * 1000).toISOString()
+  const dateObj = new Date(date * 1000)
+  const dateString = dateObj.toLocaleString()
+  const isoDate = dateObj.toISOString()
   const historicRate = useHistoricalRate(`${currencyCode}_${isoFiatCurrencyCode}`, isoDate)
   const historicFiat = historicRate * Number(absExchangeAmount)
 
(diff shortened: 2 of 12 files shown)


paullinator:
Screenshot 2024-03-07 at 22 15 40

paullinator:
Screenshot 2024-03-07 at 22 26 38

Jon-edge:
I don't agree with the reorganization of the tx details. The only other place recipient makes sense to me is in the sender card so that the card represents from/to. If we are going to organize these cards, we should be able to label what category of thing describes each card. With the proposed changes I still can't attach a label to each card.

I propose: 
Card 1: "from/to": sender/receiver

Card 2: "TX explorer stuff": txid, amount, fiat values, fee used, tx bytes, explorer
OR split into two cards:
Card 2: "Novice tx stuff": txId, amount, fiat values
Card 2.5: "Technical/advanced tx stuff":  fee used, tx bytes, explorer

Card 3: User specified information: Category, notes

src/components/scenes/TransactionDetailsScene.tsx
```diff
@@ -125,6 +125,7 @@ const TransactionDetailsComponent = (props: Props) => {
         returnKeyType="done"
         keyboardType="numeric"
         submitLabel={lstrings.string_save}
+        textSizeRem={1.5}
```
Jon-edge, line 128: What makes this field deserve a different text size? I don't think our input modals should even have a prop like this to begin with, we should use one consistent size or else this just looks messy.

src/components/cards/PoweredByCard.tsx
```diff
@@ -54,6 +54,7 @@ const getStyles = cacheStyles((theme: Theme) => ({
     flexDirection: 'column'
   },
   poweredByContainer: {
+    marginHorizontal: theme.rem(0.25),
```
Jon-edge, line 57: Though the margins were increased, they're still less than the rest of the app and it makes this stand out in a bad way. I said this before in an earlier PR, I don't think reducing margins is appropriate to "de-emphasize" a card. To me it doesn't convey any sort of de-emphasis, but looks messy and inconsistent when the padding is different. Madison also brought this up without bringing it up to her and expressed the same views.

I would really prefer just using the same UI4 components we already have for this component instead of a custom one-off implementation. If we really want something different, then we should design a different component instead. 

Honestly, I don't think there is even a need to de-emphasize this at all.

Jon-edge (changes requested):

src/components/scenes/TransactionDetailsScene.tsx
paullinator: That's what we use in the FiatAmount scene so it matches. The text field looks bizarringly small without this change

src/components/cards/PoweredByCard.tsx
paullinator: So let's take up the discussion on whether or not it "needs de-emphasis" as a separate task/sidebar. This commit is to address the inconsistent padding for that button so that the top/bottom match left/right.

src/components/scenes/TransactionDetailsScene.tsx
paullinator: image

Screenshot 2024-03-09 at 18 49 27

Jon-edge: To me, `FiatPluginEnterAmount` scene seems weird since it's inconsistent with our app for various things like spacing and field width/height. I don't think FPEA should be the basis that the rest of the app adheres to, but rather the other way around. If we have reasons for it to be different, I think it needs to be appear _obviously_ different. Right now I think it looks like unintentional inconsistency since it's so slightly different. Also, this modal isn't even seen from the `FiatPluginEnterAmount` scene so there's no way for the user to even make that association.

In a vacuum, though, if input modals in general are too small, I agree with the change on that basis if we are also comfortable changing ALL text modals, and to remove this prop. Consistency I think is less debatable between like components.

Jon-edge:
I'm gonna go ahead to approve because though I strongly disagree on some of these changes, they're not showstoppers and it seems I don't know the full context of the requirements. I definitely want to continue discussing this further in the future, though.

Jon-edge (approved):

PR #4898, opened 2024-03-09
Refactor exchange scenes

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206578167289576


src/components/themed/CryptoExchangeFlipInput.tsx
```diff
@@ -0,0 +1,193 @@
...
+  const cryptoAmount = useMemo(() => {
+    if (wallet == null || tokenId == null) return
+    const balance = wallet.balanceMap.get(tokenId) ?? '0'
+    const cryptoAmountRaw: string = convertNativeToDenomination(displayDenomination.multiplier)(balance)
+    return formatNumber(add(cryptoAmountRaw, '0'))
+  }, [displayDenomination.multiplier, tokenId, wallet])
+
+  const guiWalletName = useMemo(() => {
+    if (wallet == null) return ''
+    return getWalletName(wallet)
```
swansontec, line 72: This may be fast enough not to benefit from memoization. It's basically just a `?:`, and the `getWalletName` is also just a `?:` internally.

src/components/themed/CryptoExchangeFlipInput.tsx
```diff
@@ -0,0 +1,193 @@
...
+  }, [wallet])
+
+  //
+  // Handlers
+  //
+
+  const handleAmountsChanged = (amounts: ExchangedFlipInputAmounts) => {
+    props.onAmountChanged(amounts)
+  }
```
swansontec, line 81: You can pass `props.onAmountChanged` straight down as a prop, without wrapping it in a handler. The ESLint rule has an exception for this case.

swansontec (changes requested):
I found a few small things, but overall this is excellent!

I did the review using the "files changed" tab on GitHub. I think this would make more sense squashed down into one commit, aside from [Move checkEnabledExchanges to SettingsActions](https://github.com/EdgeApp/edge-react-gui/pull/4898/commits/0bf3237f42ffb8e871ca27695637d9cf918937ab) which could stay separate.

I was hoping we would rename "ExchangeScene" to "SwapScene" and similarly for the other component names. I guess that's out of scope for this PR, but at least it would be waaaay simpler to do that now.

src/components/scenes/CryptoExchangeQuoteProcessingScene.tsx
```diff
@@ -47,9 +47,15 @@ export function CryptoExchangeQuoteProcessingScene(props: Props) {
...
 
+  const fromCurrencyCode = getCurrencyCode(swapRequest.fromWallet, swapRequest.fromTokenId)
+  const toCurrencyCode = getCurrencyCode(swapRequest.toWallet, swapRequest.toTokenId)
+
   const dispatch = useDispatch()
   const account = useSelector(state => state.core.account)
 
+  const fromDenomination = useSelector(state => selectDisplayDenomByCurrencyCode(state, swapRequest.fromWallet.currencyConfig, fromCurrencyCode))
+  const toDenomination = useSelector(state => selectDisplayDenomByCurrencyCode(state, swapRequest.toWallet.currencyConfig, toCurrencyCode))
+
```
swansontec: If you use the new `selectDisplayDenom` or even `useDisplayDenom` hooks, those take the tokenId natively, which helps with our currency code elimination.

swansontec (approved):
Approved, but there is one more optional cleanup if you want to hit that before merging.

PR #4906, opened 2024-03-14
Fix noChangeFees handling

peachbits:


src/components/ui4/RowUi4.tsx
```diff
@@ -76,7 +76,7 @@ export const RowUi4 = (props: Props) => {
   })
 
   const rightButtonVisible = rightButtonType !== 'none'
-  const isTappable = onPress != null || onLongPress != null
+  const isTappable = rightButtonVisible && (onPress != null || onLongPress != null)
```
Jon-edge, line 79: This is wrong. A tappable button is not a requirement for a row to be tappable. You can have a `rightButtonType === 'none'` while being tappable.

Jon-edge (changes requested):

Jon-edge (approved):

PR #4907, opened 2024-03-14
Jon/fix/notif-slider-conflict

Jon-edge:
- Simplify all the weird spacing assignments. The scene simply needs insets for the main scroll view - the notification height is already provided through those insets and was actually previously doubling up on those insets. There's no observed need to add any extra bottom padding as nothing gets covered and the inset works as expected.
- Small dynamic adjustment added to the slider placement depending on the existence of notifications

image
image

Proof hard-coded extra 5 rem was not required and nothing gets covered:
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206832602251889

### DIFF
--- a/src/components/scenes/SendScene2.tsx
+++ b/src/components/scenes/SendScene2.tsx
@@ -176,6 +176,8 @@ const SendComponent = (props: Props) => {
   const pinSpendingLimitsEnabled = useSelector<boolean>(state => state.ui.settings.spendingLimits.transaction.isEnabled)
   const pinSpendingLimitsAmount = useSelector<number>(state => state.ui.settings.spendingLimits.transaction.amount ?? 0)
   const defaultIsoFiat = useSelector<string>(state => state.ui.settings.defaultIsoFiat)
+  const hasNotifications = useSelector(state => state.ui.notificationHeight > 0)
+
   const currencyWallets = useWatch(account, 'currencyWallets')
   const [tokenId, setTokenId] = useState<EdgeTokenId>(spendInfo.tokenId ?? tokenIdProp)
   const coreWallet = currencyWallets[walletId]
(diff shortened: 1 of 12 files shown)


src/components/scenes/SendScene2.tsx
```diff
@@ -1025,8 +1025,7 @@ const SendComponent = (props: Props) => {
       {({ insetStyle }) => (
         <>
           <StyledKeyboardAwareScrollView
-            notificationHeight={insetStyle.paddingBottom}
-            contentContainerStyle={{ ...insetStyle, paddingTop: 0, paddingBottom: theme.rem(5) + insetStyle.paddingBottom }}
```
Jon-edge, line 1029: Hard-coded extra 5rem not needed: 
image

src/components/scenes/SendScene2.tsx
```diff
@@ -1025,8 +1025,7 @@ const SendComponent = (props: Props) => {
...
+            contentContainerStyle={{ ...insetStyle, paddingTop: 0 }}
```
paullinator, line 1028: This ignores the padding needed to ensure the slider doesn't overlap the content. That's what the 5 rem was for in the first place. It wasn't for the notification card. Try adding 5 target addresses and see what happens

paullinator (changes requested):

src/components/scenes/SendScene2.tsx
Jon-edge: The screenshot I posted as a code comment was meant to illustrate this - the slider does not cover any content already and the extra 5 rem isn't needed

paullinator: I just tested the code and here's a screenshot with the lowest tile occluded verifying my comment

Screenshot 2024-03-15 at 17 55 29

Jon-edge: image
image

Jon-edge:
Some info on that dynamic slider margin: it looks too spaced out when notifications are visible if we keep the 2 rem bottom spacing in all cases. However there was a bug in the implementation:

The `notificationHeight` prop name was misleading in the `StyledSliderView` - the value wasn't actually the notification height, but the value of all combined bottom insets, which may include non-notification margins. This was addressed in the fixup to correctly pick out the notification visibility condition.

paullinator (approved):

PR #4909, opened 2024-03-16
Jon/detected-token-highlight

Jon-edge:
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206097714609113

### DIFF
--- a/src/components/scenes/ManageTokensScene.tsx
+++ b/src/components/scenes/ManageTokensScene.tsx
@@ -1,10 +1,11 @@
 import { EdgeCurrencyWallet } from 'edge-core-js'
 import * as React from 'react'
-import { View } from 'react-native'
+import { SectionList, View } from 'react-native'
 import { FlatList } from 'react-native-gesture-handler'
 
 import { PREFERRED_TOKENS, SPECIAL_CURRENCY_INFO } from '../../constants/WalletAndCurrencyConstants'
 import { useHandler } from '../../hooks/useHandler'
+import { useRowLayout } from '../../hooks/useRowLayout'
 import { useWalletName } from '../../hooks/useWalletName'
 import { useWatch } from '../../hooks/useWatch'
 import { lstrings } from '../../locales/strings'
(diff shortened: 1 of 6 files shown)


src/components/scenes/ManageTokensScene.tsx
```diff
@@ -21,19 +22,32 @@ import { FilledTextInput } from '../themed/FilledTextInput'
...
 import { Title } from '../themed/Title'
+import { WalletListSectionHeader } from '../themed/WalletListSectionHeader'
 import { ButtonsViewUi4 } from '../ui4/ButtonsViewUi4'
 import { CryptoIconUi4 } from '../ui4/CryptoIconUi4'
 
+/**
+ * walletId: ID of the wallet whose tokens we are managing
+ * newTokenIds: We routed from a notification that was triggered from detecting
+ * these new tokenIds having nonzero balance. Show these at the top of the list.
+ */
```
swansontec, line 33: I would annotate the members themselves, so you get the tooltip when you hover the exact item:

```ts
export interface ManageTokensParams {
  /** The wallet whose tokens we are managing */
  walletId: string

  /**
   * Show these at the top of the list, since we routed from
   * the new-token notification pop-up.
   */
  newTokenIds?: string[]
}
```

src/components/scenes/ManageTokensScene.tsx
```diff
@@ -97,6 +111,35 @@ function ManageTokensSceneComponent(props: Props) {
     })
   }, [allTokens, searchValue, sortedTokenIds])
 
+  // Split the list of tokens based on if there were auto-detected tokens given
+  const autoDetectedTokenIds = React.useMemo(
+    () => (newTokenIds ? filteredTokenIds.filter(filteredTokenId => newTokenIds.includes(filteredTokenId)) : []),
+    [filteredTokenIds, newTokenIds]
+  )
+
+  const extraData = React.useMemo(() => ({ allTokens, enabledTokenSet, customTokens }), [allTokens, enabledTokenSet, customTokens])
```
swansontec, line 120: If you switch `renderRow` to a `useCallback`, ESLint will help you keep this list up-to-date and correct. You won't need to pass the `extraData` param anymore, because the callback itself will change (`!==`) when its dependencies change.

swansontec (approved):
This looks good, but I have two optional cleanups.

PR #4912, opened 2024-03-18
New Swap Scene

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206755917708631


samholmes:
Changes are needed based on new designs from yesterday

swansontec (commented):
I cannot review this atomic commit style, since the forest is hidden in the trees.

I would like to see:

- Rename CryptoExchange* components
- Rename exchange scene route names
- Update text input colors (new commit, with changes to edgeDark.tsx)
- Literally everything else, in one commit

Notice that the first commit for `ifLoggedIn` should not be included, for reasons given below.

I am going to review it locally in my diff tool, and then leave comments manually.

src/components/scenes/SwapCreateScene.tsx
```diff
@@ -47,38 +47,31 @@ export interface SwapErrorDisplayInfo {
...
+  nativeAmount: string
+  nativeAmountFor: 'from' | 'to'
 }
 
 const defaultState: State = {
-  whichWalletFocus: 'from',
-  fromAmountNative: '',
-  toAmountNative: '',
-  paddingBottom: 0
+  nativeAmount: '',
```
swansontec, line 58: The default is "0" in the flip handler. Should we make it "0" here too?

src/components/scenes/SwapCreateScene.tsx
```diff
@@ -279,41 +277,32 @@ export const CryptoExchangeScene = (props: Props) => {
...
   })
 
   //
   // Render
   //
 
   const renderButton = () => {
-    const primaryNativeAmount = state.whichWalletFocus === 'from' ? state.fromAmountNative : state.toAmountNative
-    const showNext = fromCurrencyCode !== '' && toCurrencyCode !== '' && !!parseFloat(primaryNativeAmount)
+    const showNext = fromCurrencyCode !== '' && toCurrencyCode !== '' && !!parseFloat(state.nativeAmount)
```
swansontec, line 305: We should probably do `> 0` instead of `!!`. The `> 0` will return false for 0 and NaN, which are the two cases we care about.

src/components/themed/SwapInputCard.tsx
```diff
@@ -0,0 +1,295 @@
...
+export interface SwapInputCardAmounts {
+  exchangeAmount: string
+  nativeAmount: string
+  fiatAmount: string
+  fieldChanged: 'fiat' | 'crypto'
+}
+
+export interface Props {
+  heading?: string
+  wallet?: EdgeCurrencyWallet
```
swansontec, line 35: The `wallet` will always be passed, because you replace the whole card with a button when the wallet is `null`. This will have a whole bunch of clean-up effects, such as getting rid of the `useMaybeFiatCurrencyCode` hook, the `disabled` prop, and so forth.

swansontec (changes requested):
This is all I have for now - I would like to see these major issues resolved (squashing commits and making the wallet mandatory in the swap input component) before I review further, because right now things are just too chaotic.

Also, please do not rename FlipInput2 to FlipInputOld. If you genuinely need to for the component for the swap scene, then call your fork `SwapFlipInput`, but FlipInput2 isn't deprecated, and shouldn't have a name like "Old".

src/components/themed/FlipInput2.tsx
```diff
@@ -199,50 +253,75 @@ export const FlipInput2 = React.forwardRef<FlipInputRef, Props>((props: Props, r
...
-      alignItems: 'center',
+      flexDirection: 'column',
+      alignItems: 'stretch',
+      margin: theme.rem(0.5),
+
       borderWidth: theme.textInputBorderWidth,
       borderRadius: theme.rem(0.5),
-      flex: 1,
-      flexDirection: 'row',
+      // flex: 1,
```
swansontec, line 293: Leftover.

src/components/themed/SwapInput.tsx
```diff
@@ -1,190 +1,283 @@
...
+        convertValue={convertValue}
+        fieldInfos={fieldInfos}
+        renderHeader={renderHeader}
+        renderIcon={renderIcon}
+        returnKeyType={returnKeyType}
+        forceFieldNum={forceFieldMap[overrideForceField]}
+        inputAccessoryViewID={inputAccessoryViewID}
+        keyboardVisible={keyboardVisible}
+        startAmounts={[initialDisplayAmount, initialFiatAmount]}
+        placeholders={[lstrings.string_tap_to_edit, lstrings.string_tap_next_for_quote]}
```
swansontec, line 237: Since you sorted the props in `FlipInput2` itself, I was half-expecting them to be sorted here too.

swansontec (approved):
I didn't find much on this second round of reviews. I think it looks pretty reasonable now, and a lot of my confusions from before are gone - now that it's all one big commit, it's obvious which files are which, and what got renamed to what.

We will really have to test the FlipInput2 component, not only on the swap scene but in other places like the send scene where we also use it.

src/components/themed/SwapInput.tsx
samholmes: Ooh, good idea. I like sorting things 😉

src/components/themed/FlipInput2.tsx
samholmes: Nice catch!

src/components/themed/FlipInput2.tsx
```diff
@@ -261,6 +263,8 @@ export const FlipInput2 = React.forwardRef<FlipInputRef, Props>((props: Props, r
           </EdgeTouchableOpacity>
         </SideContainer>
       </InputContainerView>
+
+      {renderFooter ? renderFooter() : null}
```
swansontec, line 267: `renderFooter != null`

swansontec (approved):
One tiny tweak needed.

PR #4913, opened 2024-03-18
Jon/fix/centered-spinner

Jon-edge:
image

image

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206844514369001

### DIFF
--- a/src/components/ui4/ButtonUi4.tsx
+++ b/src/components/ui4/ButtonUi4.tsx
@@ -9,7 +9,7 @@ import LinearGradient from 'react-native-linear-gradient'
 import { cacheStyles } from 'react-native-patina'
 
 import { usePendingPress } from '../../hooks/usePendingPress'
-import { fixSides, mapSides, sidesToMargin } from '../../util/sides'
+import { fixSides, mapSides, sidesToPadding } from '../../util/sides'
 import { EdgeTouchableOpacity } from '../common/EdgeTouchableOpacity'
 import { Theme, useTheme } from '../services/ThemeContext'
 import { EdgeText } from '../themed/EdgeText'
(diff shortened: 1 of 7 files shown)


src/components/ui4/ButtonUi4.tsx
```diff
@@ -106,28 +106,32 @@ export function ButtonUi4(props: Props) {
...
+    if (marginRem == null) return undefined
+
+    const { marginLeft, marginRight, marginTop, marginBottom } = sidesToMargin(mapSides(fixSides(marginRem, 0), theme.rem))
+    return {
+      // Use margin as padding to increase tappable area
+      paddingLeft: marginLeft,
+      paddingRight: marginRight,
+      paddingTop: marginTop,
+      paddingBottom: marginBottom
+    }
```
swansontec, line 121: Why rename all the fields? Just use `sidesToPadding` to begin with. That's what the utility is for.

src/components/ui4/ButtonUi4.tsx
```diff
@@ -145,6 +149,8 @@ export function ButtonUi4(props: Props) {
     return retStyle
   }, [disabled, hideContent, layout, mini, styles, type])
 
+  const spinnerStyle = React.useMemo(() => [styles.spinner, customMarginPadding], [customMarginPadding, styles.spinner])
```
swansontec, line 152: This `useMemo` does nothing. You can read the source code for `ActivityIndicator` yourself:

https://github.com/facebook/react-native/blob/5761ad6437b29dc559fce0914f2a8e71d02d25b3/Libraries/Components/ActivityIndicator/ActivityIndicator.js#L111

As you can see, the styles go through a `StyleSheet.compose` call before they head down to native code. On top of that, the `ActivityIndicator` itself doesn't use `React.memo` or `React.PureComponent` in any way.

Since the `useMemo` call isn't free, fake "optimizations" like these actually slow things down. So, just do it the easy way and write `<Xyz style={[a, b, c]} />` when you need multiple styles in one. The simplest option is also the fastest.

swansontec (approved):
One thing to fix, but otherwise this is good.

src/components/ui4/ButtonUi4.tsx
Jon-edge: Ok I'll keep that in mind for the future. I'm pretty sure I've gotten comments of this nature from Paul before when trying to do simply `<Xyz style={[a, b, c]} />`.... Case specific? Either way I'll keep an eye out

PR #4921, opened 2024-03-21
Display new `'failed'` case for transaction confirmation status

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206663127267137

### DIFF
--- a/src/components/themed/TransactionListRow.tsx
+++ b/src/components/themed/TransactionListRow.tsx
@@ -150,21 +150,23 @@ export function TransactionListRow(props: Props) {
 
   // Pending Text and Style
   const currentConfirmations = transaction.confirmations
-  const isConfirmed = currentConfirmations === 'confirmed'
 
-  const unconfirmedOrTimeText = isConfirmed
-    ? unixToLocaleDateTime(transaction.date).time
-    : !isSentTransaction && canReplaceByFee && currentConfirmations === 'unconfirmed'
-    ? lstrings.fragment_transaction_list_unconfirmed_rbf
-    : currentConfirmations === 'unconfirmed'
-    ? lstrings.fragment_wallet_unconfirmed
-    : currentConfirmations === 'dropped'
-    ? lstrings.fragment_transaction_list_tx_dropped
-    : typeof currentConfirmations === 'number'
-    ? sprintf(lstrings.fragment_transaction_list_confirmation_progress, currentConfirmations, requiredConfirmations)
-    : lstrings.fragment_transaction_list_tx_synchronizing
+  const unconfirmedOrTimeText =
+    currentConfirmations === 'confirmed'
+      ? unixToLocaleDateTime(transaction.date).time
+      : !isSentTransaction && canReplaceByFee && currentConfirmations === 'unconfirmed'
+      ? lstrings.fragment_transaction_list_unconfirmed_rbf
+      : currentConfirmations === 'unconfirmed'
+      ? lstrings.fragment_wallet_unconfirmed
+      : currentConfirmations === 'dropped'
+      ? lstrings.fragment_transaction_list_tx_dropped
+      : currentConfirmations === 'failed'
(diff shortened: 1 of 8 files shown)


src/components/themed/TransactionListRow.tsx
```diff
@@ -303,9 +305,10 @@ const getStyles = cacheStyles((theme: Theme) => ({
     color: theme.secondaryText
   },
   unconfirmedText: {
-    flexShrink: 1,
-    fontSize: theme.rem(0.75),
     color: theme.warningText,
     marginRight: theme.rem(1)
+  },
+  failedText: {
+    color: theme.dangerText
```
swansontec, line 312: If the `marginRight: 1` isn't needed here, shouldn't we remove it from  `unconfirmedText` too?

src/components/themed/TransactionListRow.tsx
```diff
@@ -150,21 +150,23 @@ export function TransactionListRow(props: Props) {
...
+      : currentConfirmations === 'dropped'
+      ? lstrings.fragment_transaction_list_tx_dropped
+      : currentConfirmations === 'failed'
+      ? lstrings.fragment_transaction_list_tx_failed
+      : typeof currentConfirmations === 'number'
+      ? sprintf(lstrings.fragment_transaction_list_confirmation_progress, currentConfirmations, requiredConfirmations)
+      : lstrings.fragment_transaction_list_tx_synchronizing
 
-  const unconfirmedOrTimeStyle = isConfirmed ? styles.secondaryText : styles.unconfirmedText
+  const confirmationStyle = currentConfirmations === 'failed' ? styles.failedText : styles.unconfirmedText
```
swansontec, line 169: With this change, wouldn't confirmed transactions appear in yellow `warningText` style now? 

I think we would want `isConfirmed ? null : currentConfirmations === 'failed' ? styles.failedText : styles.unconfirmedText`, so we don't apply any color tweaks to the fully-confirmed case.

src/components/themed/TransactionListRow.tsx
```diff
@@ -150,21 +150,23 @@ export function TransactionListRow(props: Props) {
...
+    currentConfirmations === 'confirmed'
+      ? unixToLocaleDateTime(transaction.date).time
+      : !isSentTransaction && canReplaceByFee && currentConfirmations === 'unconfirmed'
+      ? lstrings.fragment_transaction_list_unconfirmed_rbf
+      : currentConfirmations === 'unconfirmed'
+      ? lstrings.fragment_wallet_unconfirmed
+      : currentConfirmations === 'dropped'
+      ? lstrings.fragment_transaction_list_tx_dropped
+      : currentConfirmations === 'failed'
+      ? lstrings.fragment_transaction_list_tx_failed
```
swansontec, line 164: I realize the code was like this already, but maybe we should take this huge ? : block and turn it into a utility function, rather than keep shoving one more thing in to an already over-stuffed situation. That way we can use a simpler `if` / `return` style:

```ts
function getConfirmationText(tx: EdgeTransaction) {
  if (tx.confirmations === 'confirmed') {
    return unixToLocaleDateTime(tx.date).time
  }

  if (tx.confirmations === 'unconfirmed') {
    if (!isSentTransaction && canReplaceByFee) {
      return lstrings.fragment_transaction_list_unconfirmed_rbf
    }
    return lstrings.fragment_wallet_unconfirmed
  }

  if (tx.confirmations === 'failed') {
  ...
}
```

swansontec (changes requested):

src/components/themed/TransactionListRow.tsx
samholmes: I thought the marginRight is needed for unconfirmed text because the text can be longer with a confirmation count. Maybe it makes sense to move this marginRight to the base style for this component.

swansontec (approved):

PR #4925, opened 2024-03-22
Paul/fiat max

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206779193613535

### DIFF
--- a/src/plugins/gui/amountQuotePlugin.ts
+++ b/src/plugins/gui/amountQuotePlugin.ts
@@ -49,6 +49,13 @@ const providerFactories = [banxaProvider, bityProvider, kadoProvider, moonpayPro
 
 const DEFAULT_FIAT_AMOUNT = '500'
 
+/**
+ * Amount that we will attempt to get a quote for if the user presses the MAX button
+ * This is an arbitrary large number that should be larger than any reasonable quote
+ * in either fiat or crypto
+ */
+const MAX_QUOTE_VALUE = '10000000000'
+
 export const amountQuoteFiatPlugin: FiatPluginFactory = async (params: FiatPluginFactoryArgs) => {
   const { account, guiPlugin, longPress = false, showUi } = params
   const { pluginId } = guiPlugin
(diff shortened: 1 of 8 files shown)


src/plugins/gui/amountQuotePlugin.ts
```diff
@@ -34,10 +36,24 @@ type PaymentTypeProviderPriorityMap = ReturnType<typeof asPaymentTypeProviderPri
...
+  convertValueInternal: (sourceFieldNum: number, value: string, stateManager: StateManager<EnterAmountState>) => Promise<ConvertValueInternalResult>
+}
+
 const providerFactories = [banxaProvider, bityProvider, kadoProvider, moonpayProvider, mtpelerinProvider, paybisProvider, simplexProvider]
 
 const DEFAULT_FIAT_AMOUNT = '500'
 
+// Amount that we will attempt to get a quote for if the user presses the MAX button
+// This is an arbitrary large number that should be larger than any reasonable quote
+// in either fiat or crypto
```
Jon-edge, line 54: /** Amount that we will attempt to get a quote for if the user presses the MAX button
This is an arbitrary large number that should be larger than any reasonable quote
in either fiat or crypto */
```suggestion
// Amount that we will attempt to get a quote for if the user presses the MAX button
// This is an arbitrary large number that should be larger than any reasonable quote
// in either fiat or crypto
```

Using this comment format instead will show the comment on hover

src/plugins/gui/amountQuotePlugin.ts
```diff
@@ -218,21 +234,106 @@ export const amountQuoteFiatPlugin: FiatPluginFactory = async (params: FiatPlugi
         label2: sprintf(lstrings.fiat_plugin_amount_currencycode, currencyCode),
         swapInputLocations: requireCrypto,
         async onChangeText() {},
+        async onMax(sourceFieldNum, stateManager) {
+          // Helper function to get a quote or error for the value specified
+          const getMaxQuoteOrError = async (
```
Jon-edge, line 239:
```suggestion
          /** Helper function to get a quote or error for the value specified */
```

Using this comment format instead will show the comment on hover

src/plugins/gui/amountQuotePlugin.ts
```diff
@@ -424,3 +527,33 @@ export const getBestQuote = (quotes: FiatProviderQuote[], priorityArray: Priorit
     if (bestQuote != null) return bestQuote
   }
 }
+
+const evm = '0x0d73358506663d484945BA85D0cd435ad610B0A0'
```
Jon-edge, line 531: Is this just some arbitrary address? Could use some documentation.

Also, move to top with other consts in caps

src/plugins/gui/scenes/FiatPluginEnterAmountScene.tsx
```diff
@@ -275,6 +306,10 @@ const getStyles = cacheStyles((theme: Theme) => {
       paddingTop: theme.rem(0.5),
       width: '100%'
     },
+    maxButton: {
+      alignItems: 'flex-end',
+      marginTop: theme.rem(-0.75)
```
Jon-edge, line 311: Defeats the spirit and purpose of `ButtonsViewUi4` but I get this is a temporary one-off thing until we do a proper design. Could have just used a loose button on its own with styling:
```
      flexDirection: 'row',
      justifyContent: 'flex-end',
      marginTop: theme.rem(-0.75)
```

Jon-edge (changes requested):
Small changes needed

src/plugins/gui/scenes/FiatPluginEnterAmountScene.tsx
paullinator: I'm only modifying the top margin. Everything else stays the same as `ButtonsUiView`

Jon-edge: The margins of `ButtonsViewUi4` is a large part of what the component is intended to be responsible for defining and managing, and what I spent a lot of time making happen app-wide. Adding new callers who arbitrarily modify and remove them is moving a step backwards in updating our app to UI4 and the advantages of reuse and consistency. 

Would prefer to have a button in a plain view instead as I described so we can more easily keep track of where we need to define new `ButtonsViewUi4` layouts in case the final scene design takes a while to happen.

We need a complete design for the scene and a corresponding specific, codified layout such as `'right-justified'` for this anyway so approving for now so we can merge asap to revisit later.

Jon-edge (approved):

PR #4926, opened 2024-03-22
Tx List Receive/Send/Earn/Trade Buttons

Jon-edge:
image
image
image
image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206700142144955


src/components/themed/TransactionListTop.tsx
```diff
@@ -181,6 +206,180 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
...
+    // Check balances for the displayed asset on this scene:
+    const sceneAssetCryptoBalance = wallet.balanceMap.get(tokenId)
+
+    // Filter owned assets (mainnet and token) by those defined by the
+    // SWAP_ASSET_PRIORITY:
+    const ownedAssets: Array<{ pluginId: string; tokenId: EdgeTokenId; walletId: string; dollarValue: number }> = []
+    SWAP_ASSET_PRIORITY.forEach(priorityAsset => {
+      // Filter the currencyWallets for nonzero crypto balances that are
+      // NOT the asset shown on the scene
+      Object.values(currencyWallets).forEach(accountWallet => {
```
paullinator, line 293: rename to `currencyWallet` That's the singular of `currencyWallets` which is convention.

src/components/themed/TransactionListTop.tsx
```diff
@@ -181,6 +206,180 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
...
+  handleTradeSell = (bridge: AirshipBridge<void>) => {
+    const { navigation, wallet, tokenId } = this.props
+    const { pluginId } = wallet.currencyInfo
+
+    navigation.navigate('sellTab', { screen: 'pluginListSell', params: { filterAsset: { pluginId, tokenId } } })
+    bridge.resolve()
+  }
+
+  handleTradeSwap = (bridge: AirshipBridge<void>) => {
+    const { account, navigation, wallet, tokenId } = this.props
```
paullinator, line 277: I won't require it in this PR, but this should have been a test driven developed routine. Might have actually been faster to write and test than doing it that way.

src/components/themed/TransactionListTop.tsx
```diff
@@ -181,6 +206,180 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
...
+      ownedAssets.sort((a, b) => b.dollarValue - a.dollarValue)
+
+      // Use the highest in the priority asset list as the other side of
+      // the swap. If we don't have exchange rates for one or more assets,
+      // we would still retain the ordering set forth by
+      // SWAP_ASSET_PRIORITY to use for this pick.
+      highestUsdBalanceAccountWalletId = ownedAssets[0].walletId
+      highestUsdBalancePriorityTokenId = ownedAssets[0].tokenId
+    } else {
+      // If no balances for the priority assets, we leave the opposing
```
paullinator, line 333: This comment doesn't match the spec. In the condition where the "other" wallet has zero balance but it is in the priority list, pick the asset highest on the priority list. Only in the case the user doesn't have ANY of the priority assets would they be required to pick one. A unit test would have made this easy to spot in development and in code review. I "think" code actually will pick an asset with zero balance, but the sort routine might not give you the first on the list since you don't know the sort algo when all entries are equal.

paullinator (changes requested):

src/components/ui4/IconButton.tsx
```diff
@@ -0,0 +1,114 @@
...
+  },
+  iconContainer: {
+    alignItems: 'center',
+    justifyContent: 'center',
+    backgroundColor: theme.cardBaseColor
+  },
+  superScriptContainer: {
+    position: 'absolute',
+    top: 0,
+    left: theme.rem(1.5),
```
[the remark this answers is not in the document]
Jon-edge: That's how I originally had it but it looked weird. Best solution would probably be to have a center origin

src/components/themed/TransactionListTop.tsx
Jon-edge: This made more sense in a prior implementation where I was juggling multiple wallets.

src/components/themed/TransactionListTop.tsx
Jon-edge: As I mentioned in the comment above, the sort retains the ordering set forth by the original SWAP_ASSET_PRIORITY array. If all things are equal, the sort algo retains the original ordering.

```
const ownedAssets: Array<{ name: string; dollarValue: number }> = [
  { name: 'Asset A', dollarValue: 0 },
  { name: 'Asset B', dollarValue: 0 },
  { name: 'Asset C', dollarValue: 0 }
]
ownedAssets.sort((a, b) => b.dollarValue - a.dollarValue)
// [{"dollarValue": 0, "name": "Asset A"}, {"dollarValue": 0, "name": "Asset B"}, {"dollarValue": 0, "name": "Asset C"}]
```

Jon-edge: I revised the description in the task because QA also needs to be on the same page and I think there's still ambiguity in the task. 

Hopefully it's clear now to everyone.

src/components/ui4/IconButton.tsx
```diff
@@ -0,0 +1,114 @@
...
+    height: theme.rem(1.25),
+    borderRadius: theme.rem(1.25 / 2),
+    alignItems: 'center',
+    justifyContent: 'center',
+    paddingVertical: 0,
+    paddingLeft: theme.rem(0.4),
```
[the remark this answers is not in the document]
Jon-edge: Things this small look a bit uneven without some additional fudging

src/components/scenes/GuiPluginListScene.tsx
```diff
@@ -203,7 +209,18 @@ class GuiPluginList extends React.PureComponent<Props, State> {
...
+      coreDisklet,
+      countryCode,
+      filterAsset,
+      stateProvinceCode,
+      deviceId,
+      disablePlugins,
+      navigation,
+      account,
+      onLogEvent,
+      onPluginOpened: onPluginClosed
```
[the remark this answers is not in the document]
Jon-edge: Incomplete "Rename Symbol" operation

src/components/themed/TransactionListTop.tsx
Jon-edge: Updated conditions for `isSceneAssetMatch` and `isPriorityAssetMatch` so that
balanances are not considered when populating `ownedAssets`. Only owning the
asset and having the token enabled is required, and the `ownedAssets.sort()`
handles the requirement as described in the prior comment

src/components/ui4/IconButton.tsx
```diff
@@ -0,0 +1,114 @@
...
+    justifyContent: 'center'
+  },
+  iconContainer: {
+    alignItems: 'center',
+    justifyContent: 'center',
+    backgroundColor: theme.cardBaseColor
+  },
+  superScriptContainer: {
+    position: 'absolute',
+    top: 0,
```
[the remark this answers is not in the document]
Jon-edge: The parent container superScriptContainer lives has space on top.

src/components/ui4/IconButton.tsx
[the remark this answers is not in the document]
Jon-edge: image
image
image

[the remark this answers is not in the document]
paullinator: I hate to say it but >100% APRs are common and users will want to know 100% vs 800% vs 2000%

As well, the smaller the APY the more precision we should give. Ie 9.5% which still takes up quite a bit of space.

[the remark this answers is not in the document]
paullinator: What logic is displayed above in the screenshot?

src/components/ui4/IconButton.tsx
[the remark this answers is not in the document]
paullinator: Upon closer look, you have screenshots with the superscript well above the icon and there’s enough clearance to look good. I retract the comment.

src/components/themed/TransactionListTop.tsx
```diff
@@ -181,6 +206,180 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
...
+      <ModalUi4
+        bridge={bridge}
+        title={sprintf(lstrings.trade_s, sceneCurrencyCode)}
+        onCancel={() => {
+          bridge.resolve()
+        }}
+      >
+        <SelectableRow
+          marginRem={0.5}
+          title={sprintf(lstrings.trade_s, sceneCurrencyCode)}
```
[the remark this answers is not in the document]
Jon-edge: image

src/components/themed/TransactionListTop.tsx
paullinator: Where are the fix up commits? Please don’t say you merged them in.

src/components/ui4/IconButton.tsx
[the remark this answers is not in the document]
paullinator: Why did you add another `View`. This should have been doable with the `LinearGradient` and avoid extra Views

Jon-edge (commented):
> Where are the fix up commits? Please don’t say you merged them in.  

My last comment describes the change I made in the fixup to satisfy the requirement...

src/components/themed/TransactionListTop.tsx
```diff
@@ -514,12 +520,18 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
...
+  /** Return the best APY found, defaulting to 1 decimal place, rounding to the
+   * nearest whole number if >= 10, and truncating to '>99%' if greater than 99%
+   * */
   getBestApy = (): string | undefined => {
     const { stakePolicies } = this.state
     if (stakePolicies == null || stakePolicies.length === 0) return
     const bestApy = stakePolicies.reduce((prev, curr) => Math.max(prev, curr.apy ?? 0), 0)
     if (bestApy === 0) return
-    return round(bestApy.toString(), -1) + '%'
+    if (bestApy > 99) return '>99%'
```
paullinator, line 531: No no. Need to support showing large APYs as I mentioned in my comment

src/components/themed/TransactionListTop.tsx
```diff
@@ -324,9 +325,10 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
       ownedAssets.sort((a, b) => b.dollarValue - a.dollarValue)
 
       // Use the highest in the priority asset list as the other side of
-      // the swap. If we don't have exchange rates for one or more assets,
-      // we would still retain the ordering set forth by
-      // SWAP_ASSET_PRIORITY to use for this pick.
+      // the swap. If we don't have exchange rates for one or more assets, or
```
paullinator, line 328: This comment contradicts the comment on ln 335.

src/components/ui4/IconButton.tsx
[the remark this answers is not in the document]
paullinator: Huh? I think your eyes are tricking you due to gradient or % sign on one side. Please share screenshots of even padding. Ideally 0.5 or 0.25rem to clamp to standard size stepping

paullinator (changes requested):

src/components/themed/TransactionListTop.tsx
Jon-edge: My last comment describes the change I made in the fixup to satisfy the requirement...

I just split up the last fixup into two to make it more obvious what the changes were.

paullinator:
Just noticed this is missing the redesign for the settings button and the wallet picker.

src/components/ui4/IconButton.tsx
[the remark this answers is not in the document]
Jon-edge: No it's not possible to implement a "centered origin" behavior for the `LinearGradient` without another `View` due to the `absolute` requirement. The only options without another view are: 
- "Right-origin"-ed (right side of the gradient is anchored, grows to the left with more content)
- "Left-origin"-ed (left side of the gradient is anchored, grows to the right with more content)

`superScriptCenterOrigin` creates an absolutely positioned invisible view with its center where we want the visible container to be centered around

src/components/themed/TransactionListTop.tsx
```diff
@@ -421,31 +420,32 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
       <>
         <View style={styles.balanceBoxWalletNameCurrencyContainer}>
           <EdgeTouchableOpacity accessible={false} style={styles.balanceBoxWalletNameContainer} onPress={this.handleOpenWalletListModal}>
-            <CryptoIconUi4 sizeRem={1.5} tokenId={tokenId} walletId={wallet.id} />
             <EdgeText accessible style={styles.balanceBoxWalletName}>
               {walletName}
             </EdgeText>
-            <Ionicons name="chevron-forward" size={theme.rem(1.5)} color={theme.iconTappable} />
+            <View style={styles.balanceBoxChevronAdjust}>
```
paullinator, line 426: remove the View and put the style on the `Ionicons`

src/components/themed/TransactionListTop.tsx
```diff
@@ -421,31 +420,32 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
...
+          <View style={styles.balanceBoxCryptoBalanceContainer}>
+            <EdgeText accessible style={styles.balanceBoxCurrency} minimumFontScale={0.3}>
+              {(isAccountBalanceVisible ? cryptoAmountFormat : lstrings.redacted_placeholder) + ' ' + displayDenomination.name}
             </EdgeText>
-          )}
+            <Ionicons
+              name={isAccountBalanceVisible ? 'eye-off-outline' : 'eye-outline'}
+              style={styles.eyeIcon}
+              color={theme.iconTappable}
+              size={theme.rem(2)}
```
paullinator, line 443: 2 rem looks huge. I think the design probably meant 1.75. Although I tried 1.5 and that looks much better

Screenshot 2024-03-26 at 17 40 04

paullinator: Also, the amount will have to wrap to 2 lines in absolutely worst case such as this one

Screenshot 2024-03-26 at 17 42 53

paullinator: Or you need to ensure the icon doesn't exceed the bounding area and try to get the text to shrink further

paullinator (changes requested):

src/components/themed/TransactionListTop.tsx
Jon-edge: Thought it didn't take styles for some reason... Removed View

src/components/themed/TransactionListTop.tsx
Jon-edge: image

Jon-edge: The best solution would be to update `CryptoText`/`CryptoAmount` to handle these edge cases more gracefully. I think the best solution really is truncation, like we do in market cap/volume in the Markets scene

Jon-edge: Updated screens in the main description to show edge cases with the scaled eye icon and center align

paullinator (approved):
Whoohoo! ready to merge

PR #4927, opened 2024-03-22
Update RequestScene to expect notifications

Jon-edge:
image

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206897264651547

### DIFF
--- a/src/components/common/SceneWrapper.tsx
+++ b/src/components/common/SceneWrapper.tsx
@@ -161,6 +161,8 @@ function SceneWrapperComponent(props: SceneWrapperProps): JSX.Element {
 
   // Calculate app insets considering the app's header, tab-bar,
   // notification area, etc:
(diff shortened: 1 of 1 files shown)


swansontec (approved):

swansontec (commented):
I am approving this because it seems to work, but I think it's the wrong solution. If the `hasNotifications` flag is set to false, the password reminder should not show in the first place - that's the real bug.

Jon-edge:
Yes you're right, I think we only wanted notifications for specific scenes, and this was clearly not one of them

samholmes (approved):
This makes sense. If the scene doesn't have notifications, then the height shouldn't be present in the bottom inset. For the request scene, this should remove the space at the bottom.

PR #4942, opened 2024-03-29
Paul/bity max

paullinator:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206620215693963

### DIFF
--- a/src/plugins/gui/providers/bityProvider.ts
+++ b/src/plugins/gui/providers/bityProvider.ts
@@ -1,5 +1,5 @@
 import { lt, toFixed } from 'biggystring'
-import { asArray, asEither, asMaybe, asNumber, asObject, asString, asValue } from 'cleaners'
+import { asArray, asEither, asMaybe, asNumber, asObject, asOptional, asString, asValue } from 'cleaners'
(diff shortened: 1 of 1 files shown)


src/plugins/gui/providers/bityProvider.ts
```diff
@@ -430,10 +430,59 @@ export const bityProvider: FiatProviderFactory = {
           })
         }
 
-        // Check for a max amount limit
+        // Because Bity only support <1k transactions w/o KYC and we have no
+        // way to KYC a user, add a 1k limit
+        if (amountType === 'fiat') {
+          if (gt(exchangeAmount, '1000')) {
```
Jon-edge, line 436: Does this need to be `gte`? Fix the comment if we only need `gt`

Jon-edge (approved):

src/plugins/gui/providers/bityProvider.ts
paullinator: comment fixed

PR #4953, opened 2024-04-04
Backup Modal Variant Testing

Jon-edge:
image
image
image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies
https://github.com/EdgeApp/edge-react-gui/pull/4960

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206548591329441


src/components/modals/ButtonsModal.tsx
```diff
@@ -23,14 +23,19 @@ export interface ButtonInfo {
 
 export interface ButtonModalProps<Buttons> {
   bridge: AirshipBridge<keyof Buttons | undefined>
-  title?: string
-  message?: string
-  children?: React.ReactNode
   buttons: Buttons
+  /** Ideally mutually exclusive with message/subtext since children are
+   * rendered in the body of the modal along with message/subtext */
```
swansontec, line 28: /** Used to pass non-text children, to be rendered after the title and message but before the buttons themselves. */

src/components/modals/ButtonsModal.tsx
swansontec, line 1: A lot of these changes seem out-of-scope for the rename. I think this `ButtonsModal` cleanup needs to be in its own commit.

src/actions/ScamWarningActions.tsx
```diff
@@ -23,7 +23,7 @@ export const triggerScamWarningModal = async (disklet: Disklet) => {
 
       return (
         <ConfirmContinueModal bridge={bridge} title={lstrings.warning_scam_title}>
-          <Paragraph isWarning>{warningMessage}</Paragraph>
+          <Paragraph fontType="warning">{warningMessage}</Paragraph>
```
swansontec, line 26: The `Paragraph` component doesn't take a `fontType` prop yet. Did this get into the wrong commit?

src/components/themed/EdgeText.tsx
```diff
@@ -4,6 +4,8 @@ import { Platform, StyleProp, Text, TextProps, TextStyle } from 'react-native'
 import { fixSides, mapSides, sidesToPadding } from '../../util/sides'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 
+export type ParagraphFontType = 'default' | 'warning' | 'finePrint'
```
swansontec, line 7: Can't we do this the same way the Web platform handles it? Define components like `<BoldText>`, `<WarningText>`, `<FinePrint>`, which can be used inline. Then it's:

```html
<Paragraph>
  <WarningText>
    Do not lose your keys, or you will <BoldText>not</BoldText> be able to get them back!
  </WarningText>
</Paragraph>
```

Your `ParagraphFontType` way might be OK too, especially if we really lock down what combinations are valid or not, but it could also end up growing out of control. It's hard to predict how these things play out.

src/components/themed/EdgeText.tsx
```diff
@@ -45,7 +45,7 @@ export const Paragraph = (props: {
    * Specific font style definitions for size and color:
    * - default: Default EdgeText style
    * - warning: Orange font color
-   * - finePrint: Smaller, dimmed font
+   * - finePrint: Smaller, orange font
```
swansontec, line 48: If it's orange, should it be called "warning" instead? Or even `<FinePrint><WarningText>{message}</WarningText></FinePrint>`?

Sometime the fine print is a warning, and sometimes it isn't, so I personally like the mix & match approach.

swansontec (changes requested):
Thanks for finally tackling out text situation. I strongly support the new `Paragraph` component, but there are one or two things we should fix up.

src/components/themed/EdgeText.tsx
Jon-edge: Yeah I was struggling trying to define the rules and wasn't aware the suggested was possible. Works way better, thanks

src/components/modals/ButtonsModal.tsx
Jon-edge: Fair point, it just kept ballooning out as I was working on this.

Jon-edge:
I created a new PR to separate the changes not specific to this backup modal task: https://github.com/EdgeApp/edge-react-gui/pull/4960

Made this backup modal PR dependent on the above.

swansontec (approved):

PR #4957, opened 2024-04-05
Add "Parent Wallet" Section to TxList Wallet Picker

Jon-edge:
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206779193613531

### DIFF
--- a/src/components/modals/WalletListModal.tsx
+++ b/src/components/modals/WalletListModal.tsx
@@ -64,6 +64,7 @@ interface Props {
   headerTitle: string
   showBankOptions?: boolean
   showCreateWallet?: boolean
+  parentWalletId?: string
 }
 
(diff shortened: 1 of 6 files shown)


src/components/modals/WalletListModal.tsx
```diff
@@ -64,6 +64,7 @@ interface Props {
   headerTitle: string
   showBankOptions?: boolean
   showCreateWallet?: boolean
+  parentWallet?: EdgeCurrencyWallet
```
paullinator, line 67: You don't actually use the full wallet. Just pass the `wallet.id` to ensure you don't access a wallet object that may have been deleted.

paullinator (approved):
Approved with one small simplification

PR #4962, opened 2024-04-08
Upgrade jest

swansontec:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1207030862093273


package.json
```diff
@@ -76,35 +76,6 @@
...
-    "randombytes": "2.1.0",
-    "regenerator-runtime": "0.13.2",
-    "ripemd160": "2.0.2",
-    "ripple-binary-codec": "0.2.3",
-    "safe-buffer": "5.1.2",
-    "scrypt.js": "0.3.0",
-    "sha.js": "2.4.11",
-    "tweetnacl": "^1.0.3",
-    "uuid": "3.3.2"
-  },
```
samholmes, line 107: Why "Remove package.json resolutions"?

samholmes (approved):

package.json
swansontec: Back when accountbased, exchange-plugins, and currency-plugins would be bundled together inside the GUI build process, we would use these resolutions to de-duplicate things. However, we were somewhat cheating by substituting semver-major changes. So if accountbased wants bip39 v3, and currency-plugins wants bip39 v4 we would just force both to use buffer v4, possibly breaking one or the other. On the other hand, with bip39 specifically being several 100K in size, this was a nice size win.

Now that accounbased and exchange-plugins provide their own pre-bundled build output, these resolutions don't do anything for our output size, but they do mean we are stuck with old & miss-matching package versions, which is no good.

PR #4967, opened 2024-04-09
Add Done button to search bar

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207014593281421

Simulator Screenshot - iPhone 14 - 2024-04-10 at 19 45 13

### DIFF
--- a/src/components/themed/SimpleTextInput.tsx
+++ b/src/components/themed/SimpleTextInput.tsx
@@ -17,7 +17,7 @@ import { useHandler } from '../../hooks/useHandler'
 import { SpaceProps, useSpaceStyle } from '../../hooks/useSpaceStyle'
 import { EdgeTouchableWithoutFeedback } from '../common/EdgeTouchableWithoutFeedback'
 import { styled, styledWithRef } from '../hoc/styled'
-import { AnimatedIconComponent, CloseIconAnimated } from '../icons/ThemedIcons'
+import { AnimatedIconComponent, ChevronBackAnimated, CloseIconAnimated } from '../icons/ThemedIcons'
 import { useTheme } from '../services/ThemeContext'
 
 const AnimatedTextInput = Animated.createAnimatedComponent(TextInput)
@@ -120,7 +120,7 @@ export const SimpleTextInput = React.forwardRef<SimpleTextInputRef, SimpleTextIn
   function clear(): void {
     if (inputRef.current != null) inputRef.current.clear()
     if (onChangeText != null) onChangeText('')
-    if (blurOnClear || !hasValue) blur()
+    if (blurOnClear) blur()
     if (onClear != null) onClear()
   }
   function focus(): void {
@@ -157,6 +157,10 @@ export const SimpleTextInput = React.forwardRef<SimpleTextInputRef, SimpleTextIn
   const handleClearPress = useHandler(() => {
     clear()
   })
+  const handleDonePress = useHandler(() => {
+    clear()
+    blur()
+  })
   const handleFocus = useHandler(() => {
     focusAnimation.value = withTiming(1, { duration: baseDuration })
     if (onFocus != null) onFocus()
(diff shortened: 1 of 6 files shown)


src/components/themed/SimpleTextInput.tsx
```diff
@@ -179,52 +184,61 @@ export const SimpleTextInput = React.forwardRef<SimpleTextInputRef, SimpleTextIn
...
+
+          <TouchContainer accessible onPress={handleClearPress} testID={`${testID}.clearIcon`}>
+            <SideContainer size={rightIconSize}>
+              <CloseIconAnimated color={iconColor} size={rightIconSize} />
+            </SideContainer>
+          </TouchContainer>
+        </InputContainerView>
+      </EdgeTouchableWithoutFeedback>
+      {isFocused ? (
+        <ButtonUi4 onPress={handleDonePress} type="tertiary" label={lstrings.string_done_cap} testID={`${testID}.doneButton`} marginRem={[0.5, 0, 0.5, 0.5]} />
```
Jon-edge, line 231: `marginRem={0.5}`

Out of the scope of this task but just noting for later - should make the defaults 0.5 as a default, and always override when using within a `ButtonsView`

src/components/themed/SimpleTextInput.tsx
```diff
@@ -179,52 +184,61 @@ export const SimpleTextInput = React.forwardRef<SimpleTextInputRef, SimpleTextIn
...
+      ) : null}
+    </ContainerView>
   )
 })
 
-const Container = styled(Animated.View)<{
+const ContainerView = styled(View)({
```
Jon-edge, line 237: `Container` is kind of redundant, the `View` already carries that meaning. Would be more descriptive to use `RowView,` then we can reuse it as an export later

Jon-edge (changes requested):

src/components/themed/SimpleTextInput.tsx
samholmes: These styled components are private and we should never export them. Container is a better name then row because this isn't a "row", it's the container for the all of the parts of the SimpleTextInput component. A row implies that it is a sibling of other rows.

src/components/themed/SimpleTextInput.tsx
samholmes: The margin is set to 0.5 on each side except for the right side

src/components/themed/SimpleTextInput.tsx
Jon-edge: The word "row" on its own has no implications towards parent/child relationships. It simply describes a row layout as opposed to a column layout.

One of my gripes with this pattern is that important information is abstracted away from the JSX. In the traditional `styles.X` pattern, often times simple styles such as `flexDirection: "row"` are named as such, so that it is easy to grok at the layout of the component without having to jump to a definition.

samholmes: Why name the component based on it's implementation details? We never required naming styles based on their properties before. Just because the container implements a row layout, doesn't mean it _is_ a row. Semantics for this is that it is a container. If it had a blue background, would we have to name it blue? If we change the theme colors, from blue to pink, do we need to change the component name? Same goes for the flex direction; if we change from row to column later perhaps if we need to add more elements and change the implementation, then now we might accidentally not rename the component and then it'd be confusing. 

I say we shouldn't ever name components based on the implementation details. I've been naming things by some higher-order semantic. In this case, a container is a component which is responsible for containing other components and that's it's purpose. It's similar in spirit to the name "wrapper", but I just preferred "container" as the word to use.

src/components/themed/SimpleTextInput.tsx
Jon-edge: Preferably we don't tile things in a way where one component handles 100% of the spacing on any one side. By splitting up the margins, we're able to minimize the amount of hard-coded margins, and allow things to be reused easier by others without knowledge of the actual implementation of the styling since they all follow the same expectations.

Jon-edge (changes requested):
Refer to the asana task for the requested changes by Madison

Jon-edge (approved):
Approved, pending final adjustments w/ Madison

PR #4968, opened 2024-04-09
Fix glitchy load behavior when searching on TransactionListScene

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206902425530802

https://github.com/EdgeApp/edge-react-gui/assets/517469/e7d12a20-727f-46a5-ab7e-047bf5b997ae

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - changed: Replace `ModalMessage` with `Paragraph`
 - changed: Improved wallet sync indicator bar on top of screen
 - changed: Allow light accounts to buy up to $50 worth of crypto at a time
+- fixed: Glitch on transaction list scene when starting search
 - removed: USPs a/b/c/d test experiment - always uses default USPs
 
 ## 4.4.0 (2024-04-09)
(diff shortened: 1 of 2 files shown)


src/components/scenes/TransactionListScene.tsx
```diff
@@ -71,7 +71,7 @@ function TransactionListComponent(props: Props) {
 
   // Watchers:
   const enabledTokenIds = useWatch(wallet, 'enabledTokenIds')
-  const transactionList = useTransactionList(wallet, tokenId, isSearching ? searchText : undefined)
+  const transactionList = useTransactionList(wallet, tokenId, isSearching ? searchText : '')
```
Jon-edge, line 74: Does the glitchy behavior happen when blur-ing the input?

Also, would prefer to see this default `''` encoded into `useTransactionList` instead so future users don't need to know about this quirk.

Jon-edge (changes requested):

src/components/scenes/TransactionListScene.tsx
samholmes: I moved the fix into the hook

Jon-edge (approved):
Approved, assuming input field blur doesn't also somehow cause glitchy behavior

PR #4975, opened 2024-04-11
Fix Kado Plain bank linking glitchiness

paullinator:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207057109616480

### DIFF
--- a/src/plugins/gui/providers/kadoProvider.ts
+++ b/src/plugins/gui/providers/kadoProvider.ts
@@ -7,7 +7,7 @@ import { SendScene2Params } from '../../../components/scenes/SendScene2'
 import { ENV } from '../../../env'
 import { lstrings } from '../../../locales/strings'
 import { CryptoAmount } from '../../../util/CryptoAmount'
-import { isHex } from '../../../util/utils'
+import { datelog, isHex } from '../../../util/utils'
 import { SendErrorBackPressed, SendErrorNoTransaction } from '../fiatPlugin'
 import { FiatDirection, FiatPaymentType, SaveTxActionParams } from '../fiatPluginTypes'
 import {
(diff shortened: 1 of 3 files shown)


src/plugins/gui/providers/kadoProvider.ts
```diff
@@ -613,10 +616,32 @@ export const kadoProvider: FiatProviderFactory = {
             }
             console.log('Launching Kado webview url=' + url.href)
 
+            // If Kado needs to launch the Plaid bank linking widget, it needs it in an external
+            // webview to prevent some glitchiness. When needed, Kado will send an onMessage
+            // trigger with the url to open. The below code is derived from Kado's sample code
+            const onMessage = (data: string) => {
+              try {
+                datelog(`**** Kado onMessage ${data}`)
```
samholmes, line 624: is data for sure not sensitive and therefore okay to log?

src/plugins/gui/providers/kadoProvider.ts
```diff
@@ -613,10 +616,32 @@ export const kadoProvider: FiatProviderFactory = {
...
+                const message = JSON.parse(data)
```
samholmes, line 625: We should probably be using cleaners for this.

src/plugins/gui/providers/kadoProvider.ts
```diff
@@ -613,10 +616,32 @@ export const kadoProvider: FiatProviderFactory = {
...
+                if (message && message?.type === 'PLAID_NEW_ACH_LINK') {
+                  const achLink = message?.payload?.link
+                  showUi.openExternalWebView({ url: achLink }).catch(e => {
```
samholmes, line 628: This archLink could be undefined and not string. How did this not error on the static type checker?

samholmes (changes requested):
This may be a "quick fix" but I think we need to be safer when handling the IO here.

src/plugins/gui/providers/kadoProvider.ts
paullinator: For sure. it's just a url link to redirect the user

src/plugins/gui/providers/kadoProvider.ts
paullinator: not sure. but fixed with cleaner

samholmes (approved):

PR #4977, opened 2024-04-12
Jon/search-create-activate

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206538221080918

### DIFF
--- a/src/components/themed/WalletListSwipeable.tsx
+++ b/src/components/themed/WalletListSwipeable.tsx
@@ -66,7 +66,7 @@ function WalletListSwipeableComponent(props: Props) {
 
   const handleCreateWallet = useHandler(async (walletId: string, tokenId: EdgeTokenId) => {
     dispatch(selectWalletToken({ navigation, walletId, tokenId }))
-      .then(() => navigation.navigate('transactionList', { walletId, tokenId }))
+      .then(activationNotRequired => activationNotRequired && navigation.navigate('transactionList', { walletId, tokenId }))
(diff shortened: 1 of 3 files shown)


src/components/themed/WalletListSwipeable.tsx
```diff
@@ -66,7 +66,7 @@ function WalletListSwipeableComponent(props: Props) {
 
   const handleCreateWallet = useHandler(async (walletId: string, tokenId: EdgeTokenId) => {
     dispatch(selectWalletToken({ navigation, walletId, tokenId }))
-      .then(() => navigation.navigate('transactionList', { walletId, tokenId }))
+      .then(() => selectWalletToken({ navigation, walletId, tokenId }))
```
peachbits, line 69: Calling an action without dispatch doesn't do anything. This is equivalent to deleting the .then block completely. 

Instead, you can keep the transactionList navigation but make it conditional based on the return value of selectWalletToken which is false for unactivated wallets and true for every other case that successfully updates redux.

peachbits (changes requested):

src/components/themed/WalletListSwipeable.tsx
Jon-edge: Oops didn't notice the duplication here

peachbits (approved):

PR #4979, opened 2024-04-16
Properly set default fiat value for non-USD amounts

paullinator:
Convert $500 to fiat denom of wallet and round to highest decimal precision. This prevents getting a default quote for excessively small or large amounts

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207000283682116

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - changed: "Split" wallet menu options consolidated into one
 - fixed: Possible to see receive address for a wallet requiring activation when creating the wallet from search
 - fixed: Glitch on transaction list scene when starting search
+- fixed: Properly set default buy/sell fiat amount for non-USD wallets
 - removed: 'X' button on modals
 
 ## 4.5.0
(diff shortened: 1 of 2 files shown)


ios/edge.xcodeproj/project.pbxproj
Jon-edge, line 1: What does this have to do with this change?

Jon-edge (changes requested):

Jon-edge (approved):
I assume the pbxproj commit was accidental, approved if so

ios/edge.xcodeproj/project.pbxproj
paullinator: accidental but I think it may be needed since the changes were automatically created upon compiling the app. Might have been a missing change that got left out of the RN 73 upgrade. If you see this change as well, then it's needed. Otherwise I'll drop it and call it a fluke.

PR #4991, opened 2024-04-19
Add Migrate to Import Flow

Jon-edge:
NOTE: Was unable to route directly to MigrateWalletCalculateFeeScene. Doing so made it think that there was still 0 max spendable for the wallets, even after waiting for wallets to fully sync...

As a workaround, routed a step earlier in the migrate flow instead.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205838484838736

### DIFF
--- a/src/components/scenes/CreateWalletCompletionScene.tsx
+++ b/src/components/scenes/CreateWalletCompletionScene.tsx
@@ -1,4 +1,4 @@
-import { EdgeCreateCurrencyWallet } from 'edge-core-js'
+import { EdgeCreateCurrencyWallet, EdgeCurrencyWallet } from 'edge-core-js'
 import * as React from 'react'
 import { ActivityIndicator, View } from 'react-native'
 import { FlatList } from 'react-native-gesture-handler'
@@ -21,6 +21,7 @@ import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 import { CreateWalletSelectCryptoRow } from '../themed/CreateWalletSelectCryptoRow'
 import { EdgeText } from '../themed/EdgeText'
 import { SceneHeader } from '../themed/SceneHeader'
+import { MigrateWalletItem } from './MigrateWalletSelectCryptoScene'
 
 export interface CreateWalletCompletionParams {
   createWalletList: WalletCreateItem[]
@@ -43,6 +44,7 @@ const CreateWalletCompletionComponent = (props: Props) => {
   const account = useSelector(state => state.core.account)
 
   const [done, setDone] = React.useState(false)
+  const [wallets, setWallets] = React.useState<EdgeCurrencyWallet[]>([])
 
   const { newWalletItems, newTokenItems } = React.useMemo(() => splitCreateWalletItems(createWalletList), [createWalletList])
 
@@ -107,6 +109,9 @@ const CreateWalletCompletionComponent = (props: Props) => {
           showError(result.error)
           setItemStatus(currentState => ({ ...currentState, [filteredCreateItemsForDisplay[i].key]: 'error' }))
         } else {
+          // Wait for wallet creation
+          await account.waitForCurrencyWallet(result.result.id)
+
           setItemStatus(currentState => ({ ...currentState, [filteredCreateItemsForDisplay[i].key]: 'complete' }))
         }
       }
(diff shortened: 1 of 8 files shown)


src/components/scenes/CreateWalletCompletionScene.tsx
```diff
@@ -160,6 +169,40 @@ const CreateWalletCompletionComponent = (props: Props) => {
...
+    })
+
+    // Navigate to the migration screen with the prepared list
+    if (migrateWalletList.length > 0) {
+      // TODO: Waiting for wallet sync still yields 0 getMaxSpendable...
+      // navigation.navigate('migrateWalletCalculateFee', {
+      //   migrateWalletList
+      // })
+
+      navigation.navigate('migrateWalletSelectCrypto', {
```
peachbits, line 198: It seems like the migrateWalletSelectCrypto scene is being used to delay the user so hopefully the wallets finish syncing by the time they get to the migrateWalletCalculateFee scene. Maybe be better to instead wait for sync ratio and then pause after it briefly to ensure the balance is ready. That seems more likely to be successful than adding the extra scene to slow them down. 

If the routing to the select scene stays or not is ultimately not that important but I dod think where a user waits for the wallet to sync is. That should be in the calculate fee scene before getMaxSpendable is called since there's already a spinner.

peachbits (changes requested):

src/components/scenes/CreateWalletCompletionScene.tsx
```diff
@@ -173,30 +173,21 @@ const CreateWalletCompletionComponent = (props: Props) => {
...
     const migrateWalletList: MigrateWalletItem[] = newWalletItems.map(createWallet => {
       // Link the createWalletIds with the created wallets
-      const { currencyCode, key, pluginId, tokenId, walletType } = createWallet
+      const { key, pluginId } = createWallet
       const wallet = wallets.find(wallet => wallet.currencyInfo.pluginId === pluginId)
 
       return {
-        createWalletIds: wallet == null ? [''] : [wallet.id],
-        currencyCode,
+        ...{ ...createWallet, createWalletIds: wallet == null ? [''] : [wallet.id] },
```
peachbits, line 180: Don't need to double spread. This should work:

```
return {
...createWallet,
createWalletIds: wallet == null ? [''] : [wallet.id] },
displayName: walletNames[key],
key,
type: 'create'
}
```

src/components/scenes/MigrateWalletCalculateFeeScene.tsx
```diff
@@ -126,6 +127,8 @@ const MigrateWalletCalculateFeeComponent = (props: Props) => {
   // track of which makeSpends are successful so we can enable the slider. A single failure from any of a wallet's assets will cast them all as failures.
   useAsyncEffect(
     async () => {
+      if (!migrateWalletsSynced) return
```
peachbits, line 130: Optional: The total wait to sync (and then wait the extra five seconds for balances to hopefully work) is kind of long so maybe it's worth adding a toast like "fragment_transaction_list__synchronizing" as an extra heads up that stuff is working under the hood.

peachbits (changes requested):

src/components/scenes/MigrateWalletCalculateFeeScene.tsx
Jon-edge: Good call

src/components/scenes/CreateWalletCompletionScene.tsx
```diff
@@ -184,7 +218,10 @@ const CreateWalletCompletionComponent = (props: Props) => {
             scrollIndicatorInsets={SCROLL_INDICATOR_INSET_FIX}
           />
 
-          <SceneButtons primary={{ label: lstrings.string_next_capitalized, disabled: !done, onPress: handleNext }} />
+          <SceneButtons
+            primary={{ label: lstrings.string_next_capitalized, disabled: !done, onPress: handleNext }}
+            secondary={{ label: lstrings.migrate_label, disabled: !done, onPress: handleMigrate }}
```
peachbits, line 224: I missed this on the first round of review. This should only be visible for the import flow so it should be dependent on the importText being present

peachbits (changes requested):

peachbits (approved):

PR #5002, opened 2024-04-26
Unify Backup Modals

Jon-edge:
Could consolidate a bit further, but avoiding extra work until after experiment is concluded.

"Original" - Settings, deleting light accounts, receiving FIO requests
image

"Original emergency hack" / "Control Variant" - Our quick and dirty message after realizing people could lose funds. Refreshed to be generally consistent with other backup modals
image

Other Variants:
image
image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-react-gui/pull/5009

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207024313887978


src/components/modals/ButtonsModal.tsx
```diff
@@ -62,42 +61,39 @@ export function ButtonsModal<Buttons extends { [key: string]: ButtonInfo }>(prop
...
-            }
-
-            return <ButtonUi4 key={key} label={label} marginRem={0.5} type={type} onPress={handlePress} layout="column" />
-          })}
-        </View>
+        <ModalButtons
+          primary={buttonInfo.length > 0 ? buttonInfo[0] : undefined}
+          secondary={buttonInfo.length > 1 ? buttonInfo[1] : undefined}
+          tertiary={buttonInfo.length > 2 ? buttonInfo[2] : undefined}
+        />
```
swansontec, line 96: This is a pretty big change! All the buttons modals in the app are going to have their buttons types affected, which will be a big QA effort. This should be its own PR, and in that PR, we'll have to audit all the places where we use this, and we will have to remove the `theme.preferPrimaryButton`.

swansontec (changes requested):
The backup modal changes look OK, but the changes to ButtonsModal are too broad-reaching to slip into this PR. Please skip those out-of-scope commits, and submit them separately.

It should be pretty easy to add `type="primary"` to the `BackupModal` buttons, so I don't think this shoud be a difficult rebase.

src/components/modals/ButtonsModal.tsx
Jon-edge: Yeah this bigger change came from when I resolved with Madison that this is the way we are going regarding modal buttons. I confirmed with Paul if it was safe for white labels to deprecate `theme.preferPrimaryButton` and this is the last remaining place we use `theme.preferPrimaryButton`.

Jon-edge: Created https://github.com/EdgeApp/edge-react-gui/pull/5009

Jon-edge (commented):
Created https://github.com/EdgeApp/edge-react-gui/pull/5009

swansontec (approved):
Thanks for getting this fixed.

PR #5003, opened 2024-04-27
Standard FilledTextInput Modal Margins

Jon-edge:
- Retained total margins for the most part. Prioritizing consistency over 100% retaining original margins.
- `PasswordReminderModal`, `TextInputModal`: Changes might look like they're losing vertical margins between `FilledTextInput` and buttons, but using the new standard `ModalButtons` takes care of that by providing extra top button margins.
- Added unit tests for changed modals that had no coverage

image
image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-react-gui/pull/5009

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206366276415248

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 - changed: Consistent light account backup modal UI
 - changed: Currency icon now shown in wallet picker button in Swap and Transaction List scenes
 - changed: Transaction List Scene displays wallet name in title
+- changed: Consistent light account backup modal UI
+- changed: Consistent margins for modal text inputs
 - fixed: Various React Native debug warnings
 - fixed: Fio domain registration supported payment currencies
 - removed: Bitpay deeplink support
--- a/src/__tests__/modals/CountryListModal.test.tsx
+++ b/src/__tests__/modals/CountryListModal.test.tsx
@@ -1,0 +1,27 @@
+import { describe, expect, it } from '@jest/globals'
+import * as React from 'react'
+import TestRenderer from 'react-test-renderer'
+
+import { CountryListModal } from '../../components/modals/CountryListModal'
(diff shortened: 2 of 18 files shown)


src/__tests__/components/FilledTextInput.test.tsx
```diff
@@ -3,14 +3,14 @@ import * as React from 'react'
 import { createRenderer } from 'react-test-renderer/shallow'
 
 import { SearchIconAnimated } from '../../components/icons/ThemedIcons'
-import { FilledTextInput } from '../../components/themed/FilledTextInput'
+import { FilledTextInputRaw } from '../../components/themed/FilledTextInput'
```
paullinator, line 6: I'm not sure if there was team consensus around this, but I would invert this name change. FilledTextInput should be the "raw" version. It should be capable of being a library used outside of Edge.

`ModalFilledTextInput` would be the preferred name for a wrapper that adds padding specific to modals or whatever use Edge needs.

paullinator (changes requested):

src/__tests__/components/FilledTextInput.test.tsx
Jon-edge: My two considerations that prompted the change were:
1. Make it more apparent changes are needed to bring a parent up to spec within our codebase, to make sure it's painfully obvious it's not preferable for callers to make up their own special styling
2. That this layout isn't specific to modals but UI4 conventions in general of splitting 0.5rem everywhere

Mainly #2 drove this change but we can just duplicate the styling I suppose between modal and scene.

paullinator (approved):

PR #5004, opened 2024-04-29
Fix InsufficientFeesModal logic

peachbits:
---
  - https://app.asana.com/0/0/1206616383733403

### DIFF
--- a/src/components/scenes/SendScene2.tsx
+++ b/src/components/scenes/SendScene2.tsx
@@ -344,7 +344,7 @@ const SendComponent = (props: Props) => {
         if (error == null) return
         console.log(error)
         const insufficientFunds = asMaybeInsufficientFundsError(error)
-        if (insufficientFunds != null && insufficientFunds.tokenId != null && spendInfo.tokenId !== insufficientFunds.tokenId) {
+        if (insufficientFunds != null && spendInfo.tokenId !== insufficientFunds.tokenId) {
           await Airship.show(bridge => <InsufficientFeesModal bridge={bridge} coreError={insufficientFunds} navigation={navigation} wallet={coreWallet} />)
         }
       })
--- a/src/components/scenes/SendScene2.tsx
+++ b/src/components/scenes/SendScene2.tsx
@@ -9,7 +9,8 @@ import {
   EdgeSpendInfo,
   EdgeSpendTarget,
   EdgeTokenId,
-  EdgeTransaction
+  EdgeTransaction,
+  InsufficientFundsError
 } from 'edge-core-js'
 import * as React from 'react'
 import { ActivityIndicator, Alert, TextInput, View } from 'react-native'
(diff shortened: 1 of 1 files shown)


src/components/scenes/SendScene2.tsx
```diff
@@ -984,6 +992,13 @@ const SendComponent = (props: Props) => {
           } else {
             e.message = sprintf(lstrings.stake_error_insufficient_s, errorCurrencyCode)
           }
+
+          if (spendInfo.tokenId !== insufficientFunds.tokenId) {
+            // Show the modal if the flip input modal is closed or save it to show when it closes later
+            const modalPromise = async () =>
+              await Airship.show(bridge => <InsufficientFeesModal bridge={bridge} coreError={insufficientFunds} navigation={navigation} wallet={coreWallet} />)
+            flipInputModalOpen.current == null ? (insufficientFeesModalPromise.current = modalPromise) : modalPromise().catch(error => showError(error))
```
swansontec, line 1000: This condition will never be true, because it's always `true | false`, not `null`.

src/components/scenes/SendScene2.tsx
```diff
@@ -189,6 +189,9 @@ const SendComponent = (props: Props) => {
   const parentExchangeDenom = getExchangeDenom(coreWallet.currencyConfig, null)
   const iconColor = useIconColor({ pluginId, tokenId })
 
+  const insufficientFeesModalPromise = React.useRef<(() => Promise<void>) | undefined>(undefined)
```
swansontec, line 192: This needs another name, because it's not a promise, but rather a function. Maybe call it `pendingInsufficientFeesModal`, or even change the type to `React.useRef<InsufficientFundsError | undefined>` and then copy / past the actual modal display logic to the two places. That way it's just plain-old data, not hidden logic.

swansontec (changes requested):
There is a logic condition that will never run, so we need to fix that.

samholmes (approved):

src/components/scenes/SendScene2.tsx
```diff
@@ -995,9 +992,13 @@ const SendComponent = (props: Props) => {
...
             // Show the modal if the flip input modal is closed or save it to show when it closes later
-            const modalPromise = async () =>
-              await Airship.show(bridge => <InsufficientFeesModal bridge={bridge} coreError={insufficientFunds} navigation={navigation} wallet={coreWallet} />)
-            flipInputModalOpen.current == null ? (insufficientFeesModalPromise.current = modalPromise) : modalPromise().catch(error => showError(error))
+            if (flipInputModalRef.current != null) {
+              pendingInsufficientFees.current = insufficientFunds
+            } else {
+              await Airship.show(bridge => (
+                <InsufficientFeesModal bridge={bridge} coreError={insufficientFunds} navigation={navigation} wallet={coreWallet} />
+              )).catch(error => showError(error))
```
samholmes, line 1000: Dedupe this call to InsufficientFeesModal to a `showInsufficientFeesModal(error: unknown)`

swansontec (approved):

PR #5009, opened 2024-05-02
Jon/fix/buttons-modal-buttons

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207208111119437

### DIFF
--- a/src/components/modals/ButtonsModal.tsx
+++ b/src/components/modals/ButtonsModal.tsx
@@ -3,15 +3,29 @@ import { View, ViewStyle } from 'react-native'
 import { AirshipBridge } from 'react-native-airship'
 
 import { useHandler } from '../../hooks/useHandler'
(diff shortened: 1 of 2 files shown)


src/components/modals/ButtonsModal.tsx
```diff
@@ -44,6 +58,7 @@ export interface ButtonModalProps<Buttons> {
 }
 
 /**
+ * @deprecated
```
swansontec, line 61: This is the wrong Javadoc syntax. The first line is always the item description, and then the @ tags go below that. So, please move the `@deprecated` to the bottom of the block and then add a message, so `@deprecated Use ButtonsModal2 instead`

swansontec (approved):
This will work. Please clean up the comment thing though.

PR #5014, opened 2024-05-06
Add spinning toast when finalizing Paybis quote

paullinator:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207214496338536

### DIFF
--- a/src/plugins/gui/fiatPlugin.tsx
+++ b/src/plugins/gui/fiatPlugin.tsx
@@ -249,6 +249,9 @@ export const executePlugin = async (params: {
     },
     exitScene: async () => {
       navigation.pop()
+    },
+    waitForAnimationFrame: async () => {
+      await new Promise(resolve => requestAnimationFrame(resolve))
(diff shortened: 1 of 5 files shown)


src/plugins/gui/fiatPlugin.tsx
```diff
@@ -249,6 +249,9 @@ export const executePlugin = async (params: {
     },
     exitScene: async () => {
       navigation.pop()
+    },
+    waitForAnimationFrame: async () => {
+      await new Promise(resolve => requestAnimationFrame(resolve))
```
swansontec, line 254: I wonder if calling this `prepareForBigCalculation` would make more sense, since it's goal is to get things in shape before clogging the CPU.

swansontec (approved):
This is pretty simple! I have one naming suggestion, but you name is fine too.

PR #5026, opened 2024-05-15
Add performance metrics to logs

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206479125385477

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,12 @@
 
 ## Unreleased
 
+- added: Add Visa/MC buy support with Paybis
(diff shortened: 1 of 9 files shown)


swansontec (approved):
The native boot-up measurements make sense, but the react-native-performance library seems overkill for logging some timestamp differences. Still, assuming we go with this library choice, it seems weird to then completely alter its API. Now we have complexity on top of complexity.

Nevertheless, I see that it works, so let's get it in.

PR #5030, opened 2024-05-16
React Navigation v7

swansontec:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205308893333750


src/components/Main.tsx
```diff
@@ -279,7 +279,14 @@ export const Main = () => {
       {experimentConfig == null ? (
         <LoadingSplashScreen />
       ) : (
-        <NavigationContainer theme={reactNavigationTheme}>
+        <NavigationContainer
+          // This lets us dive into nested navigators using just a route name.
+          // We need to port to using the new syntax,
+          // so do `navigate('stackName', { screen: 'nestedSceneName' }`
+          // instead of `navigate('nestedSceneName')`:
+          navigationInChildEnabled
```
samholmes, line 287: Is this necessary on all instances of `Stack.Navigator` or other navigator component instances perhaps?

samholmes (approved):
This seems like a simple upgrade even though it's a major version bump from an API level change. However, the behavior implications could only be known and resolved by testing, so let's get this in QAs hands asap. Hopefully they're navigation test suite covers all navigation calls.

src/components/Main.tsx
swansontec: Probably?

PR #5037, opened 2024-05-25
Jon/fix/multiline-text-input

Jon-edge:
Original implementation (saved in `jon/wip/multiline-input-dyn-maxheight`)
https://github.com/EdgeApp/edge-react-gui/assets/90650827/8d141a5d-53f5-4e1c-9bf5-108530f6aab9

New simpler implementation:
https://github.com/EdgeApp/edge-react-gui/assets/90650827/ca0aafdf-aea9-4550-bb67-320732d0dd69

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1205865555810653

### DIFF
--- a/src/components/themed/FilledTextInput.tsx
+++ b/src/components/themed/FilledTextInput.tsx
@@ -245,7 +245,7 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
       : keyboardType
 
   return (
-    <View style={marginRemStyle}>
+    <OuterContainer multiline={multiline} marginRemStyle={marginRemStyle}>
       <EdgeTouchableWithoutFeedback accessible={false} testID={testID} onPress={() => focus()}>
         <Container disableAnimation={disableAnimation} focusAnimation={focusAnimation} multiline={multiline} scale={scale}>
           <SideContainer scale={leftIconSize}>{LeftIcon == null ? null : <LeftIcon color={iconColor} size={leftIconSize} />}</SideContainer>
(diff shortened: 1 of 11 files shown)


src/components/themed/FilledTextInput.tsx
```diff
@@ -556,15 +576,24 @@ const StyledNumericInput = styledWithRef(NumericInput)<{
...
   // Need 2 pixels of shift given a 16 point rem settings
   // This is due to Android rendering a text input vertically lower
   // than a Text field by ~2 pixels
   const androidVShift = isAndroid ? rem / 8 : 0
 
+  // For animating height changes in multiline input
+  const animatedHeightStyle = useAnimatedStyle(() => {
+    if (dynamicHeight.value == null) return {}
+    return {
+      height: withTiming(dynamicHeight.value, { duration: 100 }) // Smooth animation for height change
```
swansontec, line 593: This works great, and feels really smooth! However, if you type in more text than fits on the screen, the text input just keeps growing forever, and never gets a scroll bar. This makes it impossible to edit the top lines of text, because they get pushed off screen.

So, I think we either need a `flexShrink` or a `maxHeight` or something like that, so the text input does not overflow the screen.

swansontec (changes requested):

src/components/themed/FilledTextInput.tsx
Jon-edge: Ok looks like I way overcomplicated it. Trying to cap the height ends up in a circular dependency nightmare, and trying to guess how tall an input can be beforehand isn't really feasible across devices.

Just gonna table the original implementation, might come in handy later. For now, just making it always full screen for multiline.

swansontec (approved):
It's not a pretty, but this new version is nice & stable.

PR #5041, opened 2024-05-30
Jon/smol-4.8

Jon-edge:
image
image
image
image
image
image
image
image

image
image
image
image
image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207225096190322

### DIFF
--- a/src/components/scenes/CreateWalletSelectCryptoScene.tsx
+++ b/src/components/scenes/CreateWalletSelectCryptoScene.tsx
@@ -19,15 +19,16 @@ import { useDispatch, useSelector } from '../../types/reactRedux'
 import { EdgeSceneProps, NavigationProp } from '../../types/routerTypes'
 import { EdgeAsset } from '../../types/types'
 import { logEvent } from '../../util/tracking'
+import { EdgeAnim } from '../common/EdgeAnim'
+import { SceneButtons } from '../common/SceneButtons'
 import { SceneWrapper } from '../common/SceneWrapper'
 import { SearchIconAnimated } from '../icons/ThemedIcons'
 import { ListModal } from '../modals/ListModal'
 import { WalletListModal, WalletListResult } from '../modals/WalletListModal'
 import { Airship, showError } from '../services/AirshipInstance'
-import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
+import { useTheme } from '../services/ThemeContext'
 import { CreateWalletSelectCryptoRow } from '../themed/CreateWalletSelectCryptoRow'
 import { EdgeText } from '../themed/EdgeText'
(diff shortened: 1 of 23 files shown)


src/components/ui4/RowUi4.tsx
```diff
@@ -42,13 +44,21 @@ export const RowUi4 = (props: Props) => {
...
 
   const margin = sidesToMargin(mapSides(fixSides(marginRem, 0.5), theme.rem))
 
   const numberOfLines = textHeights[maximumHeight]
 
+  // TODO: Merge styles.containerTemp into styles.container permanently.
+  // Explicitly setting the container style here to avoid unit test snapshot
+  // diffs.
+  const containerStyle: StyleProp<ViewStyle> = React.useMemo(
+    () => (shrinkBody ? [styles.container, margin, styles.containerTemp] : [styles.container, margin]),
```
paullinator, line 58: should be explicit boolean check `shrinkBody === true`

src/components/scenes/CreateWalletSelectCryptoScene.tsx
```diff
@@ -288,7 +288,7 @@ const CreateWalletSelectCryptoComponent = (props: Props) => {
     () => (
       <Fade noFadeIn={defaultSelection.length > 0} visible={selectedItems.size > 0} duration={300}>
         <View style={styles.bottomButton}>
-          <ButtonUi4 label={lstrings.string_next_capitalized} onPress={handleNextPress} />
+          <ButtonUi4 label={lstrings.string_next_capitalized} testID="createWalletCryptoNext" onPress={handleNextPress} />
```
paullinator, line 291: Did you test that this works? Usually you need to force the parent container to not takeover taps by putting
`accessible={false}` on the parent.

Or the problem is due to absolute positioning.

paullinator (changes requested):

src/components/scenes/CreateWalletSelectCryptoScene.tsx
Jon-edge: `accessible={false}` had no impact to the behavior. There was no Maestro tappability issues through this View when disabling the fade animation for Maestro builds.

Kind of related: `isMaestro()` itself is broken - does not return true for Maestro builds.

Jon-edge: Use of `EdgeAnim` helps, but only to a degree. Nesting buttons within a hard-coded styled `View` as done before doesn't work no matter what. `ButtonsView` flavors such as `SceneButtons` works perfectly and makes the spacing consistent.

paullinator (approved):

PR #5049, opened 2024-05-31
Pause iOS Piratechain wallets

peachbits:
---
  - https://app.asana.com/0/0/1207457481301092


src/components/services/WalletLifecycle.ts
```diff
@@ -114,7 +114,10 @@ export class WalletLifecycleComponent extends React.Component<Props> {
       if (walletItem.type !== 'asset') continue
       const { token, tokenId, wallet } = walletItem
       if (token != null || tokenId != null) continue
-      if (isKeysOnlyPlugin(wallet.currencyInfo.pluginId)) continue
+      if (isKeysOnlyPlugin(wallet.currencyInfo.pluginId)) {
+        wallet.changePaused(true).catch(showError)
```
swansontec, line 118: This should not be necessary. All wallets start off paused, and we never start keys-only wallets. Why add it?

Assuming this line is necessary for some reason (I cannot imagine why), do `.catch(error => showError(error)`, instead of `.catch(showError)`. By inserting a little arrow function in here, we get a better stack trace.

swansontec (changes requested):

swansontec (approved):

PR #5052, opened 2024-06-01
Properly size and flex the side menu

Jon-edge:
image

Unclear what dependencies changed to make this issue surface, but we just got lucky before.
This change properly flexes the side menu

### Update:
Updates to RN/RN Navigation are likely the culprit. Likely some defaults were changed, and in combination with our 
Since the `Drawer` itself is absolutely positioned and does translations to `right` under the hood, we cannot rely on flex and must set a specific width. 

Without setting an explicit width, though the side menu appears properly sized, the drawer itself takes up extra invisible horizontal space to the left of the menu. This extra horizontal space is not tappable and prevents the side menu from closing when tapping outside of the menu.
image

Below are after the updates:
image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207409357193836

### DIFF
--- a/src/components/Main.tsx
+++ b/src/components/Main.tsx
@@ -348,7 +348,7 @@ const EdgeApp = () => {
       screenOptions={{
         drawerPosition: 'right',
         drawerType: 'front',
-        drawerStyle: { backgroundColor: 'transparent', bottom: 0 },
+        drawerStyle: { backgroundColor: 'transparent', bottom: 0, width: '66%' },
         headerShown: false
       }}
     >
--- a/src/components/themed/SideMenu.tsx
+++ b/src/components/themed/SideMenu.tsx
@@ -30,6 +30,7 @@ import { NavigationBase } from '../../types/routerTypes'
 import { parseDeepLink } from '../../util/DeepLinkParser'
 import { IONIA_SUPPORTED_FIATS } from '../cards/VisaCardCard'
 import { EdgeTouchableOpacity } from '../common/EdgeTouchableOpacity'
+import { styled } from '../hoc/styled'
 import { ButtonsModal } from '../modals/ButtonsModal'
 import { ScanModal } from '../modals/ScanModal'
 import { LoadingSplashScreen } from '../progress-indicators/LoadingSplashScreen'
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - added: Dynamic height for multiline text inputs
 - added: Sentry SDK for crash reporting
 - changed: Add EUID to Simplex quoting API calls
+- fixed: Excessive side menu width
```
paullinator, line 8: No need as this wasn't broken before 4.7

paullinator (changes requested):

src/components/themed/SideMenu.tsx
```diff
@@ -476,6 +477,13 @@ const getStyles = cacheStyles((theme: Theme) => ({
   }
 }))
 
+// TODO: Refactor more of this component into styled components
+const OuterView = styled(View)<{ insets: { top: number; bottom: number } }>(() => props => ({
+  flexShrink: 1,
```
samholmes, line 482: This `flexShrink` isn't needed.

src/components/Main.tsx
```diff
@@ -348,7 +348,7 @@ const EdgeApp = () => {
       screenOptions={{
         drawerPosition: 'right',
         drawerType: 'front',
-        drawerStyle: { backgroundColor: 'transparent', bottom: 0 },
+        drawerStyle: { backgroundColor: 'transparent', bottom: 0, flexDirection: 'row', justifyContent: 'flex-end' },
```
samholmes, line 351: The problem is that the component has a set width. This can be fixed with `width: undefined` and then we need to position the component to the right with `right: 0`. Add a `alignItems: 'stretch'` for the SideMenu to be stretched out (or add `flexGrow: 1` back to the SideMenu, perhaps). The only problem is there is 56 pixels mis-alignement for some reason when the Drawer is open for some reason, so `right: 56` is needed.

This approach will not only position the SideMenu visually but also the drawer itself is positioned correctly. Otherwise the drawer continues to have the same width and this effects tap areas.

Alternatively, if we don't want to change the width and mess with that, and we're okay with having the tap-areas remain as they are, then the simpler fix is:

```
drawerStyle: { backgroundColor: 'transparent', bottom: 0, alignItems: 'flex-end' }
```

No need to change the orientation because we can use `align` instead of `justify`.

samholmes (changes requested):

samholmes (approved):

PR #5056, opened 2024-06-04
Jon/fix/debounce-2

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206854054598716

### DIFF
--- a/src/components/scenes/GuiPluginListScene.tsx
+++ b/src/components/scenes/GuiPluginListScene.tsx
@@ -101,7 +101,7 @@ interface StateProps {
   disablePlugins: NestedDisableMap
   insetStyle: InsetStyle
   forcedWalletResult?: WalletListResult
-  onCountryPress: () => void
+  onCountryPress: () => Promise<void>
   onPluginOpened: () => void
   onLogEvent: OnLogEvent
   onScroll: SceneScrollHandler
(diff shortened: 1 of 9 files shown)


src/components/scenes/GuiPluginListScene.tsx
```diff
@@ -522,8 +522,8 @@ export const GuiPluginListScene = React.memo((props: OwnProps) => {
     dispatch(logEvent(event, values))
   })
 
-  const handleCountryPress = useHandler(() => {
-    dispatch(
+  const handleCountryPress = useHandler(async () => {
+    await dispatch(
```
swansontec, line 526: This works at run-time, but there seems to be a type mis-match. The return type of `showCountrySelectionModal` should be `ThunkAction<Promise<void>>` instead of `ThunkAction<void>`. If you fix that, then a few more red squiggles appear at other places that have this same debounce issue.

src/components/scenes/DevTestScene.tsx
```diff
@@ -113,7 +113,7 @@ export function DevTestScene(props: Props) {
   let balance = coreWallet?.balanceMap.get(tokenId) ?? ''
   if (eq(balance, '0')) balance = ''
   const headerText = 'Select Wallet'
-  const headerCallback = () => console.log('Header pressed')
+  const headerCallback = async () => console.log('Header pressed')
```
swansontec, line 116: This `async` doesn't do anything, because there is nothing to `await`. Maybe the type below should be `headerCallback?: () => (void | Promise<void>)`.

That's how most of our tappable components work - if you return a promise, we handle it, and if you return void, we handle that too.

swansontec (changes requested):
A few really small changes needed.

swansontec (approved):

PR #5066, opened 2024-06-05
2FA/IP Validation Notification Card

Jon-edge:
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206859452306575

### DIFF
--- a/src/components/notification/NotificationView.tsx
+++ b/src/components/notification/NotificationView.tsx
@@ -6,6 +6,7 @@ import { sprintf } from 'sprintf-js'
 
 import { showBackupModal } from '../../actions/BackupModalActions'
 import { getDeviceSettings, writeDeviceNotifDismissInfo } from '../../actions/DeviceSettingsActions'
+import { getLocalAccountSettings, writeNotifDismissInfo } from '../../actions/LocalSettingsActions'
 import { useAsyncEffect } from '../../hooks/useAsyncEffect'
 import { useHandler } from '../../hooks/useHandler'
 import { useWatch } from '../../hooks/useWatch'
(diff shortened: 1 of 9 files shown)


src/constants/constantSettings.ts
```diff
@@ -1,6 +1,7 @@
 export const AAVE_WELCOME = 'aaveWelcome.json'
 export const FIRST_OPEN = 'firstOpen3.json'
 export const LOCAL_EXPERIMENT_CONFIG = 'remoteConfigSticky.json'
+export const NOTIFS_DISMISSED = 'notifsDismissed.json' // To track state of shown notifications
```
swansontec, line 4: I don't think filenames should be public constants. This allows any random code in the app to read or write files that don't belong to it. Instead, I would like to see these constants as *private* variables in the source that's responsible for cleaning / reading / writing this particular file. So, this should move to NotificationViewActions.ts and remove the `export`.

src/types/types.ts
```diff
@@ -166,7 +166,7 @@ const asDeviceSettingsInner = asObject({
   defaultScreen: asMaybe(asDefaultScreen, 'home'),
   developerPluginUri: asMaybe(asString),
   disableAnimations: asMaybe(asBoolean, false),
-  hasInteractedWithBackupModal: asMaybe(asBoolean, false)
+  hasInteractedWithIpValidationNotif: asMaybe(asBoolean, false)
```
swansontec, line 169: Does this belong in a later commit?

src/components/notification/NotificationView.tsx
```diff
@@ -128,6 +127,15 @@ const NotificationViewComponent = (props: Props) => {
...
   )
 
+  // Initialize the local state for the LocalNotifDismissInfo
+  useAsyncEffect(
+    async () => {
+      setLocalNotifDismissInfo(await getLocalNotifDismissInfo())
+    },
+    [],
+    'notifDismissInfo'
+  )
```
swansontec: Just delete this whole block. There is no need to "ensure the file exists", since we'll just return defaults if it's not there.

swansontec (changes requested):
I would like to request a sidebar on the way we are handling disk here. I think we need a better strategy here.

PR #5070, opened 2024-06-06
True Dev Mode: Force Light Account Creation

Jon-edge:
image

- Add new "Force Light Account Creation"
- Slight regrouping to keep developer mode settings together

### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

(https://github.com/EdgeApp/edge-login-ui-rn/pull/192)

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207269419160399


src/__tests__/scenes/SettingsScene.test.tsx
```diff
@@ -42,20 +42,20 @@ describe('MyComponent', () => {
           supportsTouchId={false}
           touchIdEnabled
           // DispatchProps:
-          dispatchUpdateEnableTouchIdEnable={async () => undefined}
-          handleClearLogs={async () => {}}
-          handleSendLogs={() => undefined}
-          lockSettings={() => undefined}
+          onUpdateTouchId={async () => undefined}
```
swansontec, line 46: This isn't "bad naming" - it's actually correct redux usage.

We use the "on" / "handle" nomenclature for true user events that should be handled outside the component. These redux actions, on the other hand, are more like internal helper functions. They aren't visible outside this file and can't be used by anybody else. Therefore, they get named after the redux action itself. Giving them the "on" nomenclature would be confusing, since then we can't tell which function props are internal and which are external.

If you want to clean up this scene (and I agree it's messy), then the better fix would have been to function-ify this whole component and just get rid of the connector. Then all these props would simply vanish.

I realize this is a big ask, though, so this would be an optional fix.

src/components/scenes/SettingsScene.tsx
```diff
@@ -426,6 +422,26 @@ export class SettingsSceneComponent extends React.Component<Props, State> {
               />
             </CardUi4>
           </>
+          {ENV.ALLOW_DEVELOPER_MODE && (
```
swansontec, line 425: Nice cleanup! This looks much prettier.

swansontec (approved):
Approved, since I only have an optional change requested, but it's a pretty important change.

PR #5082, opened 2024-06-11
Token Warning Updates

Jon-edge:
Token warning now:
- Does NOT show after wallet creation completes
- Shows immediately when toggling to enable a token
- Shows once per chain/currency plugin instead of once per wallet

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207532762643127


swansontec (approved):
This looks like it will work.

PR #5091, opened 2024-06-14
Jon/post-install-survey

Jon-edge:
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-login-ui-rn/pull/193

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207270007924847


src/util/tracking.ts
```diff
@@ -215,6 +217,7 @@ export function logEvent(event: TrackingEventName, values: TrackingValues = {}):
         const notificationPermission = await checkNotifications()
         params.notificationStatus = notificationPermission.status
 
+        console.debug({ event, params })
```
swansontec, line 220: Leftover logging?

src/components/modals/ButtonsModal.tsx
```diff
@@ -12,6 +12,8 @@ import { ModalUi4 } from '../ui4/ModalUi4'
 export interface ButtonInfo {
   label: string
 
+  disabled?: boolean
```
swansontec, line 15: Rejected. The buttons modal is for getting simple yes / no style questions. If you need more complicated interactions, you need to build your own modal. It's not complicated.

Jon, you of all people should understand this. You have been campaigning for re-usable components that are designed for a single purpose (`ModalFilledTextInput` comes to mind). You have also been fighting against cheap hacks, where a UI is "close enough" so we just tweak it instead of building the right thing.

We don't need to turn the buttons modal into a dumpster fire for some one-off feature. Just take your reusable buttons components, your reusable text components, and our reusable themed modal, and do it the right way.

src/components/modals/ButtonsModal.tsx
```diff
@@ -40,6 +42,9 @@ export interface ButtonModalProps<Buttons> {
   /** @deprecated. Does nothing. */
   // eslint-disable-next-line react/no-unused-prop-types
   closeArrow?: boolean
+
+  /** Called when modal is dismissed */
+  onCancel?: () => void | Promise<void>
```
swansontec, line 47: Also rejected.

swansontec (changes requested):

Jon-edge:
Other changes:
- Our `shuffleArray` function performs better so using that instead.
- Move the header out of the scroll area
- `SmallText` subtext to align with our backup modal style

swansontec (approved):
Thanks for fixing this! I can actually understand the code now.

PR #5098, opened 2024-06-17
WalletConnect: Select wallet before navigating to connect scene

peachbits:
Passing a wallet into the scene means we don't need to use redux selected wallet ID

---
  - https://app.asana.com/0/0/1207574551270982

### DIFF
--- a/src/components/scenes/WcConnectScene.tsx
+++ b/src/components/scenes/WcConnectScene.tsx
@@ -102,8 +102,8 @@ export const WcConnectScene = (props: Props) => {
     handleWalletListModal().catch(err => showError(err))
   })
 
-  useUnmount(() => {
-    if (!connected.current) wallet.otherMethods.rejectSession(proposal)
+  useUnmount(async () => {
+    if (!connected.current) await walletConnect.rejectSession(proposal)
   })
 
   const renderWalletSelect = () => {
--- a/src/components/scenes/WcConnectScene.tsx
+++ b/src/components/scenes/WcConnectScene.tsx
@@ -4,21 +4,18 @@ import { ScrollView, View } from 'react-native'
 import FastImage from 'react-native-fast-image'
 import { sprintf } from 'sprintf-js'
 
-import { selectWalletToken } from '../../actions/WalletActions'
 import { SCROLL_INDICATOR_INSET_FIX } from '../../constants/constantSettings'
 import { MAX_ADDRESS_CHARACTERS } from '../../constants/WalletAndCurrencyConstants'
 import { useAsyncEffect } from '../../hooks/useAsyncEffect'
 import { useHandler } from '../../hooks/useHandler'
-import { useMount } from '../../hooks/useMount'
 import { useUnmount } from '../../hooks/useUnmount'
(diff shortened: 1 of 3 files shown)


src/components/scenes/WcConnectionsScene.tsx
```diff
@@ -40,6 +41,7 @@ export const WcConnectionsScene = (props: Props) => {
   const styles = getStyles(theme)
   const [connections, setConnections] = React.useState<WcConnectionInfo[]>([])
   const [connecting, setConnecting] = React.useState(false)
+  const [sessionProposal, setSessionProposal] = React.useState<{ [uri: string]: Web3WalletTypes.SessionProposal }>({})
```
swansontec, line 44: Please use a map for this! Since the URL is attacker-controlled input, we absolutely cannot use it as an object key.

src/components/scenes/WcConnectScene.tsx
```diff
@@ -90,33 +85,19 @@ export const WcConnectScene = (props: Props) => {
...
-      return <SelectableRow title={lstrings.wc_confirm_select_wallet} onPress={handleWalletListModal} />
-    } else {
-      const walletNameStr = truncateString(walletName || '', MAX_ADDRESS_CHARACTERS)
-      const walletImage = (
-        <CryptoIconUi4 pluginId={wallet.currencyInfo.pluginId} tokenId={getTokenIdForced(account, wallet.currencyInfo.pluginId, selectedWallet.currencyCode)} />
-      )
-      const walletAddressStr = truncateString(walletAddress, MAX_ADDRESS_CHARACTERS, true)
-      return <SelectableRow icon={walletImage} subTitle={walletAddressStr} title={walletNameStr} onPress={handleWalletListModal} />
-    }
+    const walletNameStr = truncateString(walletName || '', MAX_ADDRESS_CHARACTERS)
```
swansontec, line 97: Get rid of the `|| ''`. The `useWalletName` hook always returns a valid string.

swansontec (changes requested):
Just two small-ish changes.

src/components/scenes/WcConnectionsScene.tsx
```diff
@@ -63,11 +63,15 @@ export const WcConnectionsScene = (props: Props) => {
...
     try {
-      let proposal: Web3WalletTypes.SessionProposal | undefined = sessionProposal[qrResult]
+      let proposal = sessionProposal.get(qrResult)
 
       if (proposal == null) {
-        proposal = await walletConnect.initSession(qrResult)
-        setSessionProposal({ [qrResult]: proposal })
+        const newProposal = await walletConnect.initSession(qrResult)
+        setSessionProposal(proposals => {
+          proposals.set(qrResult, newProposal)
```
swansontec, line 71: When storing a map in state, you need to avoid modifying the original object, but rather a copy:

```ts
const out = new Map(proposals)
out.set(qrResult, newProposal)
return out
```

Otherwise rendering could get messed up, as React doesn't see that anything changed.

swansontec (changes requested):

swansontec (approved):

PR #5102, opened 2024-06-18
Sam/floating-nav-bug

samholmes:
After many attempts to throwing many "setTimeout" hacks at this bug in
order to jolt the scene to update its styles, I gave up and found an
alternative solution. The reason for giving up was because setTimeout
had nasty side-effects and caused the animation to become glitchy in
various ways. I wasn't confident that even if I got the "setTimeout"
stategry working, that it wouldn't cause some other unforseen
regression.

So, the alternative is to disable then re-enable the keyboard controller
during scene transition. By doing this, it causes the internal react
context for the keyboard controller library to update. The context
updating causes all dependencies on the reanimated shared value to be
resestablished and this jolts the scene back into the correct state.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207590963112004


samholmes:
We've decided to just upgrade reanimated to an RC release: 5e151c544bfd85b5db75d8dc023c2fe72e6bea32

PR #5107, opened 2024-06-20
Add Sweep Private key flow

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-core-js/pull/599
https://github.com/EdgeApp/edge-currency-accountbased/pull/788

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1207435895725678


src/components/scenes/SweepPrivateKeySelectCryptoScene.tsx
```diff
@@ -0,0 +1,165 @@
...
+    )
+  })
+
+  const renderNextButton = React.useMemo(
+    () => (
+      <Fade noFadeIn={numSelected > 0} visible={numSelected > 0} duration={300}>
+        <View style={styles.bottomButton}>
+          <MainButton label={lstrings.string_next_capitalized} type="primary" marginRem={[0, 0, 0.75]} onPress={handleNext} />
+        </View>
+      </Fade>
```
swansontec, line 123: Just put this inline below, OR make it a normal function that just gets called as part of the render (`{ renderNextButton() }`), instead of a `useMemo`.

swansontec (changes requested):

src/components/scenes/SweepPrivateKeyCompletionScene.tsx
```diff
@@ -183,7 +143,7 @@ const SweepPrivateKeyCompletionComponent = (props: Props) => {
     )
   }, [done, navigation, styles.bottomButton])
 
-  const keyExtractor = useHandler((item: SweepPrivateKeyItem) => item.key)
+  const keyExtractor = useHandler((spendInfo: EdgeTransaction) => spendInfo.tokenId ?? '')
```
swansontec, line 146: Shouldn't this be called `tx`, if it's a transaction?

src/components/scenes/SweepPrivateKeyCompletionScene.tsx
```diff
@@ -152,19 +107,24 @@ const SweepPrivateKeyCompletionComponent = (props: Props) => {
     'SweepPrivateKeyCompletionComponent'
   )
 
-  const renderStatus = useHandler((item: SweepPrivateKeyItem) => {
+  const renderStatus = useHandler((spendInfo: EdgeTransaction) => {
```
swansontec, line 110: Shouldn't `spendInfo` be called `tx`, if it's a transaction?

src/components/scenes/SweepPrivateKeyCompletionScene.tsx
swansontec: Also, if we have `sweepPrivateKeyList` up above, that suggests we have multiple keys, and if we have multiple keys, then potentially we could have overlapping tokenId's (such as the `null` ones). Would it make sense for the key to be `tx.walletId + tx.tokenId` or something like that?

swansontec (changes requested):
Most of the fixes look great! I only found some tiny stuff in the final commit. Let's get this squashed together before doing the fixes, so we only have a commit or two on the end to look at.

swansontec (approved):

PR #5108, opened 2024-06-21
Jon/fix/balance-loading

Jon-edge:
https://github.com/EdgeApp/edge-react-gui/assets/90650827/5cf0c3e2-ce75-4a24-a386-5a1210aa7516

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207604271105270

### DIFF
--- a/src/components/ui4/BalanceCardUi4.tsx
+++ b/src/components/ui4/BalanceCardUi4.tsx
@@ -6,11 +6,12 @@ import IonIcon from 'react-native-vector-icons/Ionicons'
 import { toggleAccountBalanceVisibility } from '../../actions/LocalSettingsActions'
 import { getFiatSymbol } from '../../constants/WalletAndCurrencyConstants'
 import { useHandler } from '../../hooks/useHandler'
+import { useWatch } from '../../hooks/useWatch'
 import { formatNumber } from '../../locales/intl'
 import { lstrings } from '../../locales/strings'
 import { useDispatch, useSelector } from '../../types/reactRedux'
 import { NavigationBase } from '../../types/routerTypes'
-import { getTotalFiatAmountFromExchangeRates, removeIsoPrefix } from '../../util/utils'
+import { getTotalFiatAmountFromExchangeRates, removeIsoPrefix, zeroString } from '../../util/utils'
 import { AnimatedNumber } from '../common/AnimatedNumber'
 import { EdgeTouchableOpacity } from '../common/EdgeTouchableOpacity'
 import { TransferModal } from '../modals/TransferModal'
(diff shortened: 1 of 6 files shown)


src/components/ui4/BalanceCardUi4.tsx
```diff
@@ -64,26 +68,31 @@ export const BalanceCardUi4 = (props: Props) => {
...
   const animateNumber = fiatAmount < MAX_ANIMATED_AMOUNT
 
+  useAsyncEffect(
+    async () => {
+      await account.waitForAllWallets()
+      setIsWalletsLoaded(true)
+    },
+    [],
+    'BalanceCard'
+  )
```
swansontec, line 86: I don't think we need `useAsyncEffect` here. Instead, we can consolidate the `exchangeRatesReady` logic with the `isWalletsLoaded` logic into a single loop:

```ts
const activeWalletIds = useWatch(account, 'activeWalletIds')
const currencyWallets = useWatch(account, 'currencyWallets')
const currencyWalletErrors = useWatch(account, 'currencyWalletErrors')

const exchangeRatesReady = React.useMemo(() =>
  activeWalletIds.every(walletId => {
    // Ignore wallets that have crashed:
    if (currencyWalletErrors[walletId]) return true

    // Both the wallet and its rate need to be loaded:
    const wallet = currencyWallets[walletId]
    return wallet != null && exchangeRates[`${wallet.currencyInfo.currencyCode}_${defaultIsoFiat}`] != null
  })
, [activeWalletIds, currencyWallets, currencyWalletErrors])
```

src/hooks/useAccountWalletsSyncProgress.tsx
```diff
@@ -0,0 +1,82 @@
...
+          currencyWalletErrors[walletId] != null
+        )
+      }).length,
+    [currencyWalletErrors, progressMap, syncableWalletIds]
+  )
+
+  const progress = React.useMemo(
+    () => (syncableWalletIds.length === 0 ? 100 : 100 * (syncedWallets / syncableWalletIds.length)),
+    [syncableWalletIds.length, syncedWallets]
+  )
```
swansontec, line 67: Don't use a memo here. The `useMemo` hook needs to loop over its dependency array to see if anything has changed, which is going to cost more than just re-doing the division on each run. Math is really fast.

swansontec: Also, I request that this hook return a value in the range of 0-1, and have the name `useAccountSyncRatio`. That makes it clearer what the units are, and mirrors the fact that the wallets themselves operate on a 0-1 scale. Visual components like the sync bar can convert the range as necessary for display purposes (like 0% - 100%).

swansontec (changes requested):

paullinator (approved):

PR #5109, opened 2024-06-21
Forcefully close the keyboard before navigation to fix floating-nav bug

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1207510119361424

### DIFF
--- a/src/components/common/SceneWrapper.tsx
+++ b/src/components/common/SceneWrapper.tsx
@@ -2,7 +2,7 @@ import { useHeaderHeight } from '@react-navigation/elements'
 import { useFocusEffect, useIsFocused, useNavigation } from '@react-navigation/native'
 import * as React from 'react'
 import { useEffect, useMemo, useState } from 'react'
-import { StyleSheet, View, ViewStyle } from 'react-native'
+import { Keyboard, StyleSheet, View, ViewStyle } from 'react-native'
 import { useKeyboardHandler, useReanimatedKeyboardAnimation } from 'react-native-keyboard-controller'
 import Reanimated, { runOnJS, useAnimatedStyle } from 'react-native-reanimated'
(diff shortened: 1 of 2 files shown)


src/components/common/SceneWrapper.tsx
```diff
@@ -372,3 +375,38 @@ const SceneFooter = styled(View)({
...
+      if (Keyboard.isVisible()) {
+        // Prevent default behavior of leaving the screen
+        e.preventDefault()
+
+        Keyboard.dismiss()
+
+        const keyboardDidHideListener = Keyboard.addListener('keyboardWillHide', () => {
+          navigation.dispatch(e.data.action)
+          keyboardDidHideListener.remove()
+        })
```
swansontec, line 404: Put the `Keyboard.dismiss()` call *after* the subscription, so there's no way we miss the event.

src/components/common/SceneWrapper.tsx
```diff
@@ -372,3 +375,38 @@ const SceneFooter = styled(View)({
...
+ *
+ * This component listens for the 'beforeRemove' event on the navigation object
+ * and prevents the default behavior of leaving the screen if the keyboard is
+ * visible. It dismisses the keyboard and waits for the 'keyboardWillHide'
+ * event before dispatching the navigation action.
+ *
+ * @returns null
+ */
+function FloatingNavFixer() {
+  const navigation = useNavigation()
```
swansontec, line 391: We generally pass `navigation` as a prop instead of using the `useNavigation` hook. It's not safe to put `useNavigation` inside an Airship modal, and we don't want components to secretly blow up if you use them in the wrong place.

swansontec (changes requested):

swansontec (approved):

PR #5118, opened 2024-06-26
Paul/sentry fixes

paullinator:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207662633993331

### DIFF
--- a/src/actions/AccountReferralActions.ts
+++ b/src/actions/AccountReferralActions.ts
@@ -110,10 +110,14 @@ export function activatePromotion(installerId: string): ThunkAction<Promise<void
       return
     }
     if (!reply.ok) {
+      const text = await reply.text()
+      console.warn(`activatePromotion:fetch ${uri} ${text}`)
+      if (reply.status === 404) {
+        console.warn(`Invalid promotion code ${installerId}`)
+        return
+      }
       console.warn(`Referral server returned status code ${reply.status}`)
-    }
-    if (reply.status === 404) {
-      throw new Error(`Invalid promotion code ${installerId}`)
+      return
     }
     const clean = asServerTweaks(await reply.json())
 
--- a/src/actions/AccountReferralActions.ts
+++ b/src/actions/AccountReferralActions.ts
@@ -330,7 +330,7 @@ export async function validatePromoCardsInner(dataStore: EdgeDataStore, cards: M
 async function saveAccountReferral(state: RootState): Promise<void> {
   const { account } = state.core
   const { accountReferral } = state.account
-  await account.disklet.setText(ACCOUNT_REFERRAL_FILE, JSON.stringify(accountReferral))
+  await account?.disklet?.setText(ACCOUNT_REFERRAL_FILE, JSON.stringify(accountReferral))
 }
 
 /**
@@ -339,7 +339,7 @@ async function saveAccountReferral(state: RootState): Promise<void> {
 async function saveReferralCache(state: RootState): Promise<void> {
   const { account } = state.core
   const { referralCache } = state.account
-  await account.localDisklet.setText(REFERRAL_CACHE_FILE, JSON.stringify(referralCache))
+  await account?.localDisklet?.setText(REFERRAL_CACHE_FILE, JSON.stringify(referralCache))
 }
 
 /**
--- a/src/actions/LoginActions.tsx
+++ b/src/actions/LoginActions.tsx
@@ -31,6 +31,9 @@ import { readLocalAccountSettings } from './LocalSettingsActions'
 import { registerNotificationsV2, updateNotificationSettings } from './NotificationActions'
 import { showScamWarningModal } from './ScamWarningActions'
 
+const PER_WALLET_TIMEOUT = 5000
+const MIN_CREATE_WALLET_TIMEOUT = 20000
+
 function getFirstActiveWalletInfo(account: EdgeAccount): { walletId: string; currencyCode: string } {
   // Find the first wallet:
   const [walletId] = account.activeWalletIds
@@ -303,7 +306,8 @@ async function createCustomWallets(account: EdgeAccount, fiatCurrencyCode: strin
 
   // Actually create the wallets:
   const options = [...optionsMap.values()]
-  const results = await runWithTimeout(account.createCurrencyWallets(options), 20000, new Error(lstrings.error_creating_wallets)).catch(error => {
+  const timeoutMs = Math.max(options.length * PER_WALLET_TIMEOUT, MIN_CREATE_WALLET_TIMEOUT)
+  const results = await runWithTimeout(account.createCurrencyWallets(options), timeoutMs, new Error(lstrings.error_creating_wallets)).catch(error => {
     dispatch(logEvent('Signup_Wallets_Created_Failed', { error }))
     throw error
   })
--- a/src/components/modals/ScanModal.tsx
+++ b/src/components/modals/ScanModal.tsx
@@ -16,7 +16,7 @@ import { logActivity } from '../../util/logger'
 import { EdgeTouchableOpacity } from '../common/EdgeTouchableOpacity'
 import { QrPeephole } from '../common/QrPeephole'
 import { TextInputModal } from '../modals/TextInputModal'
-import { Airship, showDevError, showError, showWarning } from '../services/AirshipInstance'
+import { Airship, showDevError, showError, showToast } from '../services/AirshipInstance'
 import { checkAndRequestPermission } from '../services/PermissionsManager'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 import { EdgeText, Paragraph } from '../themed/EdgeText'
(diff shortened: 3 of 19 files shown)


src/actions/LoginActions.tsx
```diff
@@ -31,6 +31,9 @@ import { readLocalAccountSettings } from './LocalSettingsActions'
 import { registerNotificationsV2, updateNotificationSettings } from './NotificationActions'
 import { showScamWarningModal } from './ScamWarningActions'
 
+const PER_WALLET_TIMEOUT = 5000
+const MIN_CREATE_WALLET_TIMEOUT = 20000
```
samholmes, line 35: Isn't this a maximum timeout?

src/components/modals/ScanModal.tsx
samholmes, line 1: commit message: "Use showToast false for benign warnings" -> "Use showToast for benign warnings"

samholmes (approved):
approved with nits

src/actions/LoginActions.tsx
paullinator: not quite. its the minimum. we wait 5s per wallet being created but at minimum 20s. so we wait the following

1 wallet = 20s
2 = 20s
3 = 20s
4 = 20s
5 = 25s
6 = 30s

samholmes: Ah, then you should use `Math.max` on line 309.

src/actions/LoginActions.tsx
```diff
@@ -303,7 +306,8 @@ async function createCustomWallets(account: EdgeAccount, fiatCurrencyCode: strin
 
   // Actually create the wallets:
   const options = [...optionsMap.values()]
-  const results = await runWithTimeout(account.createCurrencyWallets(options), 20000, new Error(lstrings.error_creating_wallets)).catch(error => {
+  const timeoutMs = Math.min(options.length * PER_WALLET_TIMEOUT, MIN_CREATE_WALLET_TIMEOUT)
```
samholmes, line 309: Use `Math.max` to make sure to pick `MIN_CREATE_WALLET_TIMEOUT` over the first expression if the expression results in a lower value.

src/components/scenes/WcConnectionsScene.tsx
```diff
@@ -83,7 +84,13 @@ export const WcConnectionsScene = (props: Props) => {
...
       }
     } catch (error: any) {
-      showError(error)
+      if (error.message === UNSUPPORTED_WC_VERSION) {
+        showToast(lstrings.wc_unsupported_version)
+      } else if (error.message === NO_WALLETS_DAPP_REQUIREMENTS) {
+        showToast(lstrings.wc_no_wallets_dapp_requirements)
+      } else {
+        showError(error)
+      }
```
samholmes, line 93: Use `String(error)` instead of `error.message` sense `error: any` is not safe.

src/actions/AccountReferralActions.ts
```diff
@@ -110,10 +110,14 @@ export function activatePromotion(installerId: string): ThunkAction<Promise<void
       return
     }
     if (!reply.ok) {
-      console.warn(`Referral server returned status code ${reply.status}`)
-    }
-    if (reply.status === 404) {
-      throw new Error(`Invalid promotion code ${installerId}`)
+      const text = await reply.text()
+      console.warn(text)
```
samholmes, line 114: Shouldn't we prefix the warn with something so the dev knows where the log is coming from?

src/actions/AccountReferralActions.ts
```diff
@@ -110,10 +110,14 @@ export function activatePromotion(installerId: string): ThunkAction<Promise<void
...
+      if (reply.status === 404) {
+        console.error(`Invalid promotion code ${installerId}`)
+        return
+      }
+      console.error(`Referral server returned status code ${reply.status}`)
+      return
```
samholmes, line 120: Let's discuss the `console.error/warn` convention.

src/components/scenes/Fio/FioAddressRegisterScene.tsx
```diff
@@ -201,7 +201,7 @@ export class FioAddressRegister extends React.Component<Props, State> {
         try {
           await checkIsDomainPublic(fioPlugin, domain)
         } catch (e: any) {
-          showError(e.message)
+          showWarning(e.message, { trackError: false })
```
samholmes, line 204: `String(error)` is safer

src/components/scenes/RequestScene.tsx
```diff
@@ -520,11 +520,11 @@ export class RequestSceneComponent extends React.Component<Props & HookProps, St
     if (wallet?.id == null || currencyCode == null) return
 
     if (!this.props.isConnected) {
-      showError(lstrings.fio_network_alert_text)
+      showError(lstrings.fio_network_alert_text, { trackError: false })
```
samholmes, line 523: Why aren't we logging this error?

src/components/scenes/Fio/FioAddressRegisterScene.tsx
samholmes: Also, why aren't we logging this warning?

src/actions/AccountReferralActions.ts
```diff
@@ -330,7 +330,7 @@ export async function validatePromoCardsInner(dataStore: EdgeDataStore, cards: M
 async function saveAccountReferral(state: RootState): Promise<void> {
   const { account } = state.core
   const { accountReferral } = state.account
-  await account.disklet.setText(ACCOUNT_REFERRAL_FILE, JSON.stringify(accountReferral))
+  await account?.disklet?.setText(ACCOUNT_REFERRAL_FILE, JSON.stringify(accountReferral))
```
samholmes, line 333: These code paths shouldn't be activated if the state is invalid... What's causing these code paths to be reached before account is ready?

src/components/themed/SideMenu.tsx
```diff
@@ -152,6 +152,7 @@ export function SideMenuComponent(props: DrawerContentComponentProps) {
     const website = `${config.website}?af=appreferred_${refId}`
 
     const shareOptions = {
+      failOnCancel: false,
```
samholmes, line 155: Shouldn't we be showing the error in the catch for this `Share.open`?

src/components/scenes/TransactionsExportScene.tsx
```diff
@@ -385,6 +386,7 @@ class TransactionsExportSceneComponent extends React.PureComponent<Props, State>
     }
 
     await Share.open({
+      failOnCancel: false,
```
samholmes, line 389: Shouldn't we use `showError`?

samholmes (changes requested):

src/actions/AccountReferralActions.ts
paullinator: account object gets cleared when the core is unmounted, likely due to webview getting killed

src/components/themed/SideMenu.tsx
paullinator: No. This prevents an error when the user simply cancelled sharing. Not an error at all.

src/components/scenes/TransactionsExportScene.tsx
paullinator: No. This prevents an error when the user simply cancelled sharing. Not an error at all.

src/components/scenes/Fio/FioAddressRegisterScene.tsx
paullinator: Per our convo. this was going to be changed to remove the setTimeout and make checkFioAddress async. However the setTimeout is there to debounce the action. Fixing this makes this a much larger change.

src/components/scenes/TransactionsExportScene.tsx
samholmes: I mean shouldn't we use `showError` instead of `console.log(error)` in the catch?

src/components/themed/SideMenu.tsx
samholmes: See other comment reply.

src/components/scenes/Fio/FioAddressRegisterScene.tsx
samholmes: I'm less concerned about the setTimeout and more just about cleaning up the showWarning to instead get the error string from the return of `checkIsDomainPublic` to set it to the error state on the component and remove the try..catch

samholmes (changes requested):

src/actions/AccountReferralActions.ts
samholmes: If the `account` object is being set to the hack object, then that means we've logged out. If we've logged out, why are these functions being called? They shouldn't be running unless maybe this is caused by unresolved promises.. Eh, this is fine for now because this is just how things are with shared state and concurrency.

src/components/themed/SideMenu.tsx
paullinator: ah ok. got it. will change

src/components/scenes/TransactionsExportScene.tsx
paullinator: ah ok. got it. will change

src/components/scenes/Fio/FioAddressRegisterScene.tsx
paullinator: understood. done

samholmes (approved):

PR #5119, opened 2024-06-27
Remove react-native-check-version

paullinator:
Fixes red error due to server outage

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207673266723881


peachbits (approved):
Seems like overkill when removing the showError would be enough

paullinator:
> Seems like overkill when removing the showError would be enough

It would be overkill but the whole library is broken since it's the one that accesses an invalid out of date server. We'll move to using our in app cards instead

PR #5121, opened 2024-06-28
Fix Markets Scenes

Jon-edge:
Fix non-usd fiat setting support.

- Fall back to USD for non-CoinGecko-supported fiat currencies
- Fix various string formattings
- Fix errors on opening the Markets scenes

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-rates-server/pull/102

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207648985179833

### DIFF
--- a/src/components/charts/SwipeChart.tsx
+++ b/src/components/charts/SwipeChart.tsx
@@ -14,8 +14,6 @@ import { formatFiatString } from '../../hooks/useFiatText'
 import { useHandler } from '../../hooks/useHandler'
 import { formatDate } from '../../locales/intl'
 import { lstrings } from '../../locales/strings'
-import { getDefaultFiat } from '../../selectors/SettingsSelectors'
-import { useSelector } from '../../types/reactRedux'
 import { MinimalButton } from '../buttons/MinimalButton'
 import { FillLoader } from '../progress-indicators/FillLoader'
 import { showWarning } from '../services/AirshipInstance'
@@ -29,6 +27,7 @@ type CoinGeckoDataPair = number[]
 interface Props {
   assetId: string // The asset's 'id' as defined by CoinGecko
   currencyCode: string
+  fiatCurrencyCode: string
 }
 interface ChartDataPoint {
   x: Date
@@ -130,11 +129,9 @@ const reduceChartData = (chartData: ChartDataPoint[], timespan: Timespan): Chart
 const SwipeChartComponent = (params: Props) => {
   const theme = useTheme()
   const styles = getStyles(theme)
-  const { assetId, currencyCode } = params
+  const { assetId, currencyCode, fiatCurrencyCode } = params
 
   // #region Chart setup
-
-  const defaultFiat = useSelector(state => getDefaultFiat(state))
 
   const [chartData, setChartData] = React.useState<ChartDataPoint[]>([])
   const [cachedTimespanChartData, setCachedChartData] = React.useState<Map<Timespan, ChartDataPoint[] | undefined>>(
@@ -152,7 +149,7 @@ const SwipeChartComponent = (params: Props) => {
   const chartWidth = React.useRef(0)
   const chartHeight = React.useRef(0)
 
-  const fiatSymbol = React.useMemo(() => getFiatSymbol(defaultFiat), [defaultFiat])
+  const fiatSymbol = React.useMemo(() => getFiatSymbol(fiatCurrencyCode), [fiatCurrencyCode])
 
   // Min/Max Price Calcs
   const prices = React.useMemo(() => chartData.map(dataPoint => dataPoint.y), [chartData])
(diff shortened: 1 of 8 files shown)


src/util/network.ts
```diff
@@ -7,7 +7,7 @@ import { getVersion } from 'react-native-device-info'
 import { config } from '../theme/appConfig'
 import { asyncWaterfall, getOsVersion, shuffleArray } from './utils'
 const INFO_SERVERS = ['https://info1.edge.app', 'https://info2.edge.app']
-const RATES_SERVERS = ['https://rates1.edge.app', 'https://rates2.edge.app']
```
paullinator, line 10: Whoops!

src/hooks/useFiatText.ts
```diff
@@ -80,8 +80,12 @@ export const formatFiatString = (props: {
 }): string => {
   const { fiatAmount, minPrecision = 2, maxPrecision = 6, autoPrecision = true, noGrouping = false } = props
 
+  // Assume any spaces means this is some truncated '1.23 Bn' or '3.45 M' string
+  const magnitudeCode = fiatAmount.includes(' ') ? ` ${fiatAmount.split(' ')[1]}` : ''
+  const fiatAmountCleanedMag = magnitudeCode.length === 0 ? fiatAmount : fiatAmount.split(' ')[0]
+
```
paullinator, line 86: Since this routine runs in A LOT of places in the app, I'm going to be nitpicky on performance.

Optimize the operations 

```
const [fiatAmountCleanedMag, magnitudeCode] = fiatAmount.split(' ')
const magnitudeCodeStr = magnitudeCode == null ? '' : ` ${magnitudeCode}`
```

Then use `magnitudeCodeStr` in the return value

src/components/charts/SwipeChart.tsx
```diff
@@ -201,7 +205,7 @@ const SwipeChartComponent = (params: Props) => {
           } else {
             const unixNow = Math.trunc(new Date().getTime() / 1000)
             const fromParam = unixNow - queryFromTimeOffset
-            const fetchPath = sprintf(DATASET_URL_4S, assetId, defaultFiat, fromParam, unixNow)
+            const fetchPath = sprintf(DATASET_URL_4S, assetId, 'USD', fromParam, unixNow)
```
paullinator, line 208: This will yield an inaccurate chart as all the chart points will be in USD. Even though you convert them live to the current fiat, their relative x/y coordinates will be off relative to the actual fiat price (usually a small amount, but still wrong). 

I would sooner query Gecko for the fiat if supported, and if not fall back to USD.

paullinator (changes requested):

src/components/charts/SwipeChart.tsx
Jon-edge: This issue extends to our prior rates server changes. Historical-related prices are being scaled with current fiat-fiat prices

Hard-coding supported fiats for now, and making a task to fully fix the rates server in the same manner.

src/components/scenes/CoinRankingScene.tsx
```diff
@@ -30,6 +30,50 @@ const coinRanking: CoinRanking = { coinRankingDatas: [] }
 const QUERY_PAGE_SIZE = 30
 const LISTINGS_REFRESH_INTERVAL = 30000
 
+// HACK: To be moved to a CoinGecko query in the info or rates server.
+// Hard-coded "vs" currencies manually filtered for fiat currency codes for now.
+const COINGECKO_SUPPORTED_FIATS = [
```
paullinator, line 35: Optional fix: make an object for faster lookup

src/components/scenes/CoinRankingScene.tsx
```diff
@@ -54,8 +98,15 @@ const CoinRankingComponent = (props: Props) => {
   const styles = getStyles(theme)
   const { navigation } = props
 
-  const defaultIsoFiat = useSelector(state => `iso:${getDefaultFiat(state)}`)
-  const [lastUsedFiat, setLastUsedFiat] = useState<string>(defaultIsoFiat)
+  /** The user's fiat setting, falling back to USD if CoinGecko does not support
+   * it. */
+  const supportedFiatSetting = useSelector(state => {
```
paullinator, line 103: Get into the habit of using lightweight selectors since all selectors run on ALL redux changes. This should just be

```
const defaultFiat = useSelector(state => getDefaultFiat(state))
const supportedFiatSetting = COINGECKO_SUPPORTED_FIATS.includes(defaultFiat) ? defaultFiat : 'USD'
```

src/components/scenes/CoinRankingScene.tsx
```diff
@@ -168,8 +223,8 @@ const CoinRankingComponent = (props: Props) => {
             start += QUERY_PAGE_SIZE
           }
           setDataSize(coinRankingDatas.length)
-          if (lastUsedFiat !== defaultIsoFiat) {
-            setLastUsedFiat(defaultIsoFiat)
+          if (lastFetchedFiat !== supportedFiatSetting && COINGECKO_SUPPORTED_FIATS.includes(supportedFiatSetting)) {
```
paullinator, line 226: the `COINGECKO_SUPPORTED_FIATS.includes(supportedFiatSetting)` check shouldn't be needed as it either is supported or it's USD

paullinator (changes requested):

paullinator (approved):

PR #5123, opened 2024-07-02
Limit EdgeAnim rerenders

peachbits:
---
  - https://app.asana.com/0/0/1207699012295326

### DIFF
--- a/src/components/common/EdgeAnim.tsx
+++ b/src/components/common/EdgeAnim.tsx
@@ -19,6 +19,7 @@ import Animated, {
 } from 'react-native-reanimated'
 
 import { getDeviceSettings } from '../../actions/DeviceSettingsActions'
+import { matchJson } from '../../util/matchJson'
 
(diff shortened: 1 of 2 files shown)


src/components/common/EdgeAnim.tsx
```diff
@@ -135,4 +140,12 @@ const EdgeAnimInner = ({ children, disableAnimation, enter, exit, visible = true
   )
 }
 
-export const EdgeAnim = React.memo(EdgeAnimInner)
+const edgeAnimPropsAreEqual = (prevProps: Props, nextProps: Props): boolean => {
+  const { children: prevChildren, ...prevRest } = prevProps
```
paullinator, line 144: I would also short cut check each prop is === to each other and return true if so. Only if they are not equal AND the props that are not === are Objects do you use matchJson

paullinator (changes requested):

paullinator (approved):

PR #5143, opened 2024-07-10
Replace native alerts with showToast and ButtonsModal

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1207761072563617

### DIFF
--- a/src/actions/CreateWalletActions.tsx
+++ b/src/actions/CreateWalletActions.tsx
@@ -157,12 +157,12 @@ export function createAccountTransaction(
             paymentWallet.saveTxMetadata({ txid: edgeTransaction.txid, tokenId, metadata: edgeMetadata }).catch(err => console.warn(err))
             navigation.navigate('walletsTab', { screen: 'walletList' })
             setTimeout(() => {
-              Airship.show(bridge => (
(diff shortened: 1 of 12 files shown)


src/actions/CreateWalletActions.tsx
```diff
@@ -136,7 +135,9 @@ export function createAccountTransaction(
           if (error) {
             console.log(error)
             setTimeout(() => {
-              Alert.alert(lstrings.create_wallet_account_error_sending_transaction)
+              Airship.show(bridge => <ButtonsModal bridge={bridge} message={lstrings.create_wallet_account_error_sending_transaction} buttons={{}} />).catch(
```
swansontec, line 138: I talked to Madison about these `buttons={}` cases, and she agrees that it looks really weird. We should replace these cases with `buttons={{ ok: { label: lstrings.string_ok_cap } }}` or similar. There is no `onPress` handler needed - the modal just resolves to `'ok' | undefined` instead of `undefined` (not that we care either way).

swansontec (changes requested):

swansontec (approved):

PR #5145, opened 2024-07-11
Create `FioServices` 

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1207692345505863

### DIFF
--- a/src/components/services/FioService.ts
+++ b/src/components/services/FioService.ts
@@ -27,14 +27,14 @@ export const FioService = (props: Props) => {
   const dispatch = useDispatch()
 
   const expiredLastChecks = React.useRef<{ [fioName: string]: Date } | undefined>()
-  const expireReminderShown = useSelector(state => state.ui.fio.expireReminderShown)
+  const expireReminderShown = React.useRef(false)
   const expiredChecking = useSelector(state => state.ui.fio.expiredChecking)
   const walletsCheckedForExpired = useSelector(state => state.ui.fio.walletsCheckedForExpired)
   const fioWallets = useSelector(state => state.ui.wallets.fioWallets)
   const disklet = useSelector(state => state.core.disklet)
 
   const refreshNamesToCheckExpired = useHandler(async () => {
-    if (expireReminderShown) return
+    if (expireReminderShown.current) return
 
     if (fioWallets.length === 0) {
       return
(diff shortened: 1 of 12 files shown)


src/components/services/FioService.ts
```diff
@@ -0,0 +1,65 @@
...
+
+      if (namesToCheck.length !== 0) {
+        dispatch({ type: 'FIO/CHECKING_EXPIRED', data: true })
+        await dispatch(checkExpiredFioDomains(navigation, namesToCheck, fioWalletsById))
+      }
+    }, EXPIRE_CHECK_TIMEOUT)
+    task.start()
+
+    return () => task.stop()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
swansontec: We don't want to stop & start the task a bunch of times, but we also don't want the loop to be stale.

So, I would do this:

```ts
const refreshNamesToCheckExpired = useHandler(()= > {
  ... the body ...
})

React.useEffect(() => {
   makePeriodicTask(
     refreshNamesToCheckExpired,
     {
      onError: ....
     }
  })
  task.start()

  return () => task.stop()
})
```

That means the `refreshNamesToCheckExpired` function will have access to the fresh values, but the periodic task will keep its identity.

swansontec (changes requested):
We need a small change to the state management.

src/components/services/FioService.ts
```diff
@@ -0,0 +1,72 @@
...
+      dispatch({ type: 'FIO/CHECKING_EXPIRED', data: true })
+      await dispatch(checkExpiredFioDomains(navigation, namesToCheck, fioWalletsById))
+    }
+  })
+
+  // Check for expired FIO domains
+  React.useEffect(() => {
+    const task = makePeriodicTask(refreshNamesToCheckExpired, EXPIRE_CHECK_TIMEOUT, {
+      onError(e: unknown) {
+        console.error('refreshNamesToCheckExpired error:', e)
```
swansontec: I think we want a `showDevError` here - these ones will show up in develop, but not in staging or production.

src/components/services/FioService.ts
```diff
@@ -0,0 +1,72 @@
...
+      }
+    })
+    task.start()
+
+    return () => task.stop()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
swansontec: I think we can remove this ESLint exception, and just put `refreshNamesToCheckExpired` in the deps list now. The `useHandler` ensures that the value will never change.

swansontec (changes requested):
Just a few really tiny details now.

swansontec (approved):

PR #5146, opened 2024-07-11
Fix logout routing for all cases

Jon-edge:
Various hack-ish fixes for iOS and Android to resolve logout routing, and a fix to the getting started button for Android no longer working somehow

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207787615031270

### DIFF
--- a/src/actions/LoginActions.tsx
+++ b/src/actions/LoginActions.tsx
@@ -84,6 +84,9 @@ export function initializeAccount(navigation: NavigationBase, account: EdgeAccou
       // during account creation
       await readLocalAccountSettings(account)
 
+      // HACK: Android needs this for some reason for proper logout
+      navigation.pop()
+
       const newAccountFlow = async (navigation: NavigationProp<'createWalletSelectCrypto'>, items: WalletCreateItem[]) => {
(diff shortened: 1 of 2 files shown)


src/actions/LoginActions.tsx
```diff
@@ -84,6 +84,9 @@ export function initializeAccount(navigation: NavigationBase, account: EdgeAccou
       // during account creation
       await readLocalAccountSettings(account)
 
+      // HACK: Android needs this for some reason for proper logout
+      navigation.pop()
```
swansontec, line 88: Would it make more sense to do `navigation.replace` a few lines down? Then we don't need to `pop` the login scene, but can instead `replace` it with the main app.

Also, this way makes a lot of sense. We should search for everywhere we call `navigate('login' | 'gettingStarted' | 'edgeApp')`, and change all those places to `replace`. Then there three scenes become mutually exclusive, and there are no more hacks.

swansontec (approved):

PR #5149, opened 2024-07-16
Fix delete-account navigation

swansontec:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207787615031264

### DIFF
--- a/src/components/themed/SideMenu.tsx
+++ b/src/components/themed/SideMenu.tsx
@@ -33,7 +33,6 @@ import { EdgeTouchableOpacity } from '../common/EdgeTouchableOpacity'
 import { styled } from '../hoc/styled'
 import { ButtonsModal } from '../modals/ButtonsModal'
 import { ScanModal } from '../modals/ScanModal'
-import { LoadingSplashScreen } from '../progress-indicators/LoadingSplashScreen'
 import { Airship, showError } from '../services/AirshipInstance'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 import { TitleText } from '../text/TitleText'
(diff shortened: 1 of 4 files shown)


src/components/themed/SideMenu.tsx
```diff
@@ -484,14 +483,3 @@ const OuterView = styled(View)<{ insets: { top: number; bottom: number } }>(() =
...
-export function SideMenu(props: DrawerContentComponentProps) {
-  const { navigation } = props
-
-  const { loggedIn } = useSelector(state => state.core.account)
-  React.useEffect(() => {
-    if (!loggedIn) navigation.navigate('login')
-  }, [loggedIn, navigation])
-
-  return loggedIn ? <SideMenuComponent {...props} /> : <LoadingSplashScreen />
-}
```
samholmes, line 497: I don't understand the motivation for leaving the SideMenu mounted while on the login screen?

src/actions/LoginActions.tsx
```diff
@@ -272,7 +272,17 @@ export function logoutRequest(navigation: NavigationBase, nextLoginId?: string):
...
+    const experimentConfig = await getExperimentConfig()
+    navigation.navigate('login', { experimentConfig })
+
+    // The floating nav-bar hack in the SceneWrapper will call
+    // `preventDefault` on the logout navigation if the keyboard is up.
+    // The hack is supposed to retry the navigation, but that's not working
+    // right for some reason. Instead, do the navigation again ourselves:
+    if (Keyboard.isVisible()) {
+      setTimeout(() => navigation.navigate('login', { experimentConfig }), 1000)
+    }
```
samholmes, line 285: Try putting a setTimeout in the SceneWrapper where the hack is located to keep things tidy. If that works, then that's preferable. If not, then we can make do with this if it works.

samholmes (changes requested):

samholmes (approved):

src/components/themed/SideMenu.tsx
swansontec: When the bug happens, the user can get stuck inside the app. All the scenes get replaced by spinners, and the side menu also gets replaced by a spinner. So, there is no button the user can press to escape this situation, aside from killing the whole app.

By letting the side menu be the side menu, even if the account is in a logged-out state, the user can press the "logout" button and escape.

PR #5154, opened 2024-07-17
Move buy / sell scenes into the right tabs

swansontec:
We need to push the buy / sell SEPA scenes into the buy / sell tabs, so the `navigation.pop` can correctly remove them. Otherwise, calling `navigation.pop` will remove the plugin list scene itself, followed by removing the main app, because the scenes we want to remove are on a different tab.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207822127480559


samholmes (approved):
I approved, but I'd like to have a sidebar on this fix.

PR #5155, opened 2024-07-18
Fix login routing 2

Jon-edge:
Ensure we always only have either login or gui stacks mounted

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207848254943124

### DIFF
--- a/src/components/scenes/GettingStartedScene.tsx
+++ b/src/components/scenes/GettingStartedScene.tsx
@@ -97,7 +97,7 @@ export const GettingStartedScene = (props: Props) => {
 
   // Route helpers
   const visitPasswordScene = (): void =>
-    navigation.navigate('login', {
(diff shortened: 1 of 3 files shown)


src/components/scenes/GettingStartedScene.tsx
```diff
@@ -97,7 +97,7 @@ export const GettingStartedScene = (props: Props) => {
 
   // Route helpers
   const visitPasswordScene = (): void =>
-    navigation.navigate('login', {
+    navigation.replace('login', {
```
swansontec, line 100: This is good, but what about `handlePressSkip`?

swansontec (approved):
What you have is good, but it looks like you might have missed a spot.

PR #5175, opened 2024-08-01
Misc Bity fixes

Jon-edge:
image
image
image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206500978761526
  - https://app.asana.com/0/0/1207807100473757

### DIFF
--- a/src/plugins/gui/scenes/AddressFormScene.tsx
+++ b/src/plugins/gui/scenes/AddressFormScene.tsx
@@ -5,12 +5,12 @@ import { Platform, ScrollView, View, ViewStyle } from 'react-native'
 import { KeyboardAwareScrollView } from 'react-native-keyboard-aware-scroll-view'
 import Animated, { Easing, interpolateColor, useAnimatedStyle, useDerivedValue, useSharedValue, withTiming } from 'react-native-reanimated'
 
+import { SceneButtons } from '../../../components/buttons/SceneButtons'
 import { EdgeTouchableOpacity } from '../../../components/common/EdgeTouchableOpacity'
 import { SceneWrapper } from '../../../components/common/SceneWrapper'
 import { cacheStyles, Theme, useTheme } from '../../../components/services/ThemeContext'
 import { EdgeText } from '../../../components/themed/EdgeText'
 import { FilledTextInputRef } from '../../../components/themed/FilledTextInput'
-import { MainButton } from '../../../components/themed/MainButton'
 import { SceneHeader } from '../../../components/themed/SceneHeader'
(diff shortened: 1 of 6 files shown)


src/plugins/gui/scenes/SepaFormScene.tsx
```diff
@@ -61,21 +61,24 @@ export const SepaFormScene = React.memo((props: Props) => {
   )
 
   return (
-    <SceneWrapper hasNotifications>
+    <SceneWrapper hasTabs hasNotifications avoidKeyboard>
+      {() => (
+        <>
```
swansontec, line 74: What does this function wrapper do? I thought Sam fixed the `SceneWrapper` so the behavior was the same, no matter if you passed a function or not.

Assuming the behavior *does* differ, wouldn't the inner scene responsible for handling the safe area insets now? I don't see that code being added.

src/plugins/gui/scenes/AddressFormScene.tsx
```diff
@@ -235,6 +235,14 @@ export const AddressFormScene = React.memo((props: Props) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [searchResults])
 
+  // Unmount cleanup
+  React.useEffect(
+    () => () => {
+      onClose()
+    },
+    [onClose]
```
swansontec, line 243: If the caller passes a different `onClose` instance, this hook will call the previous `onClose` instance before installing the new instance, even if the scene is still mounted. This is kinda unsafe (even though we don't have a way to trigger the bug currently).

I would have expected this to be attached to the navigation event, like this:

```ts
React.useEffect(() => {
    return navigation.addListener('beforeRemove', () => {
      if (onClose != null) onClose()
    });
  },
  [navigation, onClose]
);
```

This will correctly handle `onClose` being updated, but without firing any callbacks until the scene is *actually* removed.

Same for SepaFormScene

swansontec (changes requested):
One question and one actual change needed.

src/plugins/gui/scenes/SepaFormScene.tsx
Jon-edge: Fine combed the styling with this in mind, basically just following implementors elsewhere.

Knowing what we know about `KeyboardAwareScrollView` and iOS, applied the same hack fix here to make sure we don't have any paste and dodgy autoscroll issues.

swansontec (approved):

PR #5185, opened 2024-08-05
Update flatList when exchange rates update

peachbits:
---
  - https://app.asana.com/0/0/1207957295481158

### DIFF
--- a/src/components/scenes/SweepPrivateKeyCalculateFeeScene.tsx
+++ b/src/components/scenes/SweepPrivateKeyCalculateFeeScene.tsx
@@ -200,6 +200,8 @@ const SweepPrivateKeyCalculateFeeComponent = (props: Props) => {
     'SweepPrivateKeyCalculateFeeComponent'
   )
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const extraData = React.useMemo(() => [], [transactionState, exchangeRates])
   const keyExtractor = useHandler((item: SweepPrivateKeyItem) => item.key)
 
   return (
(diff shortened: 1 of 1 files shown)


src/components/scenes/SweepPrivateKeyCalculateFeeScene.tsx
```diff
@@ -201,7 +201,7 @@ const SweepPrivateKeyCalculateFeeComponent = (props: Props) => {
         <FlatList
           automaticallyAdjustContentInsets={false}
           data={sweepPrivateKeyList}
-          extraData={transactionState}
+          extraData={exchangeRates}
```
samholmes, line 204: Include both exchangeRates and transactionState in the extraData as memoized state

samholmes (changes requested):

src/components/scenes/SweepPrivateKeyCalculateFeeScene.tsx
```diff
@@ -200,6 +200,8 @@ const SweepPrivateKeyCalculateFeeComponent = (props: Props) => {
     'SweepPrivateKeyCalculateFeeComponent'
   )
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const extraData = React.useMemo(() => [], [transactionState, exchangeRates])
```
samholmes, line 204: Why not just return `({ transactionState, exchangeRates})` in the memo function? I get that you're just trying to get the FlatList to re-render, but why not propagate the data just to avoid more questions when someone reads this code later?

samholmes (approved):

PR #5192, opened 2024-08-09
Deep link refactor

swansontec:
We can potentially split off the first commit as its own PR. This is the "bare minimum" needed to solve the duplicated recovery modals.

The other commits get deep links working again in general, for things such as payment links that need a logged-in account. This would be "nice to have" so marketing doesn't need to wait yet another month for their initiatives to get off the ground. Right now the links are unusable for their purposes.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207954986092378


src/components/services/DeepLinkingManager.ts
```diff
@@ -1,104 +1,106 @@
...
-       * Fires when the app is in foreground and receives a notification
-       * */
-      const unsubscribe = messaging().onMessage(remoteMessage => {
-        // do nothing for now except return the unsubscribe function
+      // Subscribe to various incoming events:
+      const linkingCleanup = Linking.addEventListener('url', event => {
+        handleDeepLink(event.url)
+      })
+      const messageCleanup = messaging().onMessage(message => {
+        handlePushMessage(message)
```
paullinator, line 82: Should we actually handle`messaging().onMessag`? The previous code didn't

paullinator: This should be handled by showing the notification as a dropdown first mimicing the OS notification, and only if the user taps on it should we call handlePushNotification. Otherwise the user would automatically get routed to a buy/sell/swap modal in the middle of whatever they're doing.

paullinator (changes requested):

paullinator (approved):

PR #5203, opened 2024-08-14
Switch to Bitpay payments for FIO names and domains

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1207426648754856

### DIFF
--- a/src/constants/WalletAndCurrencyConstants.ts
+++ b/src/constants/WalletAndCurrencyConstants.ts
@@ -152,7 +152,6 @@ interface SpecialCurrencyInfo {
   noChangeMiningFee?: boolean
   noMaxSpend?: boolean
   keysOnlyMode?: boolean
-  isPaymentProtocolSupported?: boolean
   isTransactionListUnsupported?: boolean
   isSplittingDisabled?: boolean
   isStakingSupported?: boolean
@@ -188,8 +187,7 @@ export const SPECIAL_CURRENCY_INFO: {
     displayBuyCrypto: false,
     dummyPublicAddress: '0x0d73358506663d484945ba85d0cd435ad610b0a0',
     initWalletName: lstrings.string_first_amoy_wallet_name,
-    isImportKeySupported: true,
-    isPaymentProtocolSupported: false
+    isImportKeySupported: true
   },
   bitcoin: {
     maxSpendTargets: UTXO_MAX_SPEND_TARGETS,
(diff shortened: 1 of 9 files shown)


src/constants/WalletAndCurrencyConstants.ts
```diff
@@ -199,17 +197,15 @@ export const SPECIAL_CURRENCY_INFO: {
...
+    isStakingSupported: true
   },
   bitcointestnet: {
     hasSegwit: true,
     maxSpendTargets: UTXO_MAX_SPEND_TARGETS,
     initWalletName: lstrings.string_first_bitcoin_testnet_wallet_name,
     chainCode: 'TESTBTC',
     displayBuyCrypto: true,
-    isImportKeySupported: true,
-    isPaymentProtocolSupported: true
```
swansontec, line 212: There are a bunch of currency codes that are not in the new `PAYMENT_PROTOCOL_MAP`:

- bitcointestnet
- bitcoinsv
- digibyte
- dash
- ...

I see that BitPay does not support these, and I don't know of other service providers that use the payment protocol, so there isn't much we can do about it. We are "losing support" for these formerly-supported coins, but I guess it's fine because they were never supported in the first place.

src/constants/WalletAndCurrencyConstants.ts
```diff
@@ -218,24 +214,21 @@ export const SPECIAL_CURRENCY_INFO: {
...
+    isStakingSupported: true
   },
   bitcoinsv: {
     maxSpendTargets: UTXO_MAX_SPEND_TARGETS,
     initWalletName: lstrings.string_first_bitcoin_sv_wallet_name,
     chainCode: 'BSV',
     keysOnlyMode: true,
-    isImportKeySupported: true,
-    isPaymentProtocolSupported: true
```
swansontec, line 230: bitcoinsv is not in the new `PAYMENT_PROTOCOL_MAP`. Not an immediate problem, since we can't send it anyhow, but perhaps we should keep the map up-to-date.

src/constants/WalletAndCurrencyConstants.ts
```diff
@@ -218,24 +214,21 @@ export const SPECIAL_CURRENCY_INFO: {
...
+    isImportKeySupported: true
   },
   digibyte: {
     maxSpendTargets: UTXO_MAX_SPEND_TARGETS,
     initWalletName: lstrings.string_first_digibyte_wallet_name,
     chainCode: 'DGB',
     displayBuyCrypto: true,
-    isImportKeySupported: true,
-    isPaymentProtocolSupported: true
```
swansontec, line 238: digibyte is not in the new `PAYMENT_PROTOCOL_MAP`.

swansontec (approved):

swansontec (approved):

PR #5205, opened 2024-08-15
Fix FIO Stake/Unstake tx display

Jon-edge:
Does NOT include fix for display issues after resync

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206787288256812

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - fixed: FIO OBT data not showing in received transaction memos
 - fixed: Fix broken Bity message signing
 - fixed: Return underLimit error for Simplex for very small quotes
+- fixed: FIO stake/unstake transactions displaying as sent/received transactions
 - removed: Removed pull-to-search feature from wallet list and transaction list scenes
 
 ## 4.11.0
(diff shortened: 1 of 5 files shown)


src/types/types.ts
```diff
@@ -5,6 +5,20 @@ import { LocaleStringKey } from '../locales/en_US'
...
+ * The `Include` type takes two generic parameters `T` and `U`. It returns a type that includes only those types from `T` that are also assignable to `U`.
+ *
+ * @example
+ * type A = string | number | boolean;
+ * type B = Include<A, number | boolean>; // Result is: number | boolean
+ *
+ * @typeparam T - The original union type.
+ * @typeparam U - The subset type to include.
+ */
+export type Include<T, U> = T extends U ? T : never
```
peachbits, line 20: This is already built into Typescript as Extract https://www.typescriptlang.org/docs/handbook/utility-types.html#extracttype-union

peachbits (changes requested):

src/types/types.ts
Jon-edge: Dropped this commit

peachbits (approved):

PR #5209, opened 2024-08-19
Always dismiss the keyboard when opening the SideMenu

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207994364310639

### DIFF
--- a/src/components/navigation/SideMenuButton.tsx
+++ b/src/components/navigation/SideMenuButton.tsx
@@ -1,5 +1,6 @@
 import { DrawerActions, useNavigation } from '@react-navigation/native'
 import * as React from 'react'
+import { Keyboard } from 'react-native'
 
 import { Fontello } from '../../assets/vector/index'
 import { useHandler } from '../../hooks/useHandler'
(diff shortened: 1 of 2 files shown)


src/components/navigation/SideMenuButton.tsx
```diff
@@ -12,6 +13,7 @@ export const SideMenuButton = () => {
   const theme = useTheme()
 
   const handlePress = useHandler(() => {
+    Keyboard.dismiss()
```
swansontec, line 16: This is fine, but it doesn't account for the swipe-to-open gesture.

You can use the [useDrawerStatus](https://reactnavigation.org/docs/7.x/drawer-navigator/#usedrawerstatus) hook inside the drawer itself:

```ts
const isOpen = useDrawerStatus() === 'open'
React.useEffect(()=> {
  if (isOpen) Keyboard.dismiss()
}, [isOpen])
```

That should work for both button press & swiping.

swansontec (changes requested):

src/components/navigation/SideMenuButton.tsx
swansontec: Actually, no, this is impossible. Swiping to open the drawer dismisses the keyboard.

swansontec (approved):

PR #5213, opened 2024-08-20
Fix Android TransactionList Scroll Gestures

Jon-edge:
Broken after removing the "drag down to search" feature.

Reinstate a gutted `refreshControl` to `Animated.FlatList` props

Several props are populated on this component, any of them could be causing some sort of Android-specific conflict.

Since we _do_ know that this broken when refreshControl was removed, just add it back with no functionality.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1208100508540956

### DIFF
--- a/src/components/scenes/TransactionListScene.tsx
+++ b/src/components/scenes/TransactionListScene.tsx
@@ -1,7 +1,7 @@
 import { EdgeCurrencyWallet, EdgeTokenId, EdgeTokenMap, EdgeTransaction } from 'edge-core-js'
 import { AssetStatus } from 'edge-info-server'
 import * as React from 'react'
-import { ListRenderItemInfo, View } from 'react-native'
+import { ListRenderItemInfo, Platform, RefreshControl, View } from 'react-native'
 import { getVersion } from 'react-native-device-info'
 import Animated from 'react-native-reanimated'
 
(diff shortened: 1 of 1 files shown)


src/components/scenes/TransactionListScene.tsx
```diff
@@ -187,6 +187,23 @@ function TransactionListComponent(props: Props) {
...
   //
 
+  /**
+   * HACK: This `RefreshControl` doesn't actually do anything visually or
+   * functionally noticeable besides making Android scroll gestures actually
+   * work for the parent `Animated.FlatList`
+   */
+  const refreshControl = React.useMemo(() => {
+    return (
+      <RefreshControl
```
swansontec, line 197: We could add `Platform.os !== 'android' ? undefined : <RefreshControl... />` to keep things cleaner on iOS

swansontec (approved):
Approved with the optional change of making it Android-only.

src/components/scenes/TransactionListScene.tsx
Jon-edge: A bit of cognative load with the negative logic. I'll do `Platform.OS === 'ios' ? undefined` so that its
easier to read while retaining your preference for putting the null/undef case first.

PR #5214, opened 2024-08-21
Jon/gui-guest-light-account

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device


src/components/themed/SideMenu.tsx
```diff
@@ -62,6 +62,8 @@ export function SideMenu(props: DrawerContentComponentProps) {
   // Maintain the list of usernames:
   const localUsers = useWatch(context, 'localUsers')
   const watchedUsername = useWatch(account, 'username')
+  const displayUsername =
+    watchedUsername == null ? sprintf(lstrings.guest_account_id_1s, account.rootLoginId.slice(account.rootLoginId.length - 3)) : watchedUsername
```
swansontec, line 66: Should this be in a common helper function, in case we ever change the logic?

src/components/themed/SideMenu.tsx
```diff
@@ -342,23 +346,23 @@ export function SideMenu(props: DrawerContentComponentProps) {
...
-              ) : null}
-            </View>
-            <View style={styles.rowBodyContainer}>
-              <TitleText style={styles.text}>{rowData.title}</TitleText>
-            </View>
-          </EdgeTouchableOpacity>
-        ))}
-      </ScrollView>
-    </View>
+      <View style={styles.rowsContainer}>
```
swansontec, line 348: It seems like the previous commit must have contained a whitespace error. Could you double-check the linter on "Split up giant JSX in the SideMenu render"?

swansontec (approved):
Please check these two cleanups, but everything looks like it will work.

I assume Madison / Michael gave their feedback on the "Guest Account (asdf)" wording, since that's not quite what I remember from the initial meetings. I'm guessing it went though some revisions in the meantime.

src/components/themed/SideMenu.tsx
Jon-edge: I considered it, but we'd need to wire it into loginui too somehow. A getter on
the account and userinfo seems most appropriate?

src/components/themed/SideMenu.tsx
Jon-edge: Yeah probably from staging partial diffs to make the commits more atomic, will check

PR #5218, opened 2024-08-23
2nd pass of unecessary showErrors on scenes

Jon-edge:
Often we are already showing some sort of error on the scene itself, or the error is attached to a simple info-only situation

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1207474082380292

### DIFF
--- a/src/components/scenes/Fio/FioRequestConfirmationScene.tsx
+++ b/src/components/scenes/Fio/FioRequestConfirmationScene.tsx
@@ -27,7 +27,7 @@ import { AddressModal } from '../../modals/AddressModal'
 import { ButtonsModal } from '../../modals/ButtonsModal'
 import { TextInputModal } from '../../modals/TextInputModal'
 import { EdgeRow } from '../../rows/EdgeRow'
-import { Airship, showError, showToast } from '../../services/AirshipInstance'
+import { Airship, showDevError, showError, showToast } from '../../services/AirshipInstance'
 import { cacheStyles, Theme, ThemeProps, useTheme } from '../../services/ThemeContext'
 import { ExchangedFlipInputAmounts } from '../../themed/ExchangedFlipInput2'
 import { SceneHeader } from '../../themed/SceneHeader'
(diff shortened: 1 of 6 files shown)


src/components/scenes/Fio/FioRequestConfirmationScene.tsx
```diff
@@ -211,7 +211,8 @@ export class FioRequestConfirmationConnected extends React.Component<Props, Stat
       } catch (error: any) {
         this.setState({ loading: false })
         this.resetSlider()
-        showError(
+        showError(lstrings.fio_request_error_header)
+        showDevError(
```
swansontec, line 215: Since this is `onConfirm`, it's a user-triggered action and we definitely should be showing the error. The lines below can be simplified to `error.json?.fields?.[0]?.error ?? ''`, which seems like a reasonable thing to show the user as well. So, let's do that cleanup and drop the `showDevError`

swansontec (changes requested):

src/components/scenes/Fio/FioRequestConfirmationScene.tsx
```diff
@@ -211,10 +211,7 @@ export class FioRequestConfirmationConnected extends React.Component<Props, Stat
       } catch (error: any) {
         this.setState({ loading: false })
         this.resetSlider()
-        showError(lstrings.fio_request_error_header)
-        showDevError(
-          `${lstrings.fio_request_error_header}. ${error.json && error.json.fields && error.json.fields[0] ? JSON.stringify(error.json.fields[0].error) : ''}`
-        )
+        showError(error.json?.fields?.[0]?.error ?? '')
```
swansontec, line 214: Let's do it this way:

```ts
const errorText = error.json?.fields?.[0]?.error ?? String(error)
showError(lstrings.fio_request_error_header + '. ' + errorText)
```

That way we still show something if JSON isn't present. Or, we could do this if we really don't want `lstrings.fio_request_error_header`:

```ts
showError(error.json?.fields?.[0]?.error ?? error)
```

swansontec (changes requested):

swansontec (approved):

PR #5225, opened 2024-08-30
Paul/fiat provider fixes

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1208187146318124


src/plugins/gui/providers/moonpayProvider.ts
```diff
@@ -295,16 +321,30 @@ export const moonpayProvider: FiatProviderFactory = {
...
 
         const fiatCode = removeIsoPrefix(params.fiatCurrencyCode).toLowerCase()
-        const paymentMethod = useIAch ? 'ach_bank_transfer' : 'credit_debit_card'
-        const url = `https://api.moonpay.com/v3/currencies/${cryptoCurrencyObj.code}/buy_quote/?apiKey=${apiKey}&quoteCurrencyCode=${cryptoCurrencyObj.code}&baseCurrencyCode=${fiatCode}&paymentMethod=${paymentMethod}&areFeesIncluded=true&${amountParam}`
+        let url
+        if (direction === 'buy') {
+          url = `https://api.moonpay.com/v3/currencies/${cryptoCurrencyObj.code}/buy_quote/?apiKey=${apiKey}&quoteCurrencyCode=${cryptoCurrencyObj.code}&baseCurrencyCode=${fiatCode}&paymentMethod=${paymentMethod}&areFeesIncluded=true&${amountParam}`
+        } else {
+          url = `https://api.moonpay.com/v3/currencies/${cryptoCurrencyObj.code}/sell_quote/?apiKey=${apiKey}&quoteCurrencyCode=${fiatCode}&payoutMethod=${paymentMethod}&areFeesIncluded=true&${amountParam}`
+        }
```
samholmes, line 321: Similar suggestion:

```ts
const url =
  direction === 'buy'
    ? `https://api.moonpay.com/v3/currencies/${cryptoCurrencyObj.code}/buy_quote/?apiKey=${apiKey}&quoteCurrencyCode=${cryptoCurrencyObj.code}&baseCurrencyCode=${fiatCode}&paymentMethod=${paymentMethod}&areFeesIncluded=true&${amountParam}`
    : `https://api.moonpay.com/v3/currencies/${cryptoCurrencyObj.code}/sell_quote/?apiKey=${apiKey}&quoteCurrencyCode=${fiatCode}&payoutMethod=${paymentMethod}&areFeesIncluded=true&${amountParam}`;
```

This isn't too terse for a ternary.

src/plugins/gui/util/initializeProviders.ts
```diff
@@ -14,6 +14,21 @@ export async function initializeProviders<T>(providerFactories: Array<FiatProvid
...
   const getTokenIdProvider = (pluginId: string, currencyCode: string) => getTokenId(account.currencyConfig[pluginId], currencyCode)
+  const getTokenIdFromContract = (params: { pluginId: string; contractAddress: string }) => {
+    const { pluginId, contractAddress } = params
+    const currencyInfo = account.currencyConfig[pluginId]
+    const edgeToken = Object.entries(currencyInfo.allTokens).find(([tokenId, token]) => {
+      const { networkLocation } = token
+      if (networkLocation == null) return false
+      if (networkLocation.contractAddress == null) return false
+      if (networkLocation.contractAddress.toLowerCase() === contractAddress.toLowerCase()) return true
+      return false
```
samholmes, line 25:
```ts
const { networkLocation } = token
      if (networkLocation == null) return false
      if (networkLocation.contractAddress == null) return false
      if (networkLocation.contractAddress.toLowerCase() === contractAddress.toLowerCase()) return true
      return false
```

->

```ts
return networkLocation?.contractAddress?.toLowerCase() === contractAddress.toLowerCase()
```

No?

src/plugins/gui/util/initializeProviders.ts
```diff
@@ -14,6 +14,21 @@ export async function initializeProviders<T>(providerFactories: Array<FiatProvid
...
+    })
+    if (edgeToken == null) return
+
+    // Return the tokenId which is the first entry in the tuple
+    return edgeToken[0]
```
samholmes, line 30: You can use `Object.values` instead of `Object.entries` to avoid the accessor (`[0]`).

src/plugins/gui/providers/kadoProvider.ts
```diff
@@ -533,12 +528,15 @@ export const kadoProvider: FiatProviderFactory = {
           throw new FiatProviderError({ providerId, errorType: 'paymentUnsupported' })
         }
 
+        const reverse = (direction === 'sell' && amountType === 'fiat') || (direction === 'buy' && amountType === 'crypto')
```
samholmes, line 531: Wouldn't the inverse of a typical quote be to "buy fiat" or "sell crypto"  (e.g. crypto -> fiat).

src/plugins/gui/providers/kadoProvider.ts
```diff
@@ -561,17 +559,31 @@ export const kadoProvider: FiatProviderFactory = {
...
-          fiatAmount = amount.toString()
+          if (amountType === 'crypto') {
+            cryptoAmount = exchangeAmount
+            fiatAmount = amount.toString()
+          } else {
+            fiatAmount = exchangeAmount
+            // The unitCount is 0 for sell reverse quotes but seems to be in
+            // the baseAmount.amount so use that
+            cryptoAmount = baseAmount.amount.toString()
+          }
```
samholmes, line 586:
```ts
const quote = asQuoteResponse(result).data
// ...
cryptoAmount = quote.baseAmount.amount.toString()
```

Removing the destructuring will keep the context of where `baseAmount` came from. 

It would also avoid the need to rename things like `minAmount` and `minUnit` when you can just use `quote.minValue.amount` and `quote.minValue.unit`. It's more to read where it's used, but it avoids needing to read paragraphs of destructuring and renaming.

samholmes (changes requested):
This all makes sense apart from some parts where I had questions (about the reverse quotes). Also, just some suggestions to make things read clearer.

src/plugins/gui/providers/moonpayProvider.ts
```diff
@@ -140,16 +149,18 @@ export const moonpayProvider: FiatProviderFactory = {
       partnerIcon,
       pluginDisplayName,
       getSupportedAssets: async ({ direction, paymentTypes, regionCode }): Promise<FiatProviderAssetMap> => {
+        const paymentType = PAYMENT_TYPE_MAP[paymentTypes[0]] ?? paymentTypes[0]
```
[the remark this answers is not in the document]
paullinator: moonpay will also have paypal and ach so I wanted it to be flexible.

src/plugins/gui/providers/moonpayProvider.ts
```diff
@@ -140,16 +149,18 @@ export const moonpayProvider: FiatProviderFactory = {
...
           throw new FiatProviderError({ providerId, errorType: 'paymentUnsupported' })
         }
         // Return nothing if paymentTypes are not supported by this provider
-        if (!paymentTypes.some(paymentType => allowedPaymentTypes[paymentType] === true))
-          throw new FiatProviderError({ providerId, errorType: 'paymentUnsupported' })
+        const assetMap = allowedCurrencyCodes[direction][paymentType]
+        if (assetMap == null) throw new FiatProviderError({ providerId, errorType: 'paymentUnsupported' })
 
-        if (isDailyCheckDue(lastChecked)) {
+        if (Object.keys(assetMap.crypto).length === 0 || isDailyCheckDue(lastChecked)) {
```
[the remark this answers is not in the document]
paullinator: there wouldn't be, but this check occurs before the query on the first mount

src/plugins/gui/providers/moonpayProvider.ts
```diff
@@ -172,6 +189,10 @@ export const moonpayProvider: FiatProviderFactory = {
             return assetMap
           }
           for (const currency of moonpayCurrencies) {
+            // if (direction === 'sell' && currency.isSellSupported !== true) {
+            //   continue
+            // }
```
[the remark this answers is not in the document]
paullinator: Once sell gets fully implemented, this should be re-enabled.

src/plugins/gui/providers/moonpayProvider.ts
```diff
@@ -283,10 +296,23 @@ export const moonpayProvider: FiatProviderFactory = {
...
+          maxFiat = fiatCurrencyObj.maxBuyAmount ?? fiatCurrencyObj.maxAmount ?? 0
+          minFiat = fiatCurrencyObj.minBuyAmount ?? fiatCurrencyObj.minAmount ?? Infinity
+          maxCrypto = cryptoCurrencyObj.maxBuyAmount ?? cryptoCurrencyObj.maxAmount ?? 0
+          minCrypto = cryptoCurrencyObj.minBuyAmount ?? cryptoCurrencyObj.minAmount ?? Infinity
+        } else {
+          maxFiat = fiatCurrencyObj.maxSellAmount ?? fiatCurrencyObj.maxAmount ?? 0
+          minFiat = fiatCurrencyObj.minSellAmount ?? fiatCurrencyObj.minAmount ?? Infinity
+          maxCrypto = cryptoCurrencyObj.maxSellAmount ?? cryptoCurrencyObj.maxAmount ?? 0
+          minCrypto = cryptoCurrencyObj.minSellAmount ?? cryptoCurrencyObj.minAmount ?? Infinity
+        }
```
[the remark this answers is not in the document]
paullinator: I hate to disagree here, but the code you wrong has a heavier cognative load for me.

src/plugins/gui/providers/moonpayProvider.ts
paullinator: Hmm. I think it's a toss up on this one. When it's a cointoss, I lean towards code where I can set break points

src/plugins/gui/util/initializeProviders.ts
paullinator: Ok. for sure this is a good change.

src/plugins/gui/util/initializeProviders.ts
paullinator: Nope, the tokenId is the key, not the value.

src/plugins/gui/providers/kadoProvider.ts
paullinator: normal quotes let you specify fiat amount to pay for a buy and crypto amount to receive for sell

reverse means

specify the crypto amount to buy and the fiat amount to sell.

src/plugins/gui/providers/kadoProvider.ts
paullinator: I agree. I left the `minAmount` and `minUnit` as is to reduce modification of the existing code. For the baseAmount, the destructuring seems harmless as there is no ambiguity as to where it came from.

src/plugins/gui/providers/moonpayProvider.ts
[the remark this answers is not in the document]
samholmes: Oh, isn't that handled by `isDailyCheckDue` though (`isDailyCheckDue(lastChecked) === false` on first mount).

src/plugins/gui/providers/moonpayProvider.ts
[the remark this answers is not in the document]
samholmes: It's okay to disagree 😉

src/plugins/gui/util/initializeProviders.ts
samholmes: Ah.

src/util/CurrencyInfoHelpers.ts
```diff
@@ -14,6 +14,57 @@ export function isKeysOnlyPlugin(pluginId: string): boolean {
...
+      console.warn(`findTokenIdByNetworkLocation: Unsupported networkLocation key: ${key}`)
+      return
+    }
+  }
+
+  for (const tokenId in allTokens) {
+    const edgeToken = allTokens[tokenId]
+    if (edgeToken?.networkLocation?.contractAddress?.toLowerCase() === networkLocation?.contractAddress?.toLowerCase()) {
+      return tokenId
+    }
```
samholmes, line 64: It looks like `contractAddress` is assumed to be the key within the query param `params.networkLocation`? Did you mean to do this?

If not, and instead you meant to use whatever keys are defined on `params.networkLocation`, then it seems like that could skip the whole "first key" validation song and dance:

```ts
export const findTokenIdByNetworkLocation = (params: FindTokenParams): EdgeTokenId | undefined => {
  const { networkLocation } = params
  let allTokens: EdgeTokenMap
  if ('allTokens' in params) {
    allTokens = params.allTokens
  } else {
    allTokens = params.account.currencyConfig[params.pluginId]?.allTokens
    if (allTokens == null) return
  }

  // eslint-disable-next-line no-labels
  outer: for (const tokenId in allTokens) {
    const edgeToken = allTokens[tokenId]
    for (const key in networkLocation) {
      const left = edgeToken?.networkLocation?.[key]
      if (typeof left !== 'string') {
        // Perhaps this should be an error throw?
        console.warn(`findTokenIdByNetworkLocation: Unsupported networkLocation key: ${key}`)
        return
      }
      const right = networkLocation[key]
      if (typeof right !== 'string') {
        // Perhaps this should be an error throw?
        console.warn(`findTokenIdByNetworkLocation: Unsupported networkLocation key: ${key}`)
        return
      }

      if (left.toLowerCase() !== right.toLowerCase()) {
        // eslint-disable-next-line no-labels
        continue outer
      }
    }
    // Found because it passed the checks in the inner loop
    return tokenId
  }
}
```

Maybe I misunderstand the function implementation?

samholmes (changes requested):
I got tripped up on the new `findTokenIdByNetworkLocation`.

src/util/CurrencyInfoHelpers.ts
paullinator: Good catch. Let me update as I had intended and just never completed.

samholmes (approved):

PR #5226, opened 2024-08-30
Implement SupportedAssetStore for bityProvider

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device
  
# Example JSON

Example is support store for bity.

```json
{
  "*": {
    "AT": true,
    "BE": true,
    "BG": true,
    "CH": true,
    "CZ": true,
    "DK": true,
    "EE": true,
    "FI": true,
    "FR": true,
    "DE": true,
    "GR": true,
    "HU": true,
    "IE": true,
    "IT": true,
    "LV": true,
    "LT": true,
    "LU": true,
    "NL": true,
    "PL": true,
    "PT": true,
    "RO": true,
    "SK": true,
    "SI": true,
    "ES": true,
    "SE": true,
    "HR": true,
    "LI": true,
    "NO": true,
    "SM": true,
    "GB": true,
    "*": {
      "*": {
        "sepa": {
          "bitcoin:null": true,
          "ethereum:null": true,
          "ethereum:a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48": true,
          "ethereum:dac17f958d2ee523a2206206994597c13d831ec7": true
        }
      }
    }
  },
  "sell": { "*": { "iso:CHF": { "sepa": true }, "iso:EUR": { "sepa": true } } }
}
```

---
  - https://app.asana.com/0/0/1208002585478670

### DIFF
--- a/src/__tests__/utils/ProviderSupportStore.test.ts
+++ b/src/__tests__/utils/ProviderSupportStore.test.ts
@@ -1,0 +1,564 @@
+import { describe, it } from '@jest/globals'
+
+import {
+  CryptoKey,
+  FiatProviderAssetMapQuery,
+  PaymentKey,
+  ProviderSupportObject,
+  ProviderSupportStore,
+  queryNodes
+} from '../../plugins/gui/providers/ProviderSupportStore'
+
+describe('ProviderSupportStore', () => {
+  const generalStore = makeGeneralStoreFixture()
+
+  it('toJsonObject, toJson, fromJsonObject, fromJson', () => {
+    const obj = generalStore.toJsonObject()
+    const json = generalStore.toJson()
+
+    const expectedObj = {
+      buy: true,
+      '*': {
+        US: true,
+        'US:CA': true,
+        UK: true,
+        '*': {
+          'iso:USD': true,
+          'iso:CAD': true,
+          'iso:GBP': true,
+          '*': {
+            ach: true,
+            sepa: true,
+            credit: true,
+            '*': {
+              'ethereum:null': true,
+              'ethereum:USDC': true,
+              'bitcoin:null': true
+            }
+          }
+        }
+      }
+    }
+    expect(obj).toEqual(expectedObj)
+
+    const expectedJson = JSON.stringify(expectedObj)
+    expect(json).toBe(expectedJson)
+
+    generalStore.fromJsonObject(obj)
+    expect(generalStore.toJsonObject()).toEqual(expectedObj)
+
+    generalStore.fromJson(json)
+    expect(generalStore.toJson()).toEqual(expectedJson)
+  })
+
+  it('toJsonObject', () => {
+    const obj = generalStore.toJsonObject()
+
+    expect(obj).toEqual({
+      buy: true,
+      '*': {
+        US: true,
+        'US:CA': true,
+        UK: true,
+        '*': {
+          'iso:USD': true,
+          'iso:CAD': true,
+          'iso:GBP': true,
+          '*': {
+            ach: true,
+            sepa: true,
+            credit: true,
+            '*': {
+              'ethereum:null': true,
+              'ethereum:USDC': true,
+              'bitcoin:null': true
+            }
+          }
+        }
+      }
+    })
+  })
+
+  it('isSupported -> true', () => {
+    expect(generalStore.is.direction('buy').supported).toBe(true)
+    expect(generalStore.is.direction('*').region('US').supported).toBe(true)
+    expect(generalStore.is.direction('buy').region('US').supported).toBe(true)
+    expect(generalStore.is.direction('buy').region('US').fiat('iso:USD').supported).toBe(true)
+    expect(generalStore.is.direction('buy').region('US').fiat('iso:USD').payment('ach').supported).toBe(true)
+    expect(generalStore.is.direction('buy').region('US').fiat('iso:USD').payment('ach').crypto('ethereum:null').supported).toBe(true)
+    expect(generalStore.is.direction('*').region('US').fiat('iso:USD').payment('ach').crypto('ethereum:null').supported).toBe(true)
+    expect(generalStore.is.direction('buy').region('*').fiat('iso:USD').payment('ach').crypto('ethereum:null').supported).toBe(true)
+    expect(generalStore.is.direction('*').region('*').fiat('iso:USD').payment('ach').crypto('ethereum:null').supported).toBe(true)
+    expect(generalStore.is.direction('buy').region('US').fiat('*').payment('ach').crypto('ethereum:null').supported).toBe(true)
+    expect(generalStore.is.direction('*').region('US').fiat('*').payment('ach').crypto('ethereum:null').supported).toBe(true)
+    expect(generalStore.is.direction('buy').region('*').fiat('*').payment('ach').crypto('ethereum:null').supported).toBe(true)
+    expect(generalStore.is.direction('*').region('*').fiat('*').payment('ach').crypto('ethereum:null').supported).toBe(true)
+    expect(generalStore.is.direction('buy').region('US').fiat('iso:USD').payment('*').crypto('ethereum:null').supported).toBe(true)
+    expect(generalStore.is.direction('*').region('US').fiat('iso:USD').payment('*').crypto('ethereum:null').supported).toBe(true)
+    expect(generalStore.is.direction('buy').region('*').fiat('iso:USD').payment('*').crypto('ethereum:null').supported).toBe(true)
+    expect(generalStore.is.direction('*').region('*').fiat('iso:USD').payment('*').crypto('ethereum:null').supported).toBe(true)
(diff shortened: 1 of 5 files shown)


paullinator (changes requested):
Request architectural change per phone convo

samholmes:
@paullinator Our phone conversation resulted in the task to come up with the proposal for the ProviderSupportStore interface. I followed up later that day with the following interface proposals:

1. [Provider Asset Store (using Maps)](https://gist.github.com/samholmes/394511d195dfb57d6b8f14c4435c864d)
2. [Provider Asset Store (using objects)](https://gist.github.com/samholmes/373b8afc5dae80eb1012005f9ce70bd3)

_Note: by using Map’s we can use keys like EdgeTokenId (which are not strictly a string type). This will require serialization to/from JSON for document storage on the info server._

Could you confirm that these proposals are approved to implement?

src/util/cleaners.ts
```diff
@@ -26,3 +26,77 @@ export const asMaybeContractLocation = asMaybe(
...
+        out[key] = asT(raw[key])
+      } catch (error) {
+        throw locateError(error, '[' + JSON.stringify(key) + ']', 0)
+      }
+    }
+    return out
+  }
+}
+
+export const asObjectInOrBoolean = <K extends string | number | symbol, T>(asKey: Cleaner<K>, asT: Cleaner<T>) => asEither(asObjectIn(asKey, asT), asBoolean)
```
paullinator, line 51: Really you want `asObjectInOrTrue` since you can't create a `false` entry using the API anyway

src/__tests__/utils/ProviderSupportStore.test.ts
```diff
@@ -0,0 +1,362 @@
...
+            ach: true,
+            sepa: true,
+            credit: true,
+            '*': {
+              'ethereum:null': true,
+              'ethereum:USDC': true,
+              'bitcoin:null': true
+            }
+          }
+        }
```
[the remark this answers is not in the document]
paullinator: Better to just not allow false since only `fromJson` would allow that to happen, not the normal api. see comments below

src/plugins/gui/providers/ProviderSupportStore.ts
```diff
@@ -0,0 +1,300 @@
...
+    if (regionNodes.length === 0) return fiatProviderAssetMap
+
+    const fiatNodes = queryNodes(regionNodes, query.region)
+    if (fiatNodes.length === 0) return fiatProviderAssetMap
+
+    for (const fiatNode of fiatNodes) {
+      for (const fiatKey of fiatNode.keys()) {
+        if (fiatKey === '*') continue
+        // Add fiat to fiat map:
+        fiatProviderAssetMap.fiat[fiatKey] = true
```
paullinator, line 141: The fiat values aren't filtered by the payment types

paullinator (changes requested):

src/plugins/gui/providers/bityProvider.ts
```diff
@@ -370,18 +364,30 @@ export const bityProvider: FiatProviderFactory = {
...
         // Return nothing if 'sepa' is not included in the props
         if (!paymentTypes.includes(supportedPaymentType)) throw new FiatProviderError({ providerId, errorType: 'paymentUnsupported' })
 
         const response = await fetch(`https://exchange.api.bity.com/v2/currencies`).catch(e => undefined)
         if (response == null || !response.ok) {
           console.error(`Bity getSupportedAssets response error: ${await response?.text()}`)
-          return allowedCurrencyCodes[direction]
+          return supportedAssets.getFiatProviderAssetMap({
+            direction,
+            region: `*`,
```
paullinator, line 388: This should be the actual region

src/plugins/gui/providers/bityProvider.ts
```diff
@@ -390,32 +396,45 @@ export const bityProvider: FiatProviderFactory = {
           bityCurrencies = asBityCurrencyResponse(result).currencies
         } catch (error: any) {
           console.error(error)
-          return allowedCurrencyCodes[direction]
+          return supportedAssets.getFiatProviderAssetMap({
+            direction,
+            region: `*`,
```
paullinator, line 401: This should be the actual region

src/plugins/gui/providers/bityProvider.ts
```diff
@@ -370,18 +364,30 @@ export const bityProvider: FiatProviderFactory = {
...
         if (!paymentTypes.includes(supportedPaymentType)) throw new FiatProviderError({ providerId, errorType: 'paymentUnsupported' })
 
         const response = await fetch(`https://exchange.api.bity.com/v2/currencies`).catch(e => undefined)
         if (response == null || !response.ok) {
           console.error(`Bity getSupportedAssets response error: ${await response?.text()}`)
-          return allowedCurrencyCodes[direction]
+          return supportedAssets.getFiatProviderAssetMap({
+            direction,
+            region: `*`,
+            payment: supportedPaymentType
```
paullinator, line 389: This should be the actually passed in `paymentTypes[0]`

src/plugins/gui/providers/bityProvider.ts
```diff
@@ -390,32 +396,45 @@ export const bityProvider: FiatProviderFactory = {
           bityCurrencies = asBityCurrencyResponse(result).currencies
         } catch (error: any) {
           console.error(error)
-          return allowedCurrencyCodes[direction]
+          return supportedAssets.getFiatProviderAssetMap({
+            direction,
+            region: `*`,
+            payment: supportedPaymentType
```
paullinator, line 402: This should be the actually passed in `paymentTypes[0]`

src/plugins/gui/providers/bityProvider.ts
```diff
@@ -390,32 +396,45 @@ export const bityProvider: FiatProviderFactory = {
...
+            const list = noKycCurrencyCodes[direction].crypto[pluginId]
+            if (list == null || !list.some(t => t.tokenId === tokenId)) {
+              continue
             }
-          }
 
-          // Unhandled combination not caught by cleaner. Skip to be safe.
-          if (!isAddCurrencySuccess) console.log('Unhandled Bity supported currency: ', currency)
+            const crypto = `${pluginId}:${tokenId}`
+            supportedAssets.add.direction('*').region('*').fiat('*').payment(supportedPaymentType).crypto(crypto)
```
paullinator, line 427: direction should be the passed in `direction`

src/plugins/gui/providers/bityProvider.ts
```diff
@@ -390,32 +396,45 @@ export const bityProvider: FiatProviderFactory = {
...
+          })
         }
+
         for (const currency of bityCurrencies) {
-          let isAddCurrencySuccess = false
           if (currency.tags.length === 1 && currency.tags[0] === 'fiat') {
-            allowedCurrencyCodes[direction].fiat['iso:' + currency.code.toUpperCase()] = currency
-            isAddCurrencySuccess = true
+            const fiatCurrencyCode = 'iso:' + currency.code.toUpperCase()
+            supportedAssets.add.direction(direction).region('*').fiat(fiatCurrencyCode).payment(supportedPaymentType)
```
paullinator, line 409: You're insistently using `*` for direction vs the actual passed in `direction`. seems like this will cause issues since you'll add the direction on the first call to `getSupportedAssets` but on a subsequent call for the opposite direction, you'll return false vs re-querying and filling in the cache

paullinator (changes requested):

src/__tests__/utils/ProviderSupportStore.test.ts
```diff
@@ -0,0 +1,362 @@
...
+    const json = generalStore.toJson()
+
+    const expectedObj = {
+      buy: true,
+      '*': {
+        US: true,
+        'US:CA': true,
+        UK: true,
+        '*': {
+          'iso:USD': true,
```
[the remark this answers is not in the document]
samholmes: So long as we never have a `"iso"` entry, then `"iso:USD"` is treated the same as `"isoUSD"`. The only difference is that you could query against `"iso:*"` or `"iso"`. We could consider using another delimiter to group, but `:` matches the same character in JSON which makes it a nice grouping character. We could keep the group character as `:` and require that plugins use `_` instead, so they'd add `iso_USD`.

src/__tests__/utils/ProviderSupportStore.test.ts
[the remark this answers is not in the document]
samholmes: The motivation to having `false`: ability to disable entries in the info server temporarily. But this is a moot case since there is no way to disable an entire object structure without removing it. So this "feature" is only applicable to leaf nodes. I no longer think the benefit for this feature is worth the tradeoff in ambiguity, so I'll remove `false` from the data type.

paullinator (approved):

PR #5231, opened 2024-09-05
Sam/kiln ada fix2

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1208229808315282

### DIFF
--- a/src/plugins/stake-plugins/generic/policyAdapters/CardanoKilnAdaptor.ts
+++ b/src/plugins/stake-plugins/generic/policyAdapters/CardanoKilnAdaptor.ts
@@ -1,11 +1,13 @@
 import { eq, gt, sub } from 'biggystring'
 import { EdgeCurrencyWallet, EdgeTransaction } from 'edge-core-js'
 
+import { lstrings } from '../../../../locales/strings'
+import { HumanFriendlyError } from '../../../../types/HumanFriendlyError'
 import { infoServerData } from '../../../../util/network'
 import { AssetId, ChangeQuote, PositionAllocation, QuoteAllocation, StakePosition } from '../../types'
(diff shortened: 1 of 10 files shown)


src/plugins/stake-plugins/generic/policyAdapters/CardanoKilnAdaptor.ts
```diff
@@ -71,7 +73,21 @@ export const makeCardanoKilnAdapter = (policyConfig: StakePolicyConfig<CardanoPo
         throw new Error('Insufficient funds')
       }
 
-      const stakeTransaction = await kiln.adaStakeTransaction(walletAddress, adapterConfig.poolId, accountId)
+      const result = await kiln.adaStakeTransaction(walletAddress, adapterConfig.poolId, accountId).catch(error => {
+        if (error instanceof Error) return error
+        throw error
+      })
+      if (result instanceof KilnError) {
```
peachbits, line 80: The KilnError class seems unnecessary since you don't need the additional .error string. You could just use an Error and .message

peachbits (approved):
I'll leave it up to you if the seemingly unnecessary KilnError or not. Otherwise it's approved.

src/plugins/stake-plugins/generic/policyAdapters/CardanoKilnAdaptor.ts
samholmes: Not necessarily. I don't know what Kiln will throw at us in the future. By modeling the error after their error response, we can handle anything they throw at us.

PR #5245, opened 2024-09-12
Fix GLIF Infinity Pool staking

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208285420971625

### DIFF
--- a/src/plugins/stake-plugins/generic/pluginInfo/filecoinGlifpool.ts
+++ b/src/plugins/stake-plugins/generic/pluginInfo/filecoinGlifpool.ts
@@ -17,8 +17,8 @@ const filecoinPolicyConfig: Array<StakePolicyConfig<GlifInfinityPoolAdapterConfi
     adapterConfig: {
       type: 'glif-infinity-pool',
       rpcProviderUrls: [`https://api.node.glif.io/rpc/v0`],
-      poolContractAddress: '0x43dAe5624445e7679D16a63211c5ff368681500c', // Pool 0
-      simpleRampContractAddress: '0x8147AccE69d711bcED176a0b7a029Ff54800d930'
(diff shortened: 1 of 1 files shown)


src/plugins/stake-plugins/generic/util/EdgeWalletSigner.ts
```diff
@@ -74,13 +74,17 @@ export class EdgeWalletSigner extends ethers.Signer {
     }
 
     const data = await transaction.data
+
     const memoHexValue = data == null ? undefined : typeof data === 'string' ? data.replace('0x', '') : base16.stringify(data)
     const memos: EdgeMemo[] = memoHexValue == null ? [] : [{ type: 'hex', value: memoHexValue }]
+
+    const customData = await transaction.customData
```
swansontec, line 81: This is weird. I would have expected `transaction.customDataPromise` as a clearer name, but I guess it's out of scope for this PR.

swansontec (approved):

src/plugins/stake-plugins/generic/util/EdgeWalletSigner.ts
swansontec: This whole first commit doesn't seem to be used, in general. Since we are super-late in staging, please move it to a new "Staking cleanups" PR on develop, so we don't take any extra risk. Or, if it's needed for some reason, please explain.

samholmes: The type is "Promise<T> | T ...", so I have to resolve it to remove the promise type

PR #5248, opened 2024-09-13
Jon/dropdown-update

Jon-edge:
image
image

Unable to clip an upper border's shadow for the `SideMenu.` We already fade out the background when the username list is dropped, anyway, so skipping shadows on this component for now. 

Given the original spec didn't really consider full-width dropdowns, we may need to reduce the shadow radius, anyway, in which case a simple diagonal shadow offset is passable for a side menu username dropdown.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1206926834012048

### DIFF
--- a/src/components/common/ExpandableList.tsx
+++ b/src/components/common/ExpandableList.tsx
@@ -1,43 +1,38 @@
-/**
- * IMPORTANT: Changes in this file MUST be synced between edge-react-gui and
- * edge-login-ui-rn!
- */
-
 import * as React from 'react'
-import { ScrollView, View } from 'react-native'
+import { Platform, ScrollView, View, ViewStyle } from 'react-native'
+import { AirshipBridge } from 'react-native-airship'
+import { ShadowedView } from 'react-native-fast-shadow'
 import { cacheStyles } from 'react-native-patina'
 import Animated, { Easing, useAnimatedStyle, useDerivedValue, useSharedValue, withTiming } from 'react-native-reanimated'
 
 import { SCROLL_INDICATOR_INSET_FIX } from '../../constants/constantSettings'
+import { useAsyncEffect } from '../../hooks/useAsyncEffect'
 import { GradientFadeOut } from '../modals/GradientFadeout'
+import { Airship } from '../services/AirshipInstance'
 import { Theme, useTheme } from '../services/ThemeContext'
 
 interface Props {
   isExpanded: boolean
   items: React.ReactNode[]
-  itemHeight: number
-  /** Defaults to 4.75. */
+  /** Defaults to 4.75 */
   maxDisplayedItems?: number
-  /** Defaults to full scene width. If specified, is left-justified with
-   * built-in 0.5rem margins on the specified width. */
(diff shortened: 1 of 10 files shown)


src/components/common/ExpandableList.tsx
```diff
@@ -1,92 +1,176 @@
...
+  )
+
+  // Measure the anchor view position and width
+  React.useLayoutEffect(() => {
+    if (containerLayout == null && anchorViewRef.current != null) {
+      anchorViewRef.current.measureInWindow((x, y, width, height) => {
+        // iOS and Android return various garbage values initially.
+        // We expect positive nonzero values. We can always safely omit 0 values
+        // because we guarantee some amount of margin under correct UI design
+        if (x <= 0 || y <= 0 || width <= 0) return
```
swansontec, line 119: It seems like we don't need `useLayoutEffect` for this. Instead, we can put the measurement on line 85, right before the `Airship.show`. This means the drop-down would remain in place once it appears, even if the scene below scrolls or re-adjusts. This should be fine, especially if touching the underlying scene dismisses the drop-down.

The `useLayoutEffect` hook is kinda expensive, so we don't want to do this unless it's needed (such as right before drop-down). This might also avoid the glitches, since everything will be laid out correctly by this point.

src/components/common/ExpandableList.tsx
```diff
@@ -1,92 +1,176 @@
...
   dropdownContainer: {
     position: 'absolute',
-    top: '100%',
-    backgroundColor: theme.modal,
+    borderRadius: theme.rem(0.5)
+  },
+  dummyMeasureContainer: {
+    position: 'absolute',
+    opacity: 0
+  },
```
swansontec, line 155: I think we want `pointerEvents="none"` on the view, so taps can pass through. Even though it's invisible, it might still affect the touch hierarchy.

swansontec (changes requested):
Minor changes. If this ends up being a rabbit hole, I guess we could merge the current version.

src/components/common/ExpandableList.tsx
Jon-edge: Ended up being a rabbit hole trying to merge it together. At least reduced this to a `useEffect` so we only fire until we get non-garbage initial values

src/components/common/ExpandableList.tsx
```diff
@@ -56,28 +56,24 @@ export const ExpandableList = (props: Props) => {
...
-  )
-
   useAsyncEffect(
     async () => {
-      if (dropdownLayoutStyle == null || itemHeight === 0) return
+      console.debug('main effect')
+      if (dropdownLayoutStyle == null || itemHeight === 0) {
+        console.debug('skipping')
+        return
+      }
```
swansontec, line 65: Leftover logging in here.

swansontec (approved):
Approved, but delete the console debugging when you merge.

PR #5249, opened 2024-09-13
Add minimum receive amount to SwapConfirmationScene

Jon-edge:
image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208231167011433
  - https://app.asana.com/0/0/1208162689994842

### DIFF
--- a/src/components/themed/ExchangeQuoteComponent.tsx
+++ b/src/components/themed/ExchangeQuoteComponent.tsx
@@ -15,6 +15,7 @@ import { DECIMAL_PRECISION, removeIsoPrefix } from '../../util/utils'
 import { EdgeCard } from '../cards/EdgeCard'
 import { CurrencyRow } from '../rows/CurrencyRow'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
+import { FiatText } from '../text/FiatText'
 import { EdgeText } from './EdgeText'
 
 interface Props {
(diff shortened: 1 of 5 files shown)


src/components/themed/ExchangeQuoteComponent.tsx
```diff
@@ -99,8 +105,20 @@ export const ExchangeQuote = (props: Props) => {
...
+      return (
+        <View style={styles.bottomContainer}>
+          {renderRow(
+            <EdgeText style={styles.bottomText}>{lstrings.swap_minimum_receive_amount}</EdgeText>,
+            <EdgeText style={styles.bottomText}>{minCryptoAmountText}</EdgeText>
+          )}
+          {renderRow(<></>, <EdgeText style={styles.bottomText}>({minFiatAmountText})</EdgeText>)}
+        </View>
+      )
+    } else {
```
swansontec, line 119: If the previous condition has a `return`, you don't logically need an `else`. This would apply not only to the `if (quote.minReceiveAmount != null)` block, but also to the `else { return null }` block below.

swansontec (approved):
One optional cleanup.

src/components/themed/ExchangeQuoteComponent.tsx
Jon-edge: Yeah I know it works either way, this makes it clearer (at least to me) that these are 3 paths in the logic at this point. 

When I read an if statement in the middle of a function, I would normally have to examine the rest above to make sure there isn't some dependent stuff or early returns between the various if blocks or which if block returns.

If I see chains of else ifs or switches I easily know there's no funny business in between the blocks

Jon-edge: Personally I always get blind-sighted by early returns

PR #5251, opened 2024-09-13
Handle foreground push notifications

Jon-edge:
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208040714093837

### DIFF
--- a/src/components/navigation/FlashNotification.tsx
+++ b/src/components/navigation/FlashNotification.tsx
@@ -1,28 +1,29 @@
 import * as React from 'react'
-import { Text, View } from 'react-native'
+import { View } from 'react-native'
 import { AirshipBridge } from 'react-native-airship'
 import AntDesignIcon from 'react-native-vector-icons/AntDesign'
 
-import { THEME } from '../../theme/variables/airbitz'
 import { AirshipDropdown } from '../common/AirshipDropdown'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
+import { Paragraph } from '../themed/EdgeText'
 
 interface Props {
   bridge: AirshipBridge<void>
   message: string
+  icon?: React.ReactNode
   onPress?: () => void
 }
 
 export function FlashNotification(props: Props) {
-  const { bridge, message, onPress = () => {} } = props
+  const { bridge, message, icon, onPress = () => {} } = props
   const theme = useTheme()
   const styles = getStyles(theme)
 
   return (
     <AirshipDropdown bridge={bridge} backgroundColor={theme.modal} onPress={onPress}>
       <View style={styles.container}>
-        <AntDesignIcon name="checkcircle" size={theme.rem(2)} style={styles.icon} />
-        <Text style={styles.text}>{message}</Text>
+        {icon ?? <AntDesignIcon name="checkcircle" size={theme.rem(2)} style={styles.icon} />}
+        <Paragraph center>{message}</Paragraph>
       </View>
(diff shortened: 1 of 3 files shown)


src/components/navigation/FlashNotification.tsx
```diff
@@ -1,48 +1,44 @@
...
   const styles = getStyles(theme)
 
   return (
     <AirshipDropdown bridge={bridge} backgroundColor={theme.modal} onPress={onPress}>
       <View style={styles.container}>
-        <AntDesignIcon name="checkcircle" size={theme.rem(2)} style={styles.icon} />
-        <Text style={styles.text}>{message}</Text>
+        {icon ?? <AntDesignIcon name="checkcircle" size={theme.rem(2)} style={styles.icon} />}
+        <Paragraph center>
+          <EdgeText>{message}</EdgeText>
```
swansontec, line 27: Do we need `EdgeText` inside `Paragraph`? I thought  the`<Paragraph>` was enough by itself

src/components/services/DeepLinkingManager.tsx
```diff
@@ -74,15 +80,35 @@ export function DeepLinkingManager(props: Props) {
         }
       }
 
+      /** Handler for push messages received while app is in the foreground. */
+      const handleForegroundPushMessage = (message: FirebaseMessagingTypes.RemoteMessage) => {
+        if (message.notification?.body == null) {
+          console.warn('FirebaseMessagingTypes.RemoteMessage (foreground push message) has no body')
+          return
+        }
```
swansontec, line 88: Note for later: In the `handleBackgroundPushMessage`, we use `parsePushMessage` to extract a link. We should do the same thing for foreground push messages. That way, the promo can include a link to a blog article or whatever. If the user taps the message, we should launch the link.

We can add this feature later, since right now `parsePushMessage` is extremely limited in what it can do. Once we beef up `parsePushMessage`, we can fix the dropdown to handle the link.

src/components/services/DeepLinkingManager.tsx
```diff
@@ -74,15 +80,35 @@ export function DeepLinkingManager(props: Props) {
...
+
+        // Show a FlashNotification:
+        Airship.show(bridge => (
+          <FlashNotification
+            bridge={bridge}
+            message={message.notification?.body ?? ''}
```
swansontec, line 94: If we do `const body = message.notification?.body; if (body == null) console.warn(...`, then we don't need the `?? ''` part here, since we can just use `body` directly.

src/components/services/DeepLinkingManager.tsx
```diff
@@ -74,15 +80,35 @@ export function DeepLinkingManager(props: Props) {
...
+            onPress={() => {
+              bridge.resolve()
+            }}
+            icon={<FontAwesomeIcon name="bell-o" size={theme.rem(2)} style={styles.icon} />}
+          />
+        )).catch(error => console.error(String(error)))
```
swansontec, line 100: `showDevError`?

swansontec (changes requested):
Small changes needed.

src/components/services/DeepLinkingManager.tsx
Jon-edge: Yeah, that's what I would think, but Typescript disagrees. Did another workaround

swansontec (approved):

PR #5261, opened 2024-09-18
Jon/fix/assetStatus2-country

Jon-edge:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device


src/components/scenes/HomeScene.tsx
```diff
@@ -95,7 +95,7 @@ export const HomeScene = (props: Props) => {
   // Set countryCode once
   useAsyncEffect(
     async () => {
-      const countryCode = await getCountryCodeByIp().catch(() => '')
+      const countryCode = await getCountryCodeByIp().catch(() => undefined)
```
paullinator, line 98: `getCountryCodeByIp` also has to be fixed as it returns `--` if it can't get a countryCode. This was my oversight.

paullinator (changes requested):

src/actions/AccountReferralActions.ts
```diff
@@ -202,15 +202,10 @@ export interface ValidateFuncs {
...
-  try {
-    const reply = await fetch(`https://pro.ip-api.com/json/?key=${apiKey}`)
-    const { countryCode } = asIpApi(await reply.json())
-    out = countryCode
-  } catch (e: any) {
-    console.error(e.message)
-  }
-  return out
+
+  const reply = await fetch(`https://pro.ip-api.com/json/?key=${apiKey}`)
```
paullinator, line 206: We don't want this routine to throw. That could cause a red error to show. Add back the try/catch and return undefined if anything throws

paullinator (changes requested):

src/actions/AccountReferralActions.ts
Jon-edge: Callers are already catching but yea that makes more sense

paullinator (approved):

PR #5262, opened 2024-09-18
Jon/detect-android-batt-saver

Jon-edge:
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207716275514867

### DIFF
--- a/src/components/services/Services.tsx
+++ b/src/components/services/Services.tsx
@@ -1,6 +1,9 @@
 import { asDate, asJSON, asObject, uncleaner } from 'cleaners'
 import { EdgeAccount } from 'edge-core-js'
 import * as React from 'react'
+import { EmitterSubscription } from 'react-native'
+import { AirshipBridge } from 'react-native-airship'
+import { powerSavingModeChanged, powerSavingOn } from 'react-native-power-saving-mode'
 
 import { updateExchangeInfo } from '../../actions/ExchangeInfoActions'
 import { refreshConnectedWallets } from '../../actions/FioActions'
@@ -11,6 +14,7 @@ import { ENV } from '../../env'
 import { useAsyncEffect } from '../../hooks/useAsyncEffect'
 import { useHandler } from '../../hooks/useHandler'
 import { useRefresher } from '../../hooks/useRefresher'
+import { lstrings } from '../../locales/strings'
 import { defaultAccount } from '../../reducers/CoreReducer'
 import { FooterAccordionEventService } from '../../state/SceneFooterState'
 import { useDispatch, useSelector } from '../../types/reactRedux'
@@ -18,6 +22,7 @@ import { NavigationBase } from '../../types/routerTypes'
 import { height, ratioHorizontal, ratioVertical, width } from '../../util/scaling'
 import { snooze } from '../../util/utils'
 import { FioCreateHandleModal } from '../modals/FioCreateHandleModal'
+import { AlertDropdown } from '../navigation/AlertDropdown'
 import { AccountCallbackManager } from './AccountCallbackManager'
 import { ActionQueueService } from './ActionQueueService'
 import { Airship } from './AirshipInstance'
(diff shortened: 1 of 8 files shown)


src/components/services/Services.tsx
```diff
@@ -123,6 +127,60 @@ export function Services(props: Props) {
...
+        if (isWarningShowing) return
+
+        isWarningShowing = true
+        await Airship.show(bridge => <AlertDropdown bridge={bridge} message={lstrings.warning_battery_saver} warning persistent />)
+        isWarningShowing = false
+      }
+
+      const checkPowerSavingMode = async () => {
+        // This method is only available for Android
+        if (powerSavingOn != null) {
```
paullinator, line 147: Just check at the top of the useAsyncEffect

`if (Platform.OS !== 'android') return`

This eliminates all the checks for powerSavingOn or powerSavingModeChanged

src/components/services/Services.tsx
```diff
@@ -123,6 +127,60 @@ export function Services(props: Props) {
...
+              await showBatterySaverWarning()
+            }
+          } catch (error) {
+            showDevError(`Error checking power-saving mode: ${String(error)}`)
+          }
+        }
+      }
+
+      const handlePowerSavingModeChanged = async (state: boolean) => {
+        if (state && !previousState) {
```
paullinator, line 161: No need to have a `previousState`. `showBatterySaverWarning` already checks if it's showing and early exits.

paullinator (changes requested):

src/components/services/Services.tsx
Jon-edge: Those functions are defined as optional by the lib. I'd need to also patch the lib itself and this seemed like the simpler approach.

src/components/services/Services.tsx
Jon-edge: Yes there's a need for a 'previousState' 

`showBatterySaverWarning` can check if the modal is showing and early exits, yes. 

This statement is saying only fire when the powerSaving mode changes specifically from off to on.

Two state mechanisms exist here: 
1. Whether the modal is showing (agnostic of the power saving state)
2. Only detect changes for power saving mode from off to on

Jon-edge:
Also fixed auto-dismissing the warning when battery saver was turned off but the user never previously manually dismissed the warning

src/components/services/Services.tsx
```diff
@@ -131,50 +132,42 @@ export function Services(props: Props) {
...
+      const handlePowerSavingModeChanged = async (isPowerSavingModeOn: boolean) => {
+        if (isPowerSavingModeOn && airshipBridge == null) {
+          await Airship.show(bridge => {
+            airshipBridge = bridge // Capture the bridge here
+            return <AlertDropdown bridge={bridge} message={lstrings.warning_battery_saver} warning persistent />
+          }).then(() => {
+            airshipBridge = undefined
+          })
+        } else if (airshipBridge != null) {
+          // Dismiss the alert when power-saving mode turns off and there's an
```
paullinator, line 148: This will cause the alert to be dismissed if `isPowerSavingModeOn` and the modal is showing. While you shouldn't get this case, you are relying on distributed logic to ensure this doesn't break. Logic in the SDK in addition to logic in your code.

This should be

```
if (isPowerSavingModeOn && airshipBridge == null) {
   // showmodal
} else if (!isPowerSavingModeOn && airshipBridge != null) {
    // close modal
}
```

paullinator (changes requested):

src/components/services/Services.tsx
Jon-edge: Yeah I originally had it like this but assumed you'd prefer it less verbose since it shouldn't happen

paullinator: verbose isn't bad if it makes the logic crystal clear.

paullinator (approved):

src/components/services/Services.tsx
Jon-edge: Good to hear, that's my preference too

PR #5274, opened 2024-09-25
Jon/fix/swap-assets

Jon-edge:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208388828536876

### DIFF
--- a/src/components/scenes/SwapCreateScene.tsx
+++ b/src/components/scenes/SwapCreateScene.tsx
@@ -1,5 +1,14 @@
 import { gt, gte } from 'biggystring'
-import { EdgeCurrencyWallet, EdgeSwapRequest, EdgeTokenId } from 'edge-core-js'
+import {
+  asMaybeInsufficientFundsError,
+  asMaybeSwapAboveLimitError,
+  asMaybeSwapBelowLimitError,
+  asMaybeSwapCurrencyError,
+  asMaybeSwapPermissionError,
+  EdgeCurrencyWallet,
+  EdgeSwapRequest,
+  EdgeTokenId
+} from 'edge-core-js'
 import * as React from 'react'
 import { useState } from 'react'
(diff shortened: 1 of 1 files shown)


src/components/scenes/SwapCreateScene.tsx
```diff
@@ -126,8 +126,12 @@ export const SwapCreateScene = (props: Props) => {
...
       clearError = true
     }
-    // Selected asset related errors
-    else if (changed === 'asset' && ['SwapPermissionError', 'SwapCurrencyError'].includes(error.name)) {
+    // Selected asset related errors (arbitrarily includes all amount-related
+    // errors as well)
+    else if (
+      changed === 'asset' &&
+      ['SwapPermissionError', 'SwapCurrencyError', 'InsufficientFundsError', 'SwapAboveLimitError', 'SwapBelowLimitError'].includes(error.name)
+    ) {
```
swansontec, line 134: I somehow missed this on the previous review, but in the lines about, we should be using the core-provided cleaners `asMaybeInsufficientFundsError`, `asMaybeSwapAboveLimitError`, etc..., instead of directly checking `error.name`.

We could do it this way:

```ts
const isSwapPermission = asMaybeSwapPermissionError(error) != null
const isInsufficentFunds = asMaybeInsufficientFundsError(error) != null
const isSwapCurrency = asMaybeSwapCurrencyError(error) != null
...etc...

if (changed === 'asset' &&
  (isSwapPermission || isSwapCurrency || isInsufficentFunds || isSwapAboveLimit || isSwapBelowLimit)) {
    clearError = true
}
```

Not sure if this change is super-duper-necessary, but it would be a bit "more correct".

swansontec (approved):
One optional change, but this looks right to me.

PR #5275, opened 2024-09-25
Dispatch `activateWalletTokens` from TransactionListScene

samholmes:
This is to resolve an issue where the wallet state hasn't completed
checking if a newly added token is activated or not. This means the
user can enter the transaction list scene without activating the wallet.
In that case, we'll at least navigate in retrospect.

Watch for changes to the `unactivedTokenIds` property on the wallet.
If the tokenId within the transaction list appears in the
`unactivedTokenids` array, then we'll properly handle the activation
user journey.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208383880746617

### DIFF
--- a/src/actions/WalletActions.tsx
+++ b/src/actions/WalletActions.tsx
@@ -14,7 +14,7 @@ import { selectDisplayDenomByCurrencyCode } from '../selectors/DenominationSelec
 import { ThunkAction } from '../types/reduxTypes'
 import { NavigationBase } from '../types/routerTypes'
 import { MapObject } from '../types/types'
-import { getCurrencyCode, getToken, isKeysOnlyPlugin } from '../util/CurrencyInfoHelpers'
+import { getCurrencyCode, isKeysOnlyPlugin } from '../util/CurrencyInfoHelpers'
 import { getWalletName } from '../util/CurrencyWalletHelpers'
 import { fetchInfo } from '../util/network'
 import { convertCurrencyFromExchangeRates } from '../util/utils'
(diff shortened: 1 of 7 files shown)


src/actions/WalletActions.tsx
```diff
@@ -147,101 +147,98 @@ export function updateMostRecentWalletsSelected(walletId: string, tokenId: EdgeT
...
-            <ButtonsModal
-              bridge={bridge}
-              title={lstrings.create_wallet_account_unfinished_activation_title}
-              message={sprintf(msg, feeString)}
-              buttons={{ ok: { label: lstrings.string_ok } }}
-            />
-          )).catch(err => showError(err))
-          navigation.pop()
-          return
+function activateWalletTokens(navigation: NavigationBase, wallet: EdgeCurrencyWallet, tokenIds?: string[]): ThunkAction<Promise<void>> {
```
swansontec: It seems like the `tokenIds` argument should be mandatory, and should be `EdgeTokenId[]`. The function is only called from two places, so this change should be safe. Within the function, we pass `tokenIds` straight to the core, so it's up to the plugins to decide if they support mainnet activation or not.

Further down, when we do `const tokensText`, we could change the following line to `const { currencyCode, displayName } = tokenId != null ? wallet.currencyConfig.allTokens[tokenId] : wallet.currencyInfo`, and even our string formatting will work for mainnet coins.

src/actions/WalletActions.tsx
```diff
@@ -147,101 +147,98 @@ export function updateMostRecentWalletsSelected(walletId: string, tokenId: EdgeT
...
+  return async (_dispatch, getState) => {
+    if (tokenIds == null) throw new Error('Activating mainnet wallets unsupported')
```
swansontec, line 152: If `tokenIds` are mandatory, this line goes away.

src/components/scenes/TransactionListScene.tsx
```diff
@@ -139,6 +143,23 @@ function TransactionListComponent(props: Props) {
...
+  // Automatically navigate to the token activation confirmation scene if
+  // the token appears in the unactivatedTokenIds list once the wallet loads
+  // this state.
+  useAsyncEffect(
+    async () => {
+      if (unactivatedTokenIds.length > 0) {
+        if (unactivatedTokenIds.some(unactivatedTokenId => unactivatedTokenId === tokenId)) {
+          // NOTE: It's safe to cast tokenId to string because we know it's in
+          // unactivatedTokenIds which is an array of strings.
+          await dispatch(activateWalletTokens(navigation, wallet, [tokenId as string]))
```
swansontec, line 155: If we fix the typing, the cast simply goes away.

swansontec (changes requested):

PR #5278, opened 2024-09-26
Sync OS contacts access settings with the app

Jon-edge:
Previously, we maintained a locally controlled `contactsPermissionOn` setting and had an elaborate means to try to sync it with the OS setting and our `ContactsPermissionModal` rules.

This change makes components dependent directly on the actual OS' contacts permission setting and simplifies the `ContactsPermissionModal` rules to just show once, wherever it appears first.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device


src/components/modals/ContactsPermissionModal.tsx
```diff
@@ -16,15 +22,52 @@ interface Props {
...
+export async function maybeShowContactsPermissionModal(): Promise<ContactsPermissionResult | undefined> {
+  const contactsPermissionOn = (await check(permissionNames.contacts).catch(_error => 'denied')) === 'granted'
+  if (contactsPermissionOn || showModalPromise != null) return undefined
+
+  showModalPromise = (async () => {
+    return await Airship.show<ContactsPermissionResult | undefined>(bridge => <ContactsPermissionModal bridge={bridge} />)
+  })()
+
+  return await showModalPromise
+}
```
swansontec, line 43: I like this new function, but the logic is a little off. Try this:

```ts
export function maybeShowContactsPermissionModal() : ThunkAction<Promise<ContactsPermissionResult | undefined>> {
    const state = getState()
    const {account} = state.core

    // Bail if we are currently showing the modal:
    if (showModalPromise != null) return

    // Bail if we have ever show the modal:
    const {contactsPermissionShown} = state.ui.settings.contactsPermissionShown
    if (contactsPermissionShown) return

    // Bail if we already have permission:
   const contactsPermissionOn = (await check(permissionNames.contacts).catch(_error => 'denied')) === 'granted'
  if (contactsPermissionOn) return

     // Show the modal:
    const showModalPromise = Airship.show<ContactsPermissionResult | undefined>(bridge => <ContactsPermissionModal bridge={bridge} />)
    await writeContactsPermissionShown(account, true)

    return await showModalPromise
  }
}
```

Changes:

- Turned this into a thunk, so we can access redux.
- Broke out & commented the reasons why the modal wouldn't show
- Got rid of the IIFE around `Airship.show`
- Writing to disk, since we can handle that out here now.

src/components/modals/ContactsPermissionModal.tsx
```diff
@@ -16,15 +22,52 @@ interface Props {
...
+
+/**
+ * Shows the modal if it hasn't been shown before, and attempts to set the
+ * system contacts permission setting
+ */
```
swansontec, line 48: We don't need to `export ContactsPermissionModal` anymore, since the only way to show it is via `maybeShowContactsPermissionModal`.

In the future, if there are scenarios where we want to show the modal a second or third time, we add a parameter like `maybeShowContactsPermissionModal(showAgain? boolean)`, since even in these cases, we would still want all the other logic that comes with the function.

swansontec (approved):

PR #5279, opened 2024-09-26
Re-enable Piratechain on iOS

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1207982832446869


patches/react-native-piratechain+0.5.1.patch
swansontec, line 1: It seems like we should probably fix the react-native-piratechain package itself, but this will do for the time being.

swansontec (approved):

patches/react-native-piratechain+0.5.1.patch
peachbits: Yeah we could, maybe next publish. I did submit a PR for them to fix the sdk directly.

PR #5280, opened 2024-09-27
Jon/fix/light-acc-default-buy

Jon-edge:
### Original Amount Calculations (Light Acc):
3 steps were originally performed resulting in:
image
image

### Changes:
image
image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208393963843488

### DIFF
--- a/src/plugins/gui/amountQuotePlugin.ts
+++ b/src/plugins/gui/amountQuotePlugin.ts
@@ -225,9 +225,6 @@ export const amountQuoteFiatPlugin: FiatPluginFactory = async (params: FiatPlugi
       let goodQuotes: FiatProviderQuote[] = []
       let lastSourceFieldNum: number
 
-      // HACK: Force EUR for sepa Bity, since Bity doesn't accept USD, a common
-      // wallet fiat selection regardless of region.
-      // TODO: Remove when Fiat selection is designed.
       const fiatCurrencyCode = forceFiatCurrencyCode ?? defaultIsoFiat
       const displayFiatCurrencyCode = removeIsoPrefix(fiatCurrencyCode)
       const isBuy = direction === 'buy'
--- a/src/plugins/gui/amountQuotePlugin.ts
+++ b/src/plugins/gui/amountQuotePlugin.ts
@@ -1,4 +1,4 @@
-import { div, eq, gt, mul, round, toFixed } from 'biggystring'
+import { div, eq, gt, round, toFixed } from 'biggystring'
 import { asNumber, asObject } from 'cleaners'
 import { sprintf } from 'sprintf-js'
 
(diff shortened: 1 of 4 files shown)


src/plugins/gui/amountQuotePlugin.ts
```diff
@@ -242,11 +242,17 @@ export const amountQuoteFiatPlugin: FiatPluginFactory = async (params: FiatPlugi
         const ratePair = `${isoFiatCurrencyCode}_iso:USD`
         const rate = await getHistoricalRate(ratePair, isoNow)
         initialValue1 = div(startingFiatAmount, String(rate), DECIMAL_PRECISION)
-        // Round out a decimals
+
+        if (isLightAccount) {
+          // Round down
+          initialValue1 = floor(initialValue1, 0)
```
paullinator, line 248: What's the point of this? The rounding should be the same for light account or not.

src/plugins/gui/amountQuotePlugin.ts
```diff
@@ -523,8 +523,13 @@ export const amountQuoteFiatPlugin: FiatPluginFactory = async (params: FiatPlugi
               quoteFiatCurrencyCode === 'iso:USD' ? '1' : (await getHistoricalRate(`${quoteFiatCurrencyCode}_iso:USD`, new Date().toISOString())).toString()
             const quoteDollarValue = mul(bestQuote.fiatAmount, quoteFiatRate)
 
-            if (gt(quoteDollarValue, '50')) {
-              stateManager.update({ statusText: { content: lstrings.fiat_plugin_purchase_limit_error, textType: 'error' } })
+            if (gt(quoteDollarValue, DEFAULT_FIAT_AMOUNT_LIGHT_ACCOUNT)) {
```
paullinator, line 526: I see this can work since in the previous commit you floored the last decimal place instead of rounding, but this is not clean logic. This is prone to break in the future if anyone touches either code location without knowing about the other. The logic of each don't even equal each other. The other commit takes $50 converted to the fiat currency then floors it. This code takes the EXACT exchange rate conversion.

Better to create a shared function that both places can call. Something like

`getDefaultFiatValue(startingAmount, isoFiatCurrencyCode, isoNow) => fiatAmount`

paullinator (changes requested):

src/plugins/gui/amountQuotePlugin.ts
Jon-edge: See the description - in the original implementation it was possible to potentially round up, so I wanted to retain that for full accounts

src/plugins/gui/amountQuotePlugin.ts
Jon-edge: It was intentional to have the exact exchange rate shown because that's ultimately what we check against. We want users to be able to buy as much as possible, but for the initial value we make it look nice with a rounder number

src/plugins/gui/amountQuotePlugin.ts
paullinator: @Jon-edge so why not for light accounts?

src/plugins/gui/amountQuotePlugin.ts
paullinator: It's a big ugly to show a value that's down to the penny. Just match the actual default fiat amount which can both go up or down based on `round` instead of `floor`

paullinator (changes requested):

src/plugins/gui/amountQuotePlugin.ts
```diff
@@ -61,6 +61,28 @@ const MAX_QUOTE_VALUE = '10000000000'
 
 const NO_PROVIDER_TOAST_DURATION_MS = 10000
 
+/** Pick a default fiat amount in the foreign fiat amount that is roughly equal
+ to the default USD fiat amount. Prioritizes rounding up/down to a nice looking
+ whole number. */
+async function getInitialFiatValue(startingFiatAmount: string, isoFiatCurrencyCode: string) {
+  let initialValue1: string | undefined
+  if (isoFiatCurrencyCode !== 'iso:USD') {
+    const isoNow = new Date().toISOString()
```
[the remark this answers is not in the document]
Jon-edge: Yeah I had it like that in a prior revision where I just reuse the initial default, but thought you wanted the most recent exchange rate each time this was called.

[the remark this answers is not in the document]
paullinator: I dont think I ever requested a fresh exchange rate. The only important thing is that it's consistent between the default and the limit

src/plugins/gui/amountQuotePlugin.ts
```diff
@@ -61,15 +61,18 @@ const MAX_QUOTE_VALUE = '10000000000'
 
 const NO_PROVIDER_TOAST_DURATION_MS = 10000
 
+/** Make sure to use the same timestamp for rates throughout the lifetime of the
+ * module. */
+const ISO_NOW = new Date().toISOString()
```
paullinator: Ack. Don't get the date once for the lifetime of the app. Get it at least when the plugin is started. The app could stay running for days and weeks at which point the rate could drift a lot. Once per plugin launch would be much better. Please see my fixup. It's untested so please verify. If good then merge.

paullinator (approved):
See my fixup (untested) and retest. Merge if good.

PR #5294, opened 2024-10-05
Call signTx after signing or broadcasting transactions from WalletConnect

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1207763168583636

### DIFF
--- a/src/components/modals/WcSmartContractModal.tsx
+++ b/src/components/modals/WcSmartContractModal.tsx
@@ -113,6 +113,7 @@ export const WcSmartContractModal = (props: Props) => {
           const tx = await wallet.makeSpend(spendInfo)
           const signTx = await wallet.signTx(tx)
           await walletConnect.approveRequest(topic, requestId, signTx.signedTx)
+          await wallet.saveTx(signTx)
           break
         }
         case 'eth_sendTransaction':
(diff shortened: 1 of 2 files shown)


src/components/modals/WcSmartContractModal.tsx
```diff
@@ -112,6 +112,7 @@ export const WcSmartContractModal = (props: Props) => {
           const spendInfo: EdgeSpendInfo = await wallet.otherMethods.txRpcParamsToSpendInfo(cleanPayload.params[0])
           const tx = await wallet.makeSpend(spendInfo)
           const signTx = await wallet.signTx(tx)
+          await wallet.saveTx(signTx)
```
swansontec, line 115: Can we put this below `walletConnect.approveRequest`? If WalletConnect fails, we don't want to call `saveTx` at all.

src/components/modals/WcSmartContractModal.tsx
```diff
@@ -122,6 +123,7 @@ export const WcSmartContractModal = (props: Props) => {
           const tx = await wallet.makeSpend(spendInfo)
           const signedTx = await wallet.signTx(tx)
           const sentTx = await wallet.broadcastTx(signedTx)
+          await wallet.saveTx(sentTx)
```
swansontec, line 126: Can we put this below `walletConnect.approveRequest`? If WalletConnect fails, we don't want to call `saveTx` at all.

swansontec (changes requested):

src/components/modals/WcSmartContractModal.tsx
peachbits: good idea, done

swansontec (approved):

PR #5295, opened 2024-10-07
Implement `EarnScene`

Jon-edge:
image
image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207269659885229


src/components/buttons/EdgeSwitch.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { EdgeText } from '../themed/EdgeText'
+
+interface EdgeSwitchProps {
+  labelA: string
+  labelB: string
+  onSelectA: () => void
+  onSelectB: () => void
+}
+
+const EdgeSwitch: React.FC<EdgeSwitchProps> = ({ labelA, labelB, onSelectA, onSelectB }) => {
```
swansontec, line 14: I somewhat prefer the `props => { const { ... } = props` style, where the destructuring is its own statement. It's not a big difference at this scale, but it becomes meaningful for bigger / more complicated components like our text inputs. Even though `EdgeSwitch` is small enough to get away with inline destructuring, it's better to keep our components consistent with each other.

src/components/buttons/EdgeSwitch.tsx
```diff
@@ -0,0 +1,68 @@
...
+  option: {
+    paddingVertical: 8,
+    paddingHorizontal: theme.rem(2),
+    alignItems: 'center',
+    justifyContent: 'center'
+  },
+  label: {}
+}))
+
+export default EdgeSwitch
```
swansontec, line 68: Never use `export default`! It is 100% banned in all Edge projects. We used named exports for everything, so do `export const EdgeSwitch` avove instead.

src/components/scenes/Staking/EarnScene.tsx
```diff
@@ -0,0 +1,168 @@
...
+                  positionWallets.push(wallet)
+                }
+              }
+            }
+
+            policyMap[wallet.id] = { stakePolicyPositions, stakePlugins }
+            setStakePolicyMap({ ...policyMap })
+            setPositionWallets(positionWallets)
+          } catch (e) {
+            console.error(String(e))
```
swansontec, line 94: First of all, you don't need `String()` around `e`. Secondly, shouldn't this be `showDevError`?

src/components/scenes/Staking/EarnScene.tsx
```diff
@@ -0,0 +1,168 @@
...
+  React.useEffect(() => {
+    if (stakePolicyMap == null) return
+
+    if (isPortfolioSelected) {
+      setDisplayWallets(positionWallets)
+    } else {
+      setDisplayWallets(wallets)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isPortfolioSelected, stakePolicyMap])
```
swansontec, line 114: The `displayWallets` can be a plain variable, not a `useState`. Just do`const displayWallets = !stakePolicyMap ? [] : isPortfolioSelected ? positionWallets : wallets`, and it should have the same result.

src/components/scenes/Staking/EarnScene.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  const renderStakeItems = React.useCallback(
+    (item: FlatListItem<EdgeCurrencyWallet>) => {
+      const { item: wallet } = item
+      if (stakePolicyMap == null) return <></>
```
swansontec, line 119: Maybe replace `return <></>` with `return null` in the few places it shows up. Should be simpler & faster.

src/components/scenes/Staking/EarnScene.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  return (
+    // TODO: Address "VirtualizedLists should never be nested inside plain
+    // ScrollViews with the same orientation because it can break windowing and
+    // other functionality - use another VirtualizedList-backed container
+    // instead." somehow, while retaining the bottom loader positioning...
+    <SceneWrapper scroll padding={theme.rem(0.5)}>
+      <EdgeSwitch labelA={lstrings.staking_discover} labelB={lstrings.staking_portfolio} onSelectA={handleSelectEarn} onSelectB={handleSelectPortfolio} />
+      <SectionHeader leftTitle={lstrings.staking_earning_pools} />
+      <FlatList data={displayWallets} renderItem={renderStakeItems} keyExtractor={wallet => wallet.id} style={styles.list} />
```
swansontec, line 154: Instead of using a `FlatList`, why not render the items directly? Since you've placed the list inside a scroll view already, the `FlatList` won't be able to optimize rendering in any case.

We shouldn't have 100's of options, so it shouldn't be a long list.  It seems like a `{displayWallets.map(wallet => renderStakeItems(wallet))}` would basaically do the same job.

swansontec (changes requested):

src/components/buttons/EdgeSwitch.tsx
Jon-edge: Whoops, missed AI copy pasta mistake, sorry. I'll try to fix my prompt

Jon-edge: Came from an AI prompt change, my bad

src/components/scenes/Staking/EarnScene.tsx
Jon-edge: I was having type errors around here earlier, I'll give it a try

src/components/buttons/EdgeSwitch.tsx
Jon-edge: Came from an AI prompted change, my bad

src/components/buttons/EdgeSwitch.tsx
swansontec: Yes, enabling `no-default-export` would be a welcome addition.

PR #5297, opened 2024-10-08
Small error handling fixes

peachbits:
---
  - https://app.asana.com/0/0/1208105233891024

### DIFF
--- a/src/plugins/gui/providers/bityProvider.ts
+++ b/src/plugins/gui/providers/bityProvider.ts
@@ -8,6 +8,7 @@ import { HomeAddress, SepaInfo } from '../../../types/FormTypes'
 import { StringMap } from '../../../types/types'
 import { utf8 } from '../../../util/encoding'
 import { removeIsoPrefix } from '../../../util/utils'
+import { SendErrorBackPressed } from '../fiatPlugin'
 import { FiatDirection, FiatPaymentType, FiatPluginUi } from '../fiatPluginTypes'
 import {
   FiatProvider,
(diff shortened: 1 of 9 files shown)


src/plugins/gui/providers/bityProvider.ts
```diff
@@ -609,10 +610,18 @@ export const bityProvider: FiatProviderFactory = {
...
+                    await completeBuyOrder(approveQuoteRes, showUi)
+                  } else {
+                    await completeSellOrder(approveQuoteRes, coreWallet, showUi, fiatCurrencyCode, tokenId)
+                  }
+                } catch (e: any) {
+                  if (e.message === SendErrorBackPressed) {
+                    // Do nothing
+                  } else {
+                    throw e
+                  }
```
swansontec, line 624: It is never safe to read `e.message`, until you have determined that this thing is actually an error. Do this instead:

```ts
catch (e: unknown) {
  if (e instanceof Error && e.message === SendErrorBackPressed) {
    return
  }
  throw e
}
```

src/plugins/gui/providers/paybisProvider.ts
```diff
@@ -761,6 +761,8 @@ export const paybisProvider: FiatProviderFactory = {
                       await openWebView()
                       if (e.message === SendErrorNoTransaction) {
                         await showUi.showToast(lstrings.fiat_plugin_sell_failed_to_send_try_again, NOT_SUCCESS_TOAST_HIDE_MS)
+                      } else if (e.message === SendErrorBackPressed) {
```
swansontec, line 764: This should also be `catch (e: unknown) {` above, in which case we need to add the `e instanceof Error` bits too.

swansontec (approved):

PR #5299, opened 2024-10-09
Prevent exchange rate fluctuations from restarting transaction list stream

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208296004541894

### DIFF
--- a/src/hooks/useTransactionList.ts
+++ b/src/hooks/useTransactionList.ts
@@ -29,6 +29,13 @@ export function useTransactionList(wallet: EdgeCurrencyWallet, tokenId: EdgeToke
 
   const requestMore = React.useRef(() => {})
 
+  // Ignore changes to the spamThreshold unless it's changing to or from 0
+  // This prevents starting the stream over when the exchange rate changes
+  const spamThresholdRef = React.useRef(spamThreshold)
+  if (spamThreshold !== spamThresholdRef.current && (spamThreshold === '0' || spamThresholdRef.current === '0')) {
+    spamThresholdRef.current = spamThreshold
+  }
+
   // The effect maintains internal mutable state,
   // and then calls `setOutput` to expose it atomically.
   const [output, setOutput] = React.useState<Omit<Output, 'requestMore'>>({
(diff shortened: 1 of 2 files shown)


src/hooks/useTransactionList.ts
```diff
@@ -25,10 +25,17 @@ interface TransactionListOptions {
...
+  const { searchString = '', spamThreshold: spamThresholdProp = '0' } = opts
 
   const requestMore = React.useRef(() => {})
 
+  // Ignore changes to the spamThreshold unless it's changing to or from 0
+  // This prevents starting the stream over when the exchange rate changes
+  const [spamThreshold, setSpamThreshold] = React.useState<string>(spamThresholdProp)
+  if (spamThresholdProp !== spamThreshold && (spamThresholdProp === '0' || spamThreshold === '0')) {
+    setSpamThreshold(spamThresholdProp)
+  }
```
swansontec, line 37: You can't update state variables during the render. It makes React complain.

Instead, `useRef` can do the same thing:

```ts
const spamThreshold = React.useRef(spamThresholdProp)
if (spamThresholdProp !== spamThreshold && (spamThresholdProp === '0' || spamThreshold === '0')) {
    spamThreshold.current = spamThreshold
}
```

Actually, since we are turning `spamThreshold` into a ref, we need to add `.current` everywhere we use it. For this reason, I propose renaming `spamThresholdProp` back to `spamThreshold`, and then naming the ref `spamThresholdRef`, so it's obvious we need the `.current` when we access it.

swansontec (changes requested):

src/hooks/useTransactionList.ts
peachbits: This is how I did it the first time and then switched it to useState. My concern at the time was not updating if the spam option was toggled but that's why the 0 checks are there so I apparently forgot I addressed that case.

src/hooks/useTransactionList.ts
```diff
@@ -160,7 +160,7 @@ export function useTransactionList(wallet: EdgeCurrencyWallet, tokenId: EdgeToke
       cleanupChanged()
       cleanupStream()
     }
-  }, [searchString, spamThreshold, tokenId, wallet])
+  }, [searchString, tokenId, wallet])
```
swansontec, line 163: I think we still need `spamThreshold.current` in here. That way we re-run the effect if the spam threshold ref changes.

swansontec (changes requested):

swansontec (approved):

PR #5300, opened 2024-10-10
Jon/restore-wallets-scene

Jon-edge:
image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207967192910249


src/components/cards/WalletRestoreCard.tsx
```diff
@@ -34,7 +34,7 @@ export const WalletRestoreCard = (props: Props) => {
 
   const [isSelected, setIsSelected] = React.useState(false)
 
-  const dateLabel = created?.toLocaleDateString() ?? lstrings.n_a
+  const dateLabel = created == null ? '' : toLocaleTime(created)
```
swansontec, line 37: Isn't this supposed to be a fixup commit, and not stand on its own?

src/actions/SettingsActions.tsx
```diff
@@ -247,7 +247,7 @@ export function showUnlockSettingsModal(): ThunkAction<Promise<void>> {
   }
 }
 
-export function showRestoreWalletsModal(navigation: NavigationBase): ThunkAction<Promise<void>> {
+export function showRestoreWalletsModal(navigation: NavigationBase, restoreKeys: EdgeWalletInfoFull[]): ThunkAction<Promise<void>> {
```
swansontec, line 250: Can we remove this action, and just put the logic on the scene? That is the only place where this logic is used. The scene should really be in charge of its own UX, including this modal.

Also, the `Promise.all` down below is a bit silly, because `account.changeWalletStates` can handle multiple keys at once. When you copy the code over, do this instead:

```ts
const states: EdgeWalletStates ={}
for (const info of restoreKeys) {
  states[info.id] = { archived: false, deleted: false }
}
await account.changeWalletStates(states)
```

Now we only call the core once, and the core only needs to access the disk once. Much faster.

src/components/scenes/WalletRestoreScene.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+export function WalletRestoreScene(props: Props) {
+  const { navigation } = props
+  const dispatch = useDispatch()
+  const theme = useTheme()
+  const styles = getStyles(theme)
+
+  const account = useSelector(state => state.core.account)
+  const allKeys = useWatch(account, 'allKeys')
+  const restoreWalletInfos = allKeys.filter(key => key.archived || key.deleted).reverse()
```
swansontec, line 34: The order of the keys is semi-random, so `reverse` isn't as useful as it seems. We should really sort the keys by date.

```ts
allKeys.filter(...).sort((a, b) => {
  const aDate = a.created?.valueOf() ?? 0
  const bDate = b.created?.valueOf() ?? 0
  return aDate - bDate
})
```

src/components/scenes/WalletRestoreScene.tsx
```diff
@@ -0,0 +1,185 @@
...
+  insetStyle: InsetStyle
+  restoreWalletInfos: EdgeWalletInfoFull[]
+  searchText: string
+  onSelectionChanged: (walletInfo: EdgeWalletInfoFull, isSelected: boolean) => void
+}) => {
+  const { insetStyle, restoreWalletInfos, searchText, onSelectionChanged } = props
+  const theme = useTheme()
+  const styles = getStyles(theme)
+
+  const currencyConfigMap = useSelector(state => state.core.account.currencyConfig)
```
swansontec, line 95: Please do `const account = useSelector(...)` on one line, and then grab `currencyConfig` on its own line.

src/components/scenes/WalletRestoreScene.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  const filteredWalletInfos = React.useMemo(() => {
+    const searchTarget = normalizeForSearch(searchText)
+
+    const out: Array<EdgeWalletInfoFull | string> = []
+    let lastSection = ''
+    for (const restoreWalletInfo of restoreWalletInfos) {
+      const { type, created } = restoreWalletInfo
+      const pluginId = type.replace('wallet:', '')
```
swansontec, line 104: I'm noticing a pattern here. Maybe we need a helper function to get the currencyConfig for a `walletInfo.type`:

```ts
function findCurrencyConfig(account: EdgeAccount, walletType: string): EdgeCurrencyConfig | undefined {
    const config = account.currencyConfig[pluginId]
    if (config.currencyInfo.walletType === walletType) return config
   }
}
```

This would be used not only here, but also earlier in the `WalletRestoreCard`.

swansontec (changes requested):

src/components/cards/WalletRestoreCard.tsx
```diff
@@ -23,18 +24,18 @@ interface Props {
...
   const theme = useTheme()
   const styles = getStyles(theme)
 
-  const pluginId = type.replace('wallet:', '')
-  const currencyConfig = useSelector(state => state.core.account.currencyConfig[pluginId])
-  const { currencyCode, displayName } = currencyConfig?.currencyInfo ?? {}
+  const account = useSelector(state => state.core.account)
+  const pluginId = Object.keys(account.currencyConfig).find(pluginId => account.currencyConfig[pluginId].currencyInfo.walletType === walletInfo.type)
+  const currencyConfig = pluginId == null ? undefined : account.currencyConfig[pluginId]
+  const { currencyCode, displayName } = findCurrencyInfo(account, walletInfo.type) ?? {}
```
swansontec, line 34: Oh, the `findCurrencyInfo` helper is pretty nice. We can use it to get the `pluginId` too:

```ts
const { currencyCode, displayName, pluginId } = findCurrencyInfo(account, walletInfo.type) ?? {}
const currencyConfig = pluginId == null ? undefined : account.currencyConfig[pluginId]
```

swansontec (approved):
One small cleanup left, but otherwise this looks really nice now.

PR #5305, opened 2024-10-11
Jon/rn-wheel-picker-android

Jon-edge:
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1207968955737616


src/components/modals/AutoLogoutModal.tsx
```diff
@@ -1,63 +1,59 @@
...
   }
 
-  const measurementOptionsStrings = {
-    seconds: lstrings.settings_seconds,
-    minutes: lstrings.settings_minutes,
-    hours: lstrings.settings_hours,
-    days: lstrings.settings_days
-  }
+  /** For linking selected labels to DisplayTime-compatible values */
+  const unitLabelValueList: Array<{ label: string; value: UnitOptionValue }> = React.useMemo(
```
swansontec, line 44: Don't use `useMemo`, since it's not saving any performance. Nothing is in JS is faster than raw constants. You could possibly move the constant outside of the render function, at the top-level of the file, to avoid re-evaluating it.

By using `useMemo`, we need to bundle up a callback function and then jump into React code that needs to do a state lookup, and array loop, and so forth. The `useMemo` thing only gives a performance boost if the contents are *really* expensive, like iterating over a giant list of wallets or doing biggystring math.

src/components/modals/AutoLogoutModal.tsx
```diff
@@ -1,63 +1,59 @@
...
+    () => [
+      { label: lstrings.settings_seconds, value: 'seconds' },
+      { label: lstrings.settings_minutes, value: 'minutes' },
+      { label: lstrings.settings_hours, value: 'hours' },
+      { label: lstrings.settings_days, value: 'days' }
+    ],
+    []
+  )
+
+  const unitOptionsAndroid: string[] = unitLabelValueList.map(option => option.label)
```
swansontec, line 54: This could also be lifted out of the render function, to the top level of the file.

src/components/modals/AutoLogoutModal.tsx
```diff
@@ -67,92 +63,88 @@ export const AutoLogoutModal = (props: Props) => {
...
   })
 
-  // Default to `hours` if it's disabled:
+  // Default to hours if it's disabled:
   React.useEffect(() => {
     if (displayTime.value === 0) setDisplayTime({ ...displayTime, measurement: 'hours' })
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 
+  const initPositionNumber = React.useMemo(
```
swansontec, line 72: You can use `const [initPositionNumber] = React.useState(displayTime.value)` to do the same thing without ESLint cheats. We'll capture the initial value into the state, and then just never update it.

src/components/modals/AutoLogoutModal.tsx
```diff
@@ -67,92 +63,88 @@ export const AutoLogoutModal = (props: Props) => {
...
+    () => displayTime.value,
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  )
+
+  const initPositionUnit = React.useMemo(
```
swansontec, line 78: Same thing here. `const [initPositionUnit] = React.useState( () => unitLabelValueList.findIndex(option => option.value === displayTime.measurement))`. Note the use of the function here - this means we will only evaluate the `findIndex` once on first render.

swansontec (approved):
I'm approving this, because all the requested changes are perf related and not correctness related. Still, please fix them.

src/components/modals/AutoLogoutModal.tsx
Jon-edge: I get a dependency loop for one of the Android datasets, but I'll remove the rest

PR #5310, opened 2024-10-16
Allow 0 amount entry in `SpendingLimitsScene`

Jon-edge:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208524802495873

### DIFF
--- a/src/components/scenes/SpendingLimitsScene.tsx
+++ b/src/components/scenes/SpendingLimitsScene.tsx
@@ -45,7 +45,7 @@ export const SpendingLimitsScene = (props: Props) => {
     const spendingLimits = {
       transaction: {
         isEnabled: transactionIsEnabled,
-        amount: parseFloat(transactionAmount)
+        amount: zeroString(transactionAmount) ? 0 : parseFloat(transactionAmount)
       }
     }
     await writeSpendingLimits(account, spendingLimits)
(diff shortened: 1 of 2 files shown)


src/components/scenes/SpendingLimitsScene.tsx
```diff
@@ -42,10 +42,11 @@ export const SpendingLimitsScene = (props: Props) => {
     const isAuthorized = await account.checkPassword(password)
     if (!isAuthorized) return showToast(lstrings.password_check_incorrect_password_title)
 
+    const isZeroAmount = zeroString(transactionAmount)
     const spendingLimits = {
       transaction: {
-        isEnabled: transactionIsEnabled,
-        amount: parseFloat(transactionAmount)
+        isEnabled: transactionIsEnabled && !isZeroAmount,
+        amount: isZeroAmount ? 0 : parseFloat(transactionAmount)
```
swansontec, line 49: We don't want to turn off spending limits if the amount is 0. If the transaction limits are enabled, and the amount is 0, then we simply want to block every spend, regardless of amount.

swansontec (changes requested):

swansontec (approved):

PR #5319, opened 2024-10-23
Jon/fix/uk-compliance

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208596052480583


src/components/Main.tsx
```diff
@@ -423,6 +425,16 @@ const EdgeTabs = () => {
 // -------------------------------------------------------------------------
 
 const EdgeAppStack = () => {
+  const [countryCode, setCountryCode] = React.useState<string | undefined>()
+
+  useAsyncEffect(
+    async () => {
+      setCountryCode(await getCountryCodeByIp())
```
swansontec, line 432: The `getCountryCodeByIp` function does a network call. We don't want to do a bunch of network calls as we navigate the app - it would be better to just do one call at boot-up, and then use the cached value everywhere. We already do this with `getFirstOpenInfo` - if we've successfully loaded the file, it just returns the cached value instead of loading it again.

Now that we are using the `getCountryCodeByIp` function in so many places, I believe we need the same treatment to avoid a major performance regression.

swansontec (changes requested):
The changes seem OK, but we will need a small tweak to get performance back up.

src/components/Main.tsx
Jon-edge: Yeah I was trying to come up with a new method to do it once and pass it around somehow without success, those are good ideas

src/actions/FirstOpenActions.tsx
```diff
@@ -27,9 +29,15 @@ export const getFirstOpenInfo = async (): Promise<FirstOpenInfo> => {
       firstOpenText = await firstOpenDisklet.getText(FIRST_OPEN)
       firstOpenInfo = asFirstOpenInfo(JSON.parse(firstOpenText))
       firstOpenInfo.isFirstOpen = 'false'
+
+      if (firstOpenInfo.countryCode == null) {
+        // Not critical if we can't get the country code
+        firstOpenInfo.countryCode = await getCountryCodeByIp().catch(() => undefined)
```
swansontec, line 35: Do we want to re-save here, so it's available on the next run?

swansontec (approved):
This will work

src/actions/FirstOpenActions.tsx
Jon-edge: No, then they're stuck as a UK-designated device if they happened to open the app in the UK when this change goes live.

PR #5322, opened 2024-10-24
Add swap error logging to Sentry

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206334009226435

### DIFF
--- a/src/util/tracking.ts
+++ b/src/util/tracking.ts
@@ -152,12 +152,14 @@ export function trackError(
     [key: string]: any
   }
 ): void {
-  let err: Error | string
-  if (error instanceof Error || typeof error === 'string') {
+  let err: Error
+  if (error instanceof Error) {
     err = error
+  } else if (typeof error === 'string') {
+    err = new Error(error)
   } else {
-    // At least send an error which should give us the callstack
-    err = 'Unknown error occurred'
+    // At least send an error which should give us the call-stack
+    err = new Error(`Unknown error occurred: ${String(error)}`)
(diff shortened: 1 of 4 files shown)


src/util/tracking.ts
```diff
@@ -152,19 +152,25 @@ export function trackError(
...
-  if (tag == null) {
-    captureException(err)
-  } else {
-    captureException(err, { event_id: tag, data: metadata })
+  let hint: { event_id?: string; data?: { [key: string]: any } } | undefined
+  if (tag != null) {
+    hint = { event_id: tag }
+  }
+  if (metadata != null) {
+    hint = { data: metadata }
```
swansontec, line 170: This doesn't let us have *both* a tag and a metadata at the same time. Maybe do something like this:

```ts
let hint: { event_id?: string; data?: { [key: string]: any } } = {}
if (tag != null) hint.event_id = tag
if (metadata != null) hint.data = metadata
captureException(err, tag != null || metadata != null ? hint : undefined)
```

swansontec (approved):

src/util/tracking.ts
swansontec: Oh, I see that the next PR completely re-does all this stuff, so nevermind.

samholmes: I think you mean to retract this feedback then. I'll resolve.

PR #5326, opened 2024-10-25
Log unhandled FiatProviderError errors

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-react-gui/pull/5322#pullrequestreview-2396571473

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206334009226436

### DIFF
--- a/src/components/services/EdgeCoreManager.tsx
+++ b/src/components/services/EdgeCoreManager.tsx
@@ -14,6 +14,7 @@ import { useAsyncEffect } from '../../hooks/useAsyncEffect'
 import { useHandler } from '../../hooks/useHandler'
 import { useIsAppForeground } from '../../hooks/useIsAppForeground'
 import { lstrings } from '../../locales/strings'
+import { addMetadataToContext } from '../../util/addMetadataToContext'
 import { allPlugins } from '../../util/corePlugins'
 import { fakeUser } from '../../util/fake-user'
 import { isMaestro } from '../../util/maestro'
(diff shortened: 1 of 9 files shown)


src/components/services/EdgeCoreManager.tsx
```diff
@@ -67,8 +67,55 @@ const crashReporter: EdgeCrashReporter = {
...
+        context[fullKey] += `\n    ${value.stack.replace(/\n/g, '\n    ')}`
+      }
+      continue
+    }
+
+    // Recurse over objects with added prefix keys:
+    if (typeof value === 'object' && value !== null) {
+      addMetadataToContext(context, value, allKeys)
+      continue
+    }
```
swansontec: Should we just `JSON.stringify` instead of recursing? Imagine you have something like this:

```
{
  fromCoin: { tokenId: 'asf', pluginId: 'asdf' },
  toCoin: { tokenId: 'dda', pluginId: 'asafadf' }
}
```

The fields in `toCoin` would stomp over the fields in `fromCoin`, and you would lose information. If we use `JSON.stringify` on the values, then you would get:

```
{
  fromCoin: '{"tokenId":"asf","pluginId":"asdf"}',
  toCoin: '{"tokenId":"dda","pluginId":"asafadf"}'
}
```

This would be more usable, I think.

swansontec (approved):
Approved, but I don't know if the recursion is a good idea.

src/components/services/EdgeCoreManager.tsx
samholmes: The reason why: JSON strings are big and when metadata fields are too big, they're filtered out by Sentry. So, I flattened the JSON objects into separate fields. Each field contains the parent key in it's name so there shoulnd't be possibility of key stomping

src/util/tracking.ts
```diff
@@ -160,11 +160,15 @@ export function trackError(
...
-  if (tag == null) {
-    captureException(err)
-  } else {
-    captureException(err, { event_id: tag, data: metadata })
+  let hint: { event_id?: string; data?: { [key: string]: any } } | undefined
+  if (tag != null) {
+    hint = { event_id: tag }
+  }
+  if (metadata != null) {
+    hint = { data: metadata }
```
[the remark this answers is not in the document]
samholmes: This logic is replaced because it turns out it didn't do what we expected it to do for the events we send to Sentry. In some cases, it harmed us.

PR #5330, opened 2024-10-28
Look for Android battery optimization setting

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208604046309472


src/components/navigation/AlertDropdown.tsx
```diff
@@ -42,13 +42,15 @@ export function AlertDropdown(props: Props) {
   })
 
   return (
-    <AirshipDropdown autoHideMs={persistent ? 0 : undefined} bridge={bridge} backgroundColor={color} onPress={handleOnPress}>
+    <AirshipDropdown autoHideMs={persistent ? 0 : undefined} bridge={bridge} backgroundColor={color}>
```
swansontec, line 45: I don't understand this change. Is the commit message wrong?

Edit: Ok, maybe the idea is that we want the X to close the modal, but tapping the text should open the link?

src/components/services/Services.tsx
```diff
@@ -128,49 +127,16 @@ export function Services(props: Props) {
...
-        if (subscription != null) {
-          subscription.remove()
-        }
-        if (airshipBridge != null) {
-          airshipBridge.resolve()
-        }
+      if (powerState.lowPowerMode) {
+        Airship.show(bridge => {
+          return <AlertDropdown bridge={bridge} message={lstrings.warning_battery_saver} warning persistent />
+        }).catch(e => console.error('Services 3 AlertDropdown', e))
```
swansontec: What is up with this `console.error`? We either use `showDevError`, or we ignore errors cleanly.

src/components/services/Services.tsx
```diff
@@ -127,14 +132,34 @@ export function Services(props: Props) {
     'Services 2'
   )
 
+  const closePowerModal = () => {
+    powerAlertRef.current?.resolve()
+    powerAlertRef.current = undefined
+  }
+
```
swansontec, line 139: I don't like the idea of closing the modal from the outside. I would much rather do the `.resolve()` from within the body of `Airship.show`. You could move the `onPress` inside the `Airship.show` pretty easily, but the other two cases of `closePowerModal()` are harder to get rid of. Why is this necessary? If it truly is necessary, maybe we could find another way to auto-close the modal, besides leaking the whole bridge out.

swansontec (changes requested):
The modal logic is a bit weird, but the native code changes make complete sense.

swansontec (approved):

PR #5333, opened 2024-10-30
Fix currencyWallets reference in create wallet paths

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208635507463521


swansontec (approved):
This is a good fix but we should really fix `WalletListResult` to include the actual `EdgeCurrencyWallet` object, instead of just the ID. That would solve the whole class of errors, and not just these specific instances.

PR #5348, opened 2024-11-07
Update Unstoppable domains support

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208623680992506

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -6,6 +6,7 @@
  */
 // import './wdyr'
 import * as Sentry from '@sentry/react-native'
+import { Buffer } from 'buffer'
 import { asObject, asString } from 'cleaners'
 import { LogBox, Text, TextInput } from 'react-native'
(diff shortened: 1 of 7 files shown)


src/app.ts
```diff
@@ -277,3 +278,5 @@ if (ENV.DEBUG_THEME) {
 
 initDeviceSettings().catch(err => console.log(err))
 initInfoServer().catch(err => console.log(err))
+
+global.Buffer = global.Buffer || Buffer
```
swansontec, line 282: This approach seems dangerous, so maybe don't touch the globals unless we need to:

```ts
if (global.Buffer == null) global.Buffer = Buffer
```

swansontec (approved):

src/app.ts
peachbits: changed

PR #5355, opened 2024-11-08
Lock Android to portrait mode

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208296057446451

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -43,6 +43,7 @@
     <activity
       android:name=".MainActivity"
       android:label="@string/app_name"
+      android:screenOrientation="portrait"
       android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
       android:exported="true"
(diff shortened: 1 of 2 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -43,6 +43,7 @@
     <activity
       android:name=".MainActivity"
       android:label="@string/app_name"
+      android:screenOrientation="portrait"
```
swansontec, line 46: Can we put comments in here? Since this is a template file that comes from React Native, we could easily delete this line during an upgrade. So, we need to say "Edge addition: Fix for resolution issue" or something like that.

swansontec (approved):

swansontec:
Per meeting, this won't work. We'd need to dynamically listen to resolution changes and refresh the theme. The Patina library has logic for handling theme changes dynamically.

PR #5356, opened 2024-11-11
More small fixes

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208741639032717

### DIFF
--- a/src/components/scenes/DefaultFiatSettingScene.tsx
+++ b/src/components/scenes/DefaultFiatSettingScene.tsx
@@ -76,8 +76,12 @@ export class DefaultFiatSettingComponent extends React.Component<Props, State> {
 
   render() {
     const filteredArray = this.props.supportedFiats.filter(entry => {
+      const key = `currency_label_${entry.value}`
+      const subTitle = lstrings[key as keyof typeof lstrings] ?? ''
+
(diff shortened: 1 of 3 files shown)


src/components/scenes/DefaultFiatSettingScene.tsx
```diff
@@ -76,8 +76,12 @@ export class DefaultFiatSettingComponent extends React.Component<Props, State> {
 
   render() {
     const filteredArray = this.props.supportedFiats.filter(entry => {
+      const key = `currency_label_${entry.value}`
+      const subTitle = lstrings[key as keyof typeof lstrings] ?? ''
+
```
swansontec, line 81: We usually build a table for this type of thing, so like:

```ts
const fiatTitles = {
  USD: lstrings.currency_label_usd,
  EUR: lstrings.currency_label_eur,
  ...
}
```

That was we can't "escape" into the broader strings table, because we do the lookup on a safe subset.

Now, while I would prefer to see it done this way, I won't insist on it because we are really close to the merge and we want to get this in.

swansontec (approved):

PR #5357, opened 2024-11-13
Fix incorrect rounded corners on iOS

Jon-edge:
This regressed as a result of changes to support rounded blur corners for iOS in the latest `NotificationCard` updates (66b4bdd).

image

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1208710423836257

### DIFF
--- a/src/components/common/BlurBackground.tsx
+++ b/src/components/common/BlurBackground.tsx
@@ -6,6 +6,7 @@ import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 
 const isAndroid = Platform.OS === 'android'
 
+/** A blur background WITH rounded corners, used for most components */
 export const BlurBackground = () => {
   const theme = useTheme()
   const styles = getStyles(theme)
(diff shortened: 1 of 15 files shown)


samholmes (changes requested):
Change the commit message:

```
Apply rounded corners only to NotificationCard

Rounded corners were applied to all implementations of BlurBackground
which caused the header and footer to have rounded corners. This was 
never apart of the design for the header and footer and accidentally 
slipped through as a bug because it never manifest itself in Android; 
it only appeared in iOS. Nonetheless, this fix will remove the rounded 
corner styles in the BlurBackground component by default, and provide a 
parameter to enable it for NotificationCard.
```

Add new lines at column 72.

src/components/common/BlurBackground.tsx
```diff
@@ -6,13 +6,21 @@ import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 
 const isAndroid = Platform.OS === 'android'
 
-export const BlurBackground = (props: { roundCorners?: boolean }) => {
-  const { roundCorners = false } = props
+export const BlurBackground = (props: {
+  /** If the parent requires rounded corners, this prop must be set true in
+   * order to ensure parity between iOS/Android */
+  noRoundedCorners?: boolean
```
samholmes, line 12: Let's go with rounded as an option instead of rounded by default. Or add a `RoundedBlurBackground` element which can be used in rounded components.

samholmes (changes requested):
Commit message comment still applies (or something like it)

samholmes (approved):
Approved. Though I wouldn't make the same assumption that _all components that would want a blurred background will be rounded_.

Jon-edge:
If that assumption fails in the future, then we just need to update the component's comment. I specifically made a generalized component name instead of naming `HeaderFooterBlur` or something

PR #5387, opened 2024-12-05
Deprecate Binance Beacon Chain

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208848967378659


swansontec (commented):
I'm a bit worried about the first commit, but since we are still early in staging, I think QA has a good chance of finding any fallout.

swansontec (approved):

PR #5389, opened 2024-12-06
Add 'numAccounts' to 'Signup_Complete' event

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208898501892435


swansontec (commented):
I see, so if the user creates a new account when they already have accounts on the device, we'd like to know about that.

PR #5391, opened 2024-12-09
Remove payment types for Banxa

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1206334009226449

### DIFF
--- a/src/plugins/gui/providers/banxaProvider.ts
+++ b/src/plugins/gui/providers/banxaProvider.ts
@@ -105,7 +105,6 @@ const asBanxaTxLimit = asObject({
 })
 
 const asBanxaPaymentType = asValue(
-  'CHECKOUTCREDIT',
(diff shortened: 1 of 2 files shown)


src/plugins/gui/providers/banxaProvider.ts
```diff
@@ -105,7 +105,6 @@ const asBanxaTxLimit = asObject({
 })
 
 const asBanxaPaymentType = asValue(
-  'CHECKOUTCREDIT',
```
paullinator, line 108: misunderstanding of the task. The goal is to remove the old paymentType but replace it with the new equivalent. See the Banxa chat

paullinator (changes requested):

paullinator (approved):

PR #5393, opened 2024-12-10
Jon/fix/empty-search

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1208782217930331

### DIFF
--- a/src/components/themed/SimpleTextInput.tsx
+++ b/src/components/themed/SimpleTextInput.tsx
@@ -255,13 +255,13 @@ export const SimpleTextInput = React.forwardRef<SimpleTextInputRef, SimpleTextIn
         </InputContainerView>
       </EdgeTouchableWithoutFeedback>
       {isIos && (isFocused || active === true) && (
-        <TouchContainer hitSlop={theme.rem(0.75)} accessible onPress={handleDonePress} testID={`${testID}.cancelButton`}>
+        <TouchableOpacity accessible onPress={handleDonePress} testID={`${testID}.cancelButton`}>
           <CancelButton>
             <CancelText numberOfLines={1} ellipsizeMode="clip">
               {lstrings.string_cancel_cap}
             </CancelText>
           </CancelButton>
-        </TouchContainer>
+        </TouchableOpacity>
       )}
     </ContainerView>
   )
--- a/src/components/themed/SimpleTextInput.tsx
+++ b/src/components/themed/SimpleTextInput.tsx
@@ -52,11 +52,20 @@ export interface SimpleTextInputProps extends MarginRemProps {
   secureTextEntry?: boolean // Defaults to 'false'
   testID?: string
 
-  // Unless 'autoFocus' is passed explicitly in the props, Search Bars 'autoFocus' and 'regular' text inputs don't.
+  /** Unless 'autoFocus' is passed explicitly in the props, Search Bars
+  'autoFocus' and 'regular' text inputs don't. */
   autoFocus?: boolean // Defaults to 'true'
 
-  // Unless 'blurOnClear' is passed explicitly in the props, Search Bars calls 'blur' when cleared and text inputs don't call 'blur' when cleared.
+  /** Unless 'blurOnClear' is passed explicitly in the props, Search Bars calls
+   * 'blur' when cleared and text inputs don't call 'blur' when cleared. */
   blurOnClear?: boolean // Defaults to 'false'
+
+  /**
+   * Manually control whether the input appears selected. This is only a
+   * visual change. It's mutually exclusive with the text input's true
+   * blur/focus state.
+   * */
+  active?: boolean
 
   // Whether the text input is disabled. If 'true', the component will be grayed out.
   disabled?: boolean // Defaults to 'false'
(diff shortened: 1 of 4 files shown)


src/components/themed/SimpleTextInput.tsx
```diff
@@ -52,12 +52,18 @@ export interface SimpleTextInputProps extends MarginRemProps {
...
+  'autoFocus' and 'regular' text inputs don't. */
   autoFocus?: boolean // Defaults to 'true'
 
-  // Unless 'blurOnClear' is passed explicitly in the props, Search Bars calls 'blur' when cleared and text inputs don't call 'blur' when cleared.
+  /** Unless 'blurOnClear' is passed explicitly in the props, Search Bars calls
+   * 'blur' when cleared and text inputs don't call 'blur' when cleared. */
   blurOnClear?: boolean // Defaults to 'false'
 
+  /** Manually control the whether the input appears selected. Visual only,
+   * mutually exclusive with the text input's true blur/focus state. */
```
samholmes, line 64: fixup to the grammar

```js
/** Manually control whether the input appears selected. Visual only; mutually exclusive with the text input's true blur/focus state. */
```

src/components/themed/SimpleTextInput.tsx
```diff
@@ -52,12 +52,18 @@ export interface SimpleTextInputProps extends MarginRemProps {
...
+  selected?: boolean
```
samholmes, line 65: I thought were were going with `active`? The problem with `selected` is that it is used as a word for selector input types, so it seems misplaced here.

src/components/themed/SimpleTextInput.tsx
```diff
@@ -104,6 +110,7 @@ export const SimpleTextInput = React.forwardRef<SimpleTextInputRef, SimpleTextIn
     maxLength,
     returnKeyType,
     secureTextEntry,
+    selected,
```
samholmes, line 113: The logic below becomes simpler if we set a default to `false`.

src/components/themed/SimpleTextInput.tsx
```diff
@@ -236,7 +251,7 @@ export const SimpleTextInput = React.forwardRef<SimpleTextInputRef, SimpleTextIn
           </TouchContainer>
         </InputContainerView>
       </EdgeTouchableWithoutFeedback>
-      {isIos && isFocused && (
+      {isIos && (isFocused || selected) && (
```
samholmes, line 254: `selected` is not always boolean (unless you take my suggestion to give it a default). So `selected === true` is lint 👌.

src/components/themed/SimpleTextInput.tsx
```diff
@@ -187,6 +197,11 @@ export const SimpleTextInput = React.forwardRef<SimpleTextInputRef, SimpleTextIn
     return disabled ? theme.textInputPlaceholderColorDisabled : isFocused ? theme.textInputPlaceholderColorFocused : theme.textInputPlaceholderColor
   }, [disabled, isFocused, theme.textInputPlaceholderColor, theme.textInputPlaceholderColorDisabled, theme.textInputPlaceholderColorFocused])
 
+  React.useEffect(() => {
+    if (selected == null) return
+    focusAnimation.value = selected ? withTiming(1, { duration: baseDuration }) : withDelay(animationDelay, withTiming(0, { duration: baseDuration }))
+  }, [selected, focusAnimation, baseDuration, animationDelay])
```
samholmes, line 203: Would `useDerivedValue` be the right choice of a hook to bind `focusAnimation` to the `selected` (hopefully `active`) prop?

samholmes (changes requested):
questions/feedback

src/components/themed/SimpleTextInput.tsx
Jon-edge: ChatGPT said this was fine the way it was

src/components/themed/SimpleTextInput.tsx
Jon-edge: I remember we agreed the other way around... Don't think selector input types was the rationale but I can't think of it now so I'll just change it.

src/components/themed/SimpleTextInput.tsx
Jon-edge: No, this was intentional  because we use the `null` state in various changes to hijack the "normal" handling in favor of the `active` prop.

src/components/themed/SimpleTextInput.tsx
Jon-edge: Keeping in mind the `null` case noop is necessary, then it would error about using its own value in its initializer if I converted this into `useDerivedValue`

src/components/themed/SimpleTextInput.tsx
samholmes: ChatGPT is wrong.

samholmes (approved):
Approved, but please fix the comment the whether the ChatGPT says it's okay or not.

PR #5397, opened 2024-12-13
ADA Staking Fixes

samholmes:
- **Remove unused staking lstrings**
- **Redesign StakeOptionScene**
- **Refactor staking state into Redux**
- **Fix stake position liveliness on StakeOverviewScene**
- **Block UI and fetch stake position after modify**

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1208312383224873

Simulator Screenshot - iPhone 15 Pro - 2024-12-13 at 13 38 48
Simulator Screenshot - iPhone 15 Pro - 2024-12-13 at 13 39 39
Simulator Screenshot - iPhone 15 Pro - 2024-12-13 at 13 39 42


src/reducers/scenes/StakingReducer.ts
```diff
@@ -0,0 +1,148 @@
...
+}
+
+export type StakingAction =
+  | AddStakePolicyAction
+  | FinishAction
+  | UpdateLockedAmountAction
+  | UpdatePluginsAction
+  | UpdatePoliciesAction
+  | UpdatePositionAction
+  | UpdatePositionsAction
```
swansontec: I would prefer to put the action types inline, like this:

```ts
export type StakingAction =
| {
  type: 'STAKING/UPDATE_POSITIONS',
  ...
}
| { 
    type: 'STAKING/UPDATE_POLICIES'
    ...
}
```

This saves us from having to invent a bunch of names that aren't used anywhere else like `UpdatePositionsAction`, and therefore reduces the cognitive load / boilerplate.

Since we are stuck with Redux for the time being, our general goal has been to reduce as much boilerplate as much as possible to make life tolerable.

src/reducers/scenes/StakingReducer.ts
```diff
@@ -0,0 +1,148 @@
...
+export const staking: Reducer<StakingState, Action> = combineReducers({
+  walletStakingMap: (state = {}, action: Action) => {
+    switch (action.type) {
+      case 'STAKING/ADD_POLICY':
+      case 'STAKING/FINISH_LOADING':
+      case 'STAKING/UPDATE_LOCKED_AMOUNT':
+      case 'STAKING/UPDATE_PLUGINS':
+      case 'STAKING/UPDATE_POLICIES':
+      case 'STAKING/UPDATE_POSITION':
+      case 'STAKING/UPDATE_POSITIONS':
```
swansontec: Optional: This could be `if (action.type.startsWith('STAKING/'))`, which might be more sustainable. This is the approach the core uses with its wallet actions, for instance.

src/reducers/scenes/StakingReducer.ts
swansontec, line 1: This whole file is in the wrong place. It should go in `src/reducers`.

src/components/themed/TransactionListTop.tsx
```diff
@@ -813,6 +747,10 @@ export function TransactionListTop(props: OwnProps) {
   const account = useSelector(state => state.core.account)
   const exchangeRates = useSelector(state => state.exchangeRates)
   const defaultIsoFiat = useSelector(state => state.ui.settings.defaultIsoFiat)
+  const walletStakingState: WalletStakingState = useSelector(
+    // Fallback to a default state using the reducer if the wallet is not found
+    state => state.staking.walletStakingMap[wallet.id] ?? walletStakingStateReducer(undefined, { type: '' } as any)
```
swansontec, line 752: Could you do `export defaultWalletStakingState = walletStakingStateReducer(undefined, { type: 'DUMMY_ACTION_PLEASE_IGNORE' })` in `StakingReducer.ts`, and then use that constant? Then we don't have to re-compute this value on every render. Also, we have the `DUMMY_ACTION_PLEASE_IGNORE` type defined for this exact use-case.

src/components/themed/TransactionListTop.tsx
```diff
@@ -113,99 +112,40 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
...
       let total = '0'
       let lockedNativeAmount = '0'
-      for (const stakePosition of Object.values(this.state.stakePositionMap)) {
+      for (const stakePosition of Object.values(this.props.walletStakingState.stakePositionMap)) {
         const { staked, earned } = getPositionAllocations(stakePosition)
         total = this.getTotalPosition(this.props.currencyCode, [...staked, ...earned])
         lockedNativeAmount = add(lockedNativeAmount, total)
       }
-      this.setState({ lockedNativeAmount })
+      this.props.dispatch({ type: 'STAKING/UPDATE_LOCKED_AMOUNT', walletId: this.props.wallet.id, lockedNativeAmount: lockedNativeAmount })
```
swansontec, line 132: `lockedNativeAmount: lockedNativeAmount` could be just `lockedNativeAmount`

src/components/themed/TransactionListTop.tsx
```diff
@@ -85,12 +86,14 @@ interface StateProps {
   balanceMap: EdgeBalanceMap
   currencyCode: string
   defaultFiat: string
+  dispatch: Dispatch
```
swansontec, line 89: This is a bit unconventional - normally we use `mapDispatchToProps` for this situation, but I can't see any reason why it's bad, so yay for finding simple solutions. 👍

src/components/scenes/Staking/StakeOptionsScene.tsx
```diff
@@ -117,7 +115,7 @@ const StakeOptionsSceneComponent = (props: Props) => {
       {({ undoInsetStyle, insetStyle }) => (
         <SceneContainer undoBottom undoInsetStyle={undoInsetStyle}>
           <FlatList
-            data={stakePolicies}
+            data={Object.values(stakePolicies)}
```
swansontec, line 118: I think this will cause a full list refresh on every render. You need to do `const policyList = React.useMemo(()=> Object.values(stakePolicies), [stakePolicies])` to cache the value, so the list only gets new `data` if `stakePolicies` actually changes.

src/components/scenes/Staking/EarnScene.tsx
```diff
@@ -290,13 +288,16 @@ export const EarnScene = (props: Props) => {
       }
 
       // User backed out of the WalletListModal
-      if (walletId == null) return
+      if (walletId == null || stakePosition == null) return
+
+      dispatch({ type: 'STAKING/ADD_POLICY', walletId, stakePolicy })
+      dispatch({ type: 'STAKING/UPDATE_POSITION', walletId, stakePolicyId: stakePolicy.stakePolicyId, stakePosition })
+      dispatch({ type: 'STAKING/FINISH_LOADING', walletId })
```
swansontec, line 295: A repeated `dispatch` calls can tank performance. Maybe make a `{ type: 'STAKING/UPDATE', stakePolicy, stakePosition, walletId }` action that can configure all three values in one shot.

src/components/themed/TransactionListTop.tsx
swansontec, line 132: `lockedNativeAmount: lockedNativeAmount` could be just `lockedNativeAmount`

src/actions/scene/StakingActions.tsx
```diff
@@ -0,0 +1,82 @@
...
+    for (const stakePlugin of stakePlugins) {
+      const stakePolicies = stakePlugin.getPolicies({ pluginId, wallet, currencyCode })
+      for (const stakePolicy of stakePolicies) {
+        stakePolicyMap[stakePolicy.stakePolicyId] = stakePolicy
+        let total: string | undefined
+        try {
+          const stakePosition = await stakePlugin.fetchStakePosition({
+            stakePolicyId: stakePolicy.stakePolicyId,
+            wallet: wallet,
+            account: account
```
swansontec, line 42: Could just be `wallet, account`

src/actions/scene/StakingActions.tsx
```diff
@@ -0,0 +1,82 @@
...
+          lockedNativeAmount = add(lockedNativeAmount, total)
+        }
+      }
+    }
+
+    dispatch({ type: 'STAKING/UPDATE_LOCKED_AMOUNT', walletId, lockedNativeAmount })
+    dispatch({ type: 'STAKING/UPDATE_PLUGINS', walletId, stakePlugins })
+    dispatch({ type: 'STAKING/UPDATE_POLICIES', walletId, stakePolicies: stakePolicyMap })
+    dispatch({ type: 'STAKING/UPDATE_POSITIONS', walletId, stakePositionMap })
+    dispatch({ type: 'STAKING/FINISH_LOADING', walletId })
```
swansontec, line 68: Same thing here. Just do one dispatch with some kind of uber-action, like `{ type: 'STAKING/SETUP', walletId, stakePlugins, stakePolicyMap, stakePositionMap }`.

Imagine that each call to dispatch has about ~100ms of overhead, so you are potentially burning half a second here.

src/components/themed/TransactionListTop.tsx
```diff
@@ -113,99 +112,40 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
...
         lockedNativeAmount = add(lockedNativeAmount, total)
       }
-      this.setState({ lockedNativeAmount })
+      this.props.dispatch({ type: 'STAKING/UPDATE_LOCKED_AMOUNT', walletId: this.props.wallet.id, lockedNativeAmount: lockedNativeAmount })
     }
   }
 
   componentDidMount() {
-    this.updatePluginsAndPolicies().catch(err => showError(err))
+    this.props.dispatch(updateStakingState(this.props.currencyCode, this.props.wallet)).catch(err => showError(err))
```
swansontec, line 137: Out of scope for this PR, but I wonder if we can be "smarter" about when we do these updates, now that the state management is more under control.

src/components/scenes/Staking/StakeOverviewScene.tsx
```diff
@@ -67,35 +68,41 @@ const StakeOverviewSceneComponent = (props: Props) => {
   const [rewardAllocations, setRewardAllocations] = React.useState<PositionAllocation[]>([])
   const [unstakedAllocations, setUnstakedAllocations] = React.useState<PositionAllocation[]>([])
 
-  // Background loop to force fetchStakePosition updates
-  const [updateCounter, setUpdateCounter] = React.useState<number>(0)
+  const updatePosition = React.useCallback(() => {
+    dispatch(updateStakingPosition(stakePlugin, stakePolicyId, wallet, account)).catch(err => {
+      console.error(err)
```
swansontec: Remove this logging, since the `showError` function handles everything.

src/components/scenes/Staking/StakeOverviewScene.tsx
```diff
@@ -67,35 +68,41 @@ const StakeOverviewSceneComponent = (props: Props) => {
...
+      updatePosition()
+    }
+  }, [isFocused, updatePosition])
+
+  // Update the position every minute
   React.useEffect(() => {
     const interval = setInterval(() => {
-      setUpdateCounter(updateCounter => updateCounter + 1)
-    }, 10 * 1000) // ten seconds
+      updatePosition()
```
swansontec, line 98: Never use use `setInterval` (in general), but *especially* when doing async work. There is no throttling, so tasks can pile up on top of each other and cause a big train-wreck.  Use `makePeriodicTask` instead, which correctly handles all the weird async race conditions.

src/components/scenes/Staking/StakeOverviewScene.tsx
```diff
@@ -67,35 +68,41 @@ const StakeOverviewSceneComponent = (props: Props) => {
   const [rewardAllocations, setRewardAllocations] = React.useState<PositionAllocation[]>([])
   const [unstakedAllocations, setUnstakedAllocations] = React.useState<PositionAllocation[]>([])
 
-  // Background loop to force fetchStakePosition updates
-  const [updateCounter, setUpdateCounter] = React.useState<number>(0)
+  const updatePosition = React.useCallback(() => {
```
swansontec, line 78: Remove this `useCallback` wrapper, since this function is just used by effects, and not passed as a prop. We only need to memoize when props are involved. In fact, you could probably just inline the `dispatch` call into the two effects, since it's simple enough.

swansontec (changes requested):
I like the direction this is going, but there are a lot of smaller things to tidy up.

src/components/themed/TransactionListTop.tsx
samholmes: I originally was passing in callbacks for each dispatch, then realized the redundancy. As this component is slowly transitioned into a function-component, then the prop goes away.

src/reducers/StakingReducer.ts
```diff
@@ -0,0 +1,157 @@
...
+export type StakingAction =
+  | {
+      type: 'STAKING/ADD_POLICY'
+      walletId: string
+      stakePolicy: StakePolicy
+    }
+  | {
+      type: 'STAKING/FINISH_LOADING'
+      walletId: string
+    }
```
swansontec, line 15: I don't think this action type is used anymore, and same for many of the others. It looks like you really just to setup, update, and update_locked_amount now.

swansontec (approved):
Approved, but you could clean up unused actions.

src/reducers/StakingReducer.ts
samholmes: This action is used for loading empty staking data (no staking enabled) it's also used to end re-loading started by START_LOADING independently from UPDATE_POSITION

PR #5412, opened 2024-12-24
Disallow deprecated wallets from parsing links

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208022684274583

### DIFF
--- a/src/actions/DeepLinkingActions.tsx
+++ b/src/actions/DeepLinkingActions.tsx
@@ -7,6 +7,7 @@ import { FundAccountModal } from '../components/modals/FundAccountModal'
 import { pickWallet } from '../components/modals/WalletListModal'
 import { Airship, showError, showToast, showToastSpinner } from '../components/services/AirshipInstance'
 import { guiPlugins } from '../constants/plugins/GuiPlugins'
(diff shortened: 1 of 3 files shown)


src/actions/DeepLinkingActions.tsx
```diff
@@ -226,6 +227,9 @@ async function handleLink(navigation: NavigationBase, dispatch: Dispatch, state:
       const parseWallets = async (): Promise<void> => {
         // Try to parse with all wallets
         for (const wallet of Object.values(currencyWallets)) {
+          // Ignore disabled wallets:
+          if (SPECIAL_CURRENCY_INFO[wallet.currencyInfo.pluginId].keysOnlyMode) return
```
swansontec, line 231: Since keysOnlyMode can be undefined, we should do `const { keysOnlyMode = false } = SPECIAL_CURRENCY_INFO`, and then you can do if (keysOnlyMode) on the next line. That makes the default value explicit.

swansontec (approved):
This makes sense

Farnoosh85 (commented):
Base-event

PR #5413, opened 2024-12-27
Avoid bridge calls for currency conversion

swansontec:
These are just a simple utility functions. There is no need to go over the bridge to access this logic.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209050250699586


src/actions/ScanActions.tsx
```diff
@@ -298,7 +298,8 @@ async function sweepPrivateKeys(state: RootState, account: EdgeAccount, navigati
 
     // Check for a $50 maximum sweep for light accounts:
     const sendNativeAmount = abs(unsignedTx.nativeAmount)
-    const sendExchangeAmount = await wallet.nativeToDenomination(sendNativeAmount, wallet.currencyInfo.currencyCode)
+    const multiplier = getCurrencyCodeMultiplier(wallet.currencyConfig, wallet.currencyInfo.currencyCode)
+    const sendExchangeAmount = div(sendNativeAmount, multiplier, multiplier.length)
```
samholmes, line 302: Is this `multiplier.length` or `multiplier.length-1`? What happens if we do too much precision (more decimals then expected?)?

src/components/cards/FiatExchangeDetailsCard.tsx
```diff
@@ -113,20 +114,17 @@ export function FiatExchangeDetailsCard(props: Props) {
...
+    const multiplier = getCurrencyCodeMultiplier(wallet.currencyConfig, currencyCode)
+    const exchangeAmount = nativeAmount != null ? div(nativeAmount, multiplier, multiplier.length) : ''
+    if (direction === 'buy') {
+      setSourceAmount(`${fiatAsset.fiatAmount} `)
+      setDestinationAmount(`${exchangeAmount} `)
+    } else {
+      setDestinationAmount(`${fiatAsset.fiatAmount} `)
+      setSourceAmount(`${exchangeAmount} `)
+    }
+  })
```
samholmes, line 127: We don't even need the "useMount" hook nor the component state for sourceAmount/destinationAmount. Instead we can just define `sourceAmount` and `destinationAmount` as local variables.

src/plugins/gui/providers/ioniaProvider.ts
```diff
@@ -422,7 +423,8 @@ export const makeIoniaProvider: FiatProviderFactory<IoniaMethods> = {
         if (quoteParams.wallet == null) throw new Error('missing wallet')
 
         const rateAmount = await getCardPurchaseRateAmount(quoteParams.displayCurrencyCode, RATE_QUOTE_CARD_AMOUNT)
-        const rateExchangeAmount = await quoteParams.wallet.nativeToDenomination(toBigNumberString(rateAmount), quoteParams.displayCurrencyCode)
+        const multiplier = getCurrencyCodeMultiplier(quoteParams.wallet.currencyConfig, quoteParams.displayCurrencyCode)
+        const rateExchangeAmount = div(toBigNumberString(rateAmount), multiplier, multiplier.length)
```
samholmes, line 427: I don't think toBigNumberString is needed anymore now that biggystring has been updated; `.toString()` should be safe because all biggystring functions should accept scientific notation strings IIRC. It's worth a double take before removing `toBigNumberString` however.

src/util/CurrencyInfoHelpers.ts
```diff
@@ -82,6 +82,26 @@ export const getTokenIdForced = (account: EdgeAccount, pluginId: string, currenc
...
+export function getCurrencyCodeMultiplier(currencyConfig: EdgeCurrencyConfig, currencyCode: string): string {
+  const { currencyInfo, allTokens } = currencyConfig
+  for (const denomination of currencyInfo.denominations) {
+    if (denomination.name === currencyCode) {
+      return denomination.multiplier
+    }
+  }
+
+  for (const tokenId of Object.keys(allTokens)) {
+    const token = allTokens[tokenId]
```
samholmes, line 94: You could just use `Object.values` and avoid the `tokenId` allocation altogether.

src/util/CurrencyInfoHelpers.ts
```diff
@@ -82,6 +82,26 @@ export const getTokenIdForced = (account: EdgeAccount, pluginId: string, currenc
...
+    for (const denomination of token.denominations) {
+      if (denomination.name === currencyCode) {
+        return denomination.multiplier
+      }
+    }
+  }
+
+  return '1'
```
samholmes, line 102: Silently failing is a bit scary for a function like this. I get why to do it (avoid dealing with nullish cases), but by being lazy, I worry about a bad bug that communicates the wrong amounts to a user, or worse (sends the wrong user input amount to an API).

samholmes: Might I suggest we implement `nativeToDenomination` and `denominationToNative` in the GUI which will wrap this function with the null-checks necessary to be safe? What did these function do in the core when an invalid currencyCode was given? If they silently failed by not converting, then I guess we're not _increasing_ bug surface area, but if they threw an error, then we should perhaps consider some exceptions in the GUI as well.

samholmes (changes requested):
Feedback to be discussed.

src/util/CurrencyInfoHelpers.ts
swansontec: This is an exact copy-paste of the core logic, including the silent failure case. See https://github.com/EdgeApp/edge-core-js/blob/25fc3744fa1275b9ebabc75e3372f05926244e03/src/core/currency/currency-selectors.ts#L8 My goal was to make this a straight move, and not to introduce any new behavior.

The *correct* solution is to make everything based on tokenId's, and not currency codes, and so the new `nativeToDenomination` and `denominationToNative` need to do that, along with the appropriate error handling. However, that will be a follow-on task at best.

src/actions/ScanActions.tsx
samholmes: Same implementation is okay by me.

src/util/CurrencyInfoHelpers.ts
samholmes: Same implementation is okay by me.

src/hooks/useFiatText.ts
```diff
@@ -107,7 +106,7 @@ export const formatFiatString = (props: {
  * Returns a localized fiat amount string
  * */
 export const displayFiatAmount = (fiatAmount?: number | string, precision: number = 2, noGrouping: boolean = true) => {
-  const fiatAmountBns = fiatAmount != null ? toBigNumberString(fiatAmount) : undefined
+  const fiatAmountBns = fiatAmount != null ? fiatAmount : undefined
```
samholmes, line 109: This becomes a noop. Also the conditions below are defunct from the type change from string to number.. Either change the condition or you could do `const fiatAmountBns = fiatAmount != null ? fiatAmount.toString() : undefined` as a lazy fix.

samholmes (changes requested):
One fixup needed. Also remember to change commit title:

```
Remove toBigNumberString
```

src/hooks/useFiatText.ts
swansontec: Sorry, my fixup accidentally got merged in. I replaced the call to `toBigNumberString` with a call to `add( , '0')`, which should do the same thing.

samholmes (approved):

PR #5414, opened 2024-12-27
Simplify Thorchain savers fetching

swansontec:
It's not clear what the problem is, so it's not clear if this PR will fix anything. What it *will* do is dramatically improve performance, since we fetch far less data now.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209000958628643

### DIFF
--- a/src/plugins/stake-plugins/thorchainSavers/tcSaversPlugin.ts
+++ b/src/plugins/stake-plugins/thorchainSavers/tcSaversPlugin.ts
@@ -145,8 +145,7 @@ const asThorNodePool = asObject({
 const asThorNodePools = asArray(asThorNodePool)
 
 type Saver = ReturnType<typeof asSaver>
-type Savers = ReturnType<typeof asSavers>
-type Pools = ReturnType<typeof asPools>
+type Pool = ReturnType<typeof asPool>
 type ExchangeInfo = ReturnType<typeof asThorchainExchangeInfo>
 type InboundAddresses = ReturnType<typeof asInboundAddresses>
 interface StakePositionWithPoolsParams {
@@ -154,8 +153,8 @@ interface StakePositionWithPoolsParams {
   currencyCode: string
   primaryAddress: string
   pluginId: string
-  pools: Pools
-  savers: Savers
+  pools: Pool[]
+  savers: Saver[]
   saver?: Saver // Don't bother looking to match from Savers if this is passed in
   wallet: EdgeCurrencyWallet
 }
@@ -325,8 +324,8 @@ const getStakePositionInner = async (
 
   const asset = edgeToTcAsset(account, wallet, currencyCode)
 
-  let pools: Pools = []
-  let savers: Savers = []
+  let pools: Pool[] = []
+  let savers: Saver[] = []
   const [saversResponse, poolsResponse] = await Promise.all([
     fetchWaterfall(thornodeServers, `thorchain/pool/${asset}/savers`, {
       headers: { 'x-client-id': ninerealmsClientId }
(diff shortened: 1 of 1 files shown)


src/plugins/stake-plugins/thorchainSavers/tcSaversPlugin.ts
```diff
@@ -317,90 +306,92 @@ const getStakePosition = async (opts: EdgeGuiPluginOptions, request: StakePositi
...
-    earnedAmount = toFixed(earnedAmount, 0, 0)
-
-    // Cap negative value to 0
-    earnedAmount = max(earnedAmount, '0')
+function saverToPosition(currencyConfig: EdgeCurrencyConfig, currencyCode: string, saver: Saver, pool: Pool): StakePosition {
+  const pluginId = currencyConfig.currencyInfo.pluginId
+
+  const multiplier = getCurrencyCodeMultiplier(currencyConfig, currencyCode)
+  function thorToNative(amount: string): string {
+    return toFixed(mul(div(amount, THOR_LIMIT_UNITS, DIVIDE_PRECISION), multiplier), 0, 0)
```
samholmes, line 383: Nice.

src/plugins/stake-plugins/thorchainSavers/tcSaversPlugin.ts
```diff
@@ -317,90 +306,92 @@ const getStakePosition = async (opts: EdgeGuiPluginOptions, request: StakePositi
...
   }
 
+  const { units, asset_deposit_value: assetDepositValue } = saver
+  const stakedAmount = thorToNative(assetDepositValue)
   const canUnstake = gt(stakedAmount, '0')
 
+  const { saversDepth, saversUnits } = pool
+  const redeemableValue = div(mul(units, saversDepth), saversUnits, DIVIDE_PRECISION)
+  const earnedThorAmount = max(sub(redeemableValue, assetDepositValue), '0')
```
samholmes, line 392: Now that the other consts use this nice `thorToNative` to clearly describe the semantic behind the const value, I kinda want to know what's going on with these two. Maybe some comments would be nice to explain all the operations.

swansontec: I have no idea. These seem to be in "Thor units", but I don't understand the redeemable value calculation at all. The earned value seems to be the difference between what we deposited and what we can redeem, which makes sense. The `max` ensures our earned amount can't be negative.

My only goal here was to preserve the logic as it was, without any changes.

src/plugins/stake-plugins/thorchainSavers/tcSaversPlugin.ts
```diff
@@ -317,90 +306,92 @@ const getStakePosition = async (opts: EdgeGuiPluginOptions, request: StakePositi
...
+  if (saver == null || pool == null) {
+    // We got a 404 error, so return an empty position:
+    return {
+      allocations: [
+        {
+          pluginId: wallet.currencyInfo.pluginId,
+          currencyCode,
+          allocationType: 'staked',
+          nativeAmount: '0'
+        }
```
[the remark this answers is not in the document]
swansontec: I noticed that other stake plugins would return a 0 staked amount, rather than an empty array, so I copied the logic out of caution.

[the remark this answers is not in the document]
samholmes: It's fine to leave it as you have it.

src/plugins/stake-plugins/thorchainSavers/tcSaversPlugin.ts
samholmes: No change needed then. If you want, you can add a WTF comment like "// TODO: Describe what this math is mathing"

PR #5421, opened 2025-01-09
Survey Modal 2.0

Jon-edge:
Avoiding too much refactoring because this is soon subject to change.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209065058519199

### DIFF
--- a/src/components/modals/SurveyModal.tsx
+++ b/src/components/modals/SurveyModal.tsx
@@ -1,4 +1,5 @@
-import React, { useState } from 'react'
+import { InstallSurvey2 } from 'edge-info-server'
+import React from 'react'
 import { Platform, View } from 'react-native'
 import { AirshipBridge } from 'react-native-airship'
 import { KeyboardAwareScrollView } from 'react-native-keyboard-aware-scroll-view'
@@ -6,6 +7,7 @@ import Animated, { Easing, useAnimatedStyle, useSharedValue, withTiming } from '
 import { sprintf } from 'sprintf-js'
 
 import { useHandler } from '../../hooks/useHandler'
+import { getLocaleOrDefaultString } from '../../locales/intl'
 import { lstrings } from '../../locales/strings'
 import { config } from '../../theme/appConfig'
 import { useDispatch } from '../../types/reactRedux'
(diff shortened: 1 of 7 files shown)


src/components/modals/SurveyModal.tsx
```diff
@@ -6,47 +6,77 @@ import Animated, { Easing, useAnimatedStyle, useSharedValue, withTiming } from '
...
-  { label: lstrings.survey_opt_youtube, selected: false },
-  { label: lstrings.survey_opt_search_engine, selected: false },
-  { label: lstrings.survey_opt_x_twitter, selected: false },
-  { label: lstrings.survey_opt_in_person_event, selected: false },
-  { label: lstrings.survey_opt_personal_referral, selected: false },
-  { label: lstrings.survey_opt_article, selected: false }
+interface SurveyCategory {
+  catKey: string
+  label: string
+  selected: boolean
```
swansontec, line 27: I'm not going to be too picky, since this is temporary anyhow, but this `selected` flag should not exist. Everything should use the `selectedIndex` instead, which is actual component state.

However, no need to change anything now. Just make sure the later rewrite includes this fix.

src/components/modals/SurveyModal.tsx
```diff
@@ -71,28 +101,149 @@ export const SurveyModal = (props: { bridge: AirshipBridge<void> }) => {
...
+  )
+}
+
+const SurveyModal2 = (props: { bridge: AirshipBridge<boolean>; category: string }) => {
+  const { bridge, category } = props
+  const dispatch = useDispatch()
+  const theme = useTheme()
+  const styles = getStyles(theme)
+
+  const [options, setOptions] = useState([
```
swansontec, line 196: You are going to re-caculate this initial value on every render. Very inefficient. Instead, do `useState(() => [...`, putting a function wrapper around the initializer. Then it will only run once at mount.

src/components/modals/SurveyModal.tsx
```diff
@@ -6,47 +6,77 @@ import Animated, { Easing, useAnimatedStyle, useSharedValue, withTiming } from '
...
   const theme = useTheme()
   const styles = getStyles(theme)
 
   // Randomize response options, excluding "Other"
-  const randomizedSurveyOpts = [
-    ...shuffleArray([...SURVEY_OPTS, ...(infoServerData.rollup?.installSurvey?.surveyOptions ?? []).map(option => ({ label: option, selected: false }))]),
-    { label: lstrings.survey_opt_other_specify, selected: false }
-  ]
-  const [options, setOptions] = useState(randomizedSurveyOpts)
+  const [options, setOptions] = useState([...shuffleArray(SURVEY_CATS), { catKey: 'other', label: lstrings.survey_opt_other_specify, selected: false }])
```
swansontec, line 71: Same thing. You are going to re-caculate this initial value on every render, so do `useState(() => [...`, to make it only run once at mount.

swansontec (approved):
Approved, since this is temporary in any case, but there are some glaring cleanups that should get done.

src/components/modals/SurveyModal.tsx
```diff
@@ -24,71 +26,59 @@ import { EdgeModal } from './EdgeModal'
...
+  /**
+   * Shuffle the categories, then append an “other” category at the end:
+   */
+  const [options] = React.useState<SurveyCategory[]>(() => [...shuffleArray(SURVEY_CATS), { catKey: 'other', label: lstrings.survey_opt_other_specify }])
+
+  /**
+   * Store the user's selection here (rather than on each item).
+   * This can be `undefined`, or one of the category keys (like "ads", "other", etc.)
+   */
+  const [selectedCatKey, setSelectedCatKey] = React.useState<string | undefined>()
```
swansontec, line 62: It could be `SurveyOption2Keys | 'other' | undefined` instead of `string | undefined`. That would eliminate the `as SurveyOption2Keys` below in `handleNextPress`.

Making this change might require other place to tighten up their types too. This could be annoying, so if you can't get it done in 15 minutes of effort or so, just leave it be.

src/components/modals/SurveyModal.tsx
```diff
@@ -187,36 +185,38 @@ export const SurveyModal = (props: { bridge: AirshipBridge<void> }) => {
   )
 }
 
-const SurveyModal2 = (props: { bridge: AirshipBridge<boolean>; category: string }) => {
+const SurveyModal2 = (props: { bridge: AirshipBridge<boolean>; category: keyof InstallSurvey2['surveyOptions2'] }) => {
```
swansontec, line 188: Your could use the `SurveyOption2Keys` type instead of `keyof InstallSurvey2['surveyOptions2']`, since it's the same thing.

src/util/network.ts
```diff
@@ -6,7 +6,7 @@ import { getVersion } from 'react-native-device-info'
 
 import { config } from '../theme/appConfig'
 import { asyncWaterfall, getOsVersion, shuffleArray } from './utils'
-const INFO_SERVERS = ['https://info1.edge.app', 'https://info2.edge.app']
+const INFO_SERVERS = ['http://localhost:8008']
```
swansontec, line 9: *Definitely* fix this before merging!!!

swansontec (approved):
Small cleanups, but don't forget to put back the info servers!!!

src/components/modals/SurveyModal.tsx
Jon-edge: Yeah I tinkered with this earlier and had some trouble, should have just added
"other" into the info server type in hindsight

PR #5428, opened 2025-01-17
Fix back navigation for Mt Pelerin quotes

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209107130480799

### DIFF
--- a/src/plugins/gui/scenes/FiatPluginWebView.tsx
+++ b/src/plugins/gui/scenes/FiatPluginWebView.tsx
@@ -1,3 +1,4 @@
+import { useNavigation } from '@react-navigation/native'
 import * as React from 'react'
 import { WebView, WebViewNavigation } from 'react-native-webview'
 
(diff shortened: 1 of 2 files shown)


src/plugins/gui/scenes/FiatPluginWebView.tsx
```diff
@@ -35,10 +37,18 @@ export function FiatPluginWebViewComponent(props: Props): JSX.Element {
...
   })
 
-  React.useEffect(() => () => {
-    // Cleanup code when scene unmounts
-    if (onClose != null) onClose()
-  })
+  // React.useEffect(() => () => {
+  //   // Cleanup code when scene unmounts
+  //   if (onClose != null) onClose()
+  // })
```
swansontec, line 43: I don't think we need this old code - the `beforeRemove` approach is clearly superior.

src/plugins/gui/scenes/FiatPluginWebView.tsx
```diff
@@ -35,10 +37,18 @@ export function FiatPluginWebViewComponent(props: Props): JSX.Element {
...
+
+  React.useEffect(() => {
+    navigation.addListener('beforeRemove', event => {
+      if (onClose != null) {
+        if (onClose() === false) event.preventDefault()
+      }
+    })
```
swansontec, line 50: The `navigation.addListener` returns a cleanup, function, which you should hand to `useEffect`. So maybe just remove the `{}` around the `navigation.addListener` call, so the return value flows out.

swansontec (changes requested):

swansontec (approved):

PR #5430, opened 2025-01-21
Add rfr param to mtpelerin

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209197620849928

### DIFF
--- a/src/plugins/gui/providers/mtpelerinProvider.ts
+++ b/src/plugins/gui/providers/mtpelerinProvider.ts
@@ -191,6 +191,11 @@ const allowedCountryCodes: { [code: string]: boolean } = {
   VA: true // Vatican City
 }
 
+const asApiKeys = asObject({
+  apiKey: asString,
+  referralCode: asOptional(asString, 'edgewallet')
+})
+
 /**
  * Cleaner for https://api.mtpelerin.com/currencies/tokens
  */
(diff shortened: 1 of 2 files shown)


paullinator (changes requested):
Get the `rfr` value from ENV so we can change it for partners. Default to `edgewallet` is of course fine.

src/plugins/gui/providers/mtpelerinProvider.ts
```diff
@@ -191,6 +191,11 @@ const allowedCountryCodes: { [code: string]: boolean } = {
   VA: true // Vatican City
 }
 
+const asApiKeys = asObject({
+  apiKey: asString,
+  referralCode: asOptional(asString)
```
swansontec, line 196: Would it make more sense to do `asOptional(asString, 'edgewallet')` here? The way you have it is fine too.

swansontec (approved):

src/plugins/gui/providers/mtpelerinProvider.ts
samholmes: I'd rather clean it and add business logic in the plugin

PR #5433, opened 2025-01-23
Jon/tx-list-chart

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209119564523914


src/components/charts/SwipeChart.tsx
```diff
@@ -614,6 +617,9 @@ const getStyles = cacheStyles((theme: Theme) => {
       alignSelf: 'flex-start',
       position: 'absolute'
     },
+    cardAdjust: {
+      marginLeft: theme.rem(-1)
+    },
```
swansontec, line 622: I don't like this. Normally components should be agnostic to how they are contained. If the chart has built-in margins, those could maybe be reduced to 0 and then move the padding to the container. Otherwise, maybe make the margin configurable, similar to how `marginRem` is passed to buttons. However, `cardAdjust` is too context-specific and hacky. These thinks shouldn't be tied together so closely.

src/components/scenes/TransactionListScene.tsx
```diff
@@ -59,6 +62,7 @@ function TransactionListComponent(props: Props) {
   const tokenId = checkToken(route.params.tokenId, wallet.currencyConfig.allTokens)
   const { pluginId } = wallet.currencyInfo
   const { currencyCode } = tokenId == null ? wallet.currencyInfo : wallet.currencyConfig.allTokens[tokenId]
+  const { displayName } = tokenId == null ? wallet.currencyInfo : wallet.currencyConfig.allTokens[tokenId]
```
swansontec, line 65: You could just delete this line, and do `const { currencyCode, displayName } = ...` above.

src/components/scenes/TransactionListScene2.tsx
swansontec, line 1: What is going on here? How do these two scenes differ? It's not like "2" is a new & improved version that's going to replace the old one eventually, like how we migrated the send scene. It seems like one is used for different purposes than another. Maybe give them better names, like "LightTransactionList" and "ChartTransactionList" or something. Then I can review this commit.

Could we somehow provide a prop or route param that allows us to have a single scene, instead of splitting it into two?

Jon-edge: I think they should still be separate scenes, just like how the home scene markets card + coinranking works.

src/components/charts/SwipeChart.tsx
Jon-edge: Yeah we normally want all components to have their own built in margins, but
this specific situation is the exception since we want it to fill up its card.

swansontec (approved):

swansontec (approved):

src/components/scenes/TransactionListScene2.tsx
```diff
@@ -0,0 +1,266 @@
...
+ * If the token gets deleted, the scene will crash.
+ * Fall back to the main currency code if this happens.
+ */
+function checkToken(tokenId: EdgeTokenId, allTokens: EdgeTokenMap): EdgeTokenId {
+  if (tokenId == null) return null
+  if (allTokens[tokenId] == null) return null
+  return tokenId
+}
+
+export const TransactionList = withWallet(TransactionListComponent)
```
samholmes, line 260: Why weren't these components named properly? All scene components should suffix with "Scene". Also, no need for "2" suffix if we're renaming the original component.

src/components/scenes/WalletDetailsScene.tsx
```diff
@@ -292,19 +274,7 @@ function TransactionListComponent(props: Props) {
...
-          <SectionHeader title={item} />
-        </EdgeAnim>
-      )
-    }
-    return (
-      <EdgeAnim disableAnimation={disableAnimation} enter={{ type: 'fadeInDown', distance: 30 * (index + 1) }}>
-        <TransactionListRow navigation={navigation as NavigationBase} transaction={item} wallet={wallet} />
-      </EdgeAnim>
-    )
+    return null // We're not using the FlatList rendering anymore
```
samholmes, line 277: If we're no longer using FlatList, why hasn't it be replaced with a scrollable SceneWrapper?

samholmes (commented):
Curious

PR #5435, opened 2025-01-24
Markets - Buy/Sell/Earn/Trade Buttons

Jon-edge:
**TODO: Fix Earn button functionality**

image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1208943014243441

### DIFF
--- a/src/components/scenes/CoinRankingDetailsScene.tsx
+++ b/src/components/scenes/CoinRankingDetailsScene.tsx
@@ -1,21 +1,38 @@
 import { useIsFocused } from '@react-navigation/native'
 import * as React from 'react'
-import { View } from 'react-native'
+import { ActivityIndicator, View } from 'react-native'
 import FastImage from 'react-native-fast-image'
+import Feather from 'react-native-vector-icons/Feather'
+import Ionicons from 'react-native-vector-icons/Ionicons'
 
+import { createWallet, getUniqueWalletName } from '../../actions/CreateWalletActions'
+import { getFirstOpenInfo } from '../../actions/FirstOpenActions'
+import { updateStakingState } from '../../actions/scene/StakingActions'
+import { Fontello } from '../../assets/vector/index'
+import { WalletListModal, WalletListResult } from '../../components/modals/WalletListModal'
+import { SPECIAL_CURRENCY_INFO } from '../../constants/WalletAndCurrencyConstants'
 import { useAsyncValue } from '../../hooks/useAsyncValue'
 import { formatFiatString } from '../../hooks/useFiatText'
+import { useHandler } from '../../hooks/useHandler'
+import { useWatch } from '../../hooks/useWatch'
 import { toLocaleDate, toPercentString } from '../../locales/intl'
 import { lstrings } from '../../locales/strings'
+import { defaultWalletStakingState } from '../../reducers/StakingReducer'
 import { getDefaultFiat } from '../../selectors/SettingsSelectors'
 import { asCoinRankingData, CoinRankingData, CoinRankingDataPercentChange } from '../../types/coinrankTypes'
-import { useSelector } from '../../types/reactRedux'
-import { EdgeAppSceneProps } from '../../types/routerTypes'
+import { useDispatch, useSelector } from '../../types/reactRedux'
+import { EdgeAppSceneProps, NavigationBase } from '../../types/routerTypes'
+import { EdgeAsset } from '../../types/types'
+import { CryptoAmount } from '../../util/CryptoAmount'
 import { fetchRates } from '../../util/network'
-import { formatLargeNumberString as formatLargeNumber } from '../../util/utils'
+import { getPluginFromPolicy } from '../../util/stakeUtils'
+import { getUkCompliantString } from '../../util/ukComplianceUtils'
+import { DECIMAL_PRECISION, formatLargeNumberString as formatLargeNumber } from '../../util/utils'
+import { IconButton } from '../buttons/IconButton'
 import { SwipeChart } from '../charts/SwipeChart'
 import { EdgeAnim, fadeInLeft } from '../common/EdgeAnim'
 import { SceneWrapper } from '../common/SceneWrapper'
+import { Airship, showError } from '../services/AirshipInstance'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 import { EdgeText } from '../themed/EdgeText'
 import { COINGECKO_SUPPORTED_FIATS } from './CoinRankingScene'
@@ -81,8 +98,18 @@ const COLUMN_RIGHT_DATA_KEYS: Array<keyof CoinRankingData> = [
 const CoinRankingDetailsSceneComponent = (props: Props) => {
   const theme = useTheme()
   const styles = getStyles(theme)
+  const dispatch = useDispatch()
   const { route, navigation } = props
   const { assetId, fiatCurrencyCode, coinRankingData: initCoinRankingData } = route.params
+
+  const account = useSelector(state => state.core.account)
+  const exchangeRates = useSelector(state => state.exchangeRates)
+  const walletStakingStateMap = useSelector(state => state.staking.walletStakingMap ?? defaultWalletStakingState)
+
+  const currencyWallets = useWatch(account, 'currencyWallets')
+  const isFocused = useIsFocused()
+
+  const [countryCode] = useAsyncValue(async () => (await getFirstOpenInfo()).countryCode)
 
   // In case the user changes their default fiat while viewing this scene, we
   // want to go back since the parent scene handles fetching data.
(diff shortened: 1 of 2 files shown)


src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -119,8 +144,36 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
...
+            if (asset.tokenId != null) {
+              const token = account.currencyConfig[asset.pluginId].allTokens[asset.tokenId]
+              return token.currencyCode.toLowerCase() === coinRankingData.currencyCode.toLowerCase()
+            }
+            return false
+          })
+        }
+
+        setEdgeAsset(matchingAsset)
+      }
```
swansontec, line 174: This block of code should be in `useMemo`, not the `useEffect`. We are using to derive an `EdgeAsset` from the `coinRankingData`, so doing it in `useEffect` will cause an extra render. Doing it in `useMemo` will save a render.

Second, do not use the `makeCurrencyCodeTable` function. It builds a giant table using information in the account, which you then loop over to find what you are looking for. However, you could just loop over the account directly, saving the table-building step.

So more like this:

```ts
const edgeAsset = React.useMemo(() => {
 // Search for mainnet coins:
   const config = account.currencyConfig[pluginId]
   if (config.currencyCode.toLowerCase() === coinRankingData.currencyCode.toLowerCase())
     return { tokenId: null, pluginId }
 }
 
 // Search for tokens:
   const config = account.currencyConfig[pluginId]
   for (const tokenId of Object.keys(config.allTokens)) {
    const token = config.allTokens[tokenId]
    if (token.currencyCode.toLowerCase() === coinRankingData.currencyCode.toLowerCase())
       return { tokenId, pluginId }
   }
 }
), [account, coinRankingData])
```

Obviously not debugged, since I just typed it, but you can see how it's not abusing `useEffect`, `Array.find`, or `setState` like the old code was.

src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -218,6 +271,176 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
     return rows
   }
 
+  const getWallet = async () => {
+    if (edgeAsset == null) return
+    const { pluginId, tokenId } = edgeAsset
```
swansontec, line 276: Do we need to enable the token? It seems like we don't handle the case where the wallet exists for the chain, but the token itself isn't active.

src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -218,6 +271,176 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
...
+
+  const isStakingAvailable = (): boolean => {
+    if (countryCode == null || edgeAsset == null) return false
+
+    // Special case for FIO because it uses it's own staking plugin
+    const isStakingSupported = SPECIAL_CURRENCY_INFO[edgeAsset.pluginId]?.isStakingSupported === true || currencyCode === 'FIO'
+    return isStakingSupported
+  }
+
+  // TODO: Broken - We need to come up with a wallet beforehand in order to get stakePlugins and stakePolicies for routing
```
swansontec, line 402: It looks like you solved this, and it's not broken anymore?

src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -218,6 +271,176 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
...
+      })
+    }
+
+    return targetWallet
+  }
+
+  const handleBuyPress = useHandler(async () => {
+    if (edgeAsset == null) return
+    const wallet = await getWallet()
+    if (wallet == null) return
```
swansontec, line 314: If the user presses the buy button, but the wallet is missing, nothing will happen. However, it seems like the function will only return `undefined` if the user cancels?

In any case, the `getWallet` function needs a documentation comment explaining the return value.

src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -218,6 +271,176 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
...
+
+  const handleTradePress = useHandler(async () => {
+    if (edgeAsset == null) return
+    const { tokenId } = edgeAsset
+
+    const targetWallet = await getWallet()
+    if (targetWallet == null) return
+
+    // Find the wallet with highest USD value to use as source
+    // TODO: Include tokenIds in the sort
```
swansontec, line 359: Are we going to address this TODO, or is it out of scope for the PR?

swansontec (changes requested):

src/components/scenes/CoinRankingDetailsScene.tsx
Jon-edge: No this is still broken, I wanted to get the review out first while I continue to tackle it

src/components/scenes/CoinRankingDetailsScene.tsx
Jon-edge: I wanted to address this as a fixup

src/components/scenes/CoinRankingDetailsScene.tsx
Jon-edge: The wallet will only be missing if the user pressed the button, was presented
with the wallet picker, and then dismissed the wallet picker without choosing anything

src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -384,36 +406,35 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
...
+    const walletStakingState = walletStakingStateMap[walletId] ?? defaultWalletStakingState
     const { stakePlugins } = walletStakingState
     const stakePolicies = Object.values(walletStakingState.stakePolicies)
 
     // Handle FIO staking
     if (currencyCode === 'FIO') {
       navigation.push('fioStakingOverview', {
         tokenId: null,
-        walletId: wallet.id
+        walletId: walletId
```
swansontec, line 437: Could just be `walletId`, since it's the same on both sides.

src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -422,19 +443,20 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
     if (stakePlugins != null && stakePolicies != null) {
       if (stakePolicies.length > 1) {
         navigation.push('stakeOptions', {
-          walletId: wallet.id,
+          walletId: walletId,
```
swansontec, line 446: Could just be `walletId`, since it's the same on both sides.

src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -447,32 +469,38 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
...
-              <Fontello name="buy" size={theme.rem(2)} color={theme.primaryText} />
-            </IconButton>
-            <IconButton label={lstrings.title_sell} onPress={handleSellPress}>
-              <Fontello name="sell" size={theme.rem(2)} color={theme.primaryText} />
-            </IconButton>
-            {isStakingAvailable() && (
-              <IconButton label={getUkCompliantString(countryCode, 'stake_earn_button_label')} onPress={handleStakePress}>
-                <Feather name="percent" size={theme.rem(2)} color={theme.primaryText} />
+          <SwipeChart assetId={coinRankingData.assetId} currencyCode={currencyCode} fiatCurrencyCode={initFiat} />
+          {edgeAssets.length > 0 && (
```
swansontec, line 475: Do `edgeAssets.length <= 0 ? null : <View ....>`, since using `&&` to guard rendering will generate an ESLint warning once we turn on strict-bookean.

swansontec (changes requested):
This is much better, but there are some small issues now.

src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -271,94 +283,104 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
...
+    if (matchingWallets.length === 0 && edgeAssets.length === 1) {
+      const walletName = getUniqueWalletName(account, edgeAssets[0].pluginId)
+      const targetWallet = await createWallet(account, {
         name: walletName,
-        walletType: `wallet:${pluginId}`,
+        walletType: `wallet:${edgeAssets[0].pluginId}`,
         fiatCurrencyCode: fiatCurrencyCode
       })
+      if (edgeAssets[0].tokenId != null) {
+        await targetWallet.changeEnabledTokenIds([edgeAssets[0].tokenId])
```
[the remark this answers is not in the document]
Jon-edge: In this situation, it's a brand new wallet, so it shouldn't matter. I'll add it anyway, though to be safe

src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -271,94 +283,104 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
...
-      targetWallet = account.currencyWallets[result.walletId]
-    }
-
-    // If no wallet exists, create one
-    if (targetWallet == null) {
-      const walletName = getUniqueWalletName(account, pluginId)
-      targetWallet = await createWallet(account, {
+  /**
+   * Returns a WalletListResult to use for the button navigation. Returns the
+   *  wallet the user chose from the wallet picker, automatically selects a
```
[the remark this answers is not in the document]
Jon-edge: My darn auto wrapper is useless, thanks

src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -98,35 +99,74 @@ const COLUMN_RIGHT_DATA_KEYS: Array<keyof CoinRankingData> = [
...
+    return out
+  }, [account.currencyConfig, coinRankingData, currencyCode])
+
   const initFiat = React.useState<string>(fiatCurrencyCode)[0]
 
+  /** Find all wallets that can hold this asset */
+  const matchingWallets = React.useMemo(
+    () => Object.values(currencyWallets).filter(wallet => edgeAssets?.some(asset => asset.pluginId === wallet.currencyInfo.pluginId)),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [edgeAssets]
```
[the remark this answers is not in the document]
Jon-edge: I thought it was unnecessary to re-render on `currencyWallet` changes but thinking about it more, we can create new wallets and stay on this scene

swansontec (approved):

PR #5439, opened 2025-01-30
Optimize WalletListModal components

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208888448558651


src/components/icons/CryptoIcon.tsx
```diff
@@ -6,85 +6,46 @@ import { ShadowedView } from 'react-native-fast-shadow'
...
 import { fixSides, mapSides, sidesToMargin } from '../../util/sides'
-import { WalletSyncCircle } from '../progress-indicators/WalletSyncCircle'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 
 interface Props {
   // Main props - If non is specified, would just render an empty view
-  pluginId?: string // Needed when walletId is not supplied and we still want to get an icon
+  pluginId: string
   tokenId: EdgeTokenId // Needed when it's a token (not the plugin's native currency)
-  walletId?: string // To allow showing the progress ratio sync circle
```
swansontec, line 20: So we still want an icon that takes a `walletId`, and automatically figures out stuff like the compromise status. I don't want to shatter this logic into a bunch of different places.

So, I propose creating a `<WalletIcon>` component, which grabs the wallet, looks up the relevant stuff, and then renders a `<CryptoIcon />` with whatever it finds. That way, we have simply split this one component into two, instead of deleting features.

swansontec (changes requested):
Just summarizing our sidebar. I'll review in more depth once the architecture gets sorted.

src/selectors/getCreateWalletList.ts
```diff
@@ -64,82 +64,104 @@ interface CreateWalletListOpts {
...
+  const excludedAssetsMap = createAssetMap(excludeAssets)
+  const allowedAssetsMap = createAssetMap(allowedAssets)
+
+  const isAllowed = (pluginId: string, tokenId: EdgeTokenId) =>
+    // if the wallet already exists, then it is not allowed
+    !existingWalletsMap.get(pluginId)?.has(tokenId) &&
+    // if allowedAssets is empty, then all assets are allowed
+    (allowedAssetsMap.size === 0 || allowedAssetsMap.get(pluginId)?.has(tokenId)) &&
+    // if excludedAssets is not empty, then the asset must not be in the excluded list
+    !(excludedAssetsMap.size > 0 && excludedAssetsMap.get(pluginId)?.has(tokenId))
```
swansontec, line 99: Applying DeMorgan's Law, we can turn `!(excludedAssetsMap.size > 0 && excludedAssetsMap.get(pluginId)?.has(tokenId))` into `(excludedAssetsMap.size === 0 || !excludedAssetsMap.get(pluginId)?.has(tokenId))`. This second version reads better, and matches the `allowedAssetsMap` logic above, except for the `!` before the `has`.

src/components/themed/TransactionListTop.tsx
```diff
@@ -416,7 +416,7 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
       <>
         <View style={styles.balanceBoxWalletNameCurrencyContainer}>
           <EdgeTouchableOpacity accessible={false} style={styles.balanceBoxWalletNameContainer} onPress={this.handleOpenWalletListModal}>
-            <CryptoIcon marginRem={[0, 0.25, 0, 0]} pluginId={wallet.currencyInfo.pluginId} sizeRem={1} tokenId={tokenId} walletId={wallet.id} />
```
swansontec, line 419: We should probably switch this back to a `<WalletIcon>`, so we can see the sync circle in the wallet details scene.

src/components/themed/TransactionListTop.tsx
```diff
@@ -409,7 +409,7 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
 
     // Fiat Balance Formatting
     const exchangeAmount = convertNativeToDenomination(exchangeDenomination.multiplier)(nativeBalance)
-    const fiatBalance = mul(exchangeAmount, exchangeRate)
+    const fiatBalance = mul(exchangeAmount, exchangeRate.toString())
```
swansontec, line 412: Should this be done as `parseFloat(exchangeAmount) * exchangeRate`? The `formatNumber` function takes `number | string`, so it's fine.

src/util/utils.ts
```diff
@@ -221,18 +221,18 @@ export function fixFiatCurrencyCode(currencyCode: string) {
 
 // multiplier / exchange rate / ( 1 / unit )
 // 100000000 / $16500 / (1/$0.001) = ~6 sats
-export const calculateSpamThreshold = (rate: string, denom: EdgeDenomination) => {
-  return div(div(denom.multiplier, rate), '1000')
+export const calculateSpamThreshold = (rate: number, denom: EdgeDenomination) => {
+  return div(div(denom.multiplier, rate.toString()), '1000')
```
swansontec, line 225: Could we do `(parseFloat(multiplier) / rate) / 1000`? Of course, then you end up returning `number` instead of `string`, so the changes would ripple out further into the spam calculation, so feel free to skip if too complicated.

src/util/utils.ts
```diff
@@ -221,18 +221,18 @@ export function fixFiatCurrencyCode(currencyCode: string) {
...
 export interface PrecisionAdjustParams {
-  exchangeSecondaryToPrimaryRatio: string
+  exchangeSecondaryToPrimaryRatio: number
   secondaryExchangeMultiplier: string
   primaryExchangeMultiplier: string
 }
 
 export function precisionAdjust(params: PrecisionAdjustParams): number {
-  const exchangeSecondaryToPrimaryRatio = parseFloat(params.exchangeSecondaryToPrimaryRatio)
+  const exchangeSecondaryToPrimaryRatio = params.exchangeSecondaryToPrimaryRatio
```
swansontec, line 235: Can do `const {exchangeSecondaryToPrimaryRatio} = params` now.

swansontec (approved):
Approving in the interests of times, but I do have a few small tidy-up things to do. The only "essential" change is to put the sync circle back on the wallet details scene.

src/util/utils.ts
peachbits: I purposely didn't make this change because I wanted to limit the change to exchange rates and this function returns a crypto amount

src/selectors/getCreateWalletList.ts
peachbits: fixed

src/components/themed/TransactionListTop.tsx
peachbits: fixed

PR #5440, opened 2025-01-31
Fix `useSelector` calls for accessing non-existent `exchangeRates`

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209259612887729


peachbits (approved):
Approved, but change the changelog entry to note that we're now using a safe fallback 0 value for rates that don't exist. Internet connectivity isn't relevant to the bug

PR #5445, opened 2025-02-07
Fix 2FA/otp Reminder Notification

Jon-edge:
- Isolate the modals to their own file
- Create a hook to ensure changes to otp disk settings are picked up by
  the `NotificationService`
- Pull all logic for otp reminder notification state into `NotificationService`

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209310723615615

### DIFF
--- a/src/components/services/NotificationService.ts
+++ b/src/components/services/NotificationService.ts
@@ -2,11 +2,10 @@ import { EdgeAccount } from 'edge-core-js'
 
 import { getLocalAccountSettings, useAccountSettings, writeAccountNotifInfo } from '../../actions/LocalSettingsActions'
 import { useAsyncEffect } from '../../hooks/useAsyncEffect.ts'
-import { useAsyncValue } from '../../hooks/useAsyncValue.ts'
 import { useWatch } from '../../hooks/useWatch.ts'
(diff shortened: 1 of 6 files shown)


src/components/services/NotificationService.ts
```diff
@@ -61,23 +60,26 @@ export const updateNotificationInfo = async (
  */
 export const NotificationService = (props: Props) => {
   const { account } = props
+
   const { notifState, accountNotifDismissInfo } = useAccountSettings()
+  const { lastChecked, dontAsk } = useOtpSettings()
```
swansontec, line 65: The variable names are pretty generic, and this component deals with a lot of unrelated stuff. We should make it clear that they pertain to OTP, specifically. I would either do:

```ts
const optSettings = useOtpSettings()

... optSettings.dontAsk ...
```

Or I would do:

```ts
const { lastChecked: otpLastChecked, dontAsk: otpDontAsk } = useOtpSettings()
```

Either way is fine.

swansontec (approved):
One cleanup needed.

PR #5446, opened 2025-02-07
Abstract L2

samholmes:
- **Fix yarn.lock**
- **Add abstract core plugin**

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209257679727561

### DIFF
--- a/src/constants/WalletAndCurrencyConstants.ts
+++ b/src/constants/WalletAndCurrencyConstants.ts
@@ -183,6 +183,19 @@ export const getSpecialCurrencyInfo = (pluginId: string): SpecialCurrencyInfo =>
 export const SPECIAL_CURRENCY_INFO: {
   [pluginId: string]: SpecialCurrencyInfo
 } = {
+  abstract: {
+    allowZeroTx: true,
+    chainCode: 'ETH',
+    chainIcon: true,
+    dummyPublicAddress: '0x0d73358506663d484945ba85d0cd435ad610b0a0',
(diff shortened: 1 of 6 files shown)


src/constants/WalletAndCurrencyConstants.ts
```diff
@@ -183,6 +183,19 @@ export const getSpecialCurrencyInfo = (pluginId: string): SpecialCurrencyInfo =>
 export const SPECIAL_CURRENCY_INFO: {
   [pluginId: string]: SpecialCurrencyInfo
 } = {
+  abstract: {
+    allowZeroTx: true,
+    chainCode: 'ETH',
```
swansontec, line 188: This is used for FIO. Does FIO support Abstract? If so, is this the code they use?

swansontec (changes requested):
The `chainCode` seems wrong. Also, please drop the first commit, or replace it with the upgrade to edge-currency-accountbased.

src/constants/WalletAndCurrencyConstants.ts
samholmes: I don't know if FIO supports Abstract or not. I copied the chainCode from all the other EVM based `SpecialCurrencyInfo`s. Do these other configs need fixing too?

swansontec: Most likely you have found a bug!

swansontec (approved):

PR #5448, opened 2025-02-08
Correct the `displayName` in `getTxActionDisplayInfo`

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208274844820487

### DIFF
--- a/src/actions/CategoriesActions.ts
+++ b/src/actions/CategoriesActions.ts
@@ -283,7 +283,7 @@ export const getTxActionDisplayInfo = (tx: EdgeTransaction, account: EdgeAccount
   const { assetAction, chainAction, chainAssetAction, metadata, savedAction, swapData, tokenId } = tx
   const { currencyConfig, currencyInfo } = wallet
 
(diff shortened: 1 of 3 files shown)


src/actions/CategoriesActions.ts
```diff
@@ -283,7 +283,7 @@ export const getTxActionDisplayInfo = (tx: EdgeTransaction, account: EdgeAccount
   const { assetAction, chainAction, chainAssetAction, metadata, savedAction, swapData, tokenId } = tx
   const { currencyConfig, currencyInfo } = wallet
 
-  const { displayName } = tokenId == null ? currencyInfo : currencyConfig.allTokens[tokenId] ?? { displayName: '' }
+  const displayName = tokenId == null ? currencyInfo.assetDisplayName : currencyConfig.allTokens[tokenId].displayName ?? ''
```
swansontec, line 286: The last `?? ''` is wrong, because `displayName` can't be undefined. I think you meant to do `currencyConfig.allTokens[tokenId]?.displayName ?? ''`, to handle the case there the token itself is missing.

swansontec (changes requested):

PR #5449, opened 2025-02-11
Fix transaction metadata loading issues

samholmes:
- **Fix yarn.lock**
- **Rename txs identifier**
- **Fix bug with relevant changedTxs set**

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209197463729195

### DIFF
--- a/src/hooks/useTransactionList.ts
+++ b/src/hooks/useTransactionList.ts
@@ -50,12 +50,12 @@ export function useTransactionList(wallet: EdgeCurrencyWallet, tokenId: EdgeToke
     // Reset our mutable state to an empty list:
     let atEnd = false
     const changedTxs = new Map<string, EdgeTransaction>()
-    const txs: EdgeTransaction[] = []
+    const streamedTxs: EdgeTransaction[] = []
 
     // Sends the mutable state to React,
     // merging the two transaction lists together:
     function requestRender() {
-      const mergedTxs = txs.filter(tx => !changedTxs.has(tx.txid))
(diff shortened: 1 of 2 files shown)


src/hooks/useTransactionList.ts
```diff
@@ -85,7 +85,7 @@ export function useTransactionList(wallet: EdgeCurrencyWallet, tokenId: EdgeToke
     const cleanupChanged = wallet.on('transactionsChanged', txs => {
       let relevant = false
       for (const tx of txs) {
-        if (tx.tokenId === tokenId) {
+        if (tx.tokenId === tokenId && streamedTxs.some(streamedTx => streamedTx.txid === tx.txid)) {
```
swansontec, line 88: This is *potentially* O(n^2), but only if the core / engines have bugs that spam onTransactionChanged events. However, spamming needless events is *already* inefficient, so let's leave this as it is.

swansontec (approved):
This PR is perfect.

swansontec:
Actually, the first commit is probably not needed, because these two fields randomly flip-flop back & forth, and we've never figured out what was the difference between the different dev's machines that was causing it. So just drop the commit, and the next time you `yarn add` something, just let them change them.

src/hooks/useTransactionList.ts
samholmes: Fixed it

PR #5450, opened 2025-02-11
Jon/wallet-details-header

Jon-edge:
For the TransactionList, keep the same title so the user knows which wallet the transactions are for.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208077272222816

### DIFF
--- a/src/components/Main.tsx
+++ b/src/components/Main.tsx
@@ -126,10 +126,10 @@ import { SweepPrivateKeyCompletionScene as SweepPrivateKeyCompletionSceneCompone
 import { SweepPrivateKeyProcessingScene as SweepPrivateKeyProcessingSceneComponent } from './scenes/SweepPrivateKeyProcessingScene'
 import { SweepPrivateKeySelectCryptoScene as SweepPrivateKeySelectCryptoSceneComponent } from './scenes/SweepPrivateKeySelectCryptoScene'
 import { TransactionDetailsScene as TransactionDetailsSceneComponent } from './scenes/TransactionDetailsScene'
-import { TransactionList as TransactionListComponent } from './scenes/TransactionListScene'
+import { TransactionList as TransactionListComponent, TransactionListTitle } from './scenes/TransactionListScene'
 import { TransactionsExportScene as TransactionsExportSceneComponent } from './scenes/TransactionsExportScene'
 import { UpgradeUsernameScene as UpgradeUsernameSceneComponent } from './scenes/UpgradeUsernameScreen'
-import { WalletDetails as WalletDetailsComponent } from './scenes/WalletDetailsScene'
+import { WalletDetails as WalletDetailsComponent, WalletDetailsTitle } from './scenes/WalletDetailsScene'
 import { WalletListScene as WalletListSceneComponent } from './scenes/WalletListScene'
 import { WalletRestoreScene as WalletRestoreSceneComponent } from './scenes/WalletRestoreScene'
 import { WcConnectionsScene as WcConnectionsSceneComponent } from './scenes/WcConnectionsScene'
@@ -277,14 +277,14 @@ const EdgeWalletsTabScreen = () => {
         name="transactionList"
         component={TransactionList}
         options={{
-          headerTitle: () => <ParamHeaderTitle<'transactionList'> fromParams={params => params.walletName} />
+          headerTitle: () => <TransactionListTitle />
         }}
       />
       <WalletsStack.Screen
         name="walletDetails"
         component={WalletDetails}
         options={{
-          headerTitle: () => <ParamHeaderTitle<'walletDetails'> fromParams={params => params.walletName} />
+          headerTitle: () => <WalletDetailsTitle />
         }}
       />
     </WalletsStack.Navigator>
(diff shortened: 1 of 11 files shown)


src/components/Main.tsx
```diff
@@ -284,7 +284,7 @@ const EdgeWalletsTabScreen = () => {
         name="walletDetails"
         component={WalletDetails}
         options={{
-          headerTitle: () => <ParamHeaderTitle<'walletDetails'> fromParams={params => params.walletName} />
+          headerTitle: () => <ParamHeaderTitle<'walletDetails'> fromParams={params => params.sceneTitle} />
```
swansontec, line 287: Delete the `sceneTitle` param, and put the string formatting in `fromParams`:

```ts
const wallet = account.currencyWallets[walletId]
return sprintf(lstrings.create_wallet_account_metadata_name, wallet?.currencyInfo.displayName)
```

Or, better yet, export a `WalletDetailsTitle` component from `WalletDetailsScene.tsx`:

```ts
export const WalletDetailsTitle = () => {
  const route = useRoute<RouteProp<'walletDetails'>>()
  const wallet = account.currencyWallets[walletId]
  const title = sprintf(lstrings.create_wallet_account_metadata_name, wallet?.currencyInfo.displayName)
  return <HeaderTitle title={title} />
}
```

Then in Main.tsx, it can just be:

```ts
options={{ headerTitle: WalletDetailsTitle })
```

src/components/scenes/WalletDetailsScene.tsx
```diff
@@ -174,7 +175,12 @@ function WalletDetailsComponent(props: Props) {
   })
 
   const handlePressSeeAll = useHandler(() => {
-    navigation.navigate('transactionList', route.params)
+    navigation.navigate('transactionList', {
+      walletId: wallet.id,
+      tokenId,
+      walletName: getWalletName(wallet),
```
swansontec, line 181: Actually, I see we have the same bad pattern here on the TransactionList. Please export a `TransactionListTitle` component like we did for the `WalletDetailsTitle`:

```ts
export const TransactionListTitle = () => {
  const route = useRoute<RouteProp<'walletDetails'>>()
  const wallet = account.currencyWallets[walletId]
  const title = wallet == null ? '' :  getWalletName(wallet)
  return <HeaderTitle title={title} />
}
```

swansontec (changes requested):

src/components/Main.tsx
Jon-edge: Regarding the string name, yes, it _should_ be that, but this is an old string. Since it's old and has translations already, we can't change the key or we reset the translations.

src/components/scenes/WalletDetailsScene.tsx
```diff
@@ -44,7 +46,7 @@ import { TransactionListTop } from '../themed/TransactionListTop'
 
 export interface WalletDetailsParams {
   walletId: string
-  sceneTitle: string
+  walletName: string
```
swansontec, line 49: So `waletName` isn't actually used anywhere. Please delete it, so the params are literally just `walletId` and `tokenId`...

It turns out that the `countryCode` route prop isn't really used either. Or, more specifically, there is only one place that actually provides it to `WalletListScene`, and that just does:

```ts
 const { countryCode } = await getFirstOpenInfo()
 navigation.navigate('walletDetails', { walletId, tokenId, countryCode })
```

The `WalletDetailsScene` doesn't use `countryCode` either. The only thing it does is pass `countryCode` to `InfoCardCarousel` and `BuyCrypto`, both of which will be buggy (showing wrong cards or wrong strings) if `countryCode` is `undefined`.

This means we should just refactor `InfoCardCarousel` and `BuyCrypto` to get their own country code:

```ts
const [ firstOpenInfo ] = useAsyncValue(async () => await getFirstOpenInfo())
const { countryCode } =  firstOpenInfo ?? {}
```

And then we can delete it the `WalletDetailsScene` `courntryCode` param.

src/components/scenes/TransactionListScene.tsx
```diff
@@ -249,6 +252,14 @@ function TransactionListComponent(props: Props) {
...
 }
 
+export const TransactionListTitle = () => {
+  const route = useRoute<RouteProp<'walletDetails'>>()
+  const account = useSelector(state => state.core.account)
+  const wallet = account.currencyWallets[route.params.walletId]
+  const title = wallet == null ? '' : getWalletName(wallet)
+  return <HeaderTitle title={title} />
+}
+
```
swansontec, line 262: Thanks. But now that this is here, the `walletName` and `countryCode` route props are unused. For the the `TransactionListScene`, the route props should just be `walletId`, `tokenId`, and `searchText`.

swansontec (changes requested):

src/components/Main.tsx
swansontec: Ok, that makes sense. We really need to figure out how to rename strings in Crowdin.

src/components/scenes/WalletDetailsScene.tsx
Jon-edge: I thought there was some reason we passed `countryCode` around instead of calling for it in the component but maybe it's no longer relevant...

swansontec (approved):

PR #5452, opened 2025-02-12
Earn button fixes

samholmes:
This include the fixes for the Earn button, but it also includes some fixes to the error handling for coingecko API queries for the chart data because those were immediately apparent while fixing the original intent.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209381036725837

### DIFF
--- a/src/components/charts/SwipeChart.tsx
+++ b/src/components/charts/SwipeChart.tsx
@@ -14,6 +14,7 @@ import { formatFiatString } from '../../hooks/useFiatText'
 import { useHandler } from '../../hooks/useHandler'
 import { formatDate } from '../../locales/intl'
 import { lstrings } from '../../locales/strings'
+import { snooze } from '../../util/utils'
 import { MinimalButton } from '../buttons/MinimalButton'
 import { FillLoader } from '../progress-indicators/FillLoader'
 import { Theme, useTheme } from '../services/ThemeContext'
@@ -213,8 +214,11 @@ const SwipeChartComponent = (params: Props) => {
                 // Rate limit error, use our API key as a fallback
                 if (!fetchUrl.includes('x_cg_pro_api_key') && ENV.COINGECKO_API_KEY !== '') {
                   fetchUrl = `${COINGECKO_URL_PRO}${fetchPath}&x_cg_pro_api_key=${ENV.COINGECKO_API_KEY}`
-                  continue
                 }
+                // Wait 2 second before retrying. It typically takes 1 minute
+                // before rate limiting is relieved, so even 2 seconds is hasty.
+                await snooze(2000)
+                continue
               }
               throw new Error(`Failed to fetch market data: ${apiError.status.error_code} ${apiError.status.error_message}`)
             }


src/components/charts/SwipeChart.tsx
```diff
@@ -196,59 +195,54 @@ const SwipeChartComponent = (params: Props) => {
...
+          let fetchUrl = `${COINGECKO_URL}${fetchPath}`
+          do {
+            // Construct the dataset query
+            const response = await fetch(fetchUrl)
+            const result = await response.json()
+            const marketChartRange = asCoinGeckoMarketApi(result)
+            if ('status' in marketChartRange) {
+              if (marketChartRange.status.error_code === 429) {
+                // Rate limit error, use our API key as a fallback
+                if (!fetchUrl.includes('x_cg_pro_api_key')) {
```
peachbits, line 222: Can you remove the dummy coingecko key from the config cleaner and add a check for it's existence here? Passing a dummy value to the api will produce a [10002 error](https://docs.coingecko.com/reference/common-errors-rate-limit) which is unhandled here and produces an error (an uglier error message with this commit's change). Instead, if the key doesn't exist we should just try again without an api key again.

src/components/themed/TransactionListTop.tsx
```diff
@@ -493,7 +493,7 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
     const styles = getStyles(theme)
     const { countryCode } = this.state
 
-    const showStaking = this.isStakingAvailable()
+    const hideStaking = !isStakingSupported(this.props.wallet.currencyInfo.pluginId, this.props.currencyCode)
```
peachbits, line 496: There's a remaining corner case that any asset with the currency code FIO (like a custom token) will show the earn button and navigate to actual FIO staking. The FIO plugin special currency info has `isStakingSupported: true` so that special case in the helper function can be removed.

peachbits (changes requested):

src/components/charts/SwipeChart.tsx
```diff
@@ -215,33 +213,33 @@ const SwipeChartComponent = (params: Props) => {
...
-            if ('status' in marketChartRange) {
-              if (marketChartRange.status.error_code === 429) {
+            const apiError = asMaybe(asCoinGeckoError)(result)
+            if (apiError != null) {
+              if (apiError.status.error_code === 429) {
                 // Rate limit error, use our API key as a fallback
-                if (!fetchUrl.includes('x_cg_pro_api_key')) {
+                if (!fetchUrl.includes('x_cg_pro_api_key') && ENV.COINGECKO_API_KEY !== '') {
                   fetchUrl = `${COINGECKO_URL_PRO}${fetchPath}&x_cg_pro_api_key=${ENV.COINGECKO_API_KEY}`
                   continue
```
peachbits, line 222: The continue should be moved outside of the apikey block so that rate-limited requests without keys can be retried as well.

peachbits (changes requested):

src/components/charts/SwipeChart.tsx
samholmes: The continue retries the request. Why should it be outside of the retry path?

peachbits: It needs to be in the 429 block but outside of the apikey block

peachbits (approved):

PR #5454, opened 2025-02-13
Wallet search performance

swansontec:
The first commit fixes the hang - something is wrong with the swipe chart, but we can work around that by providing an explicit width.

The next two commits are necessary cleanups.

The final commit fixes some of the UX issues I was experiencing when using this scene.

We really only need the first commit, but the next 3 are "nice to have".

### CHANGELOG

- [ ] Yes
- [x] No (bugfix to unreleased feature)

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209379825602483


src/components/charts/SwipeChart.tsx
```diff
@@ -179,6 +170,13 @@ const SwipeChartComponent = (params: Props) => {
   const minPriceDataPoint = React.useMemo(() => chartData.find(point => point.y === minPrice), [chartData, minPrice])
   const maxPriceDataPoint = React.useMemo(() => chartData.find(point => point.y === maxPrice), [chartData, maxPrice])
 
+  // The chart component defaults to the phone width.
+  // To fit the chart into its parent view,
+  // we measure the parent and pass that width in.
+  // The chart will freeze the whole app if the width is too narrow,
+  // so start with the window width:
+  const [width, setWidth] = React.useState(Dimensions.get('window').width)
```
samholmes, line 178: `const [slideAreaChartWidth, setSlideAreaChartWidth]`

src/components/scenes/WalletDetailsScene.tsx
```diff
@@ -328,36 +247,90 @@ function WalletDetailsComponent(props: Props) {
...
-            ListEmptyComponent={emptyComponent}
-            ListHeaderComponent={topArea}
-            onEndReachedThreshold={0.5}
-            renderItem={renderItem}
-            onScroll={handleScroll}
-            scrollIndicatorInsets={SCROLL_INDICATOR_INSET_FIX}
-            // Android scroll gestures break without refreshControl given the
-            // combination of props we use on this Animated.FlatList.
-            refreshControl={refreshControl}
+        <Reanimated.ScrollView
```
samholmes, line 250: Nice! I was about to make this change too when I noticed this left over. Thanks for doing this.

However, why wouldn't we use the SceneWrapper's scroll view instead of a new `Reanimated.ScrollView`? That way we can remove all the "insetStyle" and "undoInsetStyle" work since we don't need a custom scroll component.

PS: In the future, I'd love the API to change to just implementing a scroll component where we need it instead of a `scroll` flag to the `SceneWrapper`. This way we can tuck away all the inset/undo-inset work into the custom scroll components we want to use. This would be a better DX, but we just haven't done the work yet.

samholmes (changes requested):

src/components/scenes/WalletDetailsScene.tsx
swansontec: We need to set some props on the scroll view that we can't get to if we use the `SceneWrapper` one.

I agree that a `SceneScrollView` or such would be nice. It could use context extract the `insetStyle` from the `SceneWrapper` so the scene doesn't need to do anything fancy to hook them up. Then we could pass in the props we need, such as on this scene.

paullinator: BTW, I don't see a need for the search bar on the wallet details scene. There are only 5 transactions visible. Search isn't necessary

src/components/charts/SwipeChart.tsx
```diff
@@ -301,14 +297,15 @@ const SwipeChartComponent = (params: Props) => {
 
   // A delayed fadein for the max/min labels, to ensure the labels don't get
   // rendered before the price line. Also hidden when gesture is active
+  const minPriceLabelY = Platform.OS === 'ios' ? chartHeight - theme.rem(2.5) : chartHeight - theme.rem(2.75)
```
samholmes, line 300: This logic needs comment.

src/components/scenes/WalletDetailsScene.tsx
samholmes: I understand now after discussion. We want search because it's convenient. Though we can improve the search feature, these are reasonable fixes to a feature that may not be fully polished up. In the future we may want to replace the whole scene with a FlatList component just like the TransactionListScene has that is searchable and scrollable. This could be thought of as a "search modal view" that shows transactions (maybe other items) in the search list when search is activated.

src/components/charts/SwipeChart.tsx
samholmes: @Jon-edge ?

samholmes (approved):

PR #5461, opened 2025-02-19
Fix crashes on undefined and 0 numbers

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device


src/actions/ReceiveDropdown.tsx
```diff
@@ -38,7 +38,7 @@ export function showReceiveDropdown(navigation: NavigationBase, transaction: Edg
     const { spamFilterOn } = state.ui.settings
     const exchangeRate = getExchangeRate(state, currencyCode, isoFiatCurrencyCode)
     const exchangeDenom = getExchangeDenom(wallet.currencyConfig, tokenId)
-    const spamThreshold = calculateSpamThreshold(exchangeRate, exchangeDenom)
+    const spamThreshold = exchangeRate !== 0 ? calculateSpamThreshold(exchangeRate, exchangeDenom) : '0'
```
swansontec, line 41: Instead of fixing these various places, could `calculateSpamThreshold` check for `0` itself? As a general rule, I would expect to see ` !== 0` anywhere we do a `div` or `/`, which would include `calculateSpamThreshold`.

swansontec (approved):
One small fix requested. I'm approving because it's small enough not to re-check.

PR #5465, opened 2025-02-20
Jon/deeplink-promo

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209119209044429

### DIFF
--- a/src/__tests__/components/PromoCards.test.tsx
+++ b/src/__tests__/components/PromoCards.test.tsx
@@ -1,7 +1,7 @@
 import { describe, expect, test } from '@jest/globals'
 import { InfoCard } from 'edge-info-server'
 
-import { filterInfoCards } from '../../components/cards/InfoCardCarousel'
+import { filterInfoCards } from '../../util/infoUtils'
 
 const dummyCard: InfoCard = {
   localeMessages: { en: 'hello' },
--- a/src/__tests__/components/PromoCards.test.tsx
+++ b/src/__tests__/components/PromoCards.test.tsx
@@ -1,7 +1,7 @@
 import { describe, expect, test } from '@jest/globals'
 import { InfoCard } from 'edge-info-server'
 
-import { filterInfoCards } from '../../util/infoUtils'
+import { getDisplayInfoCards } from '../../util/infoUtils'
 
 const dummyCard: InfoCard = {
   localeMessages: { en: 'hello' },
@@ -32,9 +32,9 @@ const currentDate = new Date('2024-06-13T20:53:33.013Z')
 
 describe('filterPromoCards', () => {
   test('No cards', () => {
-    const cards: InfoCard[] = []
-    const result = filterInfoCards({
-      cards,
+    const testCards: InfoCard[] = []
+    const result = getDisplayInfoCards({
+      testCards,
       countryCode: 'US',
       buildNumber,
       osType,
@@ -45,9 +45,9 @@ describe('filterPromoCards', () => {
     expect(result.length).toBe(0)
   })
   test('Card no filters', () => {
-    const cards: InfoCard[] = [{ ...dummyCard }]
-    const result = filterInfoCards({
-      cards,
+    const testCards: InfoCard[] = [{ ...dummyCard }]
+    const result = getDisplayInfoCards({
+      testCards,
       countryCode: 'US',
       buildNumber,
       osType,
@@ -59,7 +59,7 @@ describe('filterPromoCards', () => {
     expect(result[0].localeMessages.en).toBe('hello')
   })
   test('Cards for iOS', () => {
-    const cards: InfoCard[] = [
+    const testCards: InfoCard[] = [
       {
         ...dummyCard,
         osTypes: ['android'],
@@ -73,8 +73,8 @@ describe('filterPromoCards', () => {
         localeMessages: { en: 'Another iOS Message' }
       }
     ]
-    const result = filterInfoCards({
-      cards,
+    const result = getDisplayInfoCards({
+      testCards,
       countryCode: 'US',
       buildNumber,
       osType: 'ios',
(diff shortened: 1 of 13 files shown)


src/__tests__/components/PromoCards.test.tsx
```diff
@@ -32,34 +32,18 @@ const currentDate = new Date('2024-06-13T20:53:33.013Z')
...
-      cards,
-      countryCode: 'US',
-      buildNumber,
-      osType,
-      version,
-      osVersion,
-      currentDate
-    })
+    const testCards: InfoCard[] = []
+    const result = getFilteredInfoCards({ testCards, countryCode: 'US', buildNumber, osType, version, osVersion, currentDate })
```
swansontec, line 36: Can you break up this big `{}`? It's hard to read on one line.

src/__tests__/components/PromoCards.test.tsx
```diff
@@ -32,34 +32,18 @@ const currentDate = new Date('2024-06-13T20:53:33.013Z')
...
-      cards,
-      countryCode: 'US',
-      buildNumber,
-      osType,
-      version,
-      osVersion,
-      currentDate
-    })
+    const testCards: InfoCard[] = [{ ...dummyCard }]
+    const result = getFilteredInfoCards({ testCards, countryCode: 'US', buildNumber, osType, version, osVersion, currentDate })
```
swansontec, line 41: Can you break up this big `{}` too?

src/util/infoUtils.ts
```diff
@@ -101,13 +103,38 @@ export function filterInfoCards(params: {
...
+  return filteredInfoData
+}
+
+/**
+ * Similar to filterInfoCards, but returns only an array of `FilteredInfoCard`
+ * display data.
+ *
+ * Ignores any info that doesn't contain display messages.
+ */
+export const getFilteredInfoCards = (props: InfoFilterProps): FilteredInfoCard[] => {
```
swansontec, line 124: I would call this `getDisplayInfoCards`, so it's more obvious how this is different from `filterInfoCards` - this new function prepares the cards for display.

src/actions/AccountReferralActions.ts
```diff
@@ -30,8 +31,12 @@ export function loadAccountReferral(account: EdgeAccount): ThunkAction<Promise<v
         account.disklet.getText(ACCOUNT_REFERRAL_FILE)
       ])
       const cache = asDiskReferralCache(JSON.parse(cacheText))
-      const referral = unpackAccountReferral(JSON.parse(referralText))
+      const referral = await unpackAccountReferral(JSON.parse(referralText))
```
swansontec, line 34: The `unpack` function should not be async - it's job is just to translate what we loaded earlier.

I think you should move the `getActivePromoIds` call out here, in the `loadAccountReferral` function. Then you can merge the active promo ID's with the return value from `unpackAccountReferral`.

Edit: Actually, it looks like `activePromotions` doesn't need to be a part of the account referral anyhow, so you don't need to merge the `getActivePromoIds` and `unpackAccountReferral` return values - they can just live next to each other.

src/actions/AccountReferralActions.ts
```diff
@@ -264,6 +301,7 @@ const asDiskAccountReferral = asObject({
   installerId: asOptional(asString),
   currencyCodes: asOptional(asArray(asCurrencyCode)),
   promotions: asOptional(asArray(asDiskPromotion), []),
+  activePromotions: asOptional(asArray(asString), []),
```
swansontec, line 304: I think we can remove this line. As soon as you load the account referral from disk, you immediately stomp the `activePromotions` with data from the info server. So, there is no reason to even write this to disk in the first place.

Also, it looks like you have a separate reducer to hold this data in memory, so already lives "on the side" anyhow. You can just edit the `ACCOUNT_REFERRAL_LOADED` action to look like this instead, so the new data flows to the right place:

```ts
dispatch({ type: 'ACCOUNT_REFERRAL_LOADED', data: {
  cache,
  referral,
  activePromotions // <- add me!
})
```

src/util/tracking.ts
```diff
@@ -224,6 +224,7 @@ export function logEvent(event: TrackingEventName, values: TrackingValues = {}):
         const { accountReferral } = account
         params.refDeviceInstallerId = deviceReferral.installerId
         params.refDeviceCurrencyCodes = deviceReferral.currencyCodes
+        params.promoIds = accountReferral.activePromotions
```
swansontec, line 227: This would become `state.account.activePromotions`, instead of `state.account.accountReferral.activePromotions`, once you make the change requested in the previous commit.

swansontec (changes requested):

src/actions/AccountReferralActions.ts
Jon-edge: How else would we persist the promos if we didn't write to disk?

I know there's related logic regarding the current account `installerId,` but that's only one piece. 

The general use case is that deeplinks containing promoIds will more easily persist their `promoId` to the device

src/actions/AccountReferralActions.ts
```diff
@@ -243,10 +249,9 @@ export const getCountryCodeByIp = async (): Promise<string | undefined> => {
  */
 async function saveAccountReferral(state: RootState): Promise<void> {
   const { account } = state.core
-  const { accountReferral, activePromotions } = state.account
+  const { accountReferral } = state.account
   const referralData = {
-    ...accountReferral,
-    activePromotions
+    ...accountReferral
```
swansontec, line 254: You can skip this copying step now. Just put this function back to the way it is in `develop`.

swansontec (approved):
Only one small cleanup left. Approving because what's here will work.

PR #5469, opened 2025-02-21
Make text input components semi-controlled

samholmes:
The reason for this change is to fix bugs on Android related to
controlled TextInput components. Cursor placement is not accurate when
entering text for a controlled TextInput on Android.

A semi-controlled component is one where the value of the input is
controlled by the component imperatively; not declaratively. This is
done via the `setNativeProps` method on the TextInput.

This is a nice solution because it abstracts away the imperative nature
of an uncontrolled component (TextInput) with a parent controlled
component (SimpleTextInput/FilledTextInput).

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209448569952172

### DIFF
--- a/src/components/themed/FilledTextInput.tsx
+++ b/src/components/themed/FilledTextInput.tsx
@@ -181,7 +181,8 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
 
   const LeftIcon = iconComponent
   const hasIcon = LeftIcon != null
-  const hasValue = value !== ''
+
+  const valueRef = React.useRef(value)
 
   const marginRemStyle = useMarginRemStyle(marginRemProps)
 
@@ -195,9 +196,9 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
     if (inputRef.current != null) inputRef.current.blur()
   }
   function clear(): void {
+    if (blurOnClear || valueRef.current === '') blur()
     if (inputRef.current != null) inputRef.current.clear()
-    if (onChangeText != null) onChangeText('')
-    if (blurOnClear || !hasValue) blur()
+    handleChangeText('')
     if (onClear != null) onClear()
   }
   function focus(): void {
@@ -209,6 +210,18 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
   function setNativeProps(nativeProps: object): void {
     if (inputRef.current != null) inputRef.current.setNativeProps(nativeProps)
   }
+
+  // This fixes RN bugs with controlled TextInput components.
+  // By avoiding the `value` prop for the TextInput, we can avoid the bugs.
+  // So an alternative to using the `value` prop, is to imperatively set the,
+  // TextInput's text value using `setNativeProps({ text: value })`.
+  // We do this only if the `value` prop from this component has changed.
+  React.useEffect(() => {
+    if (inputRef.current != null && value !== valueRef.current) {
+      valueRef.current = value
+      inputRef.current.setNativeProps({ text: value })
+    }
+  }, [inputRef, value])
 
   React.useImperativeHandle(ref, () => ({
     blur,
@@ -237,6 +250,7 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
     if (onBlur != null) onBlur()
   })
   const handleChangeText = useHandler((value: string) => {
+    valueRef.current = value
     if (autoSelect) {
       setNativeProps({
         selection: {}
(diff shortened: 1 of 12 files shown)


src/components/themed/FilledTextInput.tsx
```diff
@@ -197,7 +198,7 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
   function clear(): void {
     if (inputRef.current != null) inputRef.current.clear()
     if (onChangeText != null) onChangeText('')
-    if (blurOnClear || !hasValue) blur()
+    if (blurOnClear || valueRef.current.length === 0) blur()
```
swansontec, line 201: Optional: `value.current === ''` would be simpler. You use the `value.current.length > 0` pattern in many places, and I feel like it's harder to read - I have to stop & think about math for a split second, vs. it literally just being `''` or not.

src/components/themed/FilledTextInput.tsx
```diff
@@ -237,6 +238,7 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
     if (onBlur != null) onBlur()
   })
   const handleChangeText = useHandler((value: string) => {
+    valueRef.current = value
```
swansontec, line 241: In the opposite direction, do we want an effect like this:

```ts
React.useEffect(() => {
  if (value !== valueRef.current) {
    setNativeProps({ value })
    valueRef.current = value
  }
}, [value])
```

That way, if somebody does `<FilledTextInput value={somethingUnexpected} />`, the text content will update. For normal typing, where the parent just catches the `onChangeText` value and sends it back into the `value` prop, this effect will be a no-op, since `valueRef` will already anticipate the new `value`, and be equal to it. On the other hand, if the parent manually sets the value to something different, such as `onMaxPress = () => setValue(maxBalance)`, the text input will properly update.

Edit: Oh, you already do this in the other component. Please copy it here too.

src/components/themed/FilledTextInput.tsx
```diff
@@ -197,7 +198,7 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
   function clear(): void {
     if (inputRef.current != null) inputRef.current.clear()
     if (onChangeText != null) onChangeText('')
```
swansontec, line 200: You never set `valueRef` to `''` inside the `clear` method. You could either do it explicitly, like `valueRef.current = ''`, or you could change the call to `onChangeText` to `handleChangeText` as you do in the other component.

src/components/themed/SimpleTextInput.tsx
```diff
@@ -148,6 +154,18 @@ export const SimpleTextInput = React.forwardRef<SimpleTextInputRef, SimpleTextIn
...
+  // By avoiding the `value` prop for the TextInput, we can avoid the bugs.
+  // So an alternative to using the `value` prop, is to imperatively set the,
+  // TextInput's text value using `setNativeProps({ text: value })`.
+  // We do this only if the `value` prop from this component has changed.
+  React.useEffect(() => {
+    if (inputRef.current != null && value !== valueRef.current) {
+      valueRef.current = value
+      inputRef.current.setNativeProps({ text: value })
+    }
+  }, [inputRef, value])
```
swansontec, line 167: Since the `inputRef` is a reference, it shouldn't be in the deps list.

src/util/corePlugins.ts
```diff
@@ -3,78 +3,78 @@ import { EdgeCorePluginsInit } from 'edge-core-js'
...
-  litecoin: ENV.LITECOIN_INIT,
-  qtum: true,
-  ravencoin: true,
-  smartcash: true,
-  ufo: true,
-  vertcoin: true,
-  zcoin: ENV.ZCOIN_INIT,
-  // edge-currency-monero:
-  monero: ENV.MONERO_INIT
+  // // // edge-currency-accountbased:
```
swansontec, line 6: This should not be committed.

swansontec (changes requested):
The FilledTextInput has some mistakes, compared to the SimpleTextInput.

src/components/themed/SimpleTextInput.tsx
samholmes: tell that to @react-hooks eslint plugin 🤷‍♂️

src/components/themed/FilledTextInput.tsx
```diff
@@ -196,9 +196,9 @@ export const FilledTextInput = React.forwardRef<FilledTextInputRef, FilledTextIn
     if (inputRef.current != null) inputRef.current.blur()
   }
   function clear(): void {
+    if (blurOnClear || valueRef.current !== '') blur()
```
swansontec, line 199: I think this should be `valueRef.current === ''`, to match how it worked before.

src/components/themed/SimpleTextInput.tsx
swansontec: Weird, but OK.

swansontec (approved):
One single-character "oops"

PR #5478, opened 2025-02-28
Jon/smol-4.24

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209101588771630


src/components/themed/SideMenu.tsx
```diff
@@ -163,18 +163,18 @@ export function SideMenuComponent(props: Props) {
...
     const refId = hashjs.sha256().update(data).digest('hex').replace('0x', '').substring(0, 10)
 
-    const website = `${config.website}?af=appreferred_${refId}`
+    const url = `${config.website}?af=appreferred_${refId}`
 
     const shareOptions = {
       failOnCancel: false,
-      message: Platform.OS === 'ios' ? message : message + website,
-      url: Platform.OS === 'ios' ? website : ''
+      title: lstrings.share_subject,
```
swansontec, line 174: The `share_subject` string has a `%s` in it, so you need to bake it out. While we're here, let's also use async / await to get the debounce effect:

```ts
const subject = sprintf(lstrings.share_subject, config.appName)
await Share.open({
  subject,
  title: subject,
  ...
})
```

swansontec (changes requested):

src/components/themed/SideMenu.tsx
```diff
@@ -162,21 +162,21 @@ export function SideMenuComponent(props: Props) {
...
       failOnCancel: false,
-      title: lstrings.share_subject,
-      subject: lstrings.share_subject,
+      title: subject,
+      subject,
       message: lstrings.share_message,
       url
     }
-    Share.open(shareOptions).catch(e => showError(e))
+    await Share.open(shareOptions).catch(e => showError(e))
```
swansontec, line 179: Remove the `.catch` - the `EdgeTouchableOpacity` deals with that as part of the debounce.

swansontec (approved):

PR #5483, opened 2025-03-01
Fix styling in QrCode

Jon-edge:
- Removed existing styling hack for the QR code itself, which then fixes the crypto icon positioning
- Changed to a dynamically calculated icon size to retain size ratios across device sizes

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device


src/components/themed/QrCode.tsx
```diff
@@ -53,9 +52,29 @@ export function QrCode(props: Props) {
...
+      position: 'absolute',
+      width: iconContainerSize,
+      height: iconContainerSize,
+      borderRadius: theme.cardBorderRadius,
+      backgroundColor: theme.qrBackgroundColor,
+      alignItems: 'center',
+      justifyContent: 'center',
+      left: '50%',
+      top: '50%',
+      transform: [{ translateX: -iconContainerSize / 2 }, { translateY: -iconContainerSize / 2 }]
```
swansontec, line 68: Optional: The transform may not be the most efficient option. You could also do it this way:

```ts
marginLeft: -iconContainerSize / 2,
marginTop: -iconContainerSize / 2
```

This will move the container up & to the left by half it's width, allowing `left: '50%'` and `top: '50%'` to correctly center it. It's very similar to the `transform` you have, but it happens at the layout stage instead of the rendering phase, which is generally perferable.

swansontec (approved):
This PR is quite good, thanks. I do have an optional suggestion.

PR #5488, opened 2025-03-05
Improve loading time for Earn button

Jon-edge:
Can probably fully remove `stakingWallets,` but making a low-impact change for now.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209408349186154

### DIFF
--- a/src/components/scenes/CoinRankingDetailsScene.tsx
+++ b/src/components/scenes/CoinRankingDetailsScene.tsx
@@ -232,10 +232,9 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
   /** Check if all the stake plugins are loaded for this asset type */
   const isStakingLoading =
     stakingWallets.length === 0 ||
-    stakingWallets.some(
+    stakingWallets.every(
       wallet =>
         walletStakingStateMap[wallet.id] == null ||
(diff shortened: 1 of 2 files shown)


src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -235,7 +235,6 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
     stakingWallets.some(
```
swansontec, line 235: What if we change `some` to `every`. As soon as even a single wallet has a policy, we call it "good enough" and show a rate. In that case, we might not even need to remove `isLoading`.

I see how removing `isLoading` speeds things up, but it doesn't feel the best. Still, I approve because it does get the job done.

swansontec (approved):
Suggesting an optional approach, but I would approve this way too if the other way doesn't work out.

src/components/scenes/CoinRankingDetailsScene.tsx
Jon-edge: Your suggestion seems to work

swansontec (approved):

PR #5492, opened 2025-03-11
Reduce Earn button loading in `TransactionListTop`

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device


src/components/themed/TransactionListTop.tsx
```diff
@@ -578,8 +578,9 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
     }
 
     // Handle StakePlugin staking
-    if (stakePlugins != null && stakePolicies != null) {
-      if (stakePolicies.length > 1) {
+    if (stakePlugins != null) {
+      // More than one option or stakePolicies is not yet loaded
+      if (stakePolicies.length > 1 || stakePolicies == null) {
```
swansontec, line 583: This logic seems wrong. If `stakePolicies == null`, then `stakePolicies.length > 1` would crash. However, since `const stakePolicies = Object.keys(...`, it can never be undefined. Just delete the `stakePolicies == null`.

swansontec (approved):

PR #5500, opened 2025-03-14
Jon/fix/stake-options-scene

Jon-edge:
`StakeOptionsScene`, `TransactionListTop`: Reduce dependency on `walletStakingMap`

`CoinRankingDetailsScene`: Now limited by only the time it takes to read `countryCode` from `getFirstOpenInfo()`

General: Avoid relying on `stakePolicy` and `stakePlugins` from `walletStakingMap.` This map takes a while to load since it also checks open positions. The majority of what's required from `stakePolicy`/`stakePlugins` doesn't require position info and can be generated quickly by other means.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209683564147110

### DIFF
--- a/src/components/scenes/CoinRankingDetailsScene.tsx
+++ b/src/components/scenes/CoinRankingDetailsScene.tsx
@@ -202,7 +202,7 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
   }, [matchingWallets.length, coinRankingData])
 
   // Get all stake policies we support
-  const [allStakePolicies] = useAsyncValue<StakePolicy[]>(async () => {
+  const [allStakePolicies = []] = useAsyncValue<StakePolicy[]>(async () => {
     const out = []
     const pluginIds = Object.keys(currencyConfigMap)
     if (currencyCode === 'FIO') {
@@ -222,25 +222,19 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
     return out
   }, [currencyCode, currencyConfigMap])
 
-  const edgeStakingAssets =
-    allStakePolicies == null
-      ? []
-      : matchingEdgeAssets.filter(
-          asset => filterStakePolicies(allStakePolicies, { pluginId: asset.pluginId, currencyCode: currencyCode.toUpperCase() }).length > 0
-        )
+  const edgeStakingAssets = matchingEdgeAssets.filter(
+    asset =>
+      filterStakePolicies(allStakePolicies, {
+        pluginId: asset.pluginId,
+        currencyCode: currencyCode.toUpperCase()
+      }).length > 0
+  )
 
-  /** Check if all the stake plugins are loaded for this asset type */
-  const isStakingLoading =
-    stakingWallets.length === 0 ||
-    stakingWallets.every(
-      wallet =>
-        walletStakingStateMap[wallet.id] == null ||
-        walletStakingStateMap[wallet.id].stakePlugins.length === 0 ||
-        Object.keys(walletStakingStateMap[wallet.id].stakePolicies).length === 0
-    ) ||
-    edgeStakingAssets.length === 0 ||
-    allStakePolicies == null ||
-    allStakePolicies.length === 0
+  /** True if currency supports staking, regardless of if wallets are owned */
+  const isEarnShown = matchingEdgeAssets.some(asset => SPECIAL_CURRENCY_INFO[asset.pluginId]?.isStakingSupported === true)
+
+  /** Check if all the stake plugins/policies are loaded for this asset type */
+  const isStakingLoading = isEarnShown && allStakePolicies.length === 0
 
   const imageUrlObject = React.useMemo(
     () => ({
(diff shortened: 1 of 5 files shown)


src/components/scenes/Staking/StakeOptionsScene.tsx
```diff
@@ -38,9 +41,19 @@ export interface StakeOptionsParams {
...
+    stakePlugin
+      .getPolicies({ pluginId })
+      .filter(
+        stakePolicy =>
+          !stakePolicy.deprecated &&
+          stakePolicy.stakeAssets.some(asset => asset.pluginId === wallet.currencyInfo.pluginId && asset.currencyCode === currencyCode)
+      )
+  )
+  const stakePositionMap = useSelector(state =>
+    state.staking.walletStakingMap[wallet.id] == null ? {} : state.staking.walletStakingMap[wallet.id].stakePositionMap
```
swansontec, line 55: This can be simplified as `state => state.staking.walletStakingMap[wallet.id]?.stakePositionMap ?? {}`

src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -222,25 +222,13 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
...
-      wallet =>
-        walletStakingStateMap[wallet.id] == null ||
-        walletStakingStateMap[wallet.id].stakePlugins.length === 0 ||
-        Object.keys(walletStakingStateMap[wallet.id].stakePolicies).length === 0
-    ) ||
-    edgeStakingAssets.length === 0 ||
-    allStakePolicies == null ||
-    allStakePolicies.length === 0
+  const edgeStakingAssets = matchingEdgeAssets.filter(
+    asset => filterStakePolicies(allStakePolicies, { pluginId: asset.pluginId, currencyCode: currencyCode.toUpperCase() }).length > 0
```
swansontec, line 226: Can we break the `{ ... }` over multiple lines, so we can read it?

src/components/scenes/CoinRankingDetailsScene.tsx
```diff
@@ -222,25 +222,13 @@ const CoinRankingDetailsSceneComponent = (props: Props) => {
...
+  )
+
+  /** Check if all the stake plugins/policies are loaded for this asset type */
```
swansontec, line 229: Since we are going with `/** */`, this comment describes the following variable, not the following code. So should be "True if this currency supports staking"

src/components/scenes/Staking/StakeOptionsScene.tsx
```diff
@@ -149,7 +149,6 @@ const StakeOptionsSceneComponent = (props: Props) => {
             }
             keyExtractor={(stakePolicy: StakePolicy) => stakePolicy.stakePolicyId}
             scrollIndicatorInsets={SCROLL_INDICATOR_INSET_FIX}
-            ListEmptyComponent={<ActivityIndicator color={theme.primaryText} size="large" />}
```
swansontec: This was added in the previous commit, but now removed. Don't add it in the first place?

src/components/themed/TransactionListTop.tsx
```diff
@@ -794,7 +804,8 @@ export function TransactionListTop(props: OwnProps) {
       theme={theme}
       walletName={walletName}
       walletStakingState={walletStakingState}
```
swansontec, line 806: The only remaining things we read from `walletStakingState` are `stakePositionMap` and `lockedBalance`. So please just pass these in directly, and get rid of the big state object.

If you do this, you will find that `stakePolicies` are still being read from the `walletStakingState` instead of your new prop - so potentially more crash locations.

swansontec (changes requested):

Jon-edge:
Note the extra commits: [Rework getPluginFromPolicy](https://github.com/EdgeApp/edge-react-gui/pull/5500/commits/16cae173f8e5f8e91c45133ef4a7f67af6c26935),  [Add TODO comments](https://github.com/EdgeApp/edge-react-gui/pull/5500/commits/7310e8e8c5aa8a345bd50373a017d443e5fa0337)

swansontec (approved):

src/components/themed/TransactionListTop.tsx
```diff
@@ -581,10 +581,9 @@ export class TransactionListTopComponent extends React.PureComponent<Props, Stat
     }
 
     // Handle StakePlugin staking
-    if (stakePolicies.length === 1) {
-      const [stakePolicy] = stakePolicies
-      const { stakePolicyId } = stakePolicy
-      const stakePlugin = getPluginFromPolicyId(stakePlugins, stakePolicy.stakePolicyId, {
+    if (Object.keys(stakePositionMap).length === 1) {
+      const stakePolicyId = Object.keys(stakePositionMap)[0]
+      const stakePlugin = getPluginFromPolicyId(stakePlugins, stakePolicyId, {
```
swansontec, line 586: Please revert this change in behavior, which seems broken. Before, we would bypass the selection page if there was only one option on that page. Now, we bypass the selection page if you have staked funds with a single provider. This means:

1. If there is only one option, but you haven't used it yet, you still get taken to the selection page.
2. If there are multiple options, but you've already staked with one, you can't select another.

Unless there's some other reason I'm not seeing, this specific change looks like a bug.

PR #5503, opened 2025-03-18
Jon/uk-compliance

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209645103166895

### DIFF
--- a/src/components/themed/BuyCrypto.tsx
+++ b/src/components/themed/BuyCrypto.tsx
@@ -3,7 +3,6 @@ import * as React from 'react'
 import { View } from 'react-native'
 import { sprintf } from 'sprintf-js'
 
-import { getFirstOpenInfo } from '../../actions/FirstOpenActions'
 import { DONE_THRESHOLD, SPECIAL_CURRENCY_INFO } from '../../constants/WalletAndCurrencyConstants'
 import { useAsyncValue } from '../../hooks/useAsyncValue'
 import { useHandler } from '../../hooks/useHandler'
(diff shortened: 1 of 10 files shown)


src/util/ukComplianceUtils.ts
```diff
@@ -27,6 +28,15 @@ const formatString = (template: string, values: string[]): string => {
   return template.replace(/%(\d+)\$s/g, (_, index) => values[parseInt(index) - 1] || '')
 }
 
+/** Whether we should hide a non-UK compliant feature */
+export const hideNonUkCompliantFeature = async () => {
```
swansontec, line 32: Please put an explicit return value here, `: Promise<boolean>`

src/components/themed/BuyCrypto.tsx
```diff
@@ -45,13 +43,13 @@ export const BuyCrypto = (props: Props) => {
 
   return (
     <>
-      {allowedPluginIds.includes(pluginId) && tokenId == null && (
+      {allowedPluginIds.includes(pluginId) && tokenId == null && !hideNonUkCompliantFeat && (
```
swansontec, line 46: Instead of expanding the bad old code, let's update it to use `?:`, so `!allowedPluginIds.includes(pluginId) || tokenId != null || hideNonUkCompliantFeat ? null : ...`

swansontec (approved):
I didn't verify that you got all the locations, but the locations I see so far are fine.

There are a few small cleanups to do in here, but I'm approving because I trust them to be done correctly.

PR #5505, opened 2025-03-18
Add exchangeRatesMap to redux for current and 24 hour rates

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209696532407295

### DIFF
--- a/src/actions/ExchangeRateActions.ts
+++ b/src/actions/ExchangeRateActions.ts
@@ -3,7 +3,7 @@ import { makeReactNativeDisklet } from 'disklet'
 import { EdgeAccount } from 'edge-core-js'
 
 import { ThunkAction } from '../types/reduxTypes'
-import { GuiExchangeRates } from '../types/types'
+import { GuiExchangeRates, GuiExchangeRatesMap } from '../types/types'
 import { fetchRates } from '../util/network'
 import { datelog } from '../util/utils'
 
@@ -32,7 +32,8 @@ const asExchangeRateCacheFile = asObject({
 })
 
 type AssetPair = ReturnType<typeof asAssetPair>
-type ExchangeRateCache = ReturnType<typeof asExchangeRateCache>
+// Exported for unit tests
+export type ExchangeRateCache = ReturnType<typeof asExchangeRateCache>
 type ExchangeRateCacheFile = ReturnType<typeof asExchangeRateCacheFile>
 
 let exchangeRateCache: ExchangeRateCacheFile | undefined
@@ -52,6 +53,8 @@ export function updateExchangeRates(): ThunkAction<Promise<void>> {
     const state = getState()
     const { account } = state.core
     const { defaultIsoFiat } = state.ui.settings
+    const now = Date.now()
+    const yesterday = getYesterdayDateRoundDownHour(now).toISOString()
 
     // If this is the first run, immediately use whatever we have on disk
     // before moving on to the potentially slow network:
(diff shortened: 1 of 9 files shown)


src/__tests__/components/__snapshots__/ExchangeQuoteComponent.test.tsx.snap
```diff
@@ -435,7 +435,7 @@ exports[`ExchangeQuote should render with loading props 1`] = `
             ]
           }
         >
-          0.00000001 BTC (&lt;0.01 EUR)
+          0 BTC (&lt;0.01 EUR)
```
swansontec, line 438: If you go into `src/util/fake/fakeRootState.ts` and add new `exchangeRatesMap` dummy data, this would get rid of these snapshot diffs.

src/actions/ExchangeRateActions.ts
```diff
@@ -32,7 +32,7 @@ const asExchangeRateCacheFile = asObject({
 })
 
 type AssetPair = ReturnType<typeof asAssetPair>
-type ExchangeRateCache = ReturnType<typeof asExchangeRateCache>
+export type ExchangeRateCache = ReturnType<typeof asExchangeRateCache>
```
swansontec, line 35: Possibly add `// Exported for unit tests`?

src/actions/ExchangeRateActions.ts
```diff
@@ -238,18 +247,45 @@ async function fetchExchangeRates(account: EdgeAccount, accountIsoFiat: string,
...
+    }
+
+    const codeAMap = outMap.get(codeA) ?? new Map()
+    outMap.set(codeA, codeAMap.set(codeB, { currentRate: rate, yesterdayRate }))
+
+    const codeBMap = outMap.get(codeB) ?? new Map()
+    outMap.set(
+      codeB,
+      codeBMap.set(codeA, { currentRate: out[reverseKey], yesterdayRate: yesterdayRate === 0 || yesterdayRate == null ? 0 : 1 / yesterdayRate })
+    )
```
swansontec, line 282: Oh, interesting. I see that we are keeping the reverse rate thing going here. I was under the impression were going to always have the outer map be the crypto, and the inner map be the fiat, and therefore get rid of the need for reverse rates, since we always know which way is which.

src/actions/ExchangeRateActions.ts
```diff
@@ -60,20 +62,29 @@ export function updateExchangeRates(): ThunkAction<Promise<void>> {
...
+    const { exchangeRates, exchangeRatesMap } = buildGuiRates(exchangeRateCache.rates, yesterday)
     dispatch({
       type: 'EXCHANGE_RATES/UPDATE_EXCHANGE_RATES',
       data: {
-        exchangeRates: buildGuiRates(exchangeRateCache.rates)
+        exchangeRates
+      }
+    })
+    dispatch({
+      type: 'EXCHANGE_RATES/UPDATE_EXCHANGE_RATES_MAP',
```
swansontec, line 85: This leftover redux action needs to be merged with the one above.

src/actions/ExchangeRateActions.ts
peachbits: I started with this assumption but there are places in the app that use crypto/crypto rates. Not sure where fiat/crypto rates would be used, but I ultimately figured it would be safer to just mirror the current rates object for now

swansontec (approved):
Nice.

PR #5512, opened 2025-03-22
Jon/fix/markets-fiat-change

Jon-edge:
- Race conditions
- Multiple fetches for the same data
- Stale periodic task timers retaining the old fiat
- Fetch indexing bugs after changing fiat
- Failure to re-render due to global data cache
- `handleOnEndReached` handling after changing fiat

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209331858493618

### DIFF
--- a/src/components/cards/MarketsCard.tsx
+++ b/src/components/cards/MarketsCard.tsx
@@ -139,11 +139,20 @@ export const MarketsCard = (props: Props) => {
   const coingeckoFiat = useSelector(state => getCoingeckoFiat(state))
 
   const [coinRankingDatas, setCoinRankingDatas] = React.useState<CoinRankingData[]>([])
+  // Detect change to fiat to allow refresh
+  const [lastFetchedFiat, setLastFetchedFiat] = React.useState<string>(coingeckoFiat)
 
   /**
    * Fetch Markets Data
    */
   React.useEffect(() => {
+    // Force an immediate refresh when currency changes
+    if (lastFetchedFiat !== coingeckoFiat) {
+      // Clear the data array to avoid showing stale data
+      setCoinRankingDatas([])
+      setLastFetchedFiat(coingeckoFiat)
+    }
+
     const task = makePeriodicTask(
       async () => {
         const fetchedData = []
@@ -162,6 +171,10 @@ export const MarketsCard = (props: Props) => {
         }
 
         setCoinRankingDatas(fetchedData)
+        // Update lastFetchedFiat after successful fetch
+        if (lastFetchedFiat !== coingeckoFiat) {
+          setLastFetchedFiat(coingeckoFiat)
+        }
       },
       LISTINGS_REFRESH_INTERVAL,
       {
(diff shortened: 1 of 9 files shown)


src/components/cards/MarketsCard.tsx
```diff
@@ -139,11 +139,20 @@ export const MarketsCard = (props: Props) => {
   const coingeckoFiat = useSelector(state => getCoingeckoFiat(state))
 
   const [coinRankingDatas, setCoinRankingDatas] = React.useState<CoinRankingData[]>([])
+  // Detect change to fiat to allow refresh
+  const [lastFetchedFiat, setLastFetchedFiat] = React.useState<string>(coingeckoFiat)
```
samholmes, line 143: A clearer approach to this commit to fix the issue.

First, make the async function for the periodic task a ref so it can depend on the most current state in the component (`coingeckoFiat` and `numRows`). This way the periodic task is always working with the most current state when executing it's routine.

Second, make the effect only depend on `coingeckoFiat` so that it only resubscribes when this changes.

Third, always reset the `coinRankingDatas` state before setting up the periodic task subscription for the component. 

With this approach, there is no need for the extra state `lastFetchedFiat` and we get less churn on the periodic task subscription.

src/components/cards/MarketsCard.tsx
```diff
@@ -175,12 +188,12 @@ export const MarketsCard = (props: Props) => {
...
     // Cleanup logic:
     return () => task.stop()
-  }, [coingeckoFiat, numRows])
+  }, [coingeckoFiat, numRows, lastFetchedFiat])
 
   return (
     <EdgeCard sections>
       {coinRankingDatas.map((coinRow, index) => (
-        <CoinRow key={coinRow.assetId} coinRow={coinRow} fiatCurrencyCode={coingeckoFiat} index={index} {...props} />
+        <CoinRow key={`${coinRow.assetId}-${coingeckoFiat}`} coinRow={coinRow} fiatCurrencyCode={coingeckoFiat} index={index} {...props} />
```
samholmes, line 196: This looks like a separate fix and therefore a separate commit.

samholmes (changes requested):
Last commit should be broken into multiple commits for each change in the itemized list in the commit body.

I feel like the code is decaying because we're dealing with unconventional use of React's API. Instead we could consider improving the implementation to fix the issues without the increasing overhead of tech-debt. Though, it's not clear what that may look like until the changes are cleared up by breaking up the commit.

samholmes (changes requested):
I have a strong feeling that this approach is a shotgun approach to solving the issue. The amount of time spent with the shotgun may have been equal or less time if spent with a handgun (refract the imperative mess) or a sniper riffle (surgical change to get it to work). The reason why I say this is because it seems like using a ref for the query function is so that way it prevents it from being a closure, but then it wants a closure on some previous state values. I feel like the right surgical change here is to pick one approach: is it a closure or is it not a closure? 

From what I can tell, it has to be a closure in order to abort it, so that's one reason to keep it as a closure. That would greatly simplify things (no more `useHandler`), no more `forceRefresh`. If it's no longer a closure (e.g. it's a handler from `useHandler`), then I'm not sure how to abort it, so it's effects should be noop'd at the end of it's routine (not ideal). So I say keep it a closure and see what you can do with that constraint. I think we'll quickly see this PR line changes go down significantly and the change as a whole be a lot more straight forward.

samholmes (changes requested):
Looking good. Just some extra comments from what I noticed while reviewing on my phone.

I’m curious to know how the testing went too.

eslint-local-rules/useAbortable-then-abort.js
```diff
@@ -2,37 +2,37 @@ module.exports = {
   meta: {
     type: 'problem',
     docs: {
-      description: 'Ensure awaited promises within useCancellable are followed by .then(safety)',
+      description: 'Ensure awaited promises within useAbortable are followed by .then(abort)',
```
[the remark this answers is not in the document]
Jon-edge: I think `maybeAbort` is better. "Check" doesn't communicate what happens as a result of that check, just says that the abort condition is being checked. "Abort" is what's potentially going to happen.

samholmes (approved):

PR #5519, opened 2025-03-25
Sam/conf text

samholmes:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device


Jon-edge:
Screenshot?

src/locales/strings/enUS.json
```diff
@@ -160,7 +160,7 @@
   "fragment_transaction_list_tx_failed": "Failed",
   "fragment_transaction_list_tx_synchronizing": "Syncing...",
   "fragment_transaction_list_confirmation_progress": "%s of %s Confirmations",
-  "fragment_transaction_list_unconfirmed_rbf": "Awaiting confirmation, may be cancelled",
+  "fragment_transaction_list_unconfirmed_rbf": "Awaiting confirmation; may be cancelled",
```
Jon-edge, line 163: Also update other applicable languages since this string is already partially localized. 

Maybe also change the string key too, because I don't know if crowdin would cache the original string.

Jon-edge (changes requested):

src/locales/en_US.ts
```diff
@@ -217,7 +217,7 @@ const strings = {
   fragment_transaction_list_tx_failed: 'Failed',
   fragment_transaction_list_tx_synchronizing: 'Syncing...',
   fragment_transaction_list_confirmation_progress: '%s of %s Confirmations',
-  fragment_transaction_list_unconfirmed_rbf: 'Awaiting confirmation, may be cancelled',
+  fragment_transaction_list_unconfirmed_rbf2: 'Awaiting confirmation; may be cancelled',
```
Jon-edge, line 220: I generally avoid using a number in the string key, and just use those opportunities to improve the key name. There's almost always an opportunity to improve the key name, i.e. `transaction_awaiting_confirmation_rbf,` since this isn't specific to the tx list, potentially can be reused elsewhere, and is not a fragment. 

Fragments aren't even really a thing anymore because we don't concatenate string fragments anymore because it doesn't make sense for all locales.

Jon-edge (approved):
Approved with optional change

PR #5520, opened 2025-03-27
Jon/usp-updates

Jon-edge:
image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209517334134374

### DIFF
--- a/src/components/scenes/GettingStartedScene.tsx
+++ b/src/components/scenes/GettingStartedScene.tsx
@@ -383,8 +383,10 @@ const WelcomeHeroMessage = styled(EdgeText)(theme => ({
 }))
 const WelcomeHeroPrompt = styled(EdgeText)(theme => ({
   fontSize: theme.rem(0.75),
+  color: theme.textLink,
   fontFamily: theme.fontFaceBold,
-  textAlign: 'center'
+  textAlign: 'center',
+  margin: theme.rem(0.5)
 }))
 
 const HeroItem = styled(Animated.View)<{
@@ -558,9 +560,8 @@ const SectionParagraph = styled(EdgeText)(theme => ({
 const Footnote = styled(EdgeText)(theme => ({
   color: theme.primaryText,
   fontFamily: theme.fontFaceDefault,
-  fontSize: theme.rem(0.6),
+  fontSize: theme.rem(0.75),
   marginBottom: theme.rem(1),
-  opacity: 0.75,
   includeFontPadding: false
 }))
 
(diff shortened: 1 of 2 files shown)


src/components/scenes/GettingStartedScene.tsx
```diff
@@ -136,9 +136,15 @@ export const GettingStartedScene = (props: Props) => {
     visitPasswordScene()
   })
 
-  const handlePressSignUp = useHandler(() => {
-    dispatch(logEvent('Signup_Welcome'))
-    visitNewAccountScene()
+  const handlePressGetStartedNext = useHandler(() => {
```
samholmes, line 139: `handleProgressButtonPress`

src/components/scenes/GettingStartedScene.tsx
```diff
@@ -136,9 +136,15 @@ export const GettingStartedScene = (props: Props) => {
...
+    // If we're at the last slide, navigate to account creation
+    if (swipeOffset.value >= sections.length) {
+      dispatch(logEvent('Signup_Welcome'))
+      handleFinalSwipe()
```
samholmes, line 143: Why is this case necessary to handle here?

src/components/scenes/GettingStartedScene.tsx
```diff
@@ -156,15 +162,44 @@ export const GettingStartedScene = (props: Props) => {
...
+
+        <EdgeAnim visible={showNextButton} enter={fadeIn} exit={fadeOut}>
+          <ButtonsView
+            layout="column"
+            primary={{
+              label: lstrings.string_next_capitalized,
+              onPress: handlePressGetStartedNext
+            }}
+          />
+        </EdgeAnim>
```
samholmes, line 201: I really wish we could just use a `Button` component and give it a `Text` component as a child so that way we can animate `Text`.

samholmes (changes requested):

src/components/scenes/GettingStartedScene.tsx
Jon-edge: It needs to behave the same as a swipe - progressing past the last slide should
complete the flow.

src/components/scenes/GettingStartedScene.tsx
Jon-edge: Yeah this is a new use case that wasn't anticipated in the original design. A
good idea for future changes

src/components/scenes/GettingStartedScene.tsx
```diff
@@ -156,15 +162,44 @@ export const GettingStartedScene = (props: Props) => {
     }
   )
 
+  // Use state to track which button should be visible
+  const [showNextButton, setShowNextButton] = React.useState(false)
```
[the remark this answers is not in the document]
Jon-edge: Declaring a derived value depending on a reanimated value does not trigger a rerender. 

There are some other ways like `useAnimatedReactiveValue` to bridge, but either way we need to jump through hoops and what's here is the cleanest out of the available options.

samholmes (approved):

PR #5531, opened 2025-04-04
Remove iOS rendering hack in `EdgeCarousel`

Jon-edge:
This affects only iOS only when there is exactly one card. 

This is the conflict between the original rendering hack that was required to get iOS to work with exactly 1 card and the recent fix for: Home - Fix blogPostsGeo (swapix article) https://app.asana.com/0/1200234943382353/1208440993939949/f

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209868735707894

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - changed: Newly created wallets are added to `mostRecentWallets`
 - fixed: `EdgeCarousel` not updating when data changes
 - fixed: Network error handling in `SwipeChart`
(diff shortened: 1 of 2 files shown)


samholmes (approved):
Removing hacks seems good if they're no longer needed. I hope this doesn't cause a regression?

Jon-edge:
The hack was clearly documented for specifically handling this case of one card not rendering immediately for iOS. Testing on physical iOS devices shows for whatever reason, this is no longer needed.

PR #5541, opened 2025-04-18
Enable ACH for moonpay

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1210025029944382

### DIFF
--- a/src/plugins/gui/providers/moonpayProvider.ts
+++ b/src/plugins/gui/providers/moonpayProvider.ts
@@ -39,6 +39,7 @@ const allowedCurrencyCodes: Record<FiatDirection, { [F in FiatPaymentType]?: Fia
     venmo: { providerId, fiat: {}, crypto: {} }
   },
   sell: {
+    ach: { providerId, fiat: {}, crypto: {}, requiredAmountType: 'crypto' },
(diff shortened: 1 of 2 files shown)


src/plugins/gui/providers/moonpayProvider.ts
```diff
@@ -163,7 +164,7 @@ const MOONPAY_PAYMENT_TYPE_MAP: Partial<Record<FiatPaymentType, MoonpayPaymentMe
   applepay: 'credit_debit_card',
   credit: 'credit_debit_card',
   googlepay: 'credit_debit_card',
-  iach: 'ach_bank_transfer',
```
samholmes, line 166: What is `iach` and why is it replaced by `ach`?

samholmes (approved):

src/plugins/gui/providers/moonpayProvider.ts
paullinator: iach is a purchase only ach method which we were using with moonpay years ago. now we're only enabling an `ach` sell method. moonpay uses the same code but iach isn't supported for sell, only buy.

PR #5542, opened 2025-04-22
Jon/fix/country-code-cache-redux

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1209994809610414


src/components/modals/InsufficientFeesModal.tsx
```diff
@@ -12,26 +12,38 @@ import { getCurrencyCode } from '../../util/CurrencyInfoHelpers'
...
+
 /**
  * Show this modal when the wallet doesn't have enough funds to cover fees.
+ * If UK and swaps are disabled, this is a noop
  */
-export function InsufficientFeesModal(props: Props) {
+export const showInsufficientFeesModal = async (props: Props) => {
+  const { countryCode } = props
+
+  if (config.disableSwaps === true && countryCode === 'GB') return
```
swansontec, line 42: In this case, would it make sense to show a simple `ButtonsModal` with a message like "Insufficient funds to cover fees". Otherwise, UK users would try to send tokens and just silently fail with no feedback. We can't help the user solve the problem, since we can't offer any swaps or buys, but we can at least tell them that there is a problem.

src/util/ukComplianceUtils.ts
```diff
@@ -1,25 +1,14 @@
...
-type CompliantStringKeys =
-  | 'buy_1s'
-  | 'buy_1s_quote'
-  | 'sell_1s'
-  | 'stake_earn_1s'
-  | 'stake_earn_button_label'
-  | 'stake_x_to_earn_y'
-  | 'insufficient_fees_2s'
-  | 'insufficient_fees_3s'
+type CompliantStringKeys = 'stake_earn_1s' | 'stake_earn_button_label' | 'stake_x_to_earn_y' | 'insufficient_fees_2s' | 'insufficient_fees_3s'
```
swansontec, line 4: If you add a `//` somewhere in here, you can force it back onto multiple lines so it's readable.

swansontec (approved):
This will work! However, I have two suggestions and one question.

src/components/modals/InsufficientFeesModal.tsx
Jon-edge: I reasoned that the scene itself would show the error, but you're right that keeping some sort of modal keeps the same flow and give more room for modified messaging

swansontec (approved):

PR #5543, opened 2025-04-23
Implement Duress Mode setting scene

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

- https://github.com/EdgeApp/edge-core-js/pull/645

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209356449454549
  
Simulator Screenshot - iPhone 15 Pro - 2025-04-23 at 11 29 31
Simulator Screenshot - iPhone 15 Pro - 2025-04-23 at 11 29 24
Simulator Screenshot - iPhone 15 Pro - 2025-04-23 at 11 29 15
Simulator Screenshot - iPhone 15 Pro - 2025-04-23 at 11 27 18
Simulator Screenshot - iPhone 15 Pro - 2025-04-23 at 11 27 16
Simulator Screenshot - iPhone 15 Pro - 2025-04-23 at 11 27 05
Simulator Screenshot - iPhone 15 Pro - 2025-04-23 at 11 27 03
Simulator Screenshot - iPhone 15 Pro - 2025-04-23 at 11 27 00


src/components/scenes/DuressModeSettingScene.tsx
```diff
@@ -0,0 +1,132 @@
...
+  return (
+    <SceneWrapper scroll>
+      <HeadingContainer>
+        <HeadingImageContainer>
+          <HeadingImage source={illustrationImageSource} />
+        </HeadingImageContainer>
+        <Heading>{lstrings.title_duress_mode}</Heading>
+      </HeadingContainer>
+
+      <BodyContainer>
```
Jon-edge, line 68: Same comment as `DuressModeHowToScene` - we should just have one view that
manages the entire outer expected margins on scene edges, since all the inner components (with
a few exceptions such as scrolling containers) all expect 0.5rem outer margins

src/components/scenes/DuressPinScene.tsx
```diff
@@ -0,0 +1,85 @@
...
+    }
+  })
+
+  const scrollViewRef = useScrollToEnd(isValidPin)
+
+  return (
+    <SceneWrapper scroll={false}>
+      <SceneContainer expand>
+        <SceneHeaderUi4 title={lstrings.title_set_duress_pin} />
+        <ScrollView ref={scrollViewRef} contentContainerStyle={styles.content} keyboardShouldPersistTaps="handled">
```
Jon-edge, line 51: See relevant comment in `DuressModeSettingScene`

src/components/scenes/inputs/DigitInput/PinDots.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react'
```
Jon-edge, line 1: Add comment: https://github.com/EdgeApp/edge-react-gui/blob/fe9d69a037ace353a92ffda75d2f3e7068ecd339/src/components/modals/EdgeModal.tsx#L1-L4

src/components/layout/SceneContainer.tsx
```diff
@@ -4,13 +4,15 @@ import { UndoInsetStyle } from '../common/SceneWrapper'
...
 interface Props {
+  expand?: boolean
   undoTop?: boolean
   undoRight?: boolean
   undoBottom?: boolean
   undoLeft?: boolean
   undoInsetStyle?: UndoInsetStyle
 }
-export const SceneContainer = styled(View)<Props>(theme => ({ undoTop, undoRight, undoBottom, undoLeft, undoInsetStyle }) => ({
+export const SceneContainer = styled(View)<Props>(theme => ({ expand, undoTop, undoRight, undoBottom, undoLeft, undoInsetStyle }) => ({
```
Jon-edge, line 14: Add a `title` prop and handle the positioning of `SceneHeaderUi4` inside here. This way we don't have to use additional views to layout the scene header along with the rest of the scene contents. `SceneContainer` can then be the sole provider of all the 0.5rem margins around the edges of the scene and no custom spacing will be required for UI4 compatible components.

src/components/scenes/DuressModeSettingScene/SelectDuressAccountModal.tsx
```diff
@@ -0,0 +1,65 @@
+import { EdgeUserInfo } from 'edge-core-js'
```
Jon-edge, line 1: File no longer used

src/components/lists/SelectableRowList.tsx
```diff
@@ -0,0 +1,50 @@
...
+  keyExtractor: (item: T, index: number) => string
+  titleExtractor: (item: T, index: number) => string
+  onPress: (item: T) => void
+}
+
+export function SelectableRowList<T>(props: Props<T>) {
+  const theme = useTheme()
+  const renderRow = useHandler(({ item, index }: ListRenderItemInfo<T>) => {
+    return (
+      // Would use SelectableRow, but it can't have a chevron icon
```
Jon-edge, line 22: File unused. 

Even if we did add this component for this feature, I would have just updated `SelectableRow`, because it represents the same visual concept. Confirmation would be required, of course, but I'd be 99% positive this would be considered the same visual concept, therefore should be an extension of the same component or the `SelectableRow` should be refactored such that common overlap is lifted out and a refined parent/child component relationship created.

If we kept it completely separate, future changes to either component by other developers would happen without the knowledge of the other component's existence, creating a potential bug in consistency. Not only that, but we might also have a visual design pattern bug that was overlooked that would be used in future designs.

Enforce a heirarchy and create dependencies whenever possible instead of creating new components to avoid potential inconsistencies in the future. Doing so also keeps a dialogue open to designers to ensure we have thorough definitions and mutual understanding of our visual design heirarchy and layout+styling patterns. We also communicate in code to other developers what "a type of thing" has been defined as already.

src/hooks/useScrollToEnd.ts
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import { ScrollView } from 'react-native'
+
+/**
+ * This hook will return a ref, which you can attach to a scroll view.
+ * When the `ready` parameter becomes true,
```
Jon-edge, line 6: Nice hook, I can see this being useful.

`ready` is pretty cryptic IMO and I'm assuming is specific to describing how the prop is currently used on its calling scene, which might not always be true in all cases. 

Let's keep it closer to what is literally happening here. How bout `triggerScrollToEnd`?

Jon-edge: Actually, why limit this to scrolling to the end? Let's make this: 
`useForceScrollPosition(position: undefined | 'start' | 'end')`

Jon-edge (changes requested):
Lots of neat new things in here that I can see being useful throughout our GUI

src/components/scenes/CreateAccountScene.tsx
```diff
@@ -0,0 +1,75 @@
+import { EdgeAccount } from 'edge-core-js'
```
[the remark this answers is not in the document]
samholmes: Actually it's not needed to create the account anymore since duress accounts get created by virtue of calling `changePin({ pin, forDuressAccount: true })`.

src/components/scenes/inputs/DigitInput.tsx
```diff
@@ -0,0 +1,74 @@
+import * as React from 'react'
```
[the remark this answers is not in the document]
samholmes: Sounds like a good idea.

src/components/scenes/DuressModeHowToScene.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  const handleDone = useHandler(() => {
+    navigation.goBack()
+  })
+
+  return (
+    <SceneWrapper scroll>
+      <SceneContainer>
+        <SceneHeaderUi4 title={lstrings.how_duress_mode_works} />
+        <BodyContainerView>
```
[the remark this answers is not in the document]
samholmes: Removed the BodyContainer

src/components/scenes/DuressModeSettingScene.tsx
samholmes: Removed the BodyContainer here as well in favor of SceneContainer

src/components/scenes/DuressPinScene.tsx
samholmes: See which comment for what?

src/components/layout/SceneContainer.tsx
samholmes: I'm going to push back on this request a bit. Why must SceneContainer extend SceneHeaderUi4? The coupling would cause more work as SceneHeaderUi4 ever changes it's API, unless you're claiming ossification?

src/hooks/useScrollToEnd.ts
samholmes: Added a lil'

```ts
/** 
  * IMPORTANT: Changes in this file MUST be synced between edge-react-gui and 
  * edge-login-ui-rn! 
  */ 
```

;)

samholmes: Added a lil'

```ts
/** 
  * IMPORTANT: Changes in this file MUST be synced between edge-react-gui and 
  * edge-login-ui-rn! 
  */ 
```

src/components/scenes/DuressPinScene.tsx
Jon-edge: There's only one comment in that file

src/components/layout/SceneContainer.tsx
Jon-edge: For business logic, sure, composition all the way. This is a visual concept that applies anywhere there is a SceneHeaderUi4, inheritance is the way to go.

There's never a situation where we want to change the title layout for one scene independent of another, it's a global visual concept. Therefore, there's no reason to give the next developer the freedom to compose.

src/components/scenes/Staking/StakeOptionsScene.tsx
```diff
@@ -138,6 +138,15 @@ const StakeOptionsSceneComponent = (props: Props) => {
             contentContainerStyle={{ paddingBottom: insetStyle.paddingBottom }}
             ListHeaderComponent={
               <>
+                {/* We have to use the SceneHeaderUi4 component here because
```
Jon-edge, line 141: Yeah I anticipated this would be the next step - scroll behaviors.

I see this not as a one-off to accept, this to me seems like a visual and/or design bug. There are other things wrong with this scene too - gesturing to scroll this list can cut off the content. 
Since there's other small things wrong with the scene that warrant changing, this is a perfect place to open a conversation with design: whether we want the scene header to scroll with the rest of the content (which IMO looks really weird the way it is now).

The decision becomes a design pattern that then applies to all similar situations. If say scene headers shouldn't be part of the content being scrolled, then we're green lit to continue using the design pattern we just made (sceneHeaderTitle prop).

If not, and the consideration is that we want to see more content, then an alternate proposal could be to use the smaller title in the top bar outside of the SceneContainer.

On the broader topic of titles in general, I personally prefer we just stick with the top nav bar and ditch our logo (avoiding "vanity" is a good design).

It's not 100% a failure of the abstraction just because something doesn't match up, it's an opportunity to question it and bring it up with the team, because we SHOULD be able to neatly define these design patters, especially something this simple.

A better action item comment here would be: "See if it's acceptable to use the `SceneContainer's` `headerTitle` prop here instead"

Jon-edge (changes requested):

Jon-edge (approved):

PR #5544, opened 2025-04-23
Jon/fix/coinrank-loading

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1210038006206752

### DIFF
--- a/src/components/scenes/CoinRankingScene.tsx
+++ b/src/components/scenes/CoinRankingScene.tsx
@@ -176,7 +176,7 @@ const CoinRankingComponent = (props: Props) => {
         startIndex += QUERY_PAGE_SIZE
       }
     } catch (e: any) {
-      console.warn(`Error during data fetch: ${e.message}`)
+      console.warn(`Error during data fetch: ${e.message}, ${coingeckoFiat}, ${startIndex}, ${requestDataSize}`)
     }
 
     setDataSize(coinRankingDatas.length)
--- a/src/components/scenes/CoinRankingScene.tsx
+++ b/src/components/scenes/CoinRankingScene.tsx
@@ -157,7 +157,7 @@ const CoinRankingComponent = (props: Props) => {
 
     try {
       // Catch up to the total required items
-      while (startIndex < requestDataSize - QUERY_PAGE_SIZE) {
+      while (startIndex <= requestDataSize - QUERY_PAGE_SIZE + 1) {
         const url = `v2/coinrank?fiatCode=iso:${coingeckoFiat}&start=${startIndex}&length=${QUERY_PAGE_SIZE}`
         const response = await fetchRates(url).then(maybeAbort)
         if (!response.ok) {
--- a/src/components/scenes/CoinRankingScene.tsx
+++ b/src/components/scenes/CoinRankingScene.tsx
@@ -18,6 +18,7 @@ import { fetchRates } from '../../util/network'
 import { EdgeAnim, MAX_LIST_ITEMS_ANIM } from '../common/EdgeAnim'
 import { EdgeTouchableOpacity } from '../common/EdgeTouchableOpacity'
 import { SceneWrapper } from '../common/SceneWrapper'
+import { FillLoader } from '../progress-indicators/FillLoader'
 import { CoinRankRow } from '../rows/CoinRankRow'
 import { showDevError } from '../services/AirshipInstance'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
@@ -26,6 +27,12 @@ import { EdgeText } from '../themed/EdgeText'
 import { SearchFooter } from '../themed/SearchFooter'
 
 const coinRanking: CoinRanking = { coinRankingDatas: [] }
+
+/**
+ * Track changes that occurred to fiat while scene is unmounted. Don't react to
+ * changes to this value.
+ */
+let lastSceneFiat: string
 
 const QUERY_PAGE_SIZE = 30
 const LISTINGS_REFRESH_INTERVAL = 30000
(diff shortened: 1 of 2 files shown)


src/components/scenes/CoinRankingScene.tsx
```diff
@@ -139,9 +142,22 @@ const CoinRankingComponent = (props: Props) => {
...
 
   React.useEffect(() => {
+    if (lastSceneFiat !== coingeckoFiat) {
+      // Clear cache if we changed the fiat while outside of this scene
+      debugLog(LOG_COINRANK, 'Clearing coinRankingDatas cache')
+      coinRanking.coinRankingDatas = []
+      lastStartIndex.current = 1
+      setDataSize(0)
+      setRequestDataSize(QUERY_PAGE_SIZE)
+    }
```
samholmes, line 153: Why not just put this in `useMount` hook?

src/components/scenes/CoinRankingScene.tsx
```diff
@@ -194,30 +212,35 @@ const CoinRankingComponent = (props: Props) => {
...
+        clearTimeout(timeoutId)
+        loopAbort()
+        pageQueryAbortRef.current()
+        coinRanking.coinRankingDatas = []
+        lastStartIndex.current = 1
+        setDataSize(0)
+        setRequestDataSize(QUERY_PAGE_SIZE)
+      }
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
samholmes, line 244: Why was this `// eslint-disable-next-line react-hooks/exhaustive-deps` added? Is there an alternative?

src/components/scenes/CoinRankingScene.tsx
```diff
@@ -139,9 +142,22 @@ const CoinRankingComponent = (props: Props) => {
...
+      lastStartIndex.current = 1
+      setDataSize(0)
+      setRequestDataSize(QUERY_PAGE_SIZE)
+    }
+
     return () => {
       mounted.current = false
+
+      // Queries should update this, but just in case:
+      lastSceneFiat = coingeckoFiat
```
samholmes, line 159: Why not put this in a `useUnmount` hook?

samholmes (changes requested):

src/components/scenes/CoinRankingScene.tsx
Jon-edge: This was the cause of the actual bug in the task - this cleanup fn is appropriate only for when the scene is still mounted, implying that the fiat changed while the scene was mounted, so clearing everything was appropriate.

It's inappropriate to run this cleanup when the scene unmounts, which was happening in the original implementation.

src/components/scenes/CoinRankingScene.tsx
Jon-edge: See comment
https://github.com/EdgeApp/edge-react-gui/pull/5544#discussion_r2058954481

We need to pass mounted.current around separately anyway, so might as well fold
the changes into this existing effect.

src/components/scenes/CoinRankingScene.tsx
Jon-edge: See comment
https://github.com/EdgeApp/edge-react-gui/pull/5544#discussion_r2058954481

We need to pass mounted.current around separately anyway, so might as well fold
the changes into this existing effect.

samholmes: Using `useMount` for the purpose of avoid the `eslint-disable-next-line` rule hack

src/components/scenes/CoinRankingScene.tsx
samholmes: What's the reason for the `eslint-disable-next-line` rule hack though? Seeing this hack is a red flag. I'm trying to understand why it's necessary and perhaps why another approach wasn't taken.

src/components/scenes/CoinRankingScene.tsx
samholmes: Why I suggest `useUnmount` is to avoid the disabled eslint rule hack.

samholmes (approved):
This looks like a good approach to solving the problem. Well done! Approved, with consideration that this is well tested to be a solution.

samholmes (approved):
This looks like a good approach to solving the problem. Well done! Approved, with consideration that this is well tested to be a solution.

src/components/scenes/CoinRankingScene.tsx
```diff
@@ -176,7 +176,7 @@ const CoinRankingComponent = (props: Props) => {
         startIndex += QUERY_PAGE_SIZE
       }
     } catch (e: any) {
-      console.warn(`Error during data fetch: ${e.message}`)
+      console.warn(`Error during data fetch: ${e.message}, ${coingeckoFiat}, ${startIndex}, ${requestDataSize}`)
```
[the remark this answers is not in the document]
Jon-edge: Ah good catch yeah I didn't update this to be publicly readable from my own testing

src/components/scenes/CoinRankingScene.tsx
```diff
@@ -157,7 +157,7 @@ const CoinRankingComponent = (props: Props) => {
 
     try {
       // Catch up to the total required items
-      while (startIndex < requestDataSize - QUERY_PAGE_SIZE) {
+      while (startIndex <= requestDataSize - QUERY_PAGE_SIZE + 1) {
```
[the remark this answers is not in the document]
Jon-edge: Yes, and this also fixes the issue where the query doesn't start immediately under some situations. 

You can see in the changes you reviewed previously, I had a condition `startIndex === 1 || ...` which resolved the "not starting immediately" issue, but didn't fix the new page requests sometimes not fetching new rows.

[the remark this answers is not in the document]
Jon-edge: I'll edit the commit message

PR #5546, opened 2025-04-23
Make bottom insets static if the scene does not `avoidKeyboard`

Jon-edge:
For scenes that don't avoid the keyboard, the scene boundaries should never change.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209438524785867
  - https://app.asana.com/0/0/1209551697973955

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - added: Add PIVX.
 - added: Monero multi output support.
 - added: Moonpay Sell support for ACH.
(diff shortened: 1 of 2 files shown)


samholmes (changes requested):
You're missing adding `avoidKeyboard={false}` to the RequestScene

samholmes (approved):
Approved cause it's `false` by default for scenes that don't have `scroll` prop.

PR #5548, opened 2025-04-25
Jon/fix/ios-carousel

Jon-edge:
In the previous change that attempted to fix flickering of the promo card, `dataLocal` as a dependency was causing an infinite loop for iOS repeatedly clearing the data every 500ms.

The rendering hacks was incorrectly verified as no longer being necessary.  and removed.

The hacks  are still necessary, but need only trigger once on mount.

Delay also reduced to be less perceptible.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1210049206021043

### DIFF
--- a/src/components/common/EdgeCarousel.tsx
+++ b/src/components/common/EdgeCarousel.tsx
@@ -1,7 +1,8 @@
 import React, { useState } from 'react'
-import { ListRenderItem, View } from 'react-native'
+import { InteractionManager, ListRenderItem, Platform, View } from 'react-native'
 import Carousel, { Pagination } from 'react-native-snap-carousel'
 
+import { useAsyncEffect } from '../../hooks/useAsyncEffect'
 import { useHandler } from '../../hooks/useHandler'
 import { cacheStyles, Theme, useTheme } from '../services/ThemeContext'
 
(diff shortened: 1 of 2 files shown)


src/components/common/EdgeCarousel.tsx
```diff
@@ -26,16 +27,47 @@ export function EdgeCarousel<T>(props: Props<T>): JSX.Element {
...
+   */
+  useAsyncEffect(
+    async () => {
+      // HACK: With 1 item, this is the only way to force a render in iOS
+      if (Platform.OS === 'ios' && dataLocal.length === 1) {
+        const tempData = [...dataLocal]
+        setDataLocal([])
+        setTimeout(() => {
+          setDataLocal(tempData)
+        }, 50)
```
swansontec, line 53: What about:

```ts
setDataLocal([])
setTimeout(() => setDataLocal(data), 50)
```

This would let the hook's dependencies actually match the deps array, and would also get rid of the `tempData` variable.

swansontec (approved):
One cleanup suggestion.

Also, please squash the two commits into one. The changes in the first commit are broken without the second commit, so there's no point in keeping them apart.

swansontec (approved):

PR #5561, opened 2025-05-07
Add Apple Ads Attribution

Jon-edge:
https://developer.apple.com/documentation/AdServices/AAAttribution/attributionToken()?changes=l_6_6_8_7_8_2

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1210036875460976

### DIFF
--- a/src/actions/FirstOpenActions.tsx
+++ b/src/actions/FirstOpenActions.tsx
@@ -1,19 +1,36 @@
+import { getAttributionToken } from '@brigad/react-native-adservices'
 import { asNumber, asObject, asOptional, asString, asValue } from 'cleaners'
 import { makeReactNativeDisklet } from 'disklet'
+import { Platform } from 'react-native'
 
 import { FIRST_OPEN } from '../constants/constantSettings'
 import { makeUuid } from '../util/rnUtils'
+import { snooze } from '../util/utils'
 import { getCountryCodeByIp } from './AccountReferralActions'
 
 export const firstOpenDisklet = makeReactNativeDisklet()
 
-const asFirstOpenInfo = asObject({
+const asAppleAdsAttribution = asObject({
+  campaignId: asOptional(asString),
+  keywordId: asOptional(asString)
+})
+type AppleAdsAttribution = ReturnType<typeof asAppleAdsAttribution>
+
+interface FirstOpenInfo {
+  isFirstOpen: 'true' | 'false'
+  deviceId: string
+  firstOpenEpoch: number
+  countryCode?: string
+  appleAdsAttribution?: AppleAdsAttribution
+}
+type FirstOpenInfoFile = Omit<FirstOpenInfo, 'appleAdsAttribution'>
+
+const asFirstOpenInfoFile = asObject<FirstOpenInfoFile>({
   isFirstOpen: asValue('true', 'false'),
   deviceId: asString,
   firstOpenEpoch: asNumber,
   countryCode: asOptional(asString)
 })
-type FirstOpenInfo = ReturnType<typeof asFirstOpenInfo>
 
 let firstOpenInfo: FirstOpenInfo
 let firstLoadPromise: Promise<FirstOpenInfo> | undefined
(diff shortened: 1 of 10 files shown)


src/actions/FirstOpenActions.tsx
```diff
@@ -1,17 +1,27 @@
...
 export const firstOpenDisklet = makeReactNativeDisklet()
 
+const asAppleAdsAttribution = asObject({
+  campaignId: asOptional(asNumber),
+  keywordId: asOptional(asNumber)
+})
+type AppleAdsAttribution = ReturnType<typeof asAppleAdsAttribution>
+
 const asFirstOpenInfo = asObject({
   isFirstOpen: asValue('true', 'false'),
```
swansontec, line 20: This should be `asBoolean`.

src/actions/FirstOpenActions.tsx
```diff
@@ -42,9 +52,11 @@ const readFirstOpenInfoFromDisk = async (): Promise<FirstOpenInfo> => {
       firstOpenText = await firstOpenDisklet.getText(FIRST_OPEN)
       firstOpenInfo = asFirstOpenInfo(JSON.parse(firstOpenText))
       firstOpenInfo.isFirstOpen = 'false'
+      firstOpenInfo.appleAdsAttribution = await getAppleAdsAttribution()
```
swansontec, line 55: IF we load this on every boot, why even save it? Please remove this from the file.

One way to do this is to separate the types for the file and the in-memory data type:

```ts
interface FirstOpenInfo {
  isFirstOpen: boolean,
  deviceId: string,
  firstOpenEpoch: number,
  countryCode?: string,
  appleAdsAttribution?: AppleAdsAttribution
}

type FirstOpenInfoFile = Omit<FirstOpenInfo, 'appleAdsAttribution'>
const asFirstOpenInfoFile = asObject<FirstOpenInfoFile>({
  ...
})
```

It might make sense to create a whole new place to put stuff like this, instead of jamming it into this grab-bag file, but I see that that's a bigger change.

swansontec (changes requested):

src/actions/FirstOpenActions.tsx
Jon-edge: For `isFirstOpen`? I think this was leftover from google analytics and we need to avoid data inconsistencies. If we change this to a boolean, the existing PostHog reports might need to all be updated

swansontec: Oh, shoot! These are strings. Nevermind, I totally missed that. We could clean this up using a `asStringBool` cleaner like this:

```ts
const asStringBool: Cleaner<boolean> = (raw) => {
  if (typeof raw === 'string'){
    if (raw === 'true') return true
    if (raw === 'false') return false
  }
  return asBoolean(raw)
}
```

This would coerce the existing data into a proper boolean on the next open. However, I see that this is out-of-scope for this PR, so let's shelve the idea for now.

src/actions/FirstOpenActions.tsx
```diff
@@ -16,14 +16,21 @@ const asAppleAdsAttribution = asObject({
...
+  isFirstOpen: boolean
+  deviceId: string
+  firstOpenEpoch: number
+  countryCode?: string
+  appleAdsAttribution?: AppleAdsAttribution
+}
+type FirstOpenInfoFile = Omit<FirstOpenInfo, 'appleAdsAttribution'>
+
+const asFirstOpenInfoFile = asObject<FirstOpenInfoFile>({
+  isFirstOpen: asBoolean,
```
swansontec, line 29: Oh, shoot! I didn't see that were quoted strings on disk. Ok, we can still be backwards compatible using a cleaner like this:

```ts
const asStringBool: Cleaner<boolean> = (raw) => {
  if (typeof raw === 'string'){
    if (raw === 'true') return true
    if (raw === 'false') return false
  }
  return asBoolean(raw)
}
```

This translate the existing data into a boolean on the next open. Otherwise, we could go back to `asValue('true', 'false')`, which is insane but what we have.

swansontec (changes requested):

src/actions/FirstOpenActions.tsx
Jon-edge: Yeah in GA4 days, we had trouble using boolean values, so we resorted to strings. We didn't want to change the local disk cache since then, so we had to keep it that way.

swansontec (approved):

PR #5563, opened 2025-05-07
Fix Android display bugs in DuressModeHowToScene

samholmes:
The problem was caused by using Text element `Paragraph` as the container element for the entire markdown block. We shouldn't nest Text elements with the exception of inline-elements necessary for emphasis, bold, etc.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1210118083810748

### DIFF
--- a/src/util/parseMarkdown.tsx
+++ b/src/util/parseMarkdown.tsx
@@ -35,7 +35,7 @@ function tokenToReactNode(token: MarkedToken, key: string): React.ReactNode {
     }
     case 'paragraph': {
       console.log('P')
-      return <Paragraph>{subTokens ?? token.text}</Paragraph>
+      return <Paragraph>{subTokens ?? <EdgeText numberOfLines={1000}>{token.text}</EdgeText>}</Paragraph>
     }
     case 'em': {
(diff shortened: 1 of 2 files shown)


src/util/parseMarkdown.tsx
```diff
@@ -67,11 +65,13 @@ function tokenToReactNode(token: MarkedToken, key: string): React.ReactNode {
 const Markdown = styled(View)(theme => ({
   flexDirection: 'column',
   alignItems: 'stretch',
-  justifyContent: 'flex-start'
+  justifyContent: 'flex-start',
+  margin: theme.rem(0.5)
 }))
 
 const Paragraph = styled(View)(theme => ({
-  // backgroundColor: 'pink',
```
Jon-edge, line 74: Remove commented styles

Jon-edge (approved):

src/util/parseMarkdown.tsx
samholmes: Was going to leave them there for debugging. Ye or ne?

PR #5572, opened 2025-05-13
Paul/nicer multiout

paullinator:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device


Jon-edge:
I'm approving because it serves the needs of our internal use case (topping off a number of cards), but for the general public, there's some room for improvement.

1. Consider the use case of someone wanting to quickly grab a batch of addresses and then split payments, and either doesn't know or wants to adjust amounts later. This linear flow doesn't really work.
2. Multi output is very hidden. There's no indication that the feature even exists. The only way for it to show up is if the user to finish entering the amount, though it's pretty easy to miss considering the person is likely already thinking about completing the flow.
3. Would consider a bug - clearing out a previous "fully populated" recipient whose payment amount is set automatically brings up the scanner again.
4. There's no way to edit amounts for a prior recipient. Yes you could remove them and re-scan them, but what if you lost the QR code or address? Don't know which one it was in a pile of cards? Not having this capability presents a real logistical challenge.
5. I found myself squinting and straining to draw the separation between completed fields and what I'm currently editing. Everything is in one big block.

These all stem from the main issue of the "amount" entry being the trigger to permanently progress the flow, which is somewhat convenient for the "loading many cards" use case, but is not intuitive or flexible for other potential use cases.

**Suggestions**
I'd prefer:

image

A button that exists by default below the last amount entry field, agnostic of how many recipients there are nor if amounts are populated. This would resolve the first 4 points above. 

a. "Add Recipient" as a secondary or tertiary control visually makes the feature obvious that it exists, from the beginning of the flow, and without "insider knowledge" that the first amount needs to be populated in order to activate the feature.
b. Not having to commit to amounts beforehand naturally gives a way to edit them at any time, without resorting to losing the address

If everything above is deemed out of scope, there's low hanging fruit that can be done as a minor change, appropriate no matter the scope of changes chosen:

There are (at least) two very distinct sets of information in the flow as implemented:
As implemented in the PR:
[Populated]
[Not Populated]

If going the route of my previous suggestions, the organization is:
[Editable recipient 1]
[Editable recipient 2]
[Editable recipient 3]
(Add Recipient Button)

The UI should communicate this separation of concerns with separate card groups.

**"Auto-scan" in general (somewhat out of scope, but would be nice to address here):**
The scanner automatically showing up at the beginning of the send flow IMO has always been incredibly invasive, and assumes the user even has a need for the scanner in the first place. Even for someone like me who is familiar with the app, it takes me a split second to think of how to enter an address manually. I can only imagine what a new user would feel. Yes the UI could be improved but that's besides the point, it's a UX flow issue.

Someone could be primarily pasting addresses from their mobile browser, or trying to rebalance between a number of browsers, yet our UX assumes all users are primarily scanning things in the physical world.

Suggest considering letting the user choose how to proceed by default. Only if they previously scanned should we bring up the scanner again automatically on the next entry into the Send scene.

Jon-edge (approved):
Approved because it's fairly hidden, but strongly suggest making at least some changes as outlined.

Jon-edge (approved):

PR #5574, opened 2025-05-13
Add Stellar memo labels

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1210233839549668


src/locales/en_US.ts
```diff
@@ -886,12 +886,16 @@ const strings = {
...
+  memo_text_title: 'Memo Text',
+  memo_id_title: 'Memo ID',
   memo_note_title: 'Transaction Note',
   memo_op_return_title: 'OP_RETURN',
   memo_payment_id_title: 'Payment ID',
   memo_comment_label: 'comment',
   memo_destination_tag_label: 'destination tag',
   memo_memo_label: 'memo',
+  memo_text_label: 'memo_text',
+  memo_id_label: 'memo_id',
```
swansontec, line 898: Do we want to show the user "memo_text"? Would it be better as just "memo text", like in the title but lower case? That would match the others.

swansontec (approved):
One question.

src/locales/en_US.ts
peachbits: Yeah i can fix that

PR #5586, opened 2025-05-19
Use `getTokenDetails` in `EditTokenScene`

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-core-js/pull/658
https://github.com/EdgeApp/edge-currency-accountbased/pull/936

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1210145057911911

### DIFF
--- a/src/components/scenes/EditTokenScene.tsx
+++ b/src/components/scenes/EditTokenScene.tsx
@@ -187,10 +187,38 @@ function EditTokenSceneComponent(props: Props) {
     }
   })
 
+  const renderCustomTokenTemplateRows = () => {
+    return customTokenTemplate
+      .sort((a, b) => (a.key === 'contractAddress' ? -1 : 1))
+      .map(item => {
+        if (item.type === 'nativeAmount') return null
+        return (
+          <FilledTextInput
+            key={item.key}
+            aroundRem={0.5}
+            autoCapitalize="none"
+            autoCorrect={false}
+            autoFocus={false}
+            placeholder={translateDescription(item.displayName)}
+            keyboardType={item.type === 'number' ? 'numeric' : 'default'}
+            value={location.get(item.key) ?? ''}
+            onChangeText={value =>
+              setLocation(location => {
+                const out = new Map(location)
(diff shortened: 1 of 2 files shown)


src/components/scenes/EditTokenScene.tsx
```diff
@@ -202,13 +226,17 @@ function EditTokenSceneComponent(props: Props) {
...
+            onChangeText={value => {
               setLocation(location => {
                 const out = new Map(location)
                 out.set(item.key, value.replace(/\s/g, ''))
                 return out
               })
-            }
+
+              if (item.key === 'contractAddress') {
+                autoCompleteToken(value).catch(console.error)
```
swansontec, line 237: Ok, so in the other PR's, we talked about hiding errors in at the plugin level. We might even do that here too (`.catch(() => [])`), so we don't bother logging the errors to the console. It could be quite spammy if the internet is down.

src/components/scenes/EditTokenScene.tsx
```diff
@@ -187,6 +190,27 @@ function EditTokenSceneComponent(props: Props) {
     }
   })
 
+  const autoCompleteToken = async (searchString: string) => {
+    // Ignore autocomplete if the scene was initialized with any of the token details prefilled,
+    if (route.params.currencyCode != null || route.params.displayName != null || route.params.multiplier != null || route.params.networkLocation != null) {
+      return
+    }
+
+    const [token] = await wallet.currencyConfig.getTokenDetails({ contractAddress: searchString })
```
swansontec, line 199: We need some way to rate-limit this call, so we don't spam the network as the user types. The simplest way is a busy flag, so if one call is pending, we don't trigger any more until after that one finishes.

peachbits: I added a pretty basic rate limiter. There are obvious race conditions but since this is a field the user would most likely paste a value in I'm not concerned about it.

PR #5593, opened 2025-05-22
Jon/fix/moonpay fiat

Jon-edge:
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209185638016387

### DIFF
--- a/src/plugins/gui/amountQuotePlugin.ts
+++ b/src/plugins/gui/amountQuotePlugin.ts
@@ -13,7 +13,14 @@ import { infoServerData } from '../../util/network'
 import { logEvent } from '../../util/tracking'
 import { DECIMAL_PRECISION, fuzzyTimeout, removeIsoPrefix } from '../../util/utils'
 import { FiatPlugin, FiatPluginFactory, FiatPluginFactoryArgs, FiatPluginStartParams } from './fiatPluginTypes'
-import { FiatProvider, FiatProviderAssetMap, FiatProviderGetQuoteParams, FiatProviderQuote } from './fiatProviderTypes'
+import {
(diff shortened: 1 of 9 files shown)


swansontec (approved):

src/plugins/gui/amountQuotePlugin.ts
```diff
@@ -476,7 +483,16 @@ export const amountQuoteFiatPlugin: FiatPluginFactory = async (params: FiatPlugi
           const noQuoteText = direction === 'buy' ? lstrings.fiat_plugin_buy_no_quote : lstrings.fiat_plugin_sell_no_quote
           if (goodQuotes.length === 0) {
             // Find the best error to surface
-            const bestError = getBestError(errors as any, sourceFieldCurrencyCode, direction)
+            const fpQuoteErrors = errors
+              .map(error => {
+                if (error instanceof FiatProviderError) {
+                  return error.quoteError
+                }
+                return error
```
swansontec, line 491: Won't this return a `!= null` value, so the filter below will let the wrong thing through?

Jon-edge: This was originally expecting `error` to be a `FiatProviderError.` In Moonpay's case, was passing a `FiatProviderQuoteError,` the child object within the former.

This change tries to maintain the original expectations should it arise for other plugins.

1. Handle `FiatProviderErrors`
2. Handle `FiatProviderQuoteErrors`
3. Fallback to a default error for other situations

swansontec (approved):

PR #5594, opened 2025-05-22
Add Revolut fiat provider

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209616327127637

### DIFF
--- a/src/plugins/gui/providers/revolutProvider.ts
+++ b/src/plugins/gui/providers/revolutProvider.ts
@@ -1,0 +1,347 @@
+import { asMaybe } from 'cleaners'
+import { EdgeTokenId } from 'edge-core-js'
+import { sprintf } from 'sprintf-js'
+
+import { lstrings } from '../../../locales/strings'
+import { CryptoAmount } from '../../../util/CryptoAmount'
+import {
+  FiatProvider,
+  FiatProviderApproveQuoteParams,
+  FiatProviderAssetMap,
+  FiatProviderError,
+  FiatProviderFactory,
+  FiatProviderFactoryParams,
+  FiatProviderGetQuoteParams,
+  FiatProviderGetSupportedAssetsParams,
+  FiatProviderGetTokenIdFromContract,
+  FiatProviderQuote
+} from '../fiatProviderTypes'
+import {
+  asRevolutCrypto,
+  asRevolutFiat,
+  fetchRevolutConfig,
+  fetchRevolutQuote,
+  fetchRevolutRedirectUrl,
+  RevolutConfig,
+  RevolutCrypto,
+  RevolutFiat
+} from '../util/fetchRevolut'
+import { makeCheckDue } from './common'
+import { ProviderSupportStore } from './ProviderSupportStore'
+
+const providerId = 'revolut'
+const partnerIcon = 'revolut.png'
+const pluginDisplayName = 'Revolut'
+
+export const revolutProvider: FiatProviderFactory = {
+  providerId,
+  storeId: providerId,
+  makeProvider: async (
+    params: FiatProviderFactoryParams
+  ): Promise<FiatProvider> => {
+    const { getTokenIdFromContract, io } = params
+    const isCheckDue = makeCheckDue(1000 * 60 * 60) // 1 hour
+
+    const supportedAssets = new ProviderSupportStore(providerId)
+    supportedAssets.add.direction('buy')
+
+    const provider: FiatProvider = {
+      providerId,
+      partnerIcon,
+      pluginDisplayName,
+      getSupportedAssets: async (
+        params: FiatProviderGetSupportedAssetsParams
+      ): Promise<FiatProviderAssetMap> => {
+        const region =
+          params.regionCode.stateProvinceCode == null
+            ? params.regionCode.countryCode
+            : `${params.regionCode.countryCode}:${params.regionCode.stateProvinceCode}`
+
+        if (isCheckDue()) {
+          const configData = await fetchRevolutConfig()
+          processRevolutConfig(
+            configData,
+            getTokenIdFromContract,
+            supportedAssets
+          )
+        }
+
+        const assetMap = supportedAssets.getFiatProviderAssetMap({
+          direction: params.direction,
+          region,
+          payment: params.paymentTypes[0]
+        })
+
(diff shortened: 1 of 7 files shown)


src/plugins/gui/providers/revolutProvider.ts
```diff
@@ -0,0 +1,278 @@
...
+  supportedAssets.addFiatInfo(fiatKey, fiat)
+}
+
+export function processRevolutConfig(
+  configData: RevolutConfig,
+  getTokenIdFromContract: FiatProviderGetTokenIdFromContract,
+  supportedAssets: ProviderSupportStore
+) {
+  configData.countries.forEach(country => {
+    supportedAssets.add.direction('*').region(country).fiat('*').payment('*')
```
paullinator, line 336: Is Revolut really fully orthogonal on their support of assets across all countries?

src/plugins/gui/providers/revolutProvider.ts
```diff
@@ -0,0 +1,278 @@
...
+        if (revolutFiat == null) {
+          throw new Error(`Revolut fiat not found: ${fiatKey}`)
+        }
+        if (revolutCrypto == null) {
+          throw new Error(`Revolut crypto not found: ${cryptoKey}`)
+        }
+
+        const quoteData = await fetchRevolutQuote({
+          fiat: revolutFiat.currency,
+          amount: params.exchangeAmount,
```
paullinator, line 105: `exchangeAmount` could be the crypto or fiat amount. Need to check `amountType` first before assuming

paullinator: if the quoting only works with fiat then set the `requiredAmountType = 'fiat'` in the `FiatProviderAssetMap`

src/plugins/gui/providers/revolutProvider.ts
```diff
@@ -0,0 +1,278 @@
...
+          providerId,
+          partnerIcon,
+          pluginDisplayName,
+          displayCurrencyCode: params.displayCurrencyCode,
+          cryptoAmount,
+          isEstimate: false,
+          fiatCurrencyCode: params.fiatCurrencyCode,
+          fiatAmount,
+          direction: params.direction,
+          expirationDate: undefined,
```
paullinator: is there really no expiration on the quote? really should set something. Maybe at least 3 minutes.

src/plugins/gui/providers/revolutProvider.ts
```diff
@@ -0,0 +1,278 @@
...
+              amount: parseFloat(fiatAmount),
+              crypto: quoteData.crypto.currencyId,
+              payment: 'revolut',
+              region: params.regionCode.countryCode,
+              wallet: walletAddress,
+              partnerRedirectUrl: successReturnURL,
+              orderId
+            })
+
+            console.log('!@! redirectUrl', redirectUrl)
```
paullinator: Prefix the log with `revolut`

src/plugins/gui/providers/revolutProvider.ts
```diff
@@ -0,0 +1,278 @@
...
+
+            await showUi.openExternalWebView({
+              url: redirectUrl,
+              providerId,
+              async deeplinkHandler(link) {
+                if (link.direction === 'sell') {
+                  throw new Error('Unexpected sell return link. Sell not supported by Revolut.')
```
paullinator: if sell isn't supported, you should throw early in `getQuote`

```
          throw new FiatProviderError({
            providerId,
            errorType: 'paymentUnsupported'
          })
```

src/plugins/gui/providers/revolutProvider.ts
```diff
@@ -0,0 +1,278 @@
...
+                }
+                const { transactionStatus } = link.query
+                if (transactionStatus === 'success') {
```
paullinator, line 173: What other `transactionStatus` values are there? Looks like we'll just do nothing if we get a different status. At minimum we need an error.

src/plugins/gui/providers/revolutProvider.ts
```diff
@@ -0,0 +1,278 @@
...
+                      conversionType: 'buy',
+                      sourceFiatCurrencyCode: params.fiatCurrencyCode,
+                      sourceFiatAmount: fiatAmount,
+                      destAmount: new CryptoAmount({
+                        currencyConfig: coreWallet.currencyConfig,
+                        currencyCode: params.displayCurrencyCode,
+                        exchangeAmount: cryptoAmount
+                      }),
+                      fiatProviderId: providerId,
+                      orderId
```
paullinator, line 185: Please absolutely triple check that the `orderId` is exactly what gets returned from their reporting API as these need to match to give affiliates credit for conversions.

src/plugins/gui/providers/revolutProvider.ts
samholmes: I believe so. I forgot since I worked on it weeks ago. There has to be a way to double check.

src/plugins/gui/providers/revolutProvider.ts
samholmes: From reading the API documented in `fetchRevolut` which is modeled from their HTTP API, there is only an `amount` field which is a fiat amount. So I added `requiredAmountType: 'fiat'`

src/plugins/gui/providers/revolutProvider.ts
samholmes: Assuming 1 minute.

paullinator (approved):

src/plugins/gui/providers/revolutProvider.ts
samholmes: According to [this configuration endpoint](https://developer.revolut.com/docs/crypto-ramp/retrieve-configuration), there is no indication of a relationship between specific countries, fiat currencies, and crypto currencies.

src/plugins/gui/providers/revolutProvider.ts
samholmes: According to [this endpoint](https://developer.revolut.com/docs/crypto-ramp/retrieve-a-redirect-url), we define the UUID for the order. Notice our use of `io.makeUuid()`

PR #5606, opened 2025-05-31
Fix FIO token codes

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

### DIFF
--- a/src/util/FioAddressUtils.ts
+++ b/src/util/FioAddressUtils.ts
@@ -12,6 +12,7 @@ import { CcWalletMap } from '../reducers/FioReducer'
 import { EdgeAsset, FioAddress, FioConnectionWalletItem, FioDomain, FioObtRecord, StringMap } from '../types/types'
 import { asIntegerString } from './cleaners/asIntegerString'
 import { getWalletName } from './CurrencyWalletHelpers'
+import { infoServerData } from './network'
 import { DECIMAL_PRECISION, truncateDecimals } from './utils'
 
 const CONNECTED_WALLETS = 'ConnectedWallets.json'
@@ -355,7 +356,10 @@ const updatePublicAddresses = async (
   try {
     edgeTx = await fioMakeSpend(fioWallet, action, {
       fioAddress,
-      publicAddresses
+      publicAddresses: publicAddresses.map(p => ({
+        ...p,
+        token_code: infoServerData.rollup?.fioTokens?.[p.token_code] ?? p.token_code.replace('.', '')
+      }))
     })
     fee = edgeTx.networkFee
   } catch (e: any) {


src/components/scenes/Fio/FioConnectWalletConfirmScene.tsx
```diff
@@ -27,222 +28,222 @@ export interface FioConnectWalletConfirmParams {
...
+  const [connectWalletsLoading, setConnectWalletsLoading] = React.useState(false)
+  const [showSlider, setShowSlider] = React.useState(true)
 
-export class FioConnectWalletConfirm extends React.Component<Props, State> {
-  state = {
-    acknowledge: false,
-    connectWalletsLoading: false,
-    showSlider: true
-  }
+  const updateConnectedWallets = useHandler((currentFioAddress: string, currentCcWalletMap: CcWalletMap) => {
```
samholmes: This doesn't need to use `useHandler` hook because it's not being used as a prop. It can just be a function declaration or arrow function. This will avoid the need for a unnecessary ref (which the useHandler hook creates).

Alternatively, you could just inline the `dispatch({ ... })` call where each `updateConnectedWallets` is invoked.

src/components/scenes/Fio/FioConnectWalletConfirmScene.tsx
```diff
@@ -27,222 +28,222 @@ export interface FioConnectWalletConfirmParams {
...
-      const newCcWalletMap = { ...ccWalletMap }
-      try {
-        const getCompatibleAddress = async (wallet: EdgeCurrencyWallet): Promise<string> => {
-          const edgeAddresses = await wallet.getAddresses({ tokenId: null })
-          const edgeAddress = edgeAddresses.find(edgeAddress => edgeAddress.publicAddress.length <= 128)
-          if (edgeAddress == null) {
-            throw new Error('Address exceeds 128 characters')
-          }
-          return edgeAddress.publicAddress
+  const resetSlider = useHandler(() => {
```
samholmes: This doesn't need to be a handler.

src/components/scenes/Fio/FioConnectWalletConfirmScene.tsx
```diff
@@ -27,222 +28,222 @@ export interface FioConnectWalletConfirmParams {
...
+    requestAnimationFrame(() => {
+      setShowSlider(true)
+    })
+  })
+
+  const handleCheck = useHandler(() => {
+    setAcknowledge(!acknowledge)
+  })
+
+  const renderWalletLine = useHandler((wallet: FioConnectionWalletItem) => {
```
samholmes: This shouldn't be a handler because it renders.

src/components/scenes/Fio/FioConnectWalletConfirmScene.tsx
```diff
@@ -27,222 +28,222 @@ export interface FioConnectWalletConfirmParams {
...
+    const label = `${wallet.name} (${wallet.currencyCode})`
+    return (
+      <EdgeText key={`${wallet.id}-${wallet.currencyCode}`} style={styles.content}>
+        {label}
+      </EdgeText>
+    )
+  })
+
+  const confirm = useHandler(async () => {
```
samholmes: Recommend rename `handleSlideComplete` or `handleConfirmSlideComplete`.

src/components/scenes/Fio/FioConnectWalletConfirmScene.tsx
```diff
@@ -27,222 +28,222 @@ export interface FioConnectWalletConfirmParams {
...
-          }
-          return edgeAddress.publicAddress
+  const resetSlider = useHandler(() => {
+    setShowSlider(false)
+    requestAnimationFrame(() => {
+      setShowSlider(true)
+    })
+  })
+
+  const handleCheck = useHandler(() => {
```
samholmes: Usually include the event name: `handleCheckPress` is my rec.

src/constants/FioConstants.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Special mapping that defines `chain_codes` and `token_codes` for FIO tx's
+ * that do not fit the typical pattern of using currency codes
+ */
+export const FIO_ASSET_MAP: { [pluginId: string]: { chainCode: string; tokenCodes: { [address: string]: string } } } = {
```
samholmes: Can the type annotation be aliased? It'd be clearer to read.

src/util/FioAddressUtils.ts
```diff
@@ -356,10 +360,28 @@ const updatePublicAddresses = async (
   try {
     edgeTx = await fioMakeSpend(fioWallet, action, {
       fioAddress,
-      publicAddresses
+      publicAddresses: publicAddresses.map(p => {
+        const fioAssets = infoServerData.rollup?.fioAssets ?? FIO_ASSET_MAP
+        const pluginId = Object.keys(fioAssets).find(pid => fioAssets[pid].chainCode === p.chain_code)
```
samholmes: I'm noticing this lookup pattern where our constraint is not only that the key in the `fioAssets` map is a match, but also the `chainCode` field matches. Why is that?

Also, could we please have `pid` change to `pluginId` since `pid` is easily confused with `process ID`? This is an optional request. I get tripped up on acronym usage unless they're well established.

src/util/FioAddressUtils.ts
```diff
@@ -356,10 +360,28 @@ const updatePublicAddresses = async (
...
+
+        let parsedTokenCode = p.token_code.replace('.', '')
+
+        if (pluginId != null) {
+          const fioAsset = fioAssets[pluginId]
+          const tokenId = getTokenId(account.currencyConfig[pluginId], p.token_code)
+
+          parsedTokenCode = tokenId == null ? parsedTokenCode : fioAsset.tokenCodes[tokenId] ?? parsedTokenCode
```
samholmes:
```
const tokenCode = fioAsset.tokenCodes[tokenId]
if (tokenId != null && tokenCode != null) {
  parsedTokenCode = tokenCode
}
```
Seems like we're casing the assignment statement under these conditions, so why not just put it in a code block me thinks.

src/util/FioAddressUtils.ts
```diff
@@ -356,10 +360,28 @@ const updatePublicAddresses = async (
...
+
+        return {
+          ...p,
+          token_code: parsedTokenCode
+        }
+      })
     })
     fee = edgeTx.networkFee
   } catch (e: any) {
+    console.error(e)
```
samholmes, line 389: Looks like left-over from debugging; to be removed.

samholmes (approved):
Approved, but please make some cleanup changes. Don't see any critical problem with this PR.

src/util/FioAddressUtils.ts
samholmes: Our error logging should happen at the top of the stack.

src/util/FioAddressUtils.ts
Jon-edge: We're going from fio "chain code" to Edge pluginId

src/util/FioAddressUtils.ts
Jon-edge: It's a weird statement because tokenId is possibly undefined.

PR #5610, opened 2025-06-02
Apple Ads Attribution (without IDFA)

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1210374719087124

### DIFF
--- a/src/util/tracking.ts
+++ b/src/util/tracking.ts
@@ -139,6 +139,7 @@ export interface TrackingValues extends LoginTrackingValues {
   numAccounts?: number // Number of full accounts saved on the device
   surveyCategory2?: string // User's answer to a survey (first tier response)
   surveyResponse2?: string // User's answer to a survey
+  appleAdsKeywordId?: string // Apple Search Ads attribution keyword ID
 
   // Conversion values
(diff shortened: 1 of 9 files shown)


ios/edge/AppDelegate.mm
```diff
@@ -8,6 +8,8 @@
 #import "ExpoModulesCore-Swift.h"
 #import "Edge-Swift.h"
 #import "RNBootSplash.h"
+#import <AdServices/AdServices.h>
+#import <AppTrackingTransparency/AppTrackingTransparency.h>
```
swansontec: Adding a header file does nothing, unless you are using the code inside. It's basically the same as an `import` statement in JavaScript.

Ok, maybe some `import` statements have side effects, and I guess maybe that's true here too, but it's pretty weird for a C-style language. I'm guessing you can delete these.

src/util/tracking.ts
```diff
@@ -201,12 +202,13 @@ export function trackError(
 /**
  * Send a raw event to all backends.
  */
+
```
swansontec: Extra newline.

swansontec (approved):
Two small cleanups. Also, please squash the whole thing into one commit. It doesn't make any sense as separate commits.

PR #5612, opened 2025-06-04
Special case Moonpay/Ripple destination tag handling

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1210446427096845

### DIFF
--- a/src/plugins/gui/providers/moonpayProvider.ts
+++ b/src/plugins/gui/providers/moonpayProvider.ts
@@ -188,7 +188,7 @@ const NETWORK_CODE_PLUGINID_MAP: StringMap = {
   polygon: 'polygon',
   ripple: 'ripple',
   solana: 'solana',
-  sonic: 'sonic',
+  s_sonic: 'sonic',
   stellar: 'stellar',
   sui: 'sui',
   tezos: 'tezos',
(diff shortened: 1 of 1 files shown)


src/plugins/gui/providers/moonpayProvider.ts
```diff
@@ -197,6 +197,24 @@ const NETWORK_CODE_PLUGINID_MAP: StringMap = {
...
+    hidden: true
+  }
+
+  switch (pluginId) {
+    case 'ripple': {
+      memo.type = 'number'
+      memo.memoName = 'destination tag'
+    }    
+  }
+  return memo
```
swansontec: This code seems odd. Try:

```ts
const createMemo = (pluginId: string, value: string): EdgeMemo => {
  const memo: EdgeMemo = {
    type: 'text',
    value,
    hidden: true
  }

  switch (pluginId) {
    case 'ripple': 
      return {  ...memo, type: 'number' }
    default:
      return memo
  }
}
```

Now all the return statements live in the `switch`, and we avoid editing data. Also, I don't think `memoName` is needed in an EdgeSpendInfo?

swansontec (approved):
Approved with optional cleanups.

PR #5620, opened 2025-06-06
GUI: Add change username setting

Jon-edge:
image
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1210036875460967


src/actions/AccountActions.tsx
```diff
@@ -28,7 +28,7 @@ interface ValidatePasswordOptions {
   warningMessage?: string
 }
 
-export function validatePassword(opts: ValidatePasswordOptions = {}): ThunkAction<Promise<boolean>> {
+export function validatePassword(opts: ValidatePasswordOptions = {}): ThunkAction<Promise<string | undefined>> {
```
swansontec: Note for next time: This change would have made sense as its own commit.

src/components/scenes/ChangeUsernameScreen.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const context = useSelector(state => state.core.context)
+
+  const handleComplete = useHandler((result?: { username?: string | undefined }) => {
+    if (result?.username == null) return
+
+    account.changeUsername({ username: result.username, password }).catch(error => console.error(error))
+    logActivity(`Username changed to: ${result.username}`)
+
+    navigation.goBack()
+  })
```
swansontec: This logic should be in login-ui, similar to how change-pin or change-password work. Those basically just use login-ui directly with nothing more than a `SceneWraper` container around it.

swansontec (changes requested):

PR #5621, opened 2025-06-06
Add "Recent Wallets" to the top of the Wallet List when searching

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1209661800501921

### DIFF
--- a/src/components/themed/WalletListSwipeable.tsx
+++ b/src/components/themed/WalletListSwipeable.tsx
@@ -11,7 +11,7 @@ import { filterWalletCreateItemListBySearchText, getCreateWalletList, WalletCrea
 import { useSceneScrollHandler } from '../../state/SceneScrollState'
 import { useDispatch, useSelector } from '../../types/reactRedux'
 import { NavigationBase, WalletsTabSceneProps } from '../../types/routerTypes'
-import { FlatListItem } from '../../types/types'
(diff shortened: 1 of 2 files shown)


src/components/themed/WalletListSwipeable.tsx
```diff
@@ -93,7 +94,33 @@ function WalletListSwipeableComponent(props: Props) {
...
+
+          const pa = recentIndex[keyA]
+          const pb = recentIndex[keyB]
+          if (pa != null && pb != null) return pa - pb
+          if (pa != null) return -1
+          if (pb != null) return 1
+        }
+        return a.index - b.index
+      })
+      .map(({ item }) => item)
```
swansontec: While this works, it's a weird way to do it. I would have done something more like this:

```ts
// Make a list of recent wallets:
let recentWallets = []
for (const {id, currencyCode} of mostRecentWallets) {
  const item = list.find(...)
  if (item != null) recentWallets.push(item)
}

// Put the recent wallets in the front:
return [
  ...recentWallets,
  ...list.filter(item => !recentWallets.includes(item))
]
```

However, I do see that your way works, so this is just a suggestion. I still approve your version.

swansontec (approved):
This is fine, but a bit weirdly complicated.

PR #5622, opened 2025-06-06
Add review triggers for user actions

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1210131153156905

### DIFF
--- a/src/__tests__/actions/RequestReviewActions.test.ts
+++ b/src/__tests__/actions/RequestReviewActions.test.ts
@@ -1,0 +1,325 @@
+import { beforeEach, describe, expect, jest } from '@jest/globals'
+import { makeMemoryDisklet } from 'disklet'
+import { EdgeAccount } from 'edge-core-js'
+import { Action, Dispatch } from 'redux'
+
+import { LOCAL_SETTINGS_FILENAME } from '../../actions/LocalSettingsActions'
+import {
+  DEPOSIT_AMOUNT_THRESHOLD,
+  FIAT_PURCHASE_COUNT_THRESHOLD,
+  markAccountUpgraded,
+  readReviewTriggerData,
+  SWAP_COUNT_DATA_FILE,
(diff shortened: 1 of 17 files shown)


src/__tests__/actions/RequestReviewActions.test.ts
```diff
@@ -0,0 +1,263 @@
...
+    mockDisklet = makeMemoryDisklet()
+    mockAccount = { disklet: mockDisklet }
+
+    // Clear any data that might have been set in the review trigger file
+    try {
+      await mockDisklet.delete(REVIEW_TRIGGER_DATA_FILE)
+    } catch (e) {
+      // File might not exist, which is fine
+    }
+  })
```
swansontec: This whole block shouldn't be needed, since it's a brand-new disklet object. There is no way it can have any contents.

src/actions/RequestReviewActions.tsx
```diff
@@ -10,16 +10,67 @@ import { lstrings } from '../locales/strings'
 import { config } from '../theme/appConfig'
 import { ThunkAction } from '../types/reduxTypes'
 
-const SWAP_COUNT_DATA_FILE = 'swapCountData.json'
-const MANY_SWAPS_TO_TRIGGER_REQUEST = 3
+// File to store all review trigger data
+export const REVIEW_TRIGGER_DATA_FILE = 'reviewTriggerData.json'
```
swansontec: New files are banned. Since this is synced account-level data, it should go in the `Settings.json` of the account.

You don't need to break out all the individual fields in the settings reducer. You can just make a single update function for the whole `ReviewTriggerData` structure. Then most of the code below stays the same, but reading `state.settings.reviewTrigger` instead of re-reading from disk every time.

swansontec (changes requested):

PR #5626, opened 2025-06-11
Translate unexpected pending transactions errors

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

### DIFF
--- a/src/components/scenes/SendScene2.tsx
+++ b/src/components/scenes/SendScene2.tsx
@@ -37,6 +37,7 @@ import { useState } from '../../types/reactHooks'
 import { useDispatch, useSelector } from '../../types/reactRedux'
 import { EdgeAppSceneProps, NavigationBase } from '../../types/routerTypes'
 import { FioRequest, GuiExchangeRates } from '../../types/types'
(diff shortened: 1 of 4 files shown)


src/components/scenes/SendScene2.tsx
```diff
@@ -1106,6 +1107,10 @@ const SendComponent = (props: Props) => {
           }
         }
 
+        if (asMaybeUnexpectedPendingTxsError(e) != null) {
```
swansontec: Suggestion:

Since accountbased hasn't defined an explicit error type here, like `class FooError extends Errror`, I would prefer to check the message more explicitly. So do `if (e instanceof Error && e.message === 'Unexpected pending transactions')` instead of defining `asMaybeUnexpectedPendingTxsError`. We don't need to pay for the abstraction here, since we aren't getting any of the benefits we normally expect.

We need to use the `asMaybeFooError` pattern when we define a new `class FooError extends Errror` type, since the normal `instanceof` operator doesn't work across yaob. Defining these error cleaners alongside the error type (in the same library) ensures that changes to the error itself (like a different message) get reflected in the cleaner. If we really did want `asMaybeUnexpectedPendingTxsError`, I'd expect accountbased to define it.

PR #5629, opened 2025-06-13
Resolve names using `wallet.otherMethods.resolveName`

samholmes:
This is to support name resolution from the plugin/wallet, specifically for Zano aliases. In the future, all name resolution can be handled by plugins via a standardized `.resolveName` method on the wallet object, but for now it's a drafted API via `otherMethods`.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1210548343287631

### DIFF
--- a/src/util/resolveName.ts
+++ b/src/util/resolveName.ts
@@ -1,0 +1,13 @@
+import { EdgeCurrencyWallet } from 'edge-core-js'
+
+import { ResolutionError } from '../types/ResolutionError'
+
+export const resolveName = async (wallet: EdgeCurrencyWallet, name: string): Promise<string> => {
+  if (typeof wallet.otherMethods.resolveName !== 'function') {
+    throw new ResolutionError('UnsupportedDomain', { domain: name })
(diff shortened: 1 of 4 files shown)


src/util/resolveName.ts
```diff
@@ -0,0 +1,13 @@
+import { EdgeCurrencyWallet } from 'edge-core-js'
+
+import { ResolutionError } from '../types/ResolutionError'
+
+export const resolveName = async (wallet: EdgeCurrencyWallet, name: string): Promise<string> => {
+  if (!('resolveName' in wallet.otherMethods)) {
+    throw new ResolutionError('UnsupportedDomain', { domain: name })
+  }
```
swansontec: `if (typeof wallet.otherMethods.resolveName !== 'function')` might be more picky.

swansontec (approved):
Approved with optional cleanup.

PR #5648, opened 2025-07-02
Modify `NotificationCard` visibility rules

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1210616634217827

### DIFF
--- a/src/components/services/NotificationService.ts
+++ b/src/components/services/NotificationService.ts
@@ -186,5 +186,21 @@ export const NotificationService = (props: Props) => {
     'NotificationServices'
   )
 
+  // Make sure the backup banner is always shown on login if needed. We do this
+  // separately in this effect so that we can hide the banner during current
+  // login session if they so choose.
+  useAsyncEffect(
+    async () => {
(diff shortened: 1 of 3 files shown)


src/components/services/NotificationService.ts
```diff
@@ -186,5 +186,21 @@ export const NotificationService = (props: Props) => {
...
+  // separately in this effect so that we can hide the banner during current
+  // login session if they so choose.
+  useAsyncEffect(
+    async () => {
+      if (!isLightAccountReminder) return
+
+      await writeAccountNotifInfo(account, 'lightAccountReminder', {
+        isBannerHidden: false,
+        isCompleted: false
+      })
```
swansontec, line 199: I'm not super happy about writing to disk if it's meant to be ephemeral, but I understand that your notification data structures are kinda linked to disk, so it would be hard to do it any other way. I guess we can just leave it for now.

swansontec (approved):

PR #5655, opened 2025-07-09
Add PIVX to CURRENCY_SETTINGS_KEYS

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1210754769992348

### DIFF
--- a/src/constants/WalletAndCurrencyConstants.ts
+++ b/src/constants/WalletAndCurrencyConstants.ts
@@ -34,27 +34,28 @@ export const FEE_STRINGS = {
  */
 export const CURRENCY_SETTINGS_KEYS = [
   'bitcoin',
+  'bitcoincash',
+  'bitcoingold',
+  'bitcoinsv',
   'bitcointestnet',
   'bitcointestnet4',
-  'bitcoincash',
-  'ethereum',
-  'ethereumclassic',
   'dash',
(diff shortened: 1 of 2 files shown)


src/constants/WalletAndCurrencyConstants.ts
```diff
@@ -34,27 +34,28 @@ export const FEE_STRINGS = {
  */
 export const CURRENCY_SETTINGS_KEYS = [
   'bitcoin',
+  'bitcoincash',
```
swansontec, line 37: I see you sorted this list. Please bear in mind that the order in this list is the order the settings appear in the app, so this is a UX change. I suppose having them sorted is good for our customers too, which is why I'm fine with it. On the other hand, it's possible that people will complain about their favorite coin being moved.

swansontec (approved):

src/constants/WalletAndCurrencyConstants.ts
samholmes: Alphabetical ordering makes sense to me. It's like when the United States option is at the bottom of the asset settings. It should be manageable. Also, user's should get to the asset settings from the wallet details scene for convenience.

PR #5664, opened 2025-07-18
Disallow unsupported FIO assets

Jon-edge:
This change doesn't cover all situations, but at least prevents new FIO actions from using unsupported assets.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device


src/util/FioValidation.ts
```diff
@@ -0,0 +1,60 @@
...
+  error?: string
+  fioChainCode?: string
+  fioTokenCode?: string
+}
+
+/**
+ * Validate if an Edge asset can be converted to FIO codes using the whitelist
+ * TODO: Integrate into the rest of FIO functionality
+ */
+export const validateFioAsset = (
```
swansontec: This should move to `fioConstants.ts`, and should use the same logic of `const fioAssets = infoServerData.rollup?.fioAssets ?? FIO_ASSET_MAP`. That way it still works if the info server is down. The info server data is supposed to be supplemental to stuff baked in the app.

Indeed, the logic *could* just be added to `tokenIdToFioCode`, which would be edited to return ` | undefined`. Then we *know* any place we might send and Edge currency to Fio-land it's actually something Fio supports.

If that's a big change, then just moving this to the right file would be enough for the merge.

Jon-edge: Yeah it was intentionally separated because moving to the existing fio<>asset utils had a lot of fallout

swansontec (approved):

PR #5665, opened 2025-07-18
Fix jest mocks

swansontec:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1210825371606926


src/constants/WalletAndCurrencyConstants.ts
```diff
@@ -926,6 +920,17 @@ export const SPECIAL_CURRENCY_INFO: {
   }
 }
 
+function isZecBroken(): boolean {
```
samholmes, line 928: Add a JSDoc comment explaining "why" this function exists.

src/components/themed/FlipInput2.tsx
```diff
@@ -426,37 +431,57 @@ const TopAmountText = styled(Text)(theme => () => [
...
+  }
+
+  return (
+    <AnimatedTextInput
+      ref={ref}
+      style={[styles.textInput, dynamicStyle, animatedStyle]}
+      {...rest}
+      {...numericProps}
+    />
+  )
```
samholmes: This

```tsx
  return (
      color: interpolateTextColor(focusAnimation, disableAnimation)	    <AnimatedTextInput
    }))	      ref={ref}
  ]	      style={[styles.textInput, dynamicStyle, animatedStyle]}
      {...rest}
      {...numericProps}
    />
  )
  ```
becomes this

```
return [styles.textInput, dynamicStyle, animatedStyle]
```

if you do this

```tsx
const AmountAnimatedNumericInput = styledWithRef(AnimatedNumericInput)<{	const AnimatedTextInput = Animated.createAnimatedComponent(TextInput)
  disableAnimation: SharedValue<number>	
  focusAnimation: SharedValue<number>	const AmountAnimatedNumericInput = React.forwardRef<
  value: string	  TextInput,
}>(theme => ({ disableAnimation, focusAnimation, value }) => {
```

instead of this

```
const AmountAnimatedNumericInput = React.forwardRef<
  value: string	  TextInput,
}>(theme => ({ disableAnimation, focusAnimation, value }) => {	  TextInputProps & {
  const isAndroid = Platform.OS === 'android'	    disableAnimation: SharedValue<number>
    focusAnimation: SharedValue<number>
    maxDecimals?: number
  }
>((props, ref) => {
```

and move the `useNumericInput` hook to the FlipInput2 component.

This way all that `AmountAnimatedNumericInput` is defines is styles and animation. The logic and usage of it can be moved to the caller (FlipInput2).

src/hooks/useNumericInput.ts
```diff
@@ -0,0 +1,63 @@
...
+  keyboardType: 'decimal-pad'
+  value: string
+  onChangeText: (text: string) => void
+}
+
+/**
+ * Creates a set of props to be passed to a `TextInput` component.
+ * These will handle edits to the text, forcing the input to be a valid number.
+ */
+export function useNumericInput(opts: NumericInputOpts): NumericInputProps {
```
samholmes: Perhaps a better name `useNumericInputProps`.  This behavior logic that can be composed with `TextInputs`; this is great!

src/components/themed/FlipInput2.tsx
```diff
@@ -546,3 +571,12 @@ function useAnimatedColorInterpolateFn(
...
 }
+
+const getStyles = cacheStyles((theme: Theme) => ({
+  textInput: {
+    includeFontPadding: false,
+    fontFamily: theme.fontFaceMedium,
+    fontSize: theme.rem(1.5),
+    padding: 0
+  }
+}))
```
samholmes: This is only used in one place, why not inline it there?

src/components/themed/FlipInput2.tsx
swansontec: Yeah, this is hard to do because the scene has two of them with separate states. I started down the path you suggested, and then realized that it was simpler to just put the logic in the component. Y'know, like components are for? It seemed silly to fight the system, when we already have a perfectly good place to put the logic hooks.

src/components/themed/FlipInput2.tsx
swansontec: Hehe! When I was working on the PR, I thought to myself, "If Sam sees these styles inlined, he's going to leave a review comment asking me to do it properly with caching." So I went the extra mile to satisfy the in-my-head version of Sam. Oops. I'll move it inline, since that's definitely simpler.

samholmes: Ah, I would have until we found out that styles aren't cached haha. If they were cached, I'd be motivated to leverage the caching in the `styled` HOC. ;)

src/components/themed/FlipInput2.tsx
samholmes: Fair enough.

samholmes (approved):
Looks great!

PR #5669, opened 2025-07-23
Support WalletConnect-provided nonces

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

### DIFF
--- a/src/components/modals/WcSmartContractModal.tsx
+++ b/src/components/modals/WcSmartContractModal.tsx
@@ -174,6 +174,12 @@ export const WcSmartContractModal = (props: Props) => {
             await wallet.otherMethods.txRpcParamsToSpendInfo(
               cleanPayload.params[0]
             )
+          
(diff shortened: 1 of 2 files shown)


src/components/modals/WcSmartContractModal.tsx
```diff
@@ -174,6 +174,12 @@ export const WcSmartContractModal = (props: Props) => {
             await wallet.otherMethods.txRpcParamsToSpendInfo(
               cleanPayload.params[0]
             )
+          
+          // If WalletConnect provided a nonce, allow chained pending transactions
+          if (cleanPayload.params[0].nonce != null) {
```
swansontec, line 179: If we do it as discussed, then simply having a nonce will be enough to trigger the behavior. So we can just drop this whole PR.

PR #5679, opened 2025-08-01
Sam/abstracted asset

samholmes:
This change only impacts Libraland which has "Libraland Dollar" as it's asset name and "Libraland" as it's chain name. This means it'll get a network label, and I'm okay with that slight regression for the deterministic benefit we get from this change.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

### DIFF
--- a/src/util/isAbstractedAssetChain.ts
+++ b/src/util/isAbstractedAssetChain.ts
@@ -13,9 +13,24 @@ export const isAssetNativeToChain = (
   currencyInfo: EdgeCurrencyInfo,
   tokenId?: EdgeTokenId
 ): boolean => {
-  return (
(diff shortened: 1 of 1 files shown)


Jon-edge:
I don't like this, the inconsistency observed means that this isn't a good abstraction.

samholmes:
> I don't like this, the inconsistency observed means that this isn't a good abstraction.

There is literally one inconsistency which is with Liberland. This abstraction doesn't need to be committed to the semantics of the core if we don't feel like committing to it. Though bare in mind, this abstraction can exist within the GUI and the definition can remain there.

PR #5685, opened 2025-08-12
Jon/fix/powered by

Jon-edge:
What changed: The state manager now uses functional updates:
- Before: setState({ ...stateManagerRef.current.state, ...partial })
- After: setState(prev => ({ ...prev, ...partial }))

Why this fixed it:
- With batched/async updates (debounced convert, spinner flips, status text), two rapid updates could read the same stale snapshot (ref.current.state) and overwrite each other. Example: A sets poweredBy, B sets spinner; both merge from S0 → last write wins, dropping poweredBy.
- Functional setState guarantees each queued update receives the latest prev state (including previously queued updates), so merges compose instead of clobbering. That prevents poweredBy from disappearing.

What the old ref pattern was trying to do:
- Provide a stable object whose .state stays in sync across renders, so callbacks don’t capture stale closures.
- However, using ref.current.state as the merge base with non-functional setState is unsafe under concurrent/batched updates. The ref is current only per render, not per queued state update.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1211009486252940

### DIFF
--- a/src/plugins/gui/hooks/useStateManager.ts
+++ b/src/plugins/gui/hooks/useStateManager.ts
@@ -14,7 +14,7 @@ export const useStateManager = <T extends object>(
   const [state, setState] = useState<T>(defaultState)
   const handleUpdate: StateManager<T>['update'] = useHandler(
     (state: Partial<T>) => {
-      setState({ ...stateManagerRef.current.state, ...state })
+      setState(prevState => ({ ...prevState, ...state }))
     }
   )
   const stateManagerRef = useRef<StateManager<T>>({


src/components/cards/PoweredByCard.tsx
```diff
@@ -14,7 +14,7 @@ interface Props {
   onPress: () => void
 }
 
-export const PoweredByCard = (props: Props) => {
+export const PoweredByCard = (props: Props): React.ReactElement => {
```
swansontec: This could be perfect for `export const PoweredByCard: React.FC<Props> = props => {`, as mentioned at the meeting today.

PR #5691, opened 2025-08-13
Sam/ramp plugins

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device


Jon-edge:
image

I think there are better ways to start this flow. The UI/UX should always to the best of our ability make it obvious what to do without reading instructions on how to use the app.

There's only one thing to do in this part of the flow: to select the region. This entire flow can actually be gated by region selection. This is one aspect of our legacy flow that I liked.

If the country selection modal is dismissed or the user backs out of the flow at this stage, THEN we can show a brief message, on the scene:

"Select your region for personalized ways to buy crypto:" (personalized "options" sounds vague)
[Select your region]

Jon-edge:
image

The "Next" button is a moving tap target. 

Anchor it to the bottom of the scene/top of the keyboard with KavButton.tsx. The initial implementation called for a full width button for KavButton, though I disagree with that. This is a good opportunity to keep our UI unified and update it to typical button dimensions we use in the rest of our app.

Jon-edge:
image
Is this exchange rate representative of the best rate? Are we sure it's the exact rate? If it's a variable rate, we should be sure to represent that scenario with a "≈" sign

src/app.ts
```diff
@@ -53,7 +53,7 @@ if (ENV.SENTRY_ORGANIZATION_SLUG.includes('SENTRY_ORGANIZATION')) {
 }
 
 // Uncomment the next line to remove popup warning/error boxes.
-// LogBox.ignoreAllLogs()
+LogBox.ignoreAllLogs()
```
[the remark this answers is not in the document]
paullinator: valid concern. why was this uncommented

src/actions/ExchangeRateActions.ts
```diff
@@ -359,3 +359,117 @@ export const closestRateForTimestamp = (
   }
   return bestRate
 }
+
```
paullinator: Just do a single call to getHistoricalRate but using the current date. This removes all this code

docs/ramp-plugin-architecture.md
```diff
@@ -0,0 +1,353 @@
...
+Plugins are initialized once when the app starts and stored in Redux:
+
+```typescript
+interface RampPluginState {
+  readonly isLoading: boolean
+  readonly plugins: Record<string, RampPlugin>
+}
+```
+
+The `RampPluginManager` component handles:
```
paullinator, line 314: Bad bad. Let's stop putting background task "managers" into the gui. these slow down startup time and add lag in the app when we're not even accessing the plugins. The borrow plugins were notorious for this. Only query what you need when the user enters the buy/sell scenes.

docs/ramp-plugin-architecture.md
```diff
@@ -0,0 +1,353 @@
...
+```
+
+## UI Integration
+
+### TradeCreateScene
+
+```typescript
+export const TradeCreateScene = () => {
+  // Get all plugins directly from Redux
+  const rampPlugins = useSelector(state => state.rampPlugins.plugins)
```
paullinator, line 140: No please. Don't add unnecessary state to redux. This slows down the app. These plugins are not global state and should only be initialized when entering buy/sell

src/hooks/useExchangeRateLoader.ts
```diff
@@ -0,0 +1,52 @@
...
+import { useDispatch, useSelector } from '../types/reactRedux'
+
+interface UseExchangeRateLoaderResult {
+  loading: boolean
+  error: Error | null
+  hasRatesForFiat: (fiatCode: string) => boolean
+  loadRatesForFiat: (fiatCode: string) => Promise<void>
+}
+
+export function useExchangeRateLoader(): UseExchangeRateLoaderResult {
```
paullinator: Not used in this PR. Why would this be needed. Each plugin should support only their list of supported currencies.

src/components/scenes/TradeCreateScene.tsx
```diff
@@ -0,0 +1,741 @@
...
+        .denominations[0]
+    }
+  }, [selectedCrypto, selectedWallet])
+
+  //  Get user's current country settings
+  const { countryCode, stateProvinceCode } = useSelector(
+    state => state.ui.settings
+  )
+
+  const countryData = React.useMemo(
```
paullinator: useMemo overhead isn't worth it. just do the find each render. or create a map one time and use that.

const countryData = COUNTRY_CODES.find(c => c['alpha-2'] === countryCode);

or

// Outside the component (global or in a util file)
const COUNTRY_MAP = new Map(COUNTRY_CODES.map(c => [c['alpha-2'], c]));

// In the component
const countryData = COUNTRY_MAP.get(countryCode);

src/components/scenes/TradeOptionSelectScene.tsx
```diff
@@ -0,0 +1,386 @@
...
+
+interface Props extends BuyTabSceneProps<'rampSelectOption'> {}
+
+export const TradeOptionSelectScene = (props: Props): React.JSX.Element => {
+  const { route } = props
+  const { rampQuoteRequest, quotes: precomputedQuotes } = route.params
+
+  const theme = useTheme()
+  const rampPlugins = useSelector(state => state.rampPlugins.plugins)
+  const isPluginsLoading = useSelector(state => state.rampPlugins.isLoading)
```
paullinator: Please use tanstack instead of redux for loading the plugins. only load when needed

src/components/scenes/TradeOptionSelectScene.tsx
```diff
@@ -0,0 +1,386 @@
...
+
+  // Use supported plugins hook only if no precomputed quotes
+  const { supportedPlugins } = useSupportedPlugins({
+    selectedWallet: rampQuoteRequest.wallet,
```
paullinator, line 61: Ack. just realized a short coming in our design. We never show the wallet the user is buying into or selling from. Please add a text line about Exchange Rate

"Buy into wallet: My Bitcoin"
"Sell from wallet: My Bitcoin"

src/locales/en_US.ts
```diff
@@ -1975,6 +1991,24 @@ const strings = {
   education: 'Education',
   enter_value: 'Enter Value',
 
+  // Trade Region Select Scene
+  trade_region_select_buy_crypto: 'Buy Crypto',
```
paullinator: Use `buy_crypto`

src/locales/en_US.ts
```diff
@@ -1931,6 +1933,20 @@ const strings = {
   see_all: 'See All',
   sell_crypto: 'Sell Crypto',
   sell_crypto_footer: 'Crypto to bank or cash',
+
+  // Trade Option Select Scene
+  trade_option_buy_title: 'Buy Cryptocurrency',
```
paullinator, line 1946: Do not prefix string with where they are used. This makes it hard to utilize identical strings in different places. Instead prefix them with what type of UI element they are used for. ie

`choose_provider_button`
`total_label`
`powered_by_button`

For longer text, you can just use the first few words of the text

 `please_try_again_no_providers`

src/locales/en_US.ts
```diff
@@ -1931,6 +1933,20 @@ const strings = {
...
+  trade_option_sell_title: 'Sell Cryptocurrency',
+  trade_option_select_payment_method: 'Select Payment Method',
+  trade_option_choose_provider: 'Choose Provider',
+  trade_option_no_quotes_title: 'No quotes available',
+  trade_option_no_quotes_body:
+    'Please try again later. No providers are currently available.',
+  trade_option_total_label: 'Total',
+  trade_option_settlement_label: 'Settlement',
+  trade_option_powered_by_label: 'Powered By:',
```
paullinator: Use `plugin_powered_by_space` I'm sure things will look fine without the `:`

src/plugins/ramps/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+### RampPluginConfig
+Configuration passed to plugin factories:
+- `initOptions`: Provider-specific initialization options
+- `store`: Optional storage interface for persistent data
+- `makeUuid`: Optional UUID generator
+- `account`: EdgeAccount for wallet operations
+- `navigation`: Navigation object for scene transitions
+- `onLogEvent`: Analytics tracking function
+- `disklet`: Disklet for file operations and permissions
```
paullinator, line 39: Provide disklet or store but not both. Whichever one is more recent and recommended by @swansontec

src/plugins/ramps/paybis/paybisRampPlugin.ts
```diff
@@ -0,0 +1,1360 @@
...
+      let partnerUserId: string
+      if (pluginConfig.store != null) {
+        partnerUserId = await pluginConfig.store
+          .getItem('partnerUserId')
+          .catch(() => '')
+        if (partnerUserId === '' && pluginConfig.makeUuid != null) {
+          partnerUserId = await pluginConfig.makeUuid()
+          await pluginConfig.store.setItem('partnerUserId', partnerUserId)
+        } else if (partnerUserId === '') {
+          partnerUserId = `edge-user-${Date.now()}-${Math.random()
```
paullinator: use makeUuid. no need to prefix with `edge-user`

src/plugins/ramps/paybis/paybisRampPlugin.ts
```diff
@@ -0,0 +1,1360 @@
...
+  const ensureStateInitialized = async (): Promise<void> => {
+    if (state == null) {
+      const { apiKey, partnerUrl, privateKeyB64 } = initOptions
+
+      let partnerUserId: string
+      if (pluginConfig.store != null) {
+        partnerUserId = await pluginConfig.store
+          .getItem('partnerUserId')
+          .catch(() => '')
+        if (partnerUserId === '' && pluginConfig.makeUuid != null) {
```
paullinator: why would `makeUuid` ever be null?

src/plugins/ramps/paybis/paybisRampPlugin.ts
```diff
@@ -0,0 +1,1360 @@
...
+    if (direction === 'buy') {
+      await initializeBuyPairs()
+    } else {
+      await initializeSellPairs()
+    }
+  }
+
+  const validateSupportRequest = (
+    regionCode?: FiatPluginRegionCode,
+    countryCode?: string
```
paullinator: country code is in the regionCode so just pass region

src/plugins/ramps/paybis/paybisRampPlugin.ts
```diff
@@ -0,0 +1,1360 @@
...
+    if (regionCode != null) {
+      try {
+        validateRegion(pluginId, regionCode, SUPPORTED_REGIONS)
+      } catch (error) {
+        return { supported: false }
+      }
+    }
+
+    // Check country-specific restrictions
+    if (countryCode === 'GB') {
```
paullinator: Just put this restriction in `SUPPORTED_REGIONS`. Then you can just call validateRegion directly and no need for `validateSupportRequest`

src/plugins/ramps/paybis/paybisRampPlugin.ts
```diff
@@ -0,0 +1,1360 @@
...
+      return { supported: false }
+    }
+
+    return undefined
+  }
+
+  const checkAssetSupport = (
+    direction: FiatDirection,
+    fiatCurrencyCode: string,
+    cryptoPluginId: string,
```
paullinator: Just access an `EdgeAsset` instead of separate pluginId,tokenId

src/plugins/ramps/paybis/paybisRampPlugin.ts
```diff
@@ -0,0 +1,1360 @@
...
+    tokenId?: string | null
+  ): { supported: false } | undefined => {
+    // Check if crypto is supported
+    const paybisCc =
+      EDGE_TO_PAYBIS_CURRENCY_MAP[`${cryptoPluginId}_${tokenId ?? ''}`]
+    if (!paybisCc) {
```
paullinator: did ai write this? need explicit null check

src/plugins/ramps/paybis/paybisRampPlugin.ts
```diff
@@ -0,0 +1,1360 @@
...
+        console.error('Paybis checkSupport error:', error)
+        throw error
+      }
+    },
+
+    fetchQuote: async (
+      request: RampQuoteRequest
+    ): Promise<RampQuoteResult[]> => {
+      await ensureStateInitialized()
+      if (!state) throw new Error('Plugin state not initialized')
```
paullinator: did AI write this? need explicit null check

src/plugins/ramps/paybis/paybisRampPlugin.ts
```diff
@@ -0,0 +1,1360 @@
...
+      if (regionResult != null) {
+        // Return empty array for unsupported regions
+        return []
+      }
+
+      // Initialize assets for the direction
+      await ensureAssetsInitialized(direction)
+
+      // Get all supported payment types for the direction
+      let allPaymentTypes = Object.keys(allowedPaymentTypes[direction]).filter(
```
paullinator, line 739: If there is a way to know which payments types are valid for a specific country, that would save you having to get quotes for all payment types. This might have to be hard coded in each plugin if they don't have an API for it.

paullinator (changes requested):

docs/ramp-plugin-architecture.md
samholmes: Instead of a RampPluginManager, we will include a `useRampPlugins` hook which will implement `useQuery` to instantiate and return the plugin instances and cache until app restart.

docs/ramp-plugin-architecture.md
samholmes: Instead of storing ramp plugins in redux state, this commit will be replaced with a `useRampPlugin` hook which will instantiate the plugins from within a `useQuery` query function so it's cached.

src/hooks/useExchangeRateLoader.ts
samholmes: This was left-over from an now-unused implementation during development which used the exchange rates from redux/rates-server. It should be removed.

src/components/scenes/TradeCreateScene.tsx
samholmes: Agree. Let's avoid `useMemo` unless we need it for performance purposes. In this case, the derived state is not expensive and so the use of `useMemo` is not needed.

src/components/scenes/TradeOptionSelectScene.tsx
samholmes: Agreed. I've made other comments already, but I'll re-iterate that we want to use a `useRampPlugin` hooks which does exactly what the RampPluginManager currently does to instantiate plugins. It will leverage `useQuery` for state caching and storing of these instantiate plugins.

src/components/scenes/TradeOptionSelectScene.tsx
samholmes: We will add a `"Buying into wallet: %s"` string and use `sprintf` to format the string with the `selectedWallet`'s `displayName`. For now, buy is the only thing that `TradeCreateScene` supports, but we should create a variant string for sell when we do support sell (i.e. "Sell from wallet: %s").

src/locales/en_US.ts
samholmes: Remove this `trade_region_select_buy_crypto ` lstring and use `buy_crypto` lstring instead.

src/locales/en_US.ts
samholmes: Rename this lstring to `buy_cryptocurrency_scene_title`.

src/locales/en_US.ts
samholmes: Remove the `":"` character from the lstring value

src/plugins/ramps/README.md
samholmes: For now we wont make any changes to this part of the PR until we get @swansontec's recommendation.

src/plugins/ramps/paybis/paybisRampPlugin.ts
samholmes: Replace:

```ts
partnerUserId = `edge-user-${Date.now()}-${Math.random()
```

with

```ts
partnerUserId = makeUuid()
```

src/plugins/ramps/paybis/paybisRampPlugin.ts
samholmes: Remove `pluginConfig.makeUuid != null` from the condition.

src/plugins/ramps/paybis/paybisRampPlugin.ts
samholmes: Remove `countryCode` parameter from `validateSupportRequest` signature and instead use `regionCode.countryCode` in its place.

src/plugins/ramps/paybis/paybisRampPlugin.ts
samholmes: The condition `countryCode === 'GB'` is already handled by `validateRegion` so it can be removed. This makes `validateSupportRequest` just a wrapper around `validateRegion` to catch the errors thrown and return `undefined` for valid region support and `{ supported: false }` for invalid region. This is a confusing signature. The wrapper isn't really necessary, we can just inline `validateRegion` in the two places where `validateSupportRequest` is invoked in the paybis plugin and handle the error catching accordingly.

src/plugins/ramps/paybis/paybisRampPlugin.ts
samholmes: So what @paullinator is saying is that we define which payment types are valid on a per-plugin basis. Though, we don't filter payment types by the region parameter within `fetchQuote` routine when we could. Because payment types are semantically region specific, this would make senses to do and optimize for. Until we create a map of payment types and their region or region to payment type, we can leave this suggestion unchanged.

src/plugins/ramps/README.md
swansontec: The store is correct, but don't just grab the core one. The store is meant to be scoped by pluginId, so we need a small wrapper to set pluginId -> storeId before passing to the plugins.

src/components/scenes/TradeCreateScene.tsx
```diff
@@ -0,0 +1,749 @@
...
+      supportedPlugins.map(plugin => [plugin.pluginId, plugin])
+    )
+  })
+
+  // Get the best quote
+  const bestQuote = sortedQuotes[0]
+
+  // Calculate exchange rate from best quote
+  const quoteExchangeRate = React.useMemo(() => {
+    if (!bestQuote?.cryptoAmount || !bestQuote.fiatAmount) return 0
```
Jon-edge, line 246: [96b0de263] Use: if (bestQuote?.cryptoAmount == null || bestQuote.fiatAmount == null)

src/components/scenes/TradeCreateScene.tsx
```diff
@@ -0,0 +1,749 @@
...
+import { FilledTextInput } from '../themed/FilledTextInput'
+
+export interface TradeCreateParams {
+  forcedWalletResult?: WalletListWalletResult
+  regionCode?: string
+}
+
+interface Props extends BuyTabSceneProps<'pluginListBuy'> {}
+
+export const TradeCreateScene = (props: Props): React.ReactElement => {
```
Jon-edge, line 56: [96b0de263] Use export form: export const TradeCreateScene: React.FC<Props> = (props: Props) => {

src/components/scenes/TradeOptionSelectScene.tsx
```diff
@@ -0,0 +1,414 @@
...
+        lstrings.string_total_amount_s,
+        `${selectedQuote.fiatAmount} ${fiatCurrencyCode} → ${selectedQuote.cryptoAmount} ${selectedQuote.displayCurrencyCode}`
+      )}
+      settlementTime={formatSettlementTime(selectedQuote.settlementRange)}
+      partner={{
+        displayName: selectedQuote.pluginDisplayName,
+        icon: { uri: selectedQuote.partnerIcon }
+      }}
+      renderRight={isBestRate ? () => <BestRateBadge /> : undefined}
+      onPress={async () => {
```
Jon-edge, line 330: [96b0de263] Avoid inline async in prop. Define const handlePress = useCallback(() => onPress(selectedQuote), [onPress, selectedQuote]) and pass onPress={handlePress}.

src/components/scenes/TradeOptionSelectScene.tsx
```diff
@@ -0,0 +1,414 @@
...
+import { EdgeText } from '../themed/EdgeText'
+
+export interface RampSelectOptionParams {
+  rampQuoteRequest: RampQuoteRequest
+  quotes?: RampQuoteResult[]
+}
+
+interface Props extends BuyTabSceneProps<'rampSelectOption'> {}
+
+export const TradeOptionSelectScene = (props: Props): React.JSX.Element => {
```
Jon-edge, line 42: [96b0de263] Use export form: export const TradeOptionSelectScene: React.FC<Props> = (props: Props) => {

src/components/cards/PaymentOptionCard.tsx
```diff
@@ -0,0 +1,140 @@
...
+  /** Whether the provider button should be disabled */
+  disableProviderButton?: boolean
+
+  // Events:
+  onPress: () => Promise<void> | void
+  onLongPress?: () => Promise<void> | void
+  onProviderPress: () => Promise<void> | void
+}
+
+export const PaymentOptionCard = (props: Props): React.JSX.Element => {
```
Jon-edge, line 35: [a7ba7b92b] Use export form: export const PaymentOptionCard: React.FC<Props> = (props: Props) => {

src/components/buttons/PillButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { EdgeText } from '../themed/EdgeText'
+
+export interface PillButtonProps {
+  label: string
+  onPress: () => void | Promise<void>
+  icon?: () => React.ReactElement | null
+  disabled?: boolean
+}
+
+export const PillButton = (props: PillButtonProps): React.ReactElement => {
```
Jon-edge, line 16: [a7ba7b92b] Use export form: export const PillButton: React.FC<PillButtonProps> = (props: PillButtonProps) => {

src/hooks/useSupportedPlugins.ts
```diff
@@ -0,0 +1,129 @@
...
+          }
+        })
+      )
+
+      // Filter only supported plugins
+      return supportChecks
+        .filter(check => check.supported)
+        .map(check => check.plugin)
+    },
+    enabled: Boolean(
```
Jon-edge, line 116: [6f940e06f] For enabled, use explicit != null checks instead of Boolean(...) to avoid treating empty strings as false if they are valid.

src/hooks/useRampQuotes.ts
```diff
@@ -0,0 +1,271 @@
...
+        async (pluginId): Promise<Result<RampQuoteResult[], QuoteError>> => {
+          const plugin = plugins[pluginId]
+          if (!plugin) return { ok: true, value: [] }
+
+          try {
+            const quotes = await plugin.fetchQuote(rampQuoteRequest)
+            // Return quotes as-is (empty array means plugin doesn't support the request)
+            return { ok: true, value: quotes }
+          } catch (error) {
+            console.warn(`Failed to get quote from ${pluginId}:`, error)
```
Jon-edge, line 165: [38449493d] Prefer the app’s logging helper over console.warn for consistency.

src/actions/DeepLinkingActions.tsx
```diff
@@ -156,7 +157,13 @@ async function handleLink(
     }
 
     case 'fiatProvider': {
-      fiatProviderDeeplinkHandler(link)
+      // Try ramp deeplink handler first
+      try {
+        rampDeeplinkManager.handleDeeplink(link)
```
Jon-edge, line 162: `const handled = rampDeeplinkManager.handleDeeplink(link)
if (!handled) fiatProviderDeeplinkHandler(link)`

src/plugins/ramps/rampDeeplinkHandler.ts
```diff
@@ -0,0 +1,59 @@
...
+    deeplinkHandler: LinkHandler
+  ): void {
+    this.listener = { direction, providerId, deeplinkHandler }
+  }
+
+  unregister(): void {
+    this.listener = null
+  }
+
+  handleDeeplink(link: FiatProviderLink): void {
```
Jon-edge, line 29:
```
handleDeeplink(link: FiatProviderLink): boolean {
  if (this.listener == null) return false
  const { direction, providerId, deeplinkHandler } = this.listener
  if (link.providerId !== providerId) return false
  if (link.direction !== direction) return false
  if (Platform.OS === 'ios') SafariView.dismiss()
  this.unregister()
  deeplinkHandler(link)
  return true
}
```

src/plugins/ramps/paybis/paybisRampPlugin.ts
```diff
@@ -0,0 +1,1333 @@
...
+                  passwordless: true,
+                  trustedKyc: false,
+                  quoteId,
+                  flow: 'sellCrypto',
+                  depositCallbackUrl: RETURN_URL_PAYMENT,
+                  paymentMethod: paymentMethodId
+                }
+              }
+
+              const privateKey = atob(state!.privateKeyB64)
```
Jon-edge, line 1009: `const privateKey = Buffer.from(state!.privateKeyB64, 'base64').toString('binary')`

src/plugins/ramps/paybis/paybisRampPlugin.ts
```diff
@@ -0,0 +1,1333 @@
...
+        const response = await paybisFetch({
+          method: 'GET',
+          url: state.partnerUrl,
+          path: `v2/public/user/${state.partnerUserId}/status`,
+          apiKey: state.apiKey
+        })
+        const { hasTransactions } = asUserStatus(response)
+        userIdHasTransactions = hasTransactions
+      } catch (e) {
+        console.log(`Paybis: Error getting user status: ${e}`)
```
Jon-edge, line 781: Prefer app logging helper over console.* for consistency.

src/components/scenes/TradeCreateScene.tsx
```diff
@@ -0,0 +1,749 @@
...
+                keyboardType="decimal-pad"
+                numeric
+                showSpinner={isFetchingQuotes && lastUsedInput === 'crypto'}
+              />
+            </InputContainer>
+          </InputRow>
+
+          {/* Bottom Input (Crypto by design) */}
+          <InputRow>
+            <DropDownInputButton onPress={handleCryptDropdown}>
```
Jon-edge, line 519: Avoid empty uri image source.
`{selectedCrypto == null || selectedWallet == null ? null : (
  <CryptoIcon sizeRem={1.5} pluginId={selectedWallet.currencyInfo.pluginId} tokenId={selectedCrypto.tokenId} />
)}`

src/components/scenes/TradeOptionSelectScene.tsx
```diff
@@ -0,0 +1,414 @@
...
+      const text = `(${amount} ${currencyCode})`
+
+      return {
+        name: quote.pluginDisplayName,
+        icon: quote.partnerIcon, // Already full path
+        text
+      }
+    })
+
+    const selectedName = await Airship.show<string | undefined>(bridge => (
```
Jon-edge, line 297: Use a stable key for radio list selection.
`selected={selectedQuote.pluginId}`
`const selectedIndex = quotes.findIndex(q => q.pluginId === selectedName)`

samholmes:
> image
> The "Next" button is a moving tap target.
> 
> Anchor it to the bottom of the scene/top of the keyboard with `KavButton.tsx`. The initial implementation called for a full width button for `KavButton`, though I disagree with that design. This is a good opportunity to keep our UI unified and update it to typical button dimensions we use in the rest of our app.

We need to use KavButton to keep the button stable on the TradeCreateScene.tsx

docs/ramp-plugin-migration-guide.md
```diff
@@ -0,0 +1,1074 @@
...
+  paymentTypes: ['credit', 'bank'],
+  regionCode: { countryCode: 'US', stateCode: 'CA' }
+})
+```
+
+The API remains the same, maintaining compatibility with existing code. Note that the method returns the asset map for the first supported payment type from the provided array.
+
+## Removing getSupportedAssets
+
+The ramp plugin architecture has been simplified by removing the separate `getSupportedAssets` method. All support checking is now done within the `fetchQuote` method.
```
Jon-edge, line 381: This seems like conflicting information around `getSupportedAssets`

src/hooks/useRampQuotes.ts
```diff
@@ -0,0 +1,271 @@
...
+    const allQuotes = quoteResults
+      .filter(
+        (result): result is { ok: true; value: RampQuoteResult[] } => result.ok
+      )
+      .flatMap(result => result.value)
+
+    // Sort by best rate (lowest fiat amount for same crypto amount)
+    return allQuotes.sort((a, b) => {
+      const rateA = parseFloat(a.fiatAmount) / parseFloat(a.cryptoAmount)
+      const rateB = parseFloat(b.fiatAmount) / parseFloat(b.cryptoAmount)
```
Jon-edge, line 248: No guards against 0.

Jon-edge:
"Fix all non deprecation lint warnings in all changed files"

src/components/scenes/TradeCreateScene.tsx
```diff
@@ -0,0 +1,747 @@
...
+  // Get the best quote
+  const bestQuote = sortedQuotes[0]
+
+  // Calculate exchange rate from best quote
+  const quoteExchangeRate = React.useMemo(() => {
+    if (!bestQuote?.cryptoAmount || !bestQuote.fiatAmount) return 0
+
+    try {
+      return (
+        parseFloat(bestQuote.fiatAmount) / parseFloat(bestQuote.cryptoAmount)
```
Jon-edge, line 250: No guard for div by 0

src/hooks/useSupportedPlugins.ts
```diff
@@ -0,0 +1,129 @@
...
+      selectedWallet && selectedCrypto && selectedFiatCurrencyCode && regionCode
+    ),
+    staleTime: 5 * 60 * 1000, // 5 minutes
+    gcTime: 10 * 60 * 1000, // 10 minutes cache time
+    refetchOnWindowFocus: false
+  })
+
+  return {
+    supportedPlugins,
+    isLoading,
```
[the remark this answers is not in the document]
samholmes: It's used in the TradeCreateScene

src/hooks/useRampQuotes.ts
Jon-edge: Not addressed

Jon-edge:
Divide by 0 comments still not addressed

Jon-edge (changes requested):
Not all divide by 0 comments are addressed

Jon-edge (approved):
Approved with one final comment about docs

src/plugins/ramps/paybis/paybisRampPlugin.ts
```diff
@@ -0,0 +1,1316 @@
...
+                        lstrings.fiat_plugin_buy_failed_try_again,
+                        NOT_SUCCESS_TOAST_HIDE_MS
+                      )
+                    } else {
+                      showError(
+                        new Error(
+                          `Paybis: Invalid transactionStatus "${transactionStatus}".`
+                        )
+                      )
+                    }
```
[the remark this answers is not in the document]
samholmes: This was a bug and is intentional.

src/plugins/gui/fiatPlugin.tsx
```diff
@@ -193,7 +194,9 @@ export const executePlugin = async (params: {
     openExternalWebView: async (params): Promise<void> => {
       const { deeplinkHandler, providerId, redirectExternal, url } = params
       datelog(
-        `**** openExternalWebView ${url} deeplinkHandler:${deeplinkHandler}`
+        `**** openExternalWebView ${url} deeplinkHandler:${JSON.stringify(
+          deeplinkHandler
```
samholmes, line 201: Wtf... Why are we logging a function? The proper fix is to remove this deeplinkHandler from the log

PR #5704, opened 2025-08-18
Add explicit file list to temporary eslint errors converted to warnings

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1211083463309498

### DIFF
--- a/src/util/runOnce.ts
+++ b/src/util/runOnce.ts
@@ -15,7 +15,7 @@ export const runOnce = async <T>(
 
   const promise = Promise.resolve()
     .then(async () => await fn())
-    .catch(error => {
+    .catch((error: unknown) => {
       showError(error, { tag: key })
       runRegistry.delete(key)
       return undefined


scripts/updateEslintWarnings.ts
```diff
@@ -0,0 +1,111 @@
...
+        )}\\2,?\\s*$`,
+        'gm'
+      ),
+      new RegExp(
+        `^(\\s*)(['"])${stagedFile.replace(
+          /[.*+?^${}()|[\]\\]/g,
+          '\\$&'
+        )}\\2,?\\s*\n`,
+        'gm'
+      )
```
swansontec, line 62: Editing JS. Wow. Yeah. OK. We could probably do it more "dumb" than this, but I see that it probably works? So yay? We can leave it as-is, since it's temporary anyhow.

I know there was talk of putting the list into JSON, and then importing that into the ESLint config, but I guess that's not really needed now.

PR #5705, opened 2025-08-18
Use rates server v3 endpoint

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---
  - https://app.asana.com/0/0/1208392845356120

### DIFF
--- a/src/__tests__/exchangeRates.test.ts
+++ b/src/__tests__/exchangeRates.test.ts
@@ -1,10 +1,6 @@
 import { beforeAll, expect, it, jest } from '@jest/globals'
 import fetch from 'node-fetch'
 
-import {
-  closestRateForTimestamp,
-  type ExchangeRateCache
-} from '../actions/ExchangeRateActions'
 import { getHistoricalCryptoRate } from '../util/exchangeRates'
 import { mswServer } from '../util/mswServer'
 import { snooze } from '../util/utils'
@@ -176,40 +172,6 @@ it('get bulk rates', async () => {
   )
 
   await Promise.all(promises)
-})
-
-it('get closest rate for timestamp', () => {
-  const targetDate = '2025-03-12T23:00:00.000Z'
-  const rateCache: ExchangeRateCache = {
-    'ETH_iso:USD_2025-03-13T23:10:00.000Z': {
-      rate: 1879.75664619359304197133,
-      expiration: 12345
-    },
-    'ETH_iso:USD_2025-03-11T22:55:00.000Z': {
-      rate: 1944.99840563707630280987,
-      expiration: 12345
-    },
-    'ETH_iso:USD_2025-03-09T20:00:00.000Z': {
-      rate: 2041.15092554349985221052,
-      expiration: 12345
-    },
-    'ETH_iso:USD_2025-03-04T19:00:00.000Z': {
-      rate: 2160.9592242148933110002,
-      expiration: 12345
-    }
-  }
-  const rateETH = closestRateForTimestamp(
-    rateCache,
-    'ETH',
-    Date.parse(targetDate)
-  )
-  expect(rateETH).toEqual(1944.99840563707630280987)
-  const rateBTC = closestRateForTimestamp(
-    rateCache,
-    'BTC',
-    Date.parse(targetDate)
-  )
-  expect(rateBTC).toEqual(undefined)
 })
 
 // main().catch(e => console.log('error'))
--- a/src/__tests__/exchangeRates.test.ts
+++ b/src/__tests__/exchangeRates.test.ts
@@ -5,7 +5,7 @@ import {
   closestRateForTimestamp,
   type ExchangeRateCache
 } from '../actions/ExchangeRateActions'
-import { getHistoricalRate } from '../util/exchangeRates'
+import { getHistoricalCryptoRate } from '../util/exchangeRates'
 import { mswServer } from '../util/mswServer'
 import { snooze } from '../util/utils'
 
@@ -29,8 +29,10 @@ it('get bulk rates', async () => {
   // async function main(): Promise<void> {
   const promises: Array<Promise<unknown>> = []
   promises.push(
-    getHistoricalRate(
-      'BTC_iso:USD',
+    getHistoricalCryptoRate(
+      'bitcoin',
+      null,
+      'iso:USD',
       '2022-06-01T04:00:00.000Z',
       TEST_MAX_QUERY_SIZE,
       fetch
@@ -40,8 +42,10 @@ it('get bulk rates', async () => {
   )
 
   promises.push(
-    getHistoricalRate(
-      'ETH_iso:USD',
+    getHistoricalCryptoRate(
+      'ethereum',
+      null,
+      'iso:USD',
       '2022-06-02T04:00:00.000Z',
       TEST_MAX_QUERY_SIZE,
       fetch
@@ -51,8 +55,10 @@ it('get bulk rates', async () => {
   )
 
   promises.push(
-    getHistoricalRate(
-      'XMR_iso:EUR',
+    getHistoricalCryptoRate(
+      'monero',
+      null,
+      'iso:EUR',
       '2022-06-03T04:00:00.000Z',
       TEST_MAX_QUERY_SIZE,
       fetch
@@ -62,8 +68,10 @@ it('get bulk rates', async () => {
   )
 
   promises.push(
-    getHistoricalRate(
-      'BTC_iso:PHP',
+    getHistoricalCryptoRate(
+      'bitcoin',
+      null,
+      'iso:PHP',
       '2022-06-04T04:00:00.000Z',
       TEST_MAX_QUERY_SIZE,
       fetch
@@ -73,8 +81,10 @@ it('get bulk rates', async () => {
   )
 
   promises.push(
-    getHistoricalRate(
-      'DOGE_iso:MXN',
+    getHistoricalCryptoRate(
+      'dogecoin',
+      null,
+      'iso:MXN',
       '2022-06-04T04:00:00.000Z',
       TEST_MAX_QUERY_SIZE,
       fetch
@@ -86,8 +96,10 @@ it('get bulk rates', async () => {
   await snooze(2000)
 
   promises.push(
-    getHistoricalRate(
-      'BTC_iso:PHP',
+    getHistoricalCryptoRate(
+      'bitcoin',
+      null,
+      'iso:PHP',
       '2022-06-04T04:00:00.000Z',
       TEST_MAX_QUERY_SIZE,
       fetch
@@ -97,8 +109,10 @@ it('get bulk rates', async () => {
   )
 
   promises.push(
-    getHistoricalRate(
-      'DOGE_iso:MXN',
+    getHistoricalCryptoRate(
+      'dogecoin',
+      null,
+      'iso:MXN',
       '2022-06-04T04:00:00.000Z',
       TEST_MAX_QUERY_SIZE,
       fetch
@@ -110,8 +124,10 @@ it('get bulk rates', async () => {
   await snooze(1000)
 
   promises.push(
-    getHistoricalRate(
-      'ETH_iso:USD',
+    getHistoricalCryptoRate(
+      'ethereum',
+      null,
+      'iso:USD',
       '2022-06-02T04:00:00.000Z',
       TEST_MAX_QUERY_SIZE,
       fetch
@@ -121,8 +137,10 @@ it('get bulk rates', async () => {
   )
 
   promises.push(
-    getHistoricalRate(
-      'XMR_iso:EUR',
+    getHistoricalCryptoRate(
+      'monero',
+      null,
+      'iso:EUR',
       '2022-06-03T04:00:00.000Z',
       TEST_MAX_QUERY_SIZE,
       fetch
@@ -132,8 +150,10 @@ it('get bulk rates', async () => {
   )
 
   promises.push(
-    getHistoricalRate(
-      'BTC_iso:PHP',
+    getHistoricalCryptoRate(
+      'bitcoin',
+      null,
+      'iso:PHP',
       '2022-06-04T04:00:00.000Z',
       TEST_MAX_QUERY_SIZE,
       fetch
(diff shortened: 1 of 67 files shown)


src/__tests__/exchangeRates.test.ts
```diff
@@ -29,176 +25,164 @@ it('get bulk rates', async () => {
...
+      'bitcoin',
+      null,
+      'iso:USD',
       '2022-06-01T04:00:00.000Z',
       TEST_MAX_QUERY_SIZE,
       fetch
     ).then(v => {
-      console.log(`31616 === ${Math.floor(v)}`)
-      expect(31616).toEqual(Math.floor(v))
+      console.log(`31865 === ${Math.floor(v)}`)
```
swansontec: Please delete all these `console.log` lines. Right now we suppress all log output during tests, but once we fix that, we *really* don't want to spam the console - Jest gives each log output a fairly verbose special treatment, which is great for debugging but horrible for noise.

src/actions/ExchangeRateActions.ts
```diff
@@ -1,61 +1,78 @@
...
 const ONE_MONTH = 1000 * 60 * 60 * 24 * 30
 
-const asAssetPair = asObject({
-  currency_pair: asString,
-  date: asOptional(asString), // Defaults to today if not specified
+const asCryptoFiatPair = asObject({
+  asset: asCryptoAsset,
+  targetFiat: asString,
+  isoDate: asOptional(asString), // Defaults to today if not specified
+  expiration: asNumber
```
swansontec, line 25: Optional: Since we are updating this file format, and don't need to be compatible anymore, we could potentially convert this to a date string instead of a number. However, I do understand that the number is going to be smaller on-disk. OTOH, the number is inconsistent with the `isoDate` string on the line above.

src/actions/ExchangeRateActions.ts
```diff
@@ -66,20 +83,22 @@ export function updateExchangeRates(): ThunkAction<Promise<void>> {
...
     // before moving on to the potentially slow network:
-    if (state.exchangeRatesMap.size === 0 || exchangeRateCache == null) {
-      exchangeRateCache = await loadExchangeRateCache().catch(error => {
-        datelog('Error loading exchange rate cache:', String(error))
-        return { assetPairs: [], rates: {} }
-      })
-      const { exchangeRates, exchangeRatesMap } = buildGuiRates(
-        exchangeRateCache.rates,
-        yesterday
+    if (exchangeRateCache == null) {
```
swansontec: The `state.exchangeRatesMap.size === 0` was intended to handle the case where you log in, log out, and log in again. You might have a cache in that situation, but redux is empty because of the logout.

src/actions/ExchangeRateActions.ts
```diff
@@ -150,51 +180,107 @@ async function fetchExchangeRates(
...
+    if (pair.isoDate != null) {
+      dateStr = `_${pair.isoDate}`
+    }
+
+    let tokenIdStr = ''
+    if (pair.asset.tokenId != null) {
+      tokenIdStr = `_${tokenIdStr}`
+    }
+    const key = `${pair.asset.pluginId}${tokenIdStr}_${pair.targetFiat}${dateStr}`
+    cryptoPairMap.set(key, pair)
```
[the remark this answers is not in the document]
swansontec: Oh, yeah, the key should use `pair.asset.tokenId`, not `tokenIdStr`.

src/actions/ExchangeRateActions.ts
```diff
@@ -203,51 +289,112 @@ async function fetchExchangeRates(
       const token = wallet.currencyConfig.allTokens[tokenId]
       if (token == null) continue
       if (token.currencyCode === currencyCode) continue
-      addAssetPair({
-        currency_pair: `${token.currencyCode}_${accountIsoFiat}`,
-        date: undefined,
+      addCryptoPair({
```
swansontec, line 294: On the line above, we have `if (token.currencyCode === currencyCode) continue`. I think we should drop this line, now that we are tokenId based. You can have a token called `ETH` on some L2's. Something like this would break things before, so we filtered it out, but now it is no problem.

We still need to keep the `if (token == null) continue` line, of course.

src/actions/TransactionExportActions.tsx
```diff
@@ -56,7 +56,12 @@ export function updateTxsFiat(
       if (amountFiat === 0) {
         const date = new Date(tx.date * 1000).toISOString()
         promises.push(
-          getHistoricalRate(`${currencyCode}_${defaultIsoFiat}`, date)
+          getHistoricalCryptoRate(
+            wallet.currencyInfo.pluginId,
+            tokenId,
+            wallet.fiatCurrencyCode,
```
swansontec: Should be `defaultIsoFiat` instead of `wallet.fiatCurrencyCode`, which is no longer used.

src/components/scenes/Fio/FioSentRequestDetailsScene.tsx
```diff
@@ -39,10 +38,18 @@ interface StateProps {
 type Props = StateProps & OwnProps & ThemeProps
 
 class FioSentRequestDetailsComponent extends React.PureComponent<Props> {
-  fiatAmount = (currencyCode: string, amount: string = '0') => {
+  fiatAmount = (
+    pluginId: string,
+    tokenId: EdgeTokenId,
+    amount: string = '0'
+  ) => {
```
swansontec: You've edited the parameter types, but haven't updated the call site. It "just happens" to line up, because they are all strings. Please fix the `amountField` method a few lines down to pass the right things.

Because we have a FIO tokenCode, not an Edge currencyCode, you might have to do `fioCodeToEdgeAsset` to extract the tokenId.

src/components/scenes/Loans/LoanCreateConfirmationScene.tsx
```diff
@@ -274,15 +273,13 @@ export const LoanCreateConfirmationScene = (props: Props) => {
...
     isoFiatCurrencyCode: srcIsoFiatCurrencyCode
   } = useTokenDisplayData({
     tokenId: srcTokenId,
     currencyConfig: srcWallet.currencyConfig
   })
   const {
-    currencyCode: feeCurrencyCode,
     denomination: feeDenom,
     isoFiatCurrencyCode: feeIsoFiatCurrencyCode
   } = useTokenDisplayData({
```
swansontec, line 285: So on the line below, we seem to be passing `tokenId: borrowEngineCurrencyCode,` which looks wrong maybe? Shouldn't it be `tokenId: null` for the fee?

src/components/scenes/Loans/LoanCreateScene.tsx
```diff
@@ -313,33 +317,47 @@ export const LoanCreateScene = (props: Props) => {
...
-          ),
-          0
-        )
+  const minSwapInputNativeAmount = useSelector(state => {
+    if (srcWallet == null) return '0'
+    const rate = convertCurrency(
+      state,
+      srcWallet.currencyInfo.pluginId,
+      srcTokenId,
+      defaultIsoFiat
```
swansontec: This was `iso:USD` before, rather than `defaultIsoFiat`. I think this is because we also hard-code `30` down below, as in, "the minimum loan amount is $30". It shouldn't be 30 yen or whatever - that's a dramatically different amount.

src/components/scenes/Loans/LoanCreateScene.tsx
```diff
@@ -656,15 +674,22 @@ const getStyles = cacheStyles((theme: Theme) => ({
...
+  pluginId: string
+  tokenId: EdgeTokenId
   isoFiatCurrencyCode?: string
 }): number => {
   return useSelector(state => {
-    if (currencyCode == null || isoFiatCurrencyCode == null) return 0
+    if (isoFiatCurrencyCode == null) return 0
     else
-      return state.exchangeRates[`${currencyCode}_${isoFiatCurrencyCode}`] ?? 0
+      return getExchangeRate(
```
swansontec: Delete the `else`, since it does nothing. We already `return` on the line above.

src/components/themed/ExchangedFlipInput2.tsx
```diff
@@ -165,8 +177,9 @@ const ExchangedFlipInput2Component = React.forwardRef<
       return { nativeAmount: '', exchangeAmount: '', displayAmount: '' }
     const exchangeAmountLong = convertCurrency(
       fiatAmount,
-      defaultIsoFiat,
-      cryptoCurrencyCode
+      pluginId,
+      tokenId,
+      defaultIsoFiat
```
swansontec, line 184: This conversion is backwards, from fiat -> crypto, but you don't have the parameters flipped around like they need to be. Your parameters are for a crypto -> fiat conversion, which is the opposite of what we want.

src/components/themed/SwapInput.tsx
```diff
@@ -152,15 +158,15 @@ const SwapInputComponent = React.forwardRef<SwapInputCardInputRef, Props>(
...
         wallet.currencyConfig,
         tokenId
       )
       const exchangeAmountLong = convertCurrency(
         fiatAmount,
-        defaultIsoFiat,
-        cryptoCurrencyCode
+        wallet.currencyInfo.pluginId,
+        tokenId,
+        defaultIsoFiat
```
swansontec, line 173: This is another backwards fiat -> crypto conversion, but you have the parameters for a crypto -> fiat conversion, so it will be wrong.

src/components/tiles/InterestRateChangeTile.tsx
```diff
@@ -44,21 +46,42 @@ const InterestRateChangeTileComponent = (props: Props) => {
...
-    return pairs
-  }, [])
+  const necessaryExchangeRates = [...debts, newDebt].reduce(
+    (pairs: EdgeTokenId[], obj) => {
+      const { tokenId } = obj
+      pairs.push(tokenId)
+      return pairs
+    },
+    []
+  )
```
swansontec: This is a miss-use of `reduce`. It should be a `map`:

```ts
const necessaryExchangeRates = [...debts, newDebt].map(obj => obj.tokenId)
```

src/util/exchangeRates.ts
```diff
@@ -1,121 +1,191 @@
...
+          }
         }
+      } else {
+        const text = await response.text()
+        throw new Error(text)
       }
-    } else {
-      const text = await response.text()
-      throw new Error(text)
+    } catch (e: any) {
```
swansontec: Shouldn't this be `e: unknown`, per the lint rule?

src/util/exchangeRates.ts
```diff
@@ -1,121 +1,191 @@
...
-        }
-        const { resolvers } = resolverMap[pairDate]
-        clog(`${n} deleting ${pairDate}`)
-        delete resolverMap[pairDate]
-        if (resolvers.length) {
-          resolvers.forEach((r, i) => {
-            r(rate)
-          })
+
+  // clog(`${n} fetching ${JSON.stringify(data, null, 2)}`)
```
swansontec: Delete this leftover line.

src/selectors/WalletSelectors.ts
```diff
@@ -41,42 +42,68 @@ export const getActiveWalletCurrencyInfos = (
...
 export const convertCurrency = (
   state: RootState,
-  fromCurrencyCode: string,
+  pluginId: string,
+  tokenId: EdgeTokenId,
   toCurrencyCode: string,
   amount: string = '1'
 ): string => {
-  const exchangeRate = getExchangeRate(state, fromCurrencyCode, toCurrencyCode)
+  const exchangeRate = getExchangeRate(
```
swansontec, line 84: This is literally the same as `convertCurrencyFromExchangeRates`, but with an extra `state.`. Do we really need both functions?

src/selectors/WalletSelectors.ts
```diff
@@ -95,7 +122,8 @@ export const calculateFiatBalance = (
   )
   const fiatValue = convertCurrencyFromExchangeRates(
     exchangeRates,
-    currencyCode,
+    wallet.currencyInfo.pluginId,
+    null,
```
swansontec: Shouldn't this be `tokenId`, not `null`?

src/actions/ExchangeRateActions.ts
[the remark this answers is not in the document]
peachbits: 🤦

src/components/themed/ExchangedFlipInput2.tsx
```diff
@@ -8,11 +8,13 @@ import { useDisplayDenom } from '../../hooks/useDisplayDenom'
 import { useHandler } from '../../hooks/useHandler'
 import { lstrings } from '../../locales/strings'
 import { getExchangeDenom } from '../../selectors/DenominationSelectors'
-import { getExchangeRate } from '../../selectors/WalletSelectors'
+import {
+  convertCurrency as convertCurrencyFromExchangeRates,
```
swansontec: Why rename the import? Was this a mistake?

src/components/themed/SwapInput.tsx
```diff
@@ -8,10 +8,12 @@ import {
   getExchangeDenom,
   selectDisplayDenom
 } from '../../selectors/DenominationSelectors'
-import { getExchangeRate } from '../../selectors/WalletSelectors'
+import {
+  convertCurrency as convertCurrencyFromExchangeRates,
```
swansontec: Why rename the import?

swansontec (approved):
Approved, but please check the AI comment (I think it's confused) and maybe address the two weird imports I mentioned.

src/components/scenes/Loans/LoanCreateConfirmationScene.tsx
```diff
@@ -105,17 +105,16 @@ export const LoanCreateConfirmationScene = (props: Props) => {
...
-          mul(
-            '5.05',
-            convertCurrency(
-              state,
-              'iso:USD',
-              'MATIC',
-              destWallet.currencyInfo.denominations[0].multiplier
-            )
-          ),
+          mul('5.05', div('1', polygonPrice === '0' ? '1' : polygonPrice)),
```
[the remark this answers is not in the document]
peachbits: The original code looked up an inverse fiat/crypto rate. since the new code is only crypto/fiat I had to invert the amount so the remaining code stayed the same

src/components/themed/ExchangedFlipInput2.tsx
peachbits: "convertCurrency" was declared elsewhere in the file. I'll change that name instead

src/components/themed/SwapInput.tsx
peachbits: "convertCurrency" was declared elsewhere in the file. I'll change that name instead

PR #5706, opened 2025-08-19
Jon/fix/same addr swap

Jon-edge:
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1211049275928353


src/components/scenes/SwapProcessingScene.tsx
```diff
@@ -33,7 +37,7 @@ export interface SwapProcessingParams {
 
 type Props = SwapTabSceneProps<'swapProcessing'>
 
-export function SwapProcessingScene(props: Props) {
+export const SwapProcessingScene: React.FC<Props> = (props: Props) => {
```
swansontec, line 40: You don't need `(props: Props) =>` anymore. Just do `props =>` now, since we already give the type on the left.

src/locales/en_US.ts
```diff
@@ -1069,6 +1069,10 @@ const strings = {
   open_settings: 'Open Settings',
   ss_geolock: 'Location restricted. Unable to complete exchange.',
   ss_unable: 'No enabled exchanges support %1$s to %2$s.',
+  ss_same_address_upgrade_created_3s:
+    '%1$s->%2$s upgrades require the same address for both wallets. %3$s has been created. Please try your upgrade again.',
+  ss_same_address_upgrade_selected_3s:
+    '%1$s->%2$s upgrades require the same address for both wallets. %3$s has been selected. Please try your upgrade again.',
```
swansontec: Would the `→` symbol be prettier?

src/components/scenes/SwapProcessingScene.tsx
```diff
@@ -61,6 +65,95 @@ export function SwapProcessingScene(props: Props) {
...
+        await Airship.show<string | undefined>(bridge => (
+          <ButtonsModal
+            bridge={bridge}
+            title={lstrings.exchange_generic_error_title}
+            message={sprintf(template, fromCurrencyCode, toCurrencyCode, name)}
+            buttons={{ ok: { label: lstrings.string_ok_cap } }}
+          />
+        ))
+        return
+      } catch (e) {
```
swansontec, line 152: Doesn't ESLint require this to be `(e: unknown)`?

swansontec (approved):
Some small things

src/components/modals/ButtonsModal.tsx
```diff
@@ -54,9 +54,9 @@ export interface ButtonModalProps<Buttons> {
  * Build a custom modal component if you need form fields, check boxes,
  * or other interactive elements.
  */
-export function ButtonsModal<Buttons extends Record<string, ButtonInfo>>(
-  props: ButtonModalProps<Buttons>
-) {
+export const ButtonsModal: React.FC<
+  ButtonModalProps<Record<string, ButtonInfo>>
+> = (props: ButtonModalProps<Record<string, ButtonInfo>>) => {
```
swansontec: You don't need `(props: ...) => {` because the types are already on the left. Just do `props =>`.

src/components/scenes/SwapProcessingScene.tsx
swansontec: Please delete the redundant `: Props`.

src/components/scenes/SwapProcessingScene.tsx
swansontec: Please fix still?

swansontec (changes requested):
Looks like not all the comments from the first review got addressed.

src/components/scenes/SwapProcessingScene.tsx
Jon-edge: I didn't get any linter warnings here...

Jon-edge: `@typescript-eslint/use-unknown-in-catch-callback-variable`, applies to `Promise .catch(...)` callback parameters, not `try { ... } catch (e) { ... }`. It will not flag a try/catch clause.

The TypeScript option `useUnknownInCatchVariables: true` (default) only changes the inferred type of the catch variable to unknown. It does not require an explicit `: unknown `annotation and does not warn at the catch site. Errors will appear only where you use e unsafely (for example, e.message without narrowing), not at catch (e).

swansontec (approved):

PR #5707, opened 2025-08-19
Jon/fix/swap wallet selection

Jon-edge:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1210989718507536

### DIFF
--- a/src/components/scenes/SwapCreateScene.tsx
+++ b/src/components/scenes/SwapCreateScene.tsx
@@ -73,7 +73,7 @@ export interface SwapErrorDisplayInfo {
 
 interface Props extends SwapTabSceneProps<'swapCreate'> {}
 
-export const SwapCreateScene = (props: Props) => {
+export const SwapCreateScene: React.FC<Props> = props => {
   const { navigation, route } = props
   const {
     fromWalletId,
@@ -224,7 +224,7 @@ export const SwapCreateScene = (props: Props) => {
     )
   }
 
-  const getQuote = (swapRequest: EdgeSwapRequest) => {
+  const getQuote = (swapRequest: EdgeSwapRequest): void => {
     if (exchangeInfo != null) {
       const disableSrc = checkDisableAsset(
         exchangeInfo.swap.disableAssets.source,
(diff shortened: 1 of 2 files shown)


src/components/scenes/SwapCreateScene.tsx
```diff
@@ -531,7 +533,9 @@ export const SwapCreateScene = (props: Props) => {
             keyboardVisible={false}
             onAmountChanged={handleFromAmountChange}
             onNext={handleNext}
-            onSelectWallet={handleFromSelectWallet}
+            onSelectWallet={() => {
+              handleFromSelectWallet().catch(() => {})
+            }}
```
swansontec: I feel like this is the wrong solution. In `SwapInput.tsx`, we should have `onSelectWallet: () => Promise<void>`, and then if we use `EdgeTouchableOpacity` for `WalletPlaceHolder`, we get the correct promise handling.

src/components/scenes/SwapCreateScene.tsx
```diff
@@ -73,7 +73,7 @@ export interface SwapErrorDisplayInfo {
 
 interface Props extends SwapTabSceneProps<'swapCreate'> {}
 
-export const SwapCreateScene = (props: Props) => {
+export const SwapCreateScene = (props: Props): React.ReactElement => {
```
swansontec: `export const SwapCreateScene: React.FC<Props> = props => { ...`?

swansontec (changes requested):
There is one additional change.

src/components/scenes/SwapCreateScene.tsx
```diff
@@ -73,7 +73,7 @@ export interface SwapErrorDisplayInfo {
 
 interface Props extends SwapTabSceneProps<'swapCreate'> {}
 
-export const SwapCreateScene = (props: Props): React.ReactElement => {
+export const SwapCreateScene: React.FC<Props> = (props: Props) => {
```
swansontec: You don't need `(props: Props) =>`. It can just be `props =>`, since we have the types on the left.

PR #5715, opened 2025-08-26
Sam/infinite ramp plugin

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device


jest.config.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
   },
   preset: 'react-native',
   setupFilesAfterEnv: ['./jestSetup.js'],
+  testMatch: ['**/?(*.)+(test|spec).[tj]s?(x)'],
```
swansontec: Why not name tests normally, instead of tweaking the rules? We strongly want to keep our project as close to upstream React Native as possible, because each difference makes upgrading more time-consuming.

Most of our mock data lives in /util/fake. Maybe rename `sampleRecords` to `fakeVaultRecords` and put it in there.

src/components/themed/EdgeText.tsx
```diff
@@ -31,11 +31,14 @@ interface ParagraphProps extends TextProps {
   children: React.ReactNode
 
   center?: boolean
+  /** Makes sure the text doesn't overflow it's container */
+  fit?: boolean
+
```
swansontec: Please revert this change, and just apply `flexShrink: 1` as needed at the call site.

This style only makes sense in a row layout, and will do the wrong thing in a column layout, so `EdgeText` is too low-level to apply a style like this. Also, this is too risky for staging.

src/components/Main.tsx
```diff
@@ -422,8 +432,8 @@ const EdgeBuyTabScreen: React.FC = () => {
         }}
       />
       <BuyStack.Screen
-        name="guiPluginContactForm"
-        component={ContactFormScene}
+        name="kycForm"
+        component={RampKycFormScene}
```
swansontec, line 444: It looks like the `ContactFormScene` file isn't deleted, and `fiatPlugin.tsx` still uses it? Unless I am reading the diff wrong.

Maybe leave the existing scene as-is, and simply add the new `kycForm` next to it?

src/components/scenes/DevTestScene.tsx
```diff
@@ -175,6 +175,30 @@ export const DevTestScene: React.FC<Props> = props => {
     })
   }
 
+  const handleKycFormPress = (): void => {
+    navigation2.navigate('buyTab', {
+      screen: 'guiPluginContactForm',
```
swansontec, line 192: Yeah, here it uses the old scene name, not the new `kycForm` name.

src/components/scenes/RampBankFormScene.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const [isSubmitting, setIsSubmitting] = React.useState(false)
+  const [error, setError] = React.useState<unknown>(null)
+
+  // Create refs for each input field
+  const bankNameRef = React.useRef<TextInput>(null)
+  const ownerFirstNameRef = React.useRef<TextInput>(null)
+  const ownerLastNameRef = React.useRef<TextInput>(null)
+  const accountNameRef = React.useRef<TextInput>(null)
+  const accountNumberRef = React.useRef<TextInput>(null)
+  const routingNumberRef = React.useRef<TextInput>(null)
```
swansontec: These all appear unused. Did you meant to implement the same auto-focus logic as in `RampKycFormScene.tsx`, but never got around to it?

src/components/scenes/RampBankRoutingDetailsScene.tsx
```diff
@@ -0,0 +1,171 @@
...
+    value
+  }: {
+    label: string
+    value: string
+  }): React.ReactElement => (
+    <InfoRowContainer>
+      <InfoLabel>{label}</InfoLabel>
+      <InfoValue selectable>{value}</InfoValue>
+    </InfoRowContainer>
+  )
```
swansontec: This is fine, but call it as a normal function, not a component:

```ts
const renderInfoRow = (label: string, value: string): React.ReactNode => {
  return <InfoRowContainer>
      <InfoLabel>{label}</InfoLabel>
      <InfoValue selectable>{value}</InfoValue>
    </InfoRowContainer>
}
```

You can call it like `{renderInfoRow('label', 'value')}`.

src/components/scenes/RampPendingKycScene.tsx
swansontec, line 1: Empty file, `RampPendingKycScene.tsx`.

src/components/scenes/RampPendingScene.tsx
```diff
@@ -0,0 +1,146 @@
...
+        setStatus(status)
+        // Step-off algorithm: double the interval each time, max 10 seconds
+        pollIntervalRef.current = Math.min(pollIntervalRef.current * 2, 10000)
+        await new Promise((resolve, reject) => {
+          timeoutId = setTimeout(() => {
+            checkStatus().then(resolve, reject)
+          }, pollIntervalRef.current)
+        })
+      }
+    }
```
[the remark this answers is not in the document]
swansontec: I agree - don't do hand-coded periodic async stuff. Instead, use `makePeriodicTask`, which we copied from the core. The core version actually includes the ability to tweak the delay, so you can use that for your exponential step-off.

src/components/services/AirshipInstance.tsx
```diff
@@ -132,10 +132,11 @@ export function showToast(message: string, autoHideMs?: number): void {
  */
 export async function showToastSpinner<T>(
   message: string,
-  activity: Promise<T>
+  activity: Promise<T> | (() => Promise<T>)
```
swansontec: Please revert this change, as it's not really needed.

src/components/themed/FilledTextInput.tsx
```diff
@@ -233,6 +234,7 @@ export const FilledTextInput = React.forwardRef<
     blurOnSubmit,
     disabled = false,
     keyboardType,
+    minLength,
```
swansontec, line 239: Can you add a doxygen comment for this prop (`/** Minimum number of input characters */`) and move it to its own commit? Then I would be comfortable accepting this to staging.

src/components/themed/FilledTextInput.tsx
```diff
@@ -847,14 +857,22 @@ const MessagesContainer = styled(Animated.View)<{ noLayoutFlow?: boolean }>(
...
-  {
-    color: props.danger === true ? theme.dangerText : theme.secondaryText,
-    fontFamily: theme.fontFaceDefault,
-    fontSize: theme.rem(0.75),
-    includeFontPadding: false
-  }
-])
+const Message = styled(UnscaledText)<{ mode?: 'danger' | 'warn' }>(
+  theme => props =>
+    [
```
swansontec: Why add the array?

src/plugins/gui/components/GuiFormField.tsx
```diff
@@ -26,12 +30,13 @@ interface Props {
   onChangeText: ((text: string) => void) | undefined
   onFocus?: (() => void) | undefined
   onBlur?: (() => void) | undefined
+  onSubmitEditing?: (() => void) | undefined
```
swansontec: Weird, the `?:` means `undefined` is allowed, so why add the `| undefined` again?

src/plugins/gui/components/GuiFormField.tsx
```diff
@@ -26,12 +30,13 @@ interface Props {
...
 }
 
 export const GuiFormField = React.memo((props: Props) => {
   const {
     fieldType = 'text',
-    key = fieldType || undefined,
+    key = fieldType ?? undefined,
```
swansontec: The `?? undefined` syntax is a no-op, unless the goal is to turn `null` into `undefined` (but that's not the case here).

src/components/scenes/RampKycFormScene.tsx
swansontec, line 1: This should go into the normal components/scenes folder, not over here. Actually, I'm surprised to see that we have a plugins/gui folder at all, but we can't clean that up during staging, so we have to just leave it for now.

src/plugins/ramps/infinite/infiniteApi.ts
```diff
@@ -0,0 +1,854 @@
...
+  getCountries: false,
+  getCurrencies: false,
+  createPrivateKey: false, // This is always local, no API call
+  signChallenge: false, // This is always local, no API call
+  getPublicKeyFromPrivate: false, // This is always local, no API call
+  clearAuth: false, // This is always local, no API call
+  getAuthState: false, // This is always local, no API call
+  saveCustomerId: false, // This is always local, no API call
+  isAuthenticated: false // This is always local, no API call
+}
```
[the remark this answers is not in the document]
swansontec: I disagree with Cursor on this one. This is critical for development since we don't want to use real funds for testing.

src/plugins/ramps/infinite/infiniteApi.ts
```diff
@@ -0,0 +1,854 @@
...
+
+// Utility to convert Uint8Array to hex string
+const bytesToHex = (bytes: Uint8Array): string => {
+  return Array.from(bytes)
+    .map(b => b.toString(16).padStart(2, '0'))
+    .join('')
+}
```
swansontec, line 70: Use `base16.stringify(bytes)` from 'rfc4648'. You can throw in a `.toLowerCase()` if you care about casing.

src/plugins/ramps/infinite/utils/navigationFlow.ts
```diff
@@ -0,0 +1,28 @@
+import type { NavigationBase } from '../../../../types/routerTypes'
+
+export interface NavigationFlow {
+  navigate: (route: string, params?: any) => void
```
swansontec: I don't like the `any`. Since we can go to scenes like `rampPending`, it seems like we are dealing with the buy / sell tab navigator, so we should type this using those route names / params.

src/plugins/ramps/infinite/workflows/confirmationWorkflow.ts
```diff
@@ -0,0 +1,150 @@
...
+          // Show deposit instructions
+          if (transfer.sourceDepositInstructions?.toAddress != null) {
+            // TODO: Show deposit address to user
+            showToast(
+              `Send ${request.displayCurrencyCode} to: ${transfer.sourceDepositInstructions.toAddress}`
+            )
+          }
+
+          resolve({ confirmed: true, transfer })
+        }
```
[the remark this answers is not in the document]
swansontec: Cursor makes a valid point - the funds could come from many utxo's, not necessarily our latest receive address (indeed, this is usually a fresh address for utxo coins). If Infinite.dev is just using this as a refund address, then it's fine. If not, we might need a rethink.

src/util/vault/edgeVault.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+    return record
+  }
+
+  return {
+    createPersonalInfo: async info => {
+      try {
+        const record = await makeRecord(info)
+        const json = wasVaultPersonalRecord(record)
+        await vaultDisklet.setText(`${info.type}/${record.uuid}.json`, json)
```
swansontec: So we are saving data in `edgeVault/{randomHex}.json`. Ok, I guess this is fine.

They don't really need to be UUID's - we could do 128 bits of base58 (about 21 characters). This would simplify creating these things, and since we don't really parse them or read into them, it wouldn't affect anybody.

src/plugins/ramps/infinite/workflows/kycWorkflow.ts
```diff
@@ -0,0 +1,302 @@
...
+              }
+            },
+            onClose: () => {
+              if (!hasResolved) {
+                hasResolved = true
+                resolve(true)
+              }
+              return true // Allow close
+            }
+          })
```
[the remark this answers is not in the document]
samholmes: I disagree here. The point is there are two code-paths that lead to the resolution of this promise: either the user closes the webview or they reach a deeplink from within a webview. Both paths leads them back to the same point in the flow. If the user follows the deeplink, but it takes longer for that deeplink to be handled in the app before they close the webview, then there's potential that both event handlers are triggered, and in such a case, only one needs to resolve the promise and it's okay if the other no-ops.

src/plugins/ramps/infinite/infiniteApi.ts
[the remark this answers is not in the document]
samholmes: It may be valid to remove this and clean up in the future. This code has early development/testing purpose, but as we rely on the sandbox, it is less relevant.

src/plugins/ramps/infinite/workflows/confirmationWorkflow.ts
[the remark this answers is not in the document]
samholmes: So sell is not going to be supported in this release. There are outstanding considerations which need to be designed around both on Infinite's part and our own. I took a good amount of time discussing this with Raj today and Paul separately. I'm deciding that we're going to make it "buy" only for this sprint until the known issues are sorted out. (See latest commits where sell is explicitly/properly disabled).

src/plugins/ramps/utils/exitUtils.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * Exit error class for gracefully exiting workflows
+ */
+export class Exit extends Error {
```
swansontec: This should probably be called `ExitError`. It's not really an error as such, because we expect it, but calling it `ExitError` makes it clear that we want to catch it.

src/plugins/ramps/infinite/workflows/kycWorkflow.ts
```diff
@@ -0,0 +1,302 @@
...
+    if (kycStatus.kycStatus === 'approved') {
+      return
+    }
+
+    // If not_started or incomplete, show KYC form
+    if (
+      kycStatus.kycStatus === 'not_started' ||
+      kycStatus.kycStatus === 'incomplete'
+    ) {
+      // Fall through to show KYC form
```
swansontec: This logic threw me for a loop. Would it make more sense:

```ts
if (
  kycStatus.kycStatus !== 'not_started' &&
   kycStatus.kycStatus !== 'incomplete'
) {
  // For all other statuses (under_review, awaiting_ubo, etc.), show pending scene
      await showKycPendingScene(...
}
```

src/plugins/ramps/infinite/infiniteRampPlugin.ts
```diff
@@ -0,0 +1,671 @@
...
+  // Plugin state
+  const state: InfinitePluginState = {}
+
+  const getPrivateKey = async (): Promise<Uint8Array> => {
+    if (state.privateKey != null) {
+      return state.privateKey
+    }
+    let privateKey: Uint8Array
+    // Try to load from storage (stored as hex string)
+    const itemIds = await account.dataStore.listItemIds(pluginId)
```
swansontec: Just do `const key: string | undefined = account.dataStore.getItem(pluginId, INFINITE_PRIVATE_KEY).catch(() => {})`. It will either return the key if we have it, or undefined. Saves a round-trip over the bridge and a bunch of expensive disk access.

src/plugins/ramps/infinite/infiniteRampPlugin.ts
```diff
@@ -0,0 +1,671 @@
...
+  }
+
+  // DEV ONLY: Export for development clearing
+  const _devOnlyClearAuthKey = async (
+    account: any,
+    pluginId: string
+  ): Promise<void> => {
+    const itemIds = await account.dataStore.listItemIds(pluginId)
+    if (itemIds.includes(INFINITE_PRIVATE_KEY) === true) {
+      await account.dataStore.deleteItem(pluginId, INFINITE_PRIVATE_KEY)
```
swansontec: Just call `account.dataStore.deleteItem(pluginId, INFINITE_PRIVATE_KEY)` directly. Deleting missing items is a no-op - it's not an error, so there's no need to guard the call.

src/plugins/ramps/infinite/infiniteRampPlugin.ts
```diff
@@ -0,0 +1,652 @@
...
+    }
+    state.privateKey = undefined
+    infiniteApi.clearAuth()
+  }
+
+  // Cache for API responses
+  let countriesCache: CacheEntry<InfiniteCountriesResponse> | null = null
+  let currenciesCache: CacheEntry<InfiniteCurrenciesResponse> | null = null
+  let normalizedCurrenciesCache: CacheEntry<NormalizedCurrenciesMap> | null =
```
swansontec: These two caches contain the same data, just in two formats. Could they be merged somehow?

```ts
let currenciesCache: CacheEntry<{
  raw: InfiniteCurrenciesResponse
  normalized: NormalizedCurrenciesMap
}> | null = null

...

const getCurrenciesWithCache =
 ...
   const currenciesData = await infiniteApi.getCurrencies()
   const data = { 
      raw: currenciesData,
      normalized: normalizeCurrencies(currenciesData)
    }
    currenciesCache = { data, timestamp: Date.now() }
    return data
```

src/plugins/ramps/infinite/infiniteRampPlugin.ts
```diff
@@ -0,0 +1,652 @@
...
+
+            // Ensure we have a bank account
+            const bankAccountResult = await bankAccountWorkflow({
+              infiniteApi,
+              navigationFlow,
+              vault
+            })
+
+            // Get fresh quote before confirmation using existing params
+            const freshQuote = await infiniteApi.createQuote(quoteParams)
```
swansontec, line 556: So we're re-quoting in case the KYC took a long time, and the old quote expired?

src/plugins/ramps/infinite/infiniteRampPlugin.ts
```diff
@@ -0,0 +1,652 @@
...
+    bytes[i / 2] = parseInt(hex.substring(i, i + 2), 16)
+  }
+  return bytes
+}
+
+const bytesToHex = (bytes: Uint8Array): string => {
+  return Array.from(bytes)
+    .map(b => b.toString(16).padStart(2, '0'))
+    .join('')
+}
```
swansontec, line 645: These should be `base16.parse` / `base16.stringify` from rfc4648

src/plugins/ramps/infinite/infiniteApi.ts
```diff
@@ -0,0 +1,854 @@
...
+    const headers: Record<string, string> = {
+      'X-Organization-ID': config.orgId,
+      'Content-Type': 'application/json'
+    }
+
+    // Determine if we should include auth
+    const shouldIncludeAuth =
+      options?.includeAuth ?? (authState.token != null && !isTokenExpired())
+
+    if (shouldIncludeAuth && authState.token != null && !isTokenExpired()) {
```
swansontec: This way should be equivalent, and simpler:

```ts
const hasAuth = authState.token != null && !isTokenExpired()
if (options?.includeAuth !== false && hasAuth) {
```

src/components/scenes/RampBankFormScene.tsx
samholmes: Only the first one wasn't used. The others are used for focusing the next input when the return key is pressed.

src/components/themed/FilledTextInput.tsx
samholmes: I've added the comment, but I still need to break up the commit into a separate commit.

src/components/themed/FilledTextInput.tsx
samholmes: No array was added, but rather preserved. It can be removed however.

src/plugins/ramps/infinite/utils/navigationFlow.ts
samholmes: There's a high cog-load to the navigation types, so I'll just make it typed to NavigationBase

src/util/vault/edgeVault.ts
samholmes: If you feel like its worth it to switch from UUID to hex or base58 encoding of random bits, then I'll do it.

src/plugins/ramps/infinite/infiniteRampPlugin.ts
samholmes: Yes. KYC process tasks many minutes.

src/plugins/ramps/infinite/utils/navigationFlow.ts
swansontec: Calling `navigation.navigate('foo')` can do different things depending on which `navigate` you have. It might be necessary to do `navigation.navigate('edgeTabs', { screen: 'foo' } })` or something else weird to make it work. This is why we try to be picky about which navigator we use - React Navigation is becoming stricter and stricter about the correct syntax with each major version bump.

The `NavigationBase` is a good start. We can roll with it for now, but be aware that we might have to fix this prior to our next React Navigation upgrade.

src/util/vault/edgeVault.ts
swansontec: I feel like we should change the name `uuid` to `recordId` or `vaultId` or such, which says what the ID is used for, not what format it is.

swansontec (approved):
All good, aside from renaming uuid to vaultId or something.

PR #5720, opened 2025-08-27
Jon/fix/zano alias display

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1210698424744262

### DIFF
--- a/src/components/scenes/SendScene2.tsx
+++ b/src/components/scenes/SendScene2.tsx
@@ -48,6 +48,7 @@ import {
   checkRecordSendFee,
   FIO_FEE_EXCEEDS_SUPPLIED_MAXIMUM,
   FIO_NO_BUNDLED_ERR_CODE,
+  FioError,
   recordSend
 } from '../../util/FioAddressUtils'
 import { logActivity } from '../../util/logger'
@@ -162,7 +163,7 @@ const MULTI_OUT_DIFF_PERCENT = '0.005'
 const PIN_MAX_LENGTH = 4
 const INFINITY_STRING = '999999999999999999999999999999999999999'
 
-const SendComponent = (props: Props) => {
+const SendComponent = (props: Props): React.ReactElement => {
   const { route, navigation } = props
   const dispatch = useDispatch()
   const theme = useTheme()
@@ -302,7 +303,7 @@ const SendComponent = (props: Props) => {
 
   if (initialMount.current) {
     if (hiddenFeaturesMap.scamWarning === false) {
-      showSendScamWarningModal(account.disklet).catch(err => {
+      showSendScamWarningModal(account.disklet).catch((err: unknown) => {
         showError(err)
       })
     }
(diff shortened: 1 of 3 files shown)


src/components/scenes/SendScene2.tsx
```diff
@@ -1042,123 +1080,125 @@ const SendComponent = (props: Props) => {
...
-            fioSender.fioAddress
+        if (payeeName != null && fioSender != null) {
+          let fioNotes = sprintf(
+            `${lstrings.sent}\n`,
+            `${lstrings.fragment_send_from_label.toLowerCase()} ${
+              fioSender.fioAddress
+            }`
+          )
+          fioNotes +=
+            fioSender.memo != null
```
swansontec: Should we include `fioSender.memo !== ''` here?

src/components/scenes/SendScene2.tsx
```diff
@@ -1387,13 +1432,13 @@ const SendComponent = (props: Props) => {
             errorCurrencyCode === 'ETH' &&
             coreWallet.currencyInfo.pluginId !== 'ethereum'
           ) {
-            e.message = sprintf(
+            ;(e as Error).message = sprintf(
```
swansontec: I don't like these casts. Instead, let's do `const error = e instanceof Error ? e : new Error(String(e))`. Then for the rest of the giant `catch` block, we now know for sure that `error` is definitely an `Error` object, so we can just search & replace e -> error.

src/components/tiles/AddressTile2.tsx
```diff
@@ -120,13 +120,13 @@ export const AddressTile2 = React.forwardRef(
...
-              !e.code ||
-              e.code !==
-                fioPlugin.currencyInfo.defaultSettings?.errorCodes
-                  .INVALID_FIO_ADDRESS
-            ) {
+          } catch (e: unknown) {
+            const invalidCode =
+              fioPlugin.currencyInfo.defaultSettings?.errorCodes
+                .INVALID_FIO_ADDRESS
+            const code = (e as { code?: unknown }).code
```
swansontec: I feel like we could do something like this:

```ts
const asCodeError = asObject({ code: asString })
const codeError asMaybe(asCodeError)(e)
if (codeError != null && codeError.code !== invalidCode) ...
```

src/components/scenes/SendScene2.tsx
```diff
@@ -1252,24 +1252,28 @@ const SendComponent = (props: Props): React.ReactElement => {
       } catch (e: unknown) {
         resetSlider()
         console.log(e)
-        if (e instanceof Error) {
-          let message = sprintf(lstrings.transaction_failure_message, e.message)
-          e.message = 'broadcastError'
-          if (e.name === 'ErrorAlgoRecipientNotActivated') {
+        const error = e instanceof Error ? e : new Error(String(e))
+        if (error != null) {
```
swansontec: We don't need the `if (error != null) {` check, because the line above guarantees that this is impossible.

swansontec (approved):
Just delete the one extra line, and it's good-to-go.

PR #5724, opened 2025-09-02
Paul/unscaled text

paullinator:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1211026079011275

### DIFF
--- a/src/components/themed/FilledTextInput.tsx
+++ b/src/components/themed/FilledTextInput.tsx
@@ -430,6 +430,7 @@ export const FilledTextInput = React.forwardRef<
             {placeholder == null ? null : (
               <Placeholder shift={focusValue}>
                 <PlaceholderText
+                  allowFontScaling={false}
                   disableAnimation={disableAnimation}
                   focusAnimation={focusAnimation}
                   scale={scale}
@@ -442,7 +443,10 @@ export const FilledTextInput = React.forwardRef<
             )}
 
             {prefix == null ? null : (
-              <PrefixAnimatedText visibility={focusValue}>
+              <PrefixAnimatedText
+                allowFontScaling={false}
+                visibility={focusValue}
+              >
                 {prefix}
               </PrefixAnimatedText>
             )}
--- a/src/__tests__/components/__snapshots__/FilledTextInput.test.tsx.snap
+++ b/src/__tests__/components/__snapshots__/FilledTextInput.test.tsx.snap
@@ -136,6 +136,7 @@ exports[`FilledTextInput should render with some props 1`] = `
       }
     >
       <Text
+        allowFontScaling={false}
         collapsable={false}
         jestAnimatedProps={
           {
@@ -278,6 +279,7 @@ exports[`FilledTextInput should render with some props 1`] = `
         }
       >
         <Text
+          allowFontScaling={false}
           collapsable={false}
           disableAnimation={0}
           focusAnimation={0}
@@ -487,6 +489,7 @@ exports[`FilledTextInput should render with some props 1`] = `
         }
       >
         <Text
+          allowFontScaling={false}
           collapsable={false}
           jestAnimatedProps={
             {
--- a/src/__tests__/modals/__snapshots__/AddressModal.test.tsx.snap
+++ b/src/__tests__/modals/__snapshots__/AddressModal.test.tsx.snap
@@ -452,6 +452,7 @@ exports[`AddressModalComponent should render with loaded props 1`] = `
             }
           >
             <Text
+              allowFontScaling={false}
               collapsable={false}
               disableAnimation={0}
               focusAnimation={0}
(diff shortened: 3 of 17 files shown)


src/components/cards/BalanceCard.tsx
```diff
@@ -27,12 +27,14 @@ import {
...
 import { Airship } from '../services/AirshipInstance'
 import { cacheStyles, type Theme, useTheme } from '../services/ThemeContext'
+import { UnscaledText } from '../text/UnscaledText'
 import { EdgeText } from '../themed/EdgeText'
 import { EdgeCard } from './EdgeCard'
 
 // Numbers larger than this are likely to overflow the display width so don't
 // use regular Text components which can auto shrink
 const MAX_ANIMATED_AMOUNT = '10000000'
+const ZERO_STRING = '0'
```
swansontec, line 37: This is because of the ESLint warning, right? Otherwise I would say that this constant is pretty useless.

src/components/themed/EdgeText.tsx
```diff
@@ -82,15 +82,15 @@ export const EdgeText = (props: LabelProps) => {
   }
 
   return (
-    <Text
+    <UnscaledText
```
swansontec: Because `EdgeText` is used in a *ton* of places, the extra level of rendering will have a small but meaninful perf impact. Can we use the native `Text` element and then directly apply `allowFontScaling={false}` here, and here alone? This should minimize the perf hit without spreading too much hard-coded scaling logic over the codebase.

src/components/themed/SimpleTextInput.tsx
```diff
@@ -27,8 +27,10 @@ import {
   CloseIconAnimated
 } from '../icons/ThemedIcons'
 import { useTheme } from '../services/ThemeContext'
+import { UnscaledText } from '../text/UnscaledText'
+import { UnscaledTextInput } from '../text/UnscaledTextInput'
 
-const AnimatedTextInput = Animated.createAnimatedComponent(TextInput)
+const AnimatedTextInput = Animated.createAnimatedComponent(UnscaledTextInput)
```
swansontec: I suspect this is going to cause perf issues, since we're animating a JS wrapper, and reanimated can't "see" in to realize which props are native or not. Therefore, Reanimated will have to loop Hermes into every frame, which will chunder. I think we need to use `Animated.createAnimatedComponent(TextInput)` directly, and then manually apply `allowFontScaling` to that.

src/components/themed/FilledTextInput.tsx
```diff
@@ -651,7 +651,7 @@ const InnerContainer = styled(Animated.View)<{
   ]
 })
 
-const PrefixAnimatedText = styled(Animated.Text)<{
+const PrefixAnimatedText = styled(AnimatedUnscaledText)<{
```
swansontec: I am pretty sure this will chunder for the same reason as the text input wrapper would. Can we double-check if that's the case? If so, we'll have to directly use the animated native text component, and manually apply `allowFontScaling` in each of these places (unfortunate but necessary).

swansontec (changes requested):
I don't see any "bugs", aside from the one the AI noticed. I do have some perf concerns - Reanimated uses some magic to notice which props it can directly manipulate in native code, and I worry that these wrappers will break the magic and rope Hermes into each frame.

src/components/cards/BalanceCard.tsx
paullinator: yep. eslint doesn't recognize UnscaledText as a Text component.

src/components/themed/EdgeText.tsx
paullinator: That's fair and I considered that. Although I didn't think we'd actually render any additional work since the wrapper only applies the single flag with no actual Views. Will this actually cause extra bridge calls?

src/components/cards/BalanceCard.tsx
swansontec: In our eslint.config.mjs, we have:

```ts
'react-native/no-raw-text': [
        'error',
        {
          skip: ['B', 'EdgeText', 'Paragraph', 'SmallText', 'WarningText']
        }
      ]
```

We should probably add UnscaledText to the list.

src/components/themed/EdgeText.tsx
swansontec: There wouldn't be any bridge calls, just time spent in the JS engine running more code.

src/components/themed/FilledTextInput.tsx
swansontec: This turns out to be fine, since since the animated text component is *inside* the wrapper. This is different from the text inputs, where we animate the *wrapper*. Animating the wrapper is bad, but wrapping the animation is fine. Carry on.

PR #5732, opened 2025-09-05
Paul/fix more scaling

paullinator:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1211228803687939

### DIFF
--- a/src/components/themed/FilledTextInput.tsx
+++ b/src/components/themed/FilledTextInput.tsx
@@ -449,6 +449,7 @@ export const FilledTextInput = React.forwardRef<
               </PrefixAnimatedText>
             )}
             <StyledAnimatedTextInput
+              allowFontScaling={false}
               accessible
               editable={!disabled}
               ref={inputRef}
--- a/src/components/themed/FlipInput2.tsx
+++ b/src/components/themed/FlipInput2.tsx
@@ -489,6 +489,7 @@ const AmountAnimatedNumericInput = React.forwardRef<
 
   return (
     <AnimatedTextInput
+      allowFontScaling={false}
       ref={ref}
       style={[style, animatedStyle]}
       {...rest}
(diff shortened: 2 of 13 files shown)


src/components/themed/FilledTextInput.tsx
```diff
@@ -224,7 +224,7 @@ export const FilledTextInput = React.forwardRef<
   const hasIcon = LeftIcon != null
 
   const displayValue = React.useMemo(() => {
-    if (numeric && value !== '') {
+    if ((numeric ?? false) && value !== '') {
```
swansontec: In cases like these with an optional boolean, the team has settled on `value === true` as being the easiest-to-read idiom. You can also do `value !== true` in the opposite case.

However, since this is a *prop*, there is an even easier way. Just give a default to get rid of the `undefined` case:

```ts
const {
  ...
  numeric = false,
  secureTextEntry = false,
  ...
} = props
```

Now instead of editing a bunch of places in this file, it becomes a simple 2-line change, and all the warnings disappear.

src/components/themed/FlipInput2.tsx
```diff
@@ -173,8 +173,8 @@ export const FlipInput2 = React.forwardRef<FlipInputRef, Props>(
             setAmounts(newAmounts)
           }
         })
-        .catch(e => {
-          showDevError(e.message)
+        .catch((e: unknown) => {
+          showDevError(e instanceof Error ? e.message : 'Unknown error')
```
swansontec: The `showDevError` handles all this logic. Just do `showDevError(e)` directly.

src/components/themed/FilledTextInput.tsx
```diff
@@ -504,7 +505,7 @@ export const FilledTextInput = React.forwardRef<
           ) : null}
 
           <TouchContainer
-            extendTappable={secureTextEntry ? 'rightOnly' : 'full'}
+            extendTappable={secureTextEntry ?? false ? 'rightOnly' : 'full'}
```
swansontec: You missed a spot for ` === true`

PR #5734, opened 2025-09-06
Jon/fix/chart gesture hack

Jon-edge:
`WalletDetailsScene` manages its own scroll, so it needed the explicit binding to disable the scene-level scrolling while the chart gesture was active.

Unable to find a workaround to reliably dynamically resolve the chart vs drawer gesture conflict, most noticeable in `WalletDetailsScene` but also still possible under some rapid interactions in `CoinRankingDetailsScene` so the drawer gesture was simply disabled for all chart scenes. Newer androids already have a built-in gesture conflict anyway with the drawer since the swipe from right gesture is reserved for "Back" in most modern Androids, so we're not missing much by going with this approach.

iOS was left without modifications to drawer gesture functionality since it's such an "edge" case (literally). There's only a tiny area of chart pointer activation where the conflict would occur.

### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

### DIFF
--- a/src/components/Main.tsx
+++ b/src/components/Main.tsx
@@ -1072,6 +1072,7 @@ const EdgeAppStack: React.FC = () => {
 const EdgeApp: React.FC = () => {
   return (
     <Drawer.Navigator
+      id="edgeDrawer"
       drawerContent={({ navigation }) => <SideMenu navigation={navigation} />}
       initialRouteName="edgeAppStack"
       screenOptions={{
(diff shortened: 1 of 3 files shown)


swansontec (approved):
Approving for now - it works. Reanimated has “useAnimatedProps”, which might be a better solution, but there is no reason to get fancy at 7 on a Friday.

PR #5748, opened 2025-09-22
Sam/logbox

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -109,7 +109,7 @@ console.log('***********************')
 console.log('App directory: ' + RNFS.DocumentDirectoryPath)
 console.log('***********************')
 
-// @ts-expect-error
+// @ts-expect-error - this needs an explanation
 global.clog = console.log
 
 if (!__DEV__) {
@@ -119,7 +119,7 @@ if (!__DEV__) {
   console.error = log
 }
 
-if (ENV.LOG_SERVER) {
+if (ENV.LOG_SERVER != null) {
   console.log = function () {
     logToServer(arguments)
   }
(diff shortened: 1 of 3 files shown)


src/envConfig.ts
```diff
@@ -422,6 +422,7 @@ export const asEnvConfig = asObject({
   DEBUG_PLUGINS: asOptional(asBoolean, false),
   DEBUG_ACCOUNTBASED: asOptional(asBoolean, false),
   DEBUG_EXCHANGES: asOptional(asBoolean, false),
+  DEBUG_LOGBOX: asOptional(asBoolean, true),
```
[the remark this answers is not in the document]
samholmes: Fixed it. It should have a default to enable now.

src/app.ts
```diff
@@ -138,12 +138,16 @@ if (PERF_LOGGING_ONLY) {
 }
 
 if (ENABLE_PERF_LOGGING) {
-  // @ts-expect-error
-  if (!global.nativePerformanceNow && window?.performance) {
-    // @ts-expect-error
-    global.nativePerformanceNow = () => window.performance.now()
+  const nativePerformanceNow = (): number => {
+    // @ts-expect-error - this is a hack to get around the fact that window is not typed
+    return window.performance.now()
```
swansontec, line 143: Take all this performance stuff, and move it to its own file. Do not attach it to the global scope, but instead `export const pnow =` and so forth for each performance function. If we want to do performance analysis in the future, we can just `import { pstart, pend, ... } from '../../perf.ts'` or whatever you call the new file, rather than grabbing it from the global scope.

src/app.ts
```diff
@@ -109,7 +109,7 @@ console.log('***********************')
 console.log('App directory: ' + RNFS.DocumentDirectoryPath)
 console.log('***********************')
 
-// @ts-expect-error
+// @ts-expect-error - this needs an explanation
```
swansontec, line 112: Indeed, it does need an explanation.

This is needed for our logging integration. When we replace `console.log` with our  own custom logging, we use `global.clog` to keep forwarding logs to the normal place.

This is dumb. Instead, our logging infrastructure should use named variable like `export const realLog = console.log`, and then call `realLog()` in the logging system. That avoids polluting the global scope, and makes it clear where this is being used (TypeScript becomes aware of what's going on).

src/app.ts
```diff
@@ -264,17 +268,22 @@ if (ENV.DEBUG_THEME) {
         } catch (e: any) {
           console.log(`Failed get theme`, e.message)
         }
+      }
+      setInterval(() => {
+        intervalCallback().catch((err: unknown) => {
+          console.error(err)
+        })
```
swansontec, line 275: Optional: This should really be `makePeriodicTask`, since that avoids overlapping work in cases where the `async` function takes more that 3s.

swansontec (changes requested):
The lint cleanups in `app.js` need to go a bit deeper. I haven't reviewed the second commit, since I'm not the assigned reviewer.

src/app.ts
paullinator: The reason why this is in global scope was for the convenience in putting it in any repo like login-ui. Admittedly since most of our code is in the webview, this doesn't help much although it would be nice to have some equivalent in there as well.

PR #5759, opened 2025-09-26
Update "Recent Wallets"

Jon-edge:
image

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1208654467278105

---

> [!NOTE]
> Separates recent wallets from the rest in the wallet picker and renames the "All Wallets" section to "Other Wallets", with handler/type refactors.
> 
> - **Wallet Picker UX**:
>   - Exclude "Most Recent Wallets" from the main list; show as separate section above "Other Wallets".
>   - Rename section header from `wallet_list_modal_header_all` ("All Wallets") to `wallet_list_modal_header_other` ("Other Wallets").
> - **Logic/Refactor**:
>   - Use `useHandler` and async `handlePress`; `onPress` now supports `Promise<void> | void` across rows.
>   - Avoid `excludeWalletIds` null issues; minor memo/var cleanups and deduping of recent assets from the rest list.
>   - Convert `WalletList` to `React.FC` and simplify row press wiring.
> - **Localization**:
>   - Update EN strings/keys to reflect "Other Wallets".
> - **Changelog**:
>   - Note UI change: recent wallets no longer appear under other wallets; section renamed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b849799a418366f1648b3ec42e719fc7400ce8c8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/components/themed/WalletList.tsx
+++ b/src/components/themed/WalletList.tsx
@@ -4,6 +4,7 @@ import type { ViewStyle } from 'react-native'
 import { FlatList } from 'react-native-gesture-handler'
 
 import { selectWalletToken } from '../../actions/WalletActions'
+import { useHandler } from '../../hooks/useHandler'
 import { lstrings } from '../../locales/strings'
 import {
   filterWalletCreateItemListBySearchText,
@@ -14,7 +15,6 @@ import { useDispatch, useSelector } from '../../types/reactRedux'
 import type { NavigationBase } from '../../types/routerTypes'
 import type { EdgeAsset, FlatListItem, WalletListItem } from '../../types/types'
 import { checkAssetFilter } from '../../util/CurrencyInfoHelpers'
-import { showError } from '../services/AirshipInstance'
 import { searchWalletList } from '../services/SortedWalletList'
 import { useTheme } from '../services/ThemeContext'
 import { ModalFooter } from './ModalParts'
@@ -40,14 +40,14 @@ interface Props {
   parentWalletId?: string
 
   // Callbacks:
-  onPress?: (walletId: string, tokenId: EdgeTokenId) => Promise<void>
+  onPress?: (walletId: string, tokenId: EdgeTokenId) => Promise<void> | void
 }
 
 /**
  * This list is used inside the wallet list modal,
  * and *only* the wallet list modal.
  */
-export function WalletList(props: Props) {
+export const WalletList: React.FC<Props> = (props: Props) => {
   const dispatch = useDispatch()
   const {
     navigation,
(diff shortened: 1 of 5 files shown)


src/components/themed/WalletList.tsx
```diff
@@ -253,14 +275,16 @@ export function WalletList(props: Props) {
             <WalletListCreateRow
               createItem={item.item}
               createWalletId={createWalletId}
-              onPress={handlePress}
+              onPress={async (walletId, tokenIdValue) => {
+                handleSafePress(walletId, tokenIdValue)
+              }}
```
swansontec: The `onPress` prop already expects an `async` function, and correctly handles the `catch` stuff. So the existing `handlePress` *should* match the type perfectly, without warnings. No changes needed on this line.

Now the `WalletListCurrencyRow` component up above expects `onPress: (...) => void`, but it would be easy to change it to `onPress: (...) => Promise<void> | void`. The internal `EdgeCard` and `EdgeTouchableOpacity` are already promise-aware, so it's a trivial upgrade with one `async` / `await` needed. The file is pretty clean already, so there's hardly any lint fixes to do either. The no-deprecated rule doesn't need to be addressed.

src/components/themed/WalletList.tsx
```diff
@@ -86,6 +87,14 @@ export function WalletList(props: Props) {
...
   )
 
+  const handleSafePress = React.useMemo(() => {
+    return (walletId: string, tokenId: EdgeTokenId): void => {
+      handlePress(walletId, tokenId).catch((error: unknown) => {
+        showError(error)
+      })
+    }
+  }, [handlePress])
+
```
swansontec: As mentioned elsewhere, this shouldn't be needed because `WalletListCurrencyRow` should just take an async function to begin with.

src/components/themed/WalletList.tsx
Jon-edge, line 280: 👍 Test reply via MCP

src/components/themed/WalletList.tsx
Jon-edge, line 97: 👍 Test reply via MCP

Jon-edge (commented):
Automated test replies to review comments.

src/components/themed/WalletListCurrencyRow.tsx
```diff
@@ -60,7 +63,14 @@ const WalletListCurrencyRowComponent = (props: Props) => {
...
     triggerHaptic('impactLight')
-    if (onPress != null) onPress(wallet.id, tokenId, customAsset)
+    if (onPress != null) {
+      const maybePromise = onPress(wallet.id, tokenId, customAsset)
+      if (typeof (maybePromise as any)?.catch === 'function') {
+        ;(maybePromise as Promise<void>).catch((error: unknown) => {
+          showError(error)
+        })
+      }
+    }
```
swansontec: So for this, do:

```ts
const handlePress = useHandler(async () => {
  triggerHaptic('impactLight')
  if (onPress != null) await onPress(...)
}
```

This works because `await` handles promises & non-promises equally.

src/components/themed/WalletList.tsx
```diff
@@ -14,7 +14,7 @@ import { useDispatch, useSelector } from '../../types/reactRedux'
 import type { NavigationBase } from '../../types/routerTypes'
 import type { EdgeAsset, FlatListItem, WalletListItem } from '../../types/types'
 import { checkAssetFilter } from '../../util/CurrencyInfoHelpers'
-import { showError } from '../services/AirshipInstance'
+// import { showError } from '../services/AirshipInstance'
```
swansontec: Leftover.

swansontec (changes requested):

src/components/themed/WalletList.tsx
```diff
@@ -78,22 +79,23 @@ export function WalletList(props: Props): React.ReactElement {
...
+  const handlePress = useHandler(
+    async (walletId: string, tokenId: EdgeTokenId) => {
+      if (onPress != null) {
+        await onPress(walletId, tokenId)
+      } else {
+        await dispatch(
+          selectWalletToken({ navigation, walletId, tokenId })
+        ).catch((error: unknown) => {
+          showError(error)
+        })
```
swansontec: We don't actually need this catch block, because the `WalletListCurrencyRow` handles errors via `EdgeTouchableOpacity`.

PR #5760, opened 2025-09-26
Sam/ramps

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device


src/components/Main.tsx
```diff
@@ -457,6 +473,15 @@ const EdgeSellTabScreen: React.FC = () => {
...
       <SellStack.Screen
         name="pluginListSell"
+        component={RampCreateSellScene}
+        options={firstSceneScreenOptions}
+      />
+      <BuyStack.Screen
+        name="rampSelectOption"
+        component={RampSelectOptionScene}
+      />
+      <SellStack.Screen
```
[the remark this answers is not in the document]
Jon-edge: Valid

src/components/scenes/RampSelectOptionScene.tsx
```diff
@@ -0,0 +1,389 @@
...
+          // If either crypto amount is zero, sort that quote to the end
+          if (cryptoAmountA === 0 && cryptoAmountB === 0) return 0
+          if (cryptoAmountA === 0) return 1
+          return -1
+        }
+
+        const rateA = parseFloat(a.fiatAmount) / cryptoAmountA
+        const rateB = parseFloat(b.fiatAmount) / cryptoAmountB
+        return rateA - rateB
+      })
```
Jon-edge, line 133: Within each payment-type group, quotes are sorted by ascending fiat/crypto rate for both directions. For sells, the best rate is the highest fiat-per-crypto, so the sort should reverse. This affects which provider is selected by default in each card.

src/components/scenes/RampSelectOptionScene.tsx
```diff
@@ -0,0 +1,389 @@
...
+
+  const handleQuotePress = async (quote: RampQuoteResult): Promise<void> => {
+    try {
+      await quote.approveQuote({
+        coreWallet: rampQuoteRequest.wallet!
+      })
+    } catch (error) {
+      console.error('Failed to approve quote:', error)
+    }
+  }
```
Jon-edge, line 101: Should `showError`. Also the non-null assertion wallet! is safe given your creation flow, but this scene’s param type makes wallet optional. Consider guarding, or make wallet required in the route params to avoid future regressions.

src/components/buttons/PillButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+  ...StyleSheet.absoluteFillObject,
+  borderRadius: theme.rem(100)
+}))
+
+const Label = styled(EdgeText)(theme => ({
+  fontSize: theme.rem(0.75),
+  lineHeight: theme.rem(1.5),
+  flexShrink: 1,
+  minWidth: 0
+}))
```
Jon-edge, line 62: You didn't follow the subtask. Disallowed styled components here, a 100rem
bordered radius, missing `hitslop`. Theres a reason we try to be DRY in the GUI,
there are always a lot of considerations in the base components that have been
abstracted. It's also a lot faster to just be DRY, we can always split things up
later if needed. It's a lot more annoying to go the other way around (consolidating).

The recommendation is to create a <IconButton mini> or <MiniIconButton> that
heavily reuses existing `IconButton` just with different horizontal margins. We
make slight exceptions to spacing like this in things like matching to a different measured width of
the tertiary button for the sake of perceived consistency.

I doubt the original intent was to introduce another label size, either, this is really small.

Our wireframes are not guaranteed to be pixel-perfect.

src/plugins/ramps/utils/getRateFromRampQuoteResult.ts
```diff
@@ -0,0 +1,20 @@
...
+import { formatNumber } from '../../../locales/intl'
+import { lstrings } from '../../../locales/strings'
+import type { RampQuoteResult } from '../rampPluginTypes'
+
+export const getRateFromRampQuoteResult = (
+  quote: RampQuoteResult,
+  fiatCode: string
+): string => {
+  const bestRate = div(quote.fiatAmount, quote.cryptoAmount, 16)
+  const localeRate = formatNumber(toFixed(bestRate, 0, 2))
```
Jon-edge, line 12: Use our existing formatting. Refer to `useCryptoText` `useFiatText` for context

src/hooks/useRampQuotes.ts
```diff
@@ -0,0 +1,138 @@
...
+        return false
+      }
+      // Remove Google Pay quotes on iOS
+      if (Platform.OS === 'ios' && quote.paymentType === 'googlepay') {
+        return false
+      }
+      return true
+    })
+
+    // Sort by best rate (lowest fiat amount for same crypto amount)
```
Jon-edge, line 104: Need to factor in buy vs sell

src/components/themed/SceneHeaderUi4.tsx
```diff
@@ -43,6 +43,7 @@ const getStyles = cacheStyles((theme: Theme) => ({
   },
   titleContainer: {
     flexDirection: 'row',
+    gap: theme.rem(1),
```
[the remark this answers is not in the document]
samholmes: This doesn't appear to be necessary at all.

src/plugins/ramps/utils/getRateFromRampQuoteResult.ts
samholmes: What am I looking for in those implementations? I copied this from our previous fiat plugins exchange rate display logic. Those other hooks look to do a lot, so I don't know what I'm looking for in those components. Do you know what specifically is missing from this implementation of display logic?

Jon-edge (changes requested):
Remove styled components, rebase and address lint warnings

src/components/themed/SceneHeaderUi4.tsx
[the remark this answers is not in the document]
samholmes: Actually, my simulator was bugging out. Looks like if I remove the `gap` property here then the wide buttons would get too close. 

The header component isn't built for margin rem of 0.5 surrounding the child component at all, so gap actually makes sense here unless we modify header component. It would make sense to modify the header component so that way any component with 0.5 rem margin can be placed here and we can follow that contract throughout our codebase.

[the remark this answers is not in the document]
samholmes: I refactored the component to always have 0.5rem as per our standard. Took a lot of time on this, so I hope you like it.

src/components/scenes/RampSelectOptionScene.tsx
samholmes: This try-catch will be removed. We're not suppose to handle these unexpected errors here.

src/components/buttons/PillButton.tsx
samholmes: I've updated PillButton as you asked and introduced `hitSlop` (good idea).

Regarding the label size, it is just directly copied from the original design of this button. All of your concerns are valid too and make for a more suitable promotional experience for this component.

src/components/scenes/RampSelectOptionScene.tsx
samholmes: Good catch, though it's actually the best fiat-per-crypto for _buys and sells_ only if the source amount is denominated in crypto (because we want to find the highest fiat for our fixed crypto amount). This is an interesting miss on my testing, because I've been testing with fiat amounts mostly which is why I never noticed the sorting issue when testing buy/sell quoting! Thanks for finding this one.

samholmes: Good catch! Thanks for seeing this!

src/hooks/useMarginRemStyle.ts
```diff
@@ -4,6 +4,11 @@ import type { ViewStyle } from 'react-native'
 import { useTheme } from '../components/services/ThemeContext'
 
 export interface MarginRemProps {
+  /** Sets the alignment of the element within its container */
```
Jon-edge, line 7: Don't reinvent the wheel here, 'baseline' isn't used nor obvious what it even means. just do `FlexAlignType`

src/hooks/useMarginRemStyle.ts
```diff
@@ -25,12 +30,18 @@ export interface MarginRemProps {
...
-  'marginTop' | 'marginBottom' | 'marginLeft' | 'marginRight'
+  'alignSelf' | 'marginTop' | 'marginBottom' | 'marginLeft' | 'marginRight'
 >
 
-export const useMarginRemStyle = (props: MarginRemProps): MarginRemStyle => {
+export const useMarginRemStyle = (
+  props: MarginRemProps,
+  // TODO: Remove this prop once all designs using this prop have been updated.
+  /** @deprecated Your design should expect the component to have 0.5rem margins */
+  defaultRem: number = 0.5
```
Jon-edge, line 40: This shouldn't be configurable, should just be a constant

src/hooks/useMarginRemStyle.ts
```diff
@@ -40,26 +51,32 @@ export const useMarginRemStyle = (props: MarginRemProps): MarginRemStyle => {
...
-  )
+    }
+    // Low impact way of adding this prop to the collection (avoids huge snapshot diff)
+    if (alignSelf != null) {
+      return {
+        alignSelf
+      }
+    }
+    return style
+  }, [alignSelf, marginBottom, marginLeft, marginRight, marginTop])
```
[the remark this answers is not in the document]
Jon-edge: Not quite sure why this is complaining here, seems fine to me... Also fix the
typo in the comment

src/components/layout/Space.tsx
```diff
@@ -89,15 +89,18 @@ export const Space = React.memo((props: SpaceProps) => {
...
-    bottomRem,
-    leftRem,
-    rightRem,
-    topRem,
-    aroundRem,
-    horizontalRem,
-    verticalRem
-  })
+  const marginRemStyle = useMarginRemStyle(
+    {
```
Jon-edge, line 93: Honestly I'd really prefer this entire file to go away. Similar to styled
components, all we get here is obfuscation.

src/components/buttons/PillButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+        {label}
+      </EdgeText>
+    </EdgeTouchableOpacity>
+  )
+}
+
+const getStyles = cacheStyles((theme: ReturnType<typeof useTheme>) => ({
+  container: {
+    alignItems: 'center',
+    borderRadius: theme.rem(100),
```
Jon-edge, line 60: Fix borderRadius

src/components/buttons/PillButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const marginStyle = useMarginRemStyle(marginProps)
+
+  const theme = useTheme()
+  const styles = getStyles(theme)
+
+  return (
+    <EdgeTouchableOpacity
+      style={[styles.container, marginStyle]}
+      disabled={disabled}
+      hitSlop={theme.rem(0.5)}
```
Jon-edge, line 33: Just bake this in as a default to the component

src/components/scenes/RampCreateScene.tsx
```diff
@@ -0,0 +1,1088 @@
...
+    margin: theme.rem(0.5),
+    padding: theme.rem(1),
+    borderWidth: theme.thinLineWidth,
+    borderColor: theme.cardBorderColor,
+    gap: theme.rem(0.5)
+  },
+  regionButtonText: {
+    flexShrink: 1,
+    color: theme.primaryText,
+    fontSize: theme.rem(1.1),
```
Jon-edge, line 1032: For 99% of situations we should be using rem values divisible by 0.25. Not
really necessary to change because I already made a task to just do away with
this entire variant of the scene but just FYI

src/hooks/useMarginRemStyle.ts
samholmes: I know, it's only there to not break other places where useMarginRemStyle is used. See the comment.

src/components/layout/Space.tsx
samholmes: A cleanup for sure. I'll mark it as deprecated.

src/components/themed/SceneHeaderUi4.tsx
```diff
@@ -42,14 +42,13 @@ const getStyles = cacheStyles((theme: Theme) => ({
     paddingBottom: theme.rem(DEFAULT_MARGIN_REM)
```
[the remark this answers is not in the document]
samholmes: The change is to support the children components which should have 0.5rem. For this PR, the component is `PillButton`.

src/components/buttons/PillButton.tsx
Jon-edge: I misspoke earlier about `IconButton` and "margins," I meant `<EdgeButton mini>` should have been extended to have different "padding" in the icon-carrying configuration.

Like I said, **"Our wireframes are not guaranteed to be pixel-perfect."** so "copying directly" without designer verification is not the correct course of action when there's such an obvious close match in our app. This should have warranted discussion because it's similar to our other buttons but not quite. It's literally an `<EdgeButton mini>` with reduced horizontal padding I can assume so that it allows the icon to pleasingly match with the radius, but the font size is definitely questionable because we already established a font size for the same size container. 

Since we're late in the game I'll let this slide, but this kind of thing really needs to be double checked before robotically copying designs in the future. If designer input is not possible, we need to err towards our closer matches first in the future and make splits into new components later as fixes if necessary.

src/components/buttons/PillButton.tsx
samholmes: This was fixed in the background gradient style. I'll remove the redundant border-radius here.

src/hooks/useMarginRemStyle.ts
Jon-edge: What about `{aroundRem: 0}`?

src/hooks/useMarginRemStyle.ts
```diff
@@ -4,6 +4,11 @@ import type { ViewStyle } from 'react-native'
 import { useTheme } from '../components/services/ThemeContext'
 
 export interface MarginRemProps {
+  /** Sets the alignment of the element within its container */
+  // TODO: rename this hook to not be specific to margins because we want to
+  // our design system around these props
```
Jon-edge, line 9: Fix comment

src/components/buttons/PillButton.tsx
samholmes: Oooooh, yeah the `mini` variant of EdgeButton would be nice to re-use, but it unfortunately doesn't look like the design.. So yeah this is one of those things that needed to be discussed to achieve consensus on. The option I went with was to save discussion time and instead implement the component from SwapScene as the new PillButton component which looks like the design. Maybe we should consider a `pill` option for `EdgeButton` in the future, or maybe redesign `mini`? That's something to do as a team to consolidate our UI components to a nice tidy set.

src/hooks/useMarginRemStyle.ts
samholmes: That would override the defaultRem (should be called `fallbackRem` actually).

src/components/themed/SceneHeaderUi4.tsx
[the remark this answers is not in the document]
Jon-edge: The `titleContainer` originally only had horizontal and bottom margins. Since you moved the margin to the title, you are adding an extra 0.5rem above the title text now.

However you do it to accomodate `PillButton` you need to make sure the header looks the same.

src/components/buttons/PillButton.tsx
samholmes: Ideally, but I don't want to impact 266 places across 68 files.

src/components/scenes/RampCreateScene.tsx
samholmes: Yeah, this was one of those "match the designs and move on to bigger things" kind of work. I expected this scene was going to have follow up.

src/hooks/useMarginRemStyle.ts
Jon-edge: I mean pass `{aroundRem: 0}` at the callsite if the `marignRem` props are not provided. Then we don't need to even have this deprecation TODO to deal with in the common hook.

src/plugins/gui/providers/kadoProvider.ts
```diff
@@ -814,7 +814,7 @@ export const kadoProvider: FiatProviderFactory = {
                     if (!response.ok) {
                       const text = await response.text()
                       console.warn(`Error fetching kado blockchains: ${text}`)
-                      return allowedCurrencyCodes
+                      return
```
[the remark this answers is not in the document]
Jon-edge: Valid?

src/components/themed/SceneHeaderUi4.tsx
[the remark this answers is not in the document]
Jon-edge: Address this

src/plugins/gui/providers/kadoProvider.ts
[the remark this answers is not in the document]
samholmes: Not valid.

src/plugins/gui/providers/kadoProvider.ts
```diff
@@ -534,7 +534,7 @@ export const kadoProvider: FiatProviderFactory = {
           for (const asset of blockchain.associatedAssets) {
             const { isNative, address } = asset
 
-            if (!asset.rampProducts?.includes(direction)) continue
+            if (asset.rampProducts?.includes(direction) !== true) continue
```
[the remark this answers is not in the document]
samholmes: `!asset.rampProducts?.includes(direction)` means `asset.rampProducts?.includes(direction) === false || asset.rampProducts?.includes(direction) == null`, so the simplification of that is `asset.rampProducts?.includes(direction) !== true`.

src/components/themed/SwapInput.tsx
```diff
@@ -225,7 +225,7 @@ const SwapInputComponent = React.forwardRef<SwapInputCardInputRef, Props>(
             exchangeAmount: '',
             nativeAmount: '',
             fiatAmount: '',
-            fieldChanged: fieldNum ? 'fiat' : 'crypto'
+            fieldChanged: fieldNum !== 0 ? 'fiat' : 'crypto'
```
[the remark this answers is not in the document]
samholmes: What?... The whole point is to fix the lint error. `fieldNum !== 0` is what `fieldNum` meant before anyway.

src/util/DeepLinkParser.ts
```diff
@@ -268,7 +282,7 @@ function parseEdgeAppLink(url: URL<string>): DeepLink {
       // Parse data in format{{REWARDS:ethereum:a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48}}
       const dataMatch = /{{([^:]+):([^:]+)(?::([^}]+))?}}/.exec(data)
 
-      if (dataMatch) {
+      if (dataMatch != null) {
```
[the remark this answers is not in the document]
samholmes: It's just to fix the lint error.

Jon-edge (approved):

PR #5766, opened 2025-10-01
Add spinner when loading Send recipient addresses

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device


src/components/tiles/AddressTile2.tsx
```diff
@@ -82,6 +83,41 @@ export const AddressTile2 = React.forwardRef(
...
+    // Full-screen spinner driven by `loading` state:
+    const spinnerResolveRef = React.useRef<(() => void) | null>(null)
+    const spinnerTokenRef = React.useRef(0)
+    React.useEffect(() => {
+      if (loading && spinnerResolveRef.current == null) {
+        let resolveFn: () => void
+        const done = new Promise<void>(resolve => {
+          resolveFn = resolve
+        })
+        spinnerResolveRef.current = resolveFn!
```
swansontec, line 95: Why not:

```ts
const done = new Promise<void>(resolve => {
          spinnerResolveRef.current = resolve
})
```

Gets rid of the temporary variable.

src/components/tiles/AddressTile2.tsx
```diff
@@ -82,6 +83,41 @@ export const AddressTile2 = React.forwardRef(
...
+        const token = ++spinnerTokenRef.current
```
swansontec, line 96: Maybe call it "id" instead of "token". I thought you meant token like the crypto asset, which was confusing.

PR #5950, opened 2026-02-18
Add home screen long-press shortcuts with localized labels

JP0P:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---
  - https://app.asana.com/0/0/1211843114452190


android/app/src/main/res/xml/shortcuts.xml
```diff
@@ -0,0 +1,23 @@
...
+<shortcuts xmlns:android="http://schemas.android.com/apk/res/android">
+  <shortcut
+    android:shortcutId="contact_support"
+    android:enabled="true"
+    android:icon="@mipmap/ic_launcher"
+    android:shortcutShortLabel="@string/shortcut_contact_support_short"
+    android:shortcutLongLabel="@string/shortcut_contact_support_long">
+    <intent
+      android:action="android.intent.action.VIEW"
+      android:data="https://support.edge.app/hc/en-us?chat=open" />
```
samholmes: **Critical:** Android shortcuts will not work — https URLs are not opened in the browser.

The iOS implementation in `AppDelegate.swift` intercepts https/http URLs and opens them directly in Safari:

```swift
if url.scheme == "https" || url.scheme == "http" {
  UIApplication.shared.open(url)
  return true
}
```

However, Android has no equivalent native handling. The intent sends the URL to `MainActivity`, which passes it through React Native's `Linking` → `parseDeepLink` → `launchDeepLink`. The `'other'` case tries to parse it as a wallet URI, fails, and shows `"No wallet for this URI"` toast.

**Recommended fix (consistent with iOS):** Add native Android handling in `MainActivity` to open https/http shortcut URLs in the browser before they reach React Native. This matches the iOS pattern where https URLs bypass `parseDeepLink`.

---

**Alternative design consideration:** If the preference is to have `parseDeepLink` be the central handler for all intent URIs, the fix would instead be:

1. Update the `'other'` case in `DeepLinkingActions.tsx` to open unrecognized https URLs in the browser as a fallback (similar to `linkReferralWithCurrencies` which already does `if (parsed.type === 'other') await Linking.openURL(uri)`)
2. Update iOS `AppDelegate.swift` to pass https URLs through `RCTLinkingManager` instead of opening them directly

This would centralize URL handling logic in one place (the deep link infrastructure) rather than having platform-specific native handling.

android/app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,7 @@
 <resources>
     <string name="app_name">Edge</string>
+    <string name="shortcut_contact_support_short">Contact Support</string>
+    <string name="shortcut_contact_support_long">Contact support for help from a live agent</string>
+    <string name="shortcut_do_not_uninstall_short">⚠️ Save 2FA First!</string>
+    <string name="shortcut_do_not_uninstall_long">⚠️ Login requires 2FA & credentials!</string>
```
samholmes: **Critical:** Unescaped ampersand in XML.

In XML, `&` must be escaped as `&amp;`. An unescaped `&` can cause parse errors or unexpected behavior.

**Fix:**
```xml
<string name="shortcut_do_not_uninstall_long">⚠️ Login requires 2FA &amp; credentials!</string>
```

android/app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,7 @@
 <resources>
     <string name="app_name">Edge</string>
+    <string name="shortcut_contact_support_short">Contact Support</string>
+    <string name="shortcut_contact_support_long">Contact support for help from a live agent</string>
```
samholmes: **Warning:** Inconsistent copy between platforms.

| Platform | Contact Support (long) |
|----------|------------------------|
| Android | "Contact support for help from a live agent" |
| iOS | "Get help from our live support agents" |

Same feature with different wording can be confusing for localization and UX.

**Recommendation:** Pick one and use it on both platforms.

android/app/src/main/res/values/strings.xml
samholmes: **Warning:** Double emoji on Android vs single on iOS.

- **Android short:** "⚠️ Save 2FA First!"
- **Android long:** "⚠️ Login requires 2FA & credentials!"
- **iOS title:** "⚠️ Save 2FA First!"
- **iOS subtitle:** "Login requires 2FA & credentials!" (no emoji)

Android uses ⚠️ in both labels; iOS only in the title. The double emoji on Android can feel redundant.

**Recommendation:** Remove ⚠️ from the Android long label to match iOS.

android/app/src/main/res/xml/shortcuts.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shortcuts xmlns:android="http://schemas.android.com/apk/res/android">
+  <shortcut
```
samholmes, line 3: **Warning:** Shortcut order differs between platforms.

- **iOS:** 2FA warning first, Contact Support second
- **Android:** Contact Support first, 2FA warning second

If the order is intentional for platform UX, add a comment explaining the rationale. Otherwise, align for consistency.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased (develop)
 
+- added: Home screen long-press shortcut to contact support
+- added: Home screen long-press shortcut warning about data loss on uninstall
```
samholmes: **Warning:** "Data loss" wording may be misleading.

The shortcut points to "Getting a new phone" and emphasizes 2FA and credentials, not funds or wallet data. "Data loss" could imply loss of crypto.

**Suggestion:**
```
- added: Home screen long-press shortcut warning about 2FA and credentials needed after uninstall
```

ios/edge/Info.plist
```diff
@@ -57,6 +57,29 @@
...
+	<key>UIApplicationShortcutItems</key>
+	<array>
+		<dict>
+			<key>UIApplicationShortcutItemTitle</key>
+			<string>⚠️ Save 2FA First!</string>
+			<key>UIApplicationShortcutItemSubtitle</key>
+			<string>Login requires 2FA & credentials!</string>
+			<key>UIApplicationShortcutItemIconSymbolName</key>
+			<string>nosign</string>
+			<key>UIApplicationShortcutItemType</key>
```
samholmes, line 69: **Suggestion:** `UIApplicationShortcutItemType` is typically an identifier (e.g. `com.edge.app.contact_support`), not a full URL.

Storing URLs there works, but it mixes identity with data. Consider using a type like `contact_support` and passing the URL via `userInfo` for clearer separation. The `handleShortcutItem` function would then switch on the type and look up the URL.

android/app/src/main/res/xml/shortcuts.xml
```diff
@@ -0,0 +1,23 @@
...
+    android:shortcutId="contact_support"
+    android:enabled="true"
+    android:icon="@mipmap/ic_launcher"
```
samholmes: **Suggestion:** Both Android shortcuts use the same app launcher icon (`@mipmap/ic_launcher`), making them visually identical.

iOS uses distinct SF Symbols (`nosign` and `message.fill`) for each shortcut. Consider using different icons for Android too to improve shortcut recognition and UX.

ios/edge/AppDelegate.swift
```diff
@@ -72,9 +77,35 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    performActionFor shortcutItem: UIApplicationShortcutItem,
+    completionHandler: @escaping (Bool) -> Void
+  ) {
+    let handled = handleShortcutItem(shortcutItem)
+    completionHandler(handled)
+  }
+
+  private func handleShortcutItem(_ shortcutItem: UIApplicationShortcutItem) -> Bool {
+    guard let url = URL(string: shortcutItem.type) else { return false }
+
```
samholmes, line 100: **Suggestion:** `UIApplication.shared.open` has a completion handler that could help with debugging or analytics.

```swift
UIApplication.shared.open(url, options: [:]) { success in
  if !success {
    // Log or handle failure
  }
}
```

Not critical, but could be useful for future troubleshooting.

ios/edge/AppDelegate.swift
```diff
@@ -11,6 +11,7 @@ import UserNotifications
 class AppDelegate: UIResponder, UIApplicationDelegate {
   var window: UIWindow?
   var securityView: UIView?
+  private var pendingShortcutItem: UIApplicationShortcutItem?
```
samholmes: **Suggestion:** Consider adding a brief comment explaining why shortcut handling is deferred.

The pattern of storing `pendingShortcutItem` and handling it after React Native setup is useful but not immediately obvious. A comment like:

```swift
// Shortcut actions are deferred until React Native is fully initialized
private var pendingShortcutItem: UIApplicationShortcutItem?
```

would clarify the intent for future readers.

ios/edge/Info.plist
```diff
@@ -57,6 +57,29 @@
 	</array>
 	<key>CFBundleVersion</key>
 	<string>99999999</string>
+	<key>UIApplicationShortcutItems</key>
```
samholmes, line 60: **Localization:** These shortcut strings are hardcoded in English only.

The app supports 13 locales via React Native (`src/locales/strings/*.json`), but these native shortcuts won't be translated.

**Recommendation:** Use `InfoPlist.strings` for localization:

1. Create `ios/edge/en.lproj/InfoPlist.strings`:
```
SHORTCUT_UNINSTALL_TITLE = "⚠️ Save 2FA First!";
SHORTCUT_SUPPORT_TITLE = "Contact Support";
```

2. Create locale-specific `.lproj` folders (e.g., `es.lproj`, `de.lproj`) with translated strings.

3. Update `Info.plist` to use the keys instead of hardcoded strings.

Similarly for Android, add `values-es/strings.xml`, `values-de/strings.xml`, etc.

This could be deferred to a follow-up PR, but should at least be documented as a known limitation.

samholmes (changes requested):
## Review Summary

**2 Critical Issues:**
1. Android shortcuts won't open https URLs in browser (inconsistent with iOS implementation)
2. Unescaped `&` in Android strings.xml

**4 Warnings:** Inconsistent copy between platforms, double emoji on Android, different shortcut order, misleading "data loss" wording in changelog

**5 Suggestions:** UIApplicationShortcutItemType convention, Android icons, completion handler, code comments, localization support

See inline comments for details.

android/app/src/main/res/values/strings.xml
JP0P: Rookie mistake. Resolved.

android/app/src/main/res/values/strings.xml
JP0P: Done

JP0P: "Get help from our live support agents"

android/app/src/main/res/values/strings.xml
JP0P: Android displays one of the other (long v short) so 2 emojis are never observed, but iOS will display both the title and subtitle together.

CHANGELOG.md
JP0P: fixed. thanks.

android/app/src/main/res/xml/shortcuts.xml
JP0P: The order of strings.xml does not impact the visual order, but flipped them for consistency anyway

ios/edge/AppDelegate.swift
JP0P: nice. done.

ios/edge/Info.plist
JP0P: Fully addressed. Info.plist now uses symbolic keys (SHORTCUT_UNINSTALL_TITLE, SHORTCUT_UNINSTALL_SUBTITLE, SHORTCUT_SUPPORT_TITLE, SHORTCUT_SUPPORT_SUBTITLE) with translations hardcoded across all 12 supported locales:

- iOS: en.lproj/InfoPlist.strings + 11 locale-specific .lproj/InfoPlist.strings files, wired into the Xcode project via PBXVariantGroup
- Android: values/strings.xml + 11 locale-specific values-*/strings.xml directories

Both platforms automatically resolve the correct language at runtime based on device settings. No Crowdin dependency — translations are self-contained in this PR.

ios/edge/Info.plist
JP0P: Addressed. Types now use reverse-DNS identifiers (co.edgesecure.app.do_not_uninstall, co.edgesecure.app.contact_support) and URLs are passed via userInfo. handleShortcutItem reads the URL from shortcutItem.userInfo?["url"]

android/app/src/main/res/xml/shortcuts.xml
JP0P: Limited in which ic icons are available since we are compatible with older versions of Android, but the new ones are an info (i) icon and a question mark icon, respectively.

ios/edge/AppDelegate.swift
JP0P: Addressed. UIApplication.shared.open now uses the completion handler with failure logging via print().

android/app/src/main/res/xml/shortcuts.xml
JP0P: Addressed with the recommended approach. Added maybeOpenInBrowser() in MainActivity.kt that intercepts https/http URLs and opens them in the default browser before they reach React Native. Registered deep link hosts (deep.edge.app, dl.edge.app, return.edge.app) are excluded so they still route through React Native's Linking as before. Intent data is cleared after handling to prevent double-processing. This mirrors the iOS pattern where https URLs bypass parseDeepLink.

j0ntz (commented):
Two new findings — the Android browser-intent dispatch lacks exception handling, and the iOS shortcut launch logic can be simplified. Existing reviewer feedback looks addressed.

android/app/src/main/java/co/edgesecure/app/MainActivity.kt
```diff
@@ -38,6 +39,43 @@ class MainActivity : ReactActivity() {
...
+  private fun maybeOpenInBrowser(intent: Intent?): Boolean {
+    val data = intent?.data ?: return false
+    val scheme = data.scheme
+    if (scheme != "https" && scheme != "http") return false
+
+    val host = data.host
+    if (DEEP_LINK_HOSTS.contains(host)) return false
+
+    val browserIntent = Intent(Intent.ACTION_VIEW, data)
+    startActivity(browserIntent)
```
j0ntz, line 69: **Warning:** `startActivity(browserIntent)` can throw `ActivityNotFoundException` if no activity is registered to handle the `ACTION_VIEW` + `https` intent (e.g., device with no default browser, stripped-down enterprise builds, browsers disabled). The exception would propagate and crash the app when the user taps a shortcut.

Wrap in try-catch and fall through to `super.onNewIntent` / RN handling, or at minimum log and return `false` so the normal path runs:

```kotlin
return try {
  startActivity(browserIntent)
  true
} catch (e: ActivityNotFoundException) {
  false
}
```

ios/edge/AppDelegate.swift
```diff
@@ -72,7 +80,37 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
       launchOptions: launchOptions
     )
 
-    return true
+    if let shortcutItem = pendingShortcutItem {
+      _ = handleShortcutItem(shortcutItem)
+      pendingShortcutItem = nil
+    }
+
+    return !launchedFromShortcut
```
j0ntz, line 88: **Suggestion:** `pendingShortcutItem` is stored as an instance property but is set and consumed synchronously within `didFinishLaunchingWithOptions` — nothing is actually deferred across an async boundary, so the "Deferred until React Native is fully initialized" comment is misleading. `launchedFromShortcut` also re-derives the same launchOptions check.

Can be collapsed to a local, e.g.:

```swift
let shortcutItem = launchOptions?[.shortcutItem] as? UIApplicationShortcutItem
// ... init Sentry/Firebase/RN ...
if let shortcutItem = shortcutItem {
  _ = handleShortcutItem(shortcutItem)
}
return shortcutItem == nil
```

Removes the ivar, the bool, and the stale comment.

android/app/src/main/res/values/strings.xml
samholmes: **Warning:** English Android long label still mismatches the iOS subtitle.

- Android (this file): `Contact support for help from a live agent`
- iOS (`en.lproj/InfoPlist.strings`): `Get help from our live support agents`

Notably, all 11 non-English Android locales (`values-de`, `values-es`, `values-fr`, etc.) were translated against the iOS wording, so the English source string in this file is the only outlier in the localization tree.

**Recommendation:** Update to match the iOS subtitle:

```xml
    <string name="shortcut_contact_support_long">Get help from our live support agents</string>
```

android/app/src/main/res/values-de/strings.xml
```diff
@@ -0,0 +1,6 @@
+<resources>
+    <string name="shortcut_do_not_uninstall_short">⚠️ 2FA sichern!</string>
+    <string name="shortcut_do_not_uninstall_long">⚠️ Login erfordert 2FA &amp; Zugangsdaten!</string>
+    <string name="shortcut_contact_support_short">Support kontakt.</string>
```
samholmes: **Optional / nit:** Cross-platform short labels diverge for several locales (same class of inconsistency as samholmes' original review, but at the short-label level rather than the long-label level).

DE example here: Android `Support kontakt.` (note the trailing period — looks truncated/abbreviated) vs iOS `Support anfragen` in `de.lproj/InfoPlist.strings`.

Other divergences:

| Locale | Android short | iOS title |
|---|---|---|
| `de` | `Support kontakt.` | `Support anfragen` |
| `es` / `es-MX` | `Contactar soporte` | `Contactar ayuda` |
| `it` | `Contatta supporto` | `Contattaci` |
| `ru` | `Связаться` (just "Contact") | `Поддержка` |

Not blocking, but worth aligning so the same shortcut shows the same label across platforms within a given locale.

ios/edge/en.lproj/InfoPlist.strings
```diff
@@ -0,0 +1,7 @@
+/* Home screen shortcut: uninstall warning */
```
samholmes, line 1: **Optional / process question:** This PR ships hand/AI-translated copy for 12 locales without going through Crowdin (which the repo normally uses for translations).

A few of the translations look like they would benefit from a real translator pass — e.g. `Support kontakt.` (DE, looks truncated), `Связаться` (RU, just "Contact" rather than "Contact Support"). It also creates a parallel translation pipeline that future shortcut additions will need to remember to update.

**Suggestion (non-blocking):** confirm with the team whether native shortcut strings should be added to Crowdin going forward, or formally accept this as a one-off carve-out and document it (e.g. in a brief note in the README or a comment in the .lproj files) so future contributors don't unknowingly let it drift.

android/app/src/main/java/co/edgesecure/app/MainActivity.kt
```diff
@@ -38,6 +39,43 @@ class MainActivity : ReactActivity() {
...
+    return true
+  }
+
+  companion object {
+    private val DEEP_LINK_HOSTS = setOf(
+      "deep.edge.app",
+      "dl.edge.app",
+      "return.edge.app"
+    )
   }
```
samholmes, line 87: **Optional / nit:** `DEEP_LINK_HOSTS` is duplicated from `AndroidManifest.xml` lines 60–62 (`deep.edge.app`, `dl.edge.app`, `return.edge.app`).

If a new deep-link host is added to the manifest, this set must be updated too — with no compile-time check to catch a mismatch. Could be brittle as the deep-link surface evolves.

**Suggestion (non-blocking):** read the registered hosts at runtime from the manifest (e.g. via `PackageManager.getActivityInfo(componentName, GET_INTENT_FILTERS)` or an Android `<string-array>` resource referenced by both files), or at minimum add a brief comment here pointing readers at the manifest as the source of truth.

ios/edge.xcodeproj/project.pbxproj
```diff
@@ -42,6 +42,7 @@
 		81AB9BB82411601600AC10FF /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 81AB9BB72411601600AC10FF /* LaunchScreen.storyboard */; };
 		BF115CD26A29F1C032E30289 /* Pods_edge.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = E261C56FB78E202F218C2DCA /* Pods_edge.framework */; };
 		E469AC702DC43791006A2530 /* AdServices.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = E469AC6F2DC43791006A2530 /* AdServices.framework */; settings = {ATTRIBUTES = (Weak, ); }; };
+		F0A1B2C32E8A100000000001 /* InfoPlist.strings in Resources */ = {isa = PBXBuildFile; fileRef = F0A1B2C32E8A100000000003 /* InfoPlist.strings */; };
```
samholmes, line 45: **Optional / nit:** Hand-crafted Xcode UUIDs.

The 12 new file references and the variant group use sequential identifiers (`F0A1B2C32E8A100000000001` … `F0A1B2C32E8A10000000001A`) rather than randomly generated 24-char hex from Xcode. They're valid and work today, but:

1. They depart from the rest of the file's pattern, so subsequent edits via Xcode will mix two styles.
2. There is a small (extremely small) chance of a future collision if Xcode happens to generate a similar low-entropy ID.

**Suggestion (non-blocking):** open the project in Xcode and let it regenerate proper random IDs (drag the `InfoPlist.strings` variant group out and back in, or re-add the localizations through the Project → Info → Localizations panel). Not necessary for correctness.

ios/edge/AppDelegate.swift
```diff
@@ -72,7 +80,37 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+  }
+
+  private func handleShortcutItem(_ shortcutItem: UIApplicationShortcutItem) -> Bool {
+    guard let urlString = shortcutItem.userInfo?["url"] as? String,
+          let url = URL(string: urlString) else { return false }
+
+    if url.scheme == "https" || url.scheme == "http" {
+      UIApplication.shared.open(url, options: [:]) { success in
+        if !success {
+          print("Failed to open shortcut URL: \(urlString)")
```
samholmes: **Optional / nit:** Failures land in `print()` only, which means production failures to open a shortcut URL won't be captured for diagnostics — they'll vanish from any user device that isn't attached to Xcode.

**Suggestion (non-blocking):** route through Sentry (`SentrySDK.capture(message:)` or a breadcrumb) so the failure is observable in production. Likely a low-frequency event, but the only way to know is to instrument it.

Alternatively, if Sentry instrumentation is intentionally skipped here, a brief comment explaining why would help future readers.

samholmes (changes requested):

PR #5979, opened 2026-03-13
Fix contacts permission prompt timing

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Requirements

- [ ] Tested on iOS device

### Description

[Asana task](https://app.asana.com/0/0/1213394471711262/f)

Remove contacts modal on first receive. The contacts permission prompt no longer appears from passive thumbnail lookup when a user receives funds. It now appears only when the user opens a transaction list with named transactions or explicitly edits the payee field from transaction details.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are isolated to UI permission-prompt timing and removes a side-effect from a thumbnail hook, with no sensitive data handling beyond existing permission requests.
> 
> **Overview**
> Adjusts *when* the contacts permission modal is shown so it no longer triggers during passive contact thumbnail lookup (e.g. first receive).
> 
> Centralizes the “show modal + request OS permission” behavior into a new `promptForContactsPermission` thunk, calls it from `ContactListModal`, and adds a guard in `TransactionListScene` to only prompt when there are named transactions.
> 
> Simplifies `useContactThumbnail` to be a pure lookup over existing contacts state (no permission checks/modals), and updates the changelog entry to reflect the new prompt timing.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 43bea65cc3b8e2299e48f7b004f82688c7286428. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/components/scenes/TransactionListScene.tsx
+++ b/src/components/scenes/TransactionListScene.tsx
@@ -36,6 +36,7 @@ import { getWalletName } from '../../util/CurrencyWalletHelpers'
 import { calculateSpamThreshold, unixToLocaleDateTime } from '../../util/utils'
 import { SceneWrapper } from '../common/SceneWrapper'
 import { withWallet } from '../hoc/withWallet'
+import { promptForContactsPermission } from '../modals/ContactsPermissionModal'
 import { HeaderTitle } from '../navigation/HeaderTitle'
 import { cacheStyles, useTheme } from '../services/ThemeContext'
 import { ExplorerCard } from '../themed/ExplorerCard'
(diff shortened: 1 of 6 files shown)


src/components/scenes/TransactionListScene.tsx
samholmes, line 1: I don't understand this commit much. Why were the type deprecated. I guess maybe a little sidebar?

samholmes (changes requested):
I think this makes sense. But I think a short review-bar is all that's needed.

src/components/scenes/TransactionListScene.tsx
j0ntz: RN navigation 7 will no longer allow `NavigationBase` and requires typed navigators only, that's why it was deprecated in the first place. We need to prepare for the transition eventually so we must do it incrementally.

On second thought, though, this commit might not be a valid workaround. We might just need to actually fix everything such that we lift navigator props out of all components using `NavigationBase`

samholmes (approved):
Approved if we drop first commit because it's not certain if it's useful

PR #5984, opened 2026-03-19
Disable Migrate Wallets when no assets to migrate

j0ntz:
### Description

[Asana task](https://app.asana.com/0/0/1211114363135206/f)

Disable the "Migrate Wallets" button in Settings when there are no assets available to migrate, matching the existing "Restore Wallets" disabled pattern. Also filter compromised-key wallet IDs through the same eligibility check so the migrate modal is never shown for wallets with no migratable assets.

Extracts the migrate-wallet eligibility logic into a shared `getMigrateWalletItemList` utility to avoid duplicating the filtering rules across SettingsScene, WalletActions, and MigrateWalletSelectCryptoScene.

Closes https://app.asana.com/1/9976422036640/project/1200972350208303/task/1211114363135206

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [ ] Tested on iOS device

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes eligibility logic that gates the wallet migration flow and affects compromised-key handling/navigation; mistakes could incorrectly hide migration or preselect wrong wallets.
> 
> **Overview**
> Disables the Settings `Migrate Wallets` row when there are no eligible non-zero assets to migrate, matching the existing disabled behavior for `Restore Wallets`.
> 
> Extracts and centralizes migration eligibility into `getMigrateWalletItemList`, and updates migration-related scenes/actions to use a single `createWalletId` field (instead of an array) and to only show the compromised-key migration modal / preselect wallets that actually have migratable assets.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 18a3c1b907b86de0262665097dd2cb358b9ef86d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/actions/WalletActions.tsx
+++ b/src/actions/WalletActions.tsx
@@ -38,6 +38,7 @@ import type { NavigationBase } from '../types/routerTypes'
 import type { MapObject } from '../types/types'
 import { getCurrencyCode, isKeysOnlyPlugin } from '../util/CurrencyInfoHelpers'
 import { getWalletName } from '../util/CurrencyWalletHelpers'
+import { getMigrateWalletItemList } from '../util/getMigrateWalletItemList'
 import { fetchInfo } from '../util/network'
 
 export interface SelectWalletTokenParams {
@@ -450,13 +451,17 @@ export function checkCompromisedKeys(
       ))
     }
 
-    // If any walletId come back true show modal to go to migration scene with affected wallets preselected
-    if (exposedWalletIds.length > 0) {
-      const walletNames = exposedWalletIds.map(walletId =>
+    const migratableItems = getMigrateWalletItemList(currencyWallets)
+    const migratableWalletIds = exposedWalletIds.filter(id =>
+      migratableItems.some(item => item.createWalletId === id)
+    )
+
+    if (migratableWalletIds.length > 0) {
+      const walletNames = migratableWalletIds.map(walletId =>
         getWalletName(currencyWallets[walletId])
       )
       const response = await MigrateWalletsModal(walletNames)
-      exposedWalletIds.forEach(walletId => {
+      migratableWalletIds.forEach(walletId => {
         const { checked, modalShown } = securityCheckedWallets[walletId]
         securityCheckedWallets[walletId] = {
           checked,
@@ -466,7 +471,7 @@ export function checkCompromisedKeys(
 
       if (response === 'yes') {
         navigation.push('migrateWalletSelectCrypto', {
-          preSelectedWalletIds: exposedWalletIds
+          preSelectedWalletIds: migratableWalletIds
         })
       }
     }
--- a/src/actions/WalletActions.tsx
+++ b/src/actions/WalletActions.tsx
@@ -68,7 +68,7 @@ export function selectWalletToken({
     // Manually un-pause the wallet, if necessary:
     const wallet: EdgeCurrencyWallet = currencyWallets[walletId]
     if (wallet.paused && !isKeysOnlyPlugin(wallet.currencyInfo.pluginId))
-      wallet.changePaused(false).catch(error => {
+      wallet.changePaused(false).catch((error: unknown) => {
         showError(error)
       })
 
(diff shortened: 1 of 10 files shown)


src/actions/WalletActions.tsx
```diff
@@ -466,7 +475,7 @@ export function checkCompromisedKeys(
 
       if (response === 'yes') {
         navigation.push('migrateWalletSelectCrypto', {
-          preSelectedWalletIds: exposedWalletIds
+          preSelectedWalletIds: migratableWalletIds
```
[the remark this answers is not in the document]
j0ntz: Intentional. Exposed wallets with no migratable assets should be re-checked on future launches in case they later receive funds. The API calls (hash + POST) are lightweight and bounded by login frequency. Incrementing modalShown for non-migratable wallets would permanently suppress the check after 2 launches, defeating the purpose.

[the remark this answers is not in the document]
j0ntz: The new flow is:

1. App login → checkCompromisedKeys fires (after a 5s delay)
2. Hashes public keys → POSTs to info server → gets back "exposed" for compromised wallets
3. Our gate: migratableWalletIds is empty → the if (migratableWalletIds.length > 0) block is skipped entirely
4. Settings are written (unchanged modalShown/checked values) → function returns

No modal, no navigation, no toast, no UI indication at all. The user sees nothing. The only cost is the lightweight background API call on each login, which was already happening before our change (the old code also hashed + queried every login for unchecked wallets — it just showed a modal that the user couldn't act on).

src/util/getMigrateWalletItemList.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (pluginId === 'ripple') continue
+
+    const walletAssetList: MigrateWalletItem[] = []
+    for (const [tokenId, bal] of Array.from(balanceMap.entries())) {
+      if (zeroString(bal)) continue
+      if (tokenId != null && !enabledTokenIds.includes(tokenId)) continue
+      const currencyCode = getCurrencyCode(wallet, tokenId)
+      walletAssetList.push({
+        type: 'create',
+        createWalletIds: [walletId],
```
samholmes: `createWalletId: walletId` makes more sense

src/actions/WalletActions.tsx
```diff
@@ -450,13 +451,21 @@ export function checkCompromisedKeys(
       ))
     }
 
-    // If any walletId come back true show modal to go to migration scene with affected wallets preselected
-    if (exposedWalletIds.length > 0) {
-      const walletNames = exposedWalletIds.map(walletId =>
+    const migratableWalletIdSet = new Set(
```
samholmes: No need for `new Set` because we'd never have duplicate walletId in our output for getMigrateWalletItemList. Unless the agent has a valid reason why it might?

samholmes (approved):
approved with nits

src/util/getMigrateWalletItemList.ts
j0ntz: Renamed to `createWalletId: string` in `MigrateWalletItem` and updated all usages across migrate scenes to use the singular form with destructure-rename (`createWalletId: walletId`) where a local alias was needed.

src/actions/WalletActions.tsx
j0ntz: Removed `new Set` — now uses `.some()` directly on the list, since there are no duplicate walletIds to deduplicate.

j0ntz:
Both nits addressed: renamed `createWalletIds` to singular `createWalletId` in `MigrateWalletItem` and removed unnecessary `new Set`.

src/util/getMigrateWalletItemList.ts
```diff
@@ -0,0 +1,54 @@
...
+    const wallet = currencyWallets[walletId]
+    const {
+      currencyInfo: { pluginId, walletType },
+      balanceMap,
+      enabledTokenIds
+    } = wallet
+
+    if (isKeysOnlyPlugin(pluginId)) continue
+    if (SPECIAL_CURRENCY_INFO[pluginId].isAccountActivationRequired === true)
+      continue
```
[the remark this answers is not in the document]
j0ntz: Added optional chaining: `SPECIAL_CURRENCY_INFO[pluginId]?.isAccountActivationRequired`.

PR #6001, opened 2026-04-25
feat: add home screen long-press shortcuts via expo-quick-actions

samholmes:
Help users about to uninstall Edge recover their accounts: add a warning shortcut linking to the "save 2FA & credentials before uninstalling" support article, and a direct "Contact Support" link.

Use expo-quick-actions so titles/subtitles flow through lstrings and Crowdin instead of duplicating translations across 12 locales of InfoPlist.strings and values-*/strings.xml. Shortcuts are registered dynamically in JS, which means they only appear after the first app launch post-install — acceptable since the audience is established users considering uninstall, not fresh installs.

The `import 'expo-modules-core'` side effect in index.ts is non- obvious but required: expo-quick-actions reads
globalThis.expo.modules.ExpoQuickActions at module-evaluation time, and that global is installed lazily by expo-modules-core's NativeModule/SharedObject/EventEmitter module imports. Without one of those evaluating first the global stays empty and every expo-quick-actions call silently no-ops, including setItems, so the shortcuts never appear.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Requirements

- [x] Tested on iOS device

---

> [!NOTE]
> **Medium Risk**
> Moderate risk due to new native module dependency (`expo-quick-actions`) and iOS pod changes, plus startup-side-effect initialization in `index.ts` that could impact app boot if misconfigured.
> 
> **Overview**
> Adds home screen long-press quick actions via `expo-quick-actions`, registering two shortcuts ("⚠️ Save 2FA First!" and "Contact Support") that open Edge support URLs.
> 
> Wires a new `QuickActionsManager` into `App` to set items on startup and handle shortcut activation, and adds an `expo-modules-core` side-effect import in `index.ts` to ensure the module registry is initialized.
> 
> Updates iOS pods (`ExpoQuickActions`), `package.json`, lockfiles, localized strings, and the `CHANGELOG` entry to reflect the new shortcuts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4bec8ad020e658fdbed7ea4dc34867b3c73fdb93. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
  - https://app.asana.com/0/0/1211843114452190

### DIFF
--- a/src/components/services/QuickActionsManager.tsx
+++ b/src/components/services/QuickActionsManager.tsx
@@ -1,0 +1,61 @@
+import * as QuickActions from 'expo-quick-actions'
+import { useQuickActionCallback } from 'expo-quick-actions/hooks'
+import type * as React from 'react'
+import { Linking, Platform } from 'react-native'
+
+import { useAsyncEffect } from '../../hooks/useAsyncEffect'
+import { useHandler } from '../../hooks/useHandler'
+import { lstrings } from '../../locales/strings'
+import { config } from '../../theme/appConfig'
+import { showError } from './AirshipInstance'
+
+export const QuickActionsManager: React.FC = () => {
+  useAsyncEffect(
+    async () => {
+      const { quickActions } = config
+      if (quickActions == null) return
+      try {
+        await QuickActions.setItems([
+          {
+            id: 'do_not_uninstall',
+            title: lstrings.shortcut_do_not_uninstall_title,
+            subtitle: lstrings.shortcut_do_not_uninstall_subtitle,
+            icon: Platform.select({
+              ios: 'symbol:nosign',
+              default: 'prohibit'
+            }),
+            params: { url: quickActions.uninstallWarningUrl }
+          },
+          {
+            id: 'contact_support',
+            title: lstrings.shortcut_contact_support_title,
+            subtitle: lstrings.shortcut_contact_support_subtitle,
+            icon: Platform.select({
+              ios: 'symbol:message.fill',
+              default: 'message'
+            }),
+            params: { url: quickActions.contactSupportUrl }
+          }
+        ])
+      } catch (error: unknown) {
+        showError(error)
+      }
+    },
+    [],
+    'QuickActionsManager'
+  )
+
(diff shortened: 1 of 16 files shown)


src/locales/en_US.ts
```diff
@@ -2551,6 +2551,12 @@ const strings = {
     'Please ensure all details are correct before making the transfer.',
   // #endregion
 
+  // Home screen long-press shortcuts (expo-quick-actions)
+  shortcut_do_not_uninstall_title: '⚠️ Save 2FA First!',
+  shortcut_do_not_uninstall_subtitle: 'Login requires 2FA & credentials!',
+  shortcut_contact_support_title: 'Contact Support',
+  shortcut_contact_support_subtitle: 'Get help from our live support agents',
```
peachbits: The "live support agents" may not be accurate for white label versions of the app. Should be more generic, something like "Get help from our support team"

src/components/services/QuickActionsManager.tsx
```diff
@@ -0,0 +1,51 @@
...
+        }),
+        params: { url: CONTACT_SUPPORT_URL }
+      }
+    ]).catch(showError)
+  }, [])
+
+  const handleQuickAction = useHandler((action: QuickActions.Action) => {
+    const url = action.params?.url
+    if (typeof url === 'string') {
+      Linking.openURL(url).catch(showError)
```
peachbits: Instead of sending the user to the browser could we launch the app with a deeplink to a special  scene? I don't think it's appropriate to put links to our support pages with our branding in the quick actions of our white label apps. If we deeplink into a scene it can remain themed and use strings with the white label name in them

peachbits (changes requested):

src/components/services/QuickActionsManager.tsx
peachbits: if we're using urls from the config, this is no longer necessary

j0ntz (approved):

peachbits (approved):

PR #5976, opened 2026-03-11
Fix Infinite push transfer payload

j0ntz:
### CHANGELOG

Does this branch warrant an entry to the CHANGELOG?

- [x] Yes
- [ ] No

### Dependencies

none

### Description

[Asana task](https://app.asana.com/0/0/1213511119302370/f)

[Asana task](https://app.asana.com/1/9976422036640/project/1200972350208303/task/1213511119302370?focus=true)

#### Context
Infinite buy confirmations were creating a push-payment onramp transfer with source details attached, which Infinite flagged as invalid for this flow.

- Split Infinite transfer requests into explicit onramp and offramp payload variants.
- Remove source details from the Infinite buy/onramp transfer request.
- Drop the now-unused flow plumbing from the confirmation call path.

### Requirements

If you have made **any** visual changes to the GUI. Make sure you have:

- [ ] Tested on iOS device
- [ ] Tested on Android device
- [ ] Tested on small-screen device (iPod Touch)
- [ ] Tested on large-screen device (tablet)

---

> [!NOTE]
> **Medium Risk**
> Touches the Infinite buy/onramp transfer creation payload and confirmation workflow, so regressions could break bank-transfer purchases or deposit-instructions navigation. Changes are scoped to the Infinite ramp and related UI strings/layout.
> 
> **Overview**
> Fixes Infinite bank-transfer *buy* flow by sending explicit `ONRAMP`/`OFFRAMP` transfer request shapes and removing invalid source fields from the onramp payload.
> 
> Improves bank routing details handling by requiring deposit instructions (throws if missing), navigating back with `popToTop`, and expanding the routing details UI to show beneficiary name, reference/memo, and recipient address (including parsing a single-line address), with updated copy strings.
> 
> Tweaks ramp confirmation error handling so the slider always resets after confirm attempts.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 089545c072a338a66b3a7b54b65ffea135a1ed00. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #6005, opened 2026-04-30
Update Infinite ONRAMP for new headless transfer API

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

[Asana task](https://app.asana.com/1/9976422036640/project/1200972350208303/task/1213511119302370?focus=true)

Replaces #5976 (closed in favor of this branch).

#### Context

Infinite updated their headless transfer API to fix the deposit-matching problem the original PR set out to solve. Each ONRAMP create now provisions a per-customer Virtual Bank Account (VBA), so funds match a transfer record by destination instead of by memo. The new shape is:

- `POST /v1/headless/transfers` with `"type": "ONRAMP"` returns `id: null` plus a stable `depositAddressId` (e.g. `vba_…`); the `tfr_…` is created later, when funds arrive.
- Source is now `{ currency: "USD", network: "ACH" }`. The user's own bank account is no longer part of the request — they push funds to the VBA we get back.
- New `GET /v1/headless/transfers/deposit-address/:depositAddressId` re-fetches the same response shape for an existing VBA.
- OFFRAMP semantics are unchanged.

#### Changes

- Update `infiniteApiTypes` for the new response shape: nullable `id`, optional `depositAddressId`, optional `bankBeneficiaryName` on `sourceDepositInstructions`. Rebuild `InfiniteOnrampTransferRequest` to match the new `{ source, destination }` payload and add `getDepositAddress` to the `InfiniteApi` surface.
- Switch the Infinite buy quote to ACH push payment (`paymentType: 'ach'`) and skip `bankAccountWorkflow` on ONRAMP since Edge no longer collects the user's bank.
- Always reset the confirmation slider once `onConfirm` settles. Throw an explicit error from the ONRAMP create flow when deposit instructions are missing so the slider error path triggers instead of the slider hanging.
- Pop the Infinite ramp stack to the top after the user dismisses the bank routing details scene (replaces the old `goBack()`-by-one).
- Make the bank routing instructions and warning text wrap and stop them from shrinking under large system fonts.
- Surface the `bankBeneficiaryName` from the create response as a copy-able row above the bank name on the routing details scene.

#### Notes

The bank-info-fields commit from the prior PR (`287928f079`, address lines / deposit-message / split instruction strings) is intentionally not pulled — those fields aren't in the spec Infinite shared and need product confirmation before adding.

### Requirements

- [ ] Tested on iOS device

### Description

[Asana task](https://app.asana.com/0/0/1213511119302370/f)

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes the buy transfer request/response contract and navigation flow for the Infinite ramp, which can affect transfer creation and user completion paths.
> 
> **Overview**
> Updates Infinite ONRAMP to the new headless transfers API: buy now uses **ACH push** and no longer sends/collects the user’s bank account; `createTransfer` returns `id: null` with a `depositAddressId` and optional `bankBeneficiaryName`, and the API surface adds `getDepositAddress`.
> 
> Improves ramp UX robustness: the confirmation slider now always resets after `onConfirm` completes (including missing-instructions errors), bank routing details can show a copyable *Beneficiary Name*, wraps instruction/warning text for large fonts, and dismissing routing details now `popToTop()` instead of stepping back through confirmation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 43403f0ef32df7f6dd3c73f77b740d9e724c598e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/plugins/ramps/infinite/infiniteRampPlugin.ts
+++ b/src/plugins/ramps/infinite/infiniteRampPlugin.ts
@@ -703,7 +703,6 @@ export const infiniteRampPlugin: RampPluginFactory = (
                 freshQuote,
                 coreWallet,
                 bankAccountId: bankAccountResult.bankAccountId,
-                flow,
                 infiniteNetwork,
                 cleanFiatCode
               }
@@ -712,6 +711,11 @@ export const infiniteRampPlugin: RampPluginFactory = (
             if (!result.confirmed || result.transfer == null) {
               return
             }
+
+            // ONRAMP create returns id: null and a depositAddressId; OFFRAMP
+            // returns a tfr_… id. Fall back through both for analytics.
+            const orderId =
+              result.transfer.id ?? result.transfer.depositAddressId ?? ''
 
             // Log the success event based on direction
             if (request.direction === 'buy') {
@@ -726,7 +730,7 @@ export const infiniteRampPlugin: RampPluginFactory = (
                     exchangeAmount: freshQuote.target.amount.toString()
                   }),
                   fiatProviderId: pluginId,
-                  orderId: result.transfer.id
+                  orderId
                 }
               })
             } else {
@@ -741,7 +745,7 @@ export const infiniteRampPlugin: RampPluginFactory = (
                     exchangeAmount: freshQuote.source.amount.toString()
                   }),
                   fiatProviderId: pluginId,
-                  orderId: result.transfer.id
+                  orderId
                 }
               })
             }
--- a/src/plugins/ramps/infinite/infiniteRampPlugin.ts
+++ b/src/plugins/ramps/infinite/infiniteRampPlugin.ts
@@ -45,8 +45,8 @@ import { kycWorkflow } from './workflows/kycWorkflow'
 const pluginId = 'infinite'
 const partnerIcon = `${EDGE_CONTENT_SERVER_URI}/infinite.png`
 const pluginDisplayName = 'Infinite'
-// Extend as more become supported:
-const DEFAULT_PAYMENT_TYPE: FiatPaymentType = 'wire'
+// ACH push payment is the only supported buy rail today.
+const BUY_PAYMENT_TYPE: FiatPaymentType = 'ach'
 
 // Storage keys
 const INFINITE_PRIVATE_KEY = 'infinite_auth_private_key'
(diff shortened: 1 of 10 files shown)


src/plugins/ramps/infinite/workflows/confirmationWorkflow.ts
```diff
@@ -87,42 +88,54 @@ export const confirmationWorkflow = async (
...
+              `Transfer ${
+                transfer.id ?? transfer.depositAddressId ?? ''
+              } created but deposit instructions are missing`
+            )
           }
 
+          navigationFlow.navigate('rampBankRoutingDetails', {
+            bank: {
+              name: instructions.bankName,
+              accountNumber: instructions.bankAccountNumber ?? '',
```
samholmes, line 103: This can silently render unusable ACH instructions. The cleaner allows `bankAccountNumber` and `bankRoutingNumber` to be null, but this maps either missing value to an empty string and the bank details screen still renders copyable Account/Routing rows. Since the user cannot complete an ACH push without both values, we should treat missing account or routing numbers the same way we treat missing `bankName` / `amount` and fail before navigating.

src/plugins/ramps/infinite/infiniteRampPlugin.ts
```diff
@@ -196,7 +196,7 @@ export const infiniteRampPlugin: RampPluginFactory = (
         }
 
         // Global constraints pre-check
-        const paymentTypes: FiatPaymentType[] = [DEFAULT_PAYMENT_TYPE]
+        const paymentTypes: FiatPaymentType[] = [BUY_PAYMENT_TYPE]
```
samholmes, line 199: Since the buy flow is now hardcoded to ACH, support needs to verify that Infinite actually supports ACH for the selected country, not just that the country has any on-ramp payment method. Right now the later `paymentMethods.length === 0` check would still allow a country whose only method is `wire`, then `createTransfer` sends `network: 'ACH'`. Please check `country.supportedPaymentMethods.onRamp` for `ach` case-insensitively in both `checkSupport` and `fetchQuotes`.

samholmes (changes requested):
Requesting changes for two runtime support/instruction issues in the Infinite ACH buy flow. Targeted ESLint on the touched files passed with only existing NavigationBase deprecation warnings; full TypeScript check is blocked locally by unrelated missing zcashname-sdk types in src/util/zns.ts.

PR #5988, opened 2026-03-27
Feat: Changelly integration

changelly:
### CHANGELOG

Does this branch warrant an entry to the CHANGELOG?

- [x] Yes
- [ ] No

### Dependencies

PR's contains no dependencies

### Requirements

If you have made **any** visual changes to the GUI. Make sure you have:

- [ ] Tested on iOS device
- [ ] Tested on Android device
- [ ] Tested on small-screen device (iPod Touch)
- [ ] Tested on large-screen device (tablet)

PR #6028, opened 2026-06-10
Feat: Changelly integration

peachbits:
Replaces #5988, reordered so the obfuscation utility lands first:

1. **Add asObfuscatedString cleaner** — a codec that decodes strings stored as arrays of XOR-masked char codes, plus `scripts/obfuscateString.ts` to generate the arrays.
2. **Feat: Changelly integration** — enables the changelly swap plugin with `CHANGELLY_INIT`: an apiKey stored XOR-obfuscated in env.json plus a plaintext `partnerId` (`edge-app`), both required with no defaults, passed to the plugin via initOptions (see EdgeApp/edge-exchange-plugins#471).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> New third-party swap integration and env credential handling; impact is limited to swap quoting/execution when config is present, not core auth or funds custody.
> 
> **Overview**
> Adds **Changelly** as a centralized swap provider, gated by new `CHANGELLY_INIT` in env config (`apiKey` via XOR-masked char arrays, plaintext `partnerId`; both optional with no defaults so incomplete config disables the plugin). The provider is registered in `swapPlugins` and gets the usual UI wiring (CDN icon, KYC/terms links in `SwapVerifyTermsModal`).
> 
> Introduces **`asObfuscatedString`** (cleaner + `scripts/obfuscateString.ts`) so sensitive env fields can be stored as obfuscated arrays instead of plain strings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c377f4eab60465e6942b365330410082407fc2a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
  - https://app.asana.com/0/0/1213981191982797

### DIFF
--- a/src/envConfig.ts
+++ b/src/envConfig.ts
@@ -19,6 +19,7 @@ import { asInitOptions as asPaybisInitOptions } from './plugins/ramps/paybis/pay
 import { asInitOptions as asRevolutInitOptions } from './plugins/ramps/revolut/revolutRampTypes'
 import { asInitOptions as asSimplexInitOptions } from './plugins/ramps/simplex/simplexRampTypes'
 import { asBase16 } from './util/cleaners/asHex'
+import { asObfuscatedString } from './util/cleaners/asObfuscatedString'
 
 function asNullable<T>(cleaner: Cleaner<T>): Cleaner<T | null> {
   return function asNullable(raw) {
@@ -272,6 +273,16 @@ export const asEnvConfig = asObject({
   CHANGEHERO_INIT: asCorePluginInit(
     asObject({
       apiKey: asOptional(asString, '')
+    }).withRest
+  ),
+  CHANGELLY_INIT: asCorePluginInit(
+    asObject({
+      // Arrays of XOR-masked char codes; see asObfuscatedString.
+      // No fallback values: the plugin itself refuses to load when either
+      // member is missing, so a partial entry disables Changelly instead of
+      // supplying placeholder credentials.
+      apiKey: asOptional(asObfuscatedString),
+      partnerId: asOptional(asString)
     }).withRest
   ),
   COREUM_INIT: asCorePluginInit(asBoolean),
--- a/src/util/corePlugins.ts
+++ b/src/util/corePlugins.ts
@@ -92,6 +92,7 @@ export const swapPlugins = {
   // Centralized Swaps
   changehero: ENV.CHANGEHERO_INIT,
   changenow: ENV.CHANGE_NOW_INIT,
+  changelly: ENV.CHANGELLY_INIT,
   exolix: ENV.EXOLIX_INIT,
   godex: ENV.GODEX_INIT,
   lifi: ENV.LIFI_INIT,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased (develop)
 
+- added: Changelly swap provider
 - added: NYM swap provider (`nymswap`)
 - added: Optional Bridgeless swap referral id via the `BRIDGELESS_INIT` env config, passed through to the swap plugin.
 - changed: Route maestro test builds to a dedicated Zealot channel so they no longer appear in the production release list.
--- a/scripts/obfuscateString.ts
+++ b/scripts/obfuscateString.ts
@@ -1,0 +1,16 @@
+// Generates the obfuscated char-code array to paste into env.json for
+// fields cleaned by asObfuscatedString.
+//
+// Usage:
+//   node -r sucrase/register scripts/obfuscateString.ts <plaintext...>
+
+import { wasObfuscatedString } from '../src/util/cleaners/asObfuscatedString'
+
+const plaintext = process.argv.slice(2).join(' ')
+
+if (plaintext === '') {
+  console.error('Usage: obfuscateString.ts <plaintext>')
+  process.exit(1)
+}
+
+console.log(JSON.stringify(wasObfuscatedString(plaintext)))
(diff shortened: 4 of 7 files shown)


src/envConfig.ts
```diff
@@ -266,6 +267,13 @@ export const asEnvConfig = asObject({
       apiKey: asOptional(asString, '')
     }).withRest
   ),
+  CHANGELLY_INIT: asCorePluginInit(
+    asObject({
+      // Arrays of XOR-masked char codes; see asObfuscatedString.
+      apiKey: asOptional(asObfuscatedString, ''),
+      secret: asOptional(asObfuscatedString, '')
```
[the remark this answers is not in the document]
peachbits: Resolved by removing `secret` from `CHANGELLY_INIT` entirely: Changelly's relay auth needs only the API key (`X-Auth` is `base64(apiKey:timestamp)`), so no partner secret ships in the app. The remaining `apiKey` keeps the `asObfuscatedString` encoding purely as an anti-grep measure ("Not a secret" per the cleaner's comment), not as a security control — and it is now required alongside a plaintext `partnerId`, with no empty-string defaults, so unconfigured builds simply run without the plugin.

src/envConfig.ts
```diff
@@ -266,6 +267,13 @@ export const asEnvConfig = asObject({
       apiKey: asOptional(asString, '')
     }).withRest
   ),
+  CHANGELLY_INIT: asCorePluginInit(
+    asObject({
+      // Arrays of XOR-masked char codes; see asObfuscatedString.
+      apiKey: asObfuscatedString,
+      partnerId: asString
```
j0ntz: CHANGELLY_INIT requires both apiKey and partnerId with no asOptional default, unlike every sibling centralized swap entry (CHANGE_NOW_INIT, CHANGEHERO_INIT, EXOLIX_INIT, GODEX_INIT, LETSEXCHANGE_INIT), all of which wrap apiKey in asOptional(asString, ''). ENV = asEnvConfig(ENV_JSON) runs eagerly at import time in src/env.ts, so a partial env.json entry for CHANGELLY_INIT (e.g. {} or a missing partnerId, the placeholder pattern used to enable other swap providers) throws and crashes the entire app at startup instead of just disabling Changelly. Consider asOptional for apiKey and partnerId to match sibling behavior.

src/util/corePlugins.ts
```diff
@@ -92,6 +92,7 @@ export const swapPlugins = {
   // Centralized Swaps
   changehero: ENV.CHANGEHERO_INIT,
   changenow: ENV.CHANGE_NOW_INIT,
+  changelly: ENV.CHANGELLY_INIT,
```
j0ntz, line 95: swapPlugins.changelly wires ENV.CHANGELLY_INIT in, but package.json still pins edge-exchange-plugins at ^2.48.3 (latest published is 2.49.0). Neither release contains a changelly plugin; the implementation looks to be in the still-open [EdgeApp/edge-exchange-plugins#471](https://github.com/EdgeApp/edge-exchange-plugins/pull/471) (same branch name, opened the same day). Until that dependency is bumped to a release registering 'changelly', this entry is inert: edge-core-js won't recognize the plugin id regardless of CHANGELLY_INIT config.

src/envConfig.ts
peachbits: Switched both members to asOptional (d6163cb64) so a partial CHANGELLY_INIT no longer throws at startup — but deliberately without the sibling '' defaults: absent members stay undefined, the plugin factory's required-string cleaner then throws, and edge-core-js catches per-plugin, so a partial entry disables Changelly instead of crashing the app or sending placeholder credentials to the relay.

src/util/corePlugins.ts
peachbits: Correct — the entry is inert until edge-exchange-plugins ships changelly and the dep moves past the current pin, and that sequencing is intentional: EdgeApp/edge-exchange-plugins#471 lands first, then a release, then the dep bump activates this. An unregistered plugin id in the init map is ignored by edge-core-js, so shipping the wiring early is harmless.

src/envConfig.ts
```diff
@@ -266,6 +267,16 @@ export const asEnvConfig = asObject({
       apiKey: asOptional(asString, '')
     }).withRest
   ),
+  CHANGELLY_INIT: asCorePluginInit(
+    asObject({
+      // Arrays of XOR-masked char codes; see asObfuscatedString.
+      // No fallback values: the plugin itself refuses to load when either
+      // member is missing, so a partial entry disables Changelly instead of
+      // supplying placeholder credentials.
+      apiKey: asOptional(asObfuscatedString),
```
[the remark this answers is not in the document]
peachbits: Intentional: the char-code-array format is the point of asObfuscatedString (keys must not sit greppable in env.json), it is documented on these lines, and scripts/obfuscateString.ts generates the arrays. A wrong-typed member failing env load at dev time is the existing repo-wide behavior for every cleaner in this file (e.g. a numeric CHANGE_NOW_INIT.apiKey throws identically) — accepting plaintext here would silently defeat the obfuscation.

src/util/corePlugins.ts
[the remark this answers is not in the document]
peachbits: Known and intentional sequencing (see the earlier thread on this line): EdgeApp/edge-exchange-plugins#471 lands first, then a release, then the dep bump activates this entry. edge-core-js ignores init entries for unregistered plugin ids, so the wiring is inert and harmless until then.

j0ntz (approved):

PR #6056, opened 2026-06-24
Fix MoonPay "Send with Edge" sell link

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Requirements

- [x] Tested on iOS device

### Description

[Asana task](https://app.asana.com/0/0/1209649066660320/f)

Asana: https://app.asana.com/1/9976422036640/project/1207384676342554/task/1209649066660320

Fixes the MoonPay "Send with Edge" sell link. When a user starts a MoonPay sell order but has not completed the crypto deposit, MoonPay shows a "complete payment" button (in their email and in the `buy.moonpay.com/trade_history` page) that reflects back the deposit redirect URL our app hands MoonPay. That button dead-ended on a static "Redirecting to Payment..." page, so the user could not finish the deposit (the reported lead case is an XRP sell that also needs a destination tag).

**Root cause (regression analysis).** The redirect URL we pass to MoonPay is `RETURN_URL_PAYMENT` in `src/plugins/gui/providers/common.ts`. It has been `https://edge.app/redirect/payment/` since it was introduced (Oct 2023); it was never a `deep.edge.app` value, so this is not a value regression. The apex `edge.app` host is not claimed as a universal link on iOS (`ios/edge/edge.entitlements`) or Android (`AndroidManifest.xml`) - only `deep.edge.app`, `dl.edge.app`, and `return.edge.app` are. The constant was designed as an in-app webview interception sentinel (`uri.startsWith(RETURN_URL_PAYMENT)`), which works in-app but cannot open the app from an external click. The buy redirects already use `deep.edge.app`; only the sell payment redirect used the unclaimed apex.

**Fix.**
- Point `RETURN_URL_PAYMENT` at the claimed `deep.edge.app` host, matching the buy redirects. In-app interception is host-agnostic (prefix `startsWith`), so the existing webview flows (MoonPay and Paybis, ramp and legacy) are unaffected.
- Parse the `redirect/payment` deep link into a new `paymentRedirect` link type and route it to the Send scene, resolving the wallet from the base currency code and pre-filling the deposit address, amount, and destination tag/memo (via `uniqueIdentifier`). Both the `deep.edge.app` and legacy `edge.app` hosts are parsed.

No new user-facing strings; reuses the existing Send scene and wallet picker.

**Testing.**
- Unit tests for the parser (real-world MoonPay URL, XRP with destination tag, `deep.edge.app` host, and missing-param browser fallback).
- `verify-repo.sh` (eslint + jest) passes; `tsc` clean.
- Driven on the iOS simulator: firing `edge://redirect/payment/?baseCurrencyCode=btc&baseCurrencyAmount=0.001&depositWalletAddress=bc1q...` opens the wallet picker (all BTC wallets) and then the Send scene pre-filled with the deposit address and 0.001 BTC. Screenshots below. The slider is not confirmed (no broadcast).

---

> [!NOTE]
> **Medium Risk**
> External deep links now drive the Send flow with provider-supplied addresses and amounts; guards reduce bad pre-fills, but users still must confirm the transaction and ticker collisions rely on wallet picker disambiguation.
> 
> **Overview**
> Fixes **MoonPay "Send with Edge"** (and similar ramp sell flows) when users tap **complete payment** links outside the app—those links used to hit an unclaimed `edge.app` host and dead-end instead of opening Edge.
> 
> **Redirect URLs** for payment, success, fail, and cancel now use the claimed **`deep.edge.app`** host (matching buy flows). **`isReturnUrl`** centralizes WebView interception so both **`deep.edge.app`** and legacy **`edge.app`** URLs still work for in-flight orders.
> 
> **Deep linking:** `/redirect/payment/` parses to a new **`paymentRedirect`** type with deposit address, optional amount, and memo/tag; invalid or blank required params become a **no-op** instead of errors or browser fallback. **`success` / `fail` / `cancel`** open the app as no-ops. Handling resolves assets by currency symbol (native + tokens, including custom), runs the wallet picker when needed, converts amounts safely, then opens **Send** pre-filled via the existing URI path.
> 
> Unit tests cover parsing, return-URL matching, and edge cases (XRP tag, bad amounts, missing params).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 77117ab049fb58c68dde8746e14f15ce347681f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/actions/DeepLinkingActions.tsx
+++ b/src/actions/DeepLinkingActions.tsx
@@ -1,3 +1,4 @@
+import { mul } from 'biggystring'
 import type { EdgeParsedUri, EdgeTokenId } from 'edge-core-js'
 import * as React from 'react'
 import { Linking } from 'react-native'
@@ -21,6 +22,7 @@ import {
   fiatProviderDeeplinkHandler
 } from '../plugins/gui/fiatPlugin'
 import { rampDeeplinkManager } from '../plugins/ramps/rampDeeplinkHandler'
+import { getExchangeDenom } from '../selectors/DenominationSelectors'
 import { config } from '../theme/appConfig'
 import type { DeepLink } from '../types/DeepLinkTypes'
 import type { Dispatch, RootState, ThunkAction } from '../types/reduxTypes'
@@ -232,6 +234,84 @@ async function handleLink(
         hideScamWarning: false
       })
       break
+
+    case 'paymentRedirect': {
+      // A provider sell-completion redirect (e.g. MoonPay "Send with Edge").
+      // Resolve the provider's base currency code to candidate assets, then
+      // open the Send scene pre-filled with the deposit address, amount, and
+      // destination tag / memo so the user can finish the sell order.
+      const { currencyCode, depositAddress, amount, addressTag } = link
+
+      // Collect every native AND token asset that shares the symbol, and let
+      // the user disambiguate via the wallet picker. A provider sell can be a
+      // token whose ticker collides with another chain's native asset (the
+      // provider disambiguates by network metadata we do not get here), so we
+      // must not exclude token matches when a native one also matches. Iterate
+      // `allTokens` (builtin + user-added custom tokens) rather than
+      // `builtinTokens`: pickWallet matches wallets by their enabled token ids,
+      // which include custom tokens, so a sell of a custom token would
+      // otherwise resolve zero assets and wrongly report "no wallet".
+      const symbol = currencyCode.split('_')[0].toUpperCase()
+      const assets: EdgeAsset[] = []
+      for (const pluginId of Object.keys(account.currencyConfig)) {
+        const currencyConfig = account.currencyConfig[pluginId]
+        if (currencyConfig.currencyInfo.currencyCode.toUpperCase() === symbol) {
+          assets.push({ pluginId, tokenId: null })
+        }
+        const { allTokens } = currencyConfig
+        for (const tokenId of Object.keys(allTokens)) {
+          if (allTokens[tokenId].currencyCode.toUpperCase() === symbol) {
+            assets.push({ pluginId, tokenId })
+          }
+        }
+      }
+
+      if (assets.length === 0) {
+        showToast(lstrings.alert_deep_link_no_wallet_for_uri)
+        break
+      }
+
+      const result = await pickWallet({
+        account,
+        assets,
+        navigation,
+        showCreateWallet: true
+      })
+      if (result?.type !== 'wallet') break
+      const { walletId, tokenId } = result
+      const wallet = account.currencyWallets[walletId]
+      if (wallet == null) break
+
+      // A token-metadata refresh race could leave the picked tokenId
+      // unresolvable, in which case getExchangeDenom silently returns a
+      // multiplier of '1' and mul() would treat a decimal amount as already
+      // native (a wildly wrong send amount). Abort with a toast rather than
+      // pre-filling a wrong amount.
+      if (
+        amount != null &&
+        tokenId != null &&
+        wallet.currencyConfig.allTokens[tokenId] == null
+      ) {
+        showToast(lstrings.alert_deep_link_no_wallet_for_uri)
+        break
+      }
+      const nativeAmount =
+        amount != null
+          ? mul(
+              amount,
+              getExchangeDenom(wallet.currencyConfig, tokenId).multiplier
+            )
+          : undefined
+
+      const parsedUri: EdgeParsedUri = {
+        publicAddress: depositAddress,
+        nativeAmount,
+        uniqueIdentifier: addressTag,
+        tokenId
+      }
+      await dispatch(handleWalletUris(navigation, wallet, parsedUri))
+      break
+    }
 
     case 'price-change': {
       const { pluginId, body } = link
--- a/src/plugins/gui/providers/common.ts
+++ b/src/plugins/gui/providers/common.ts
@@ -7,10 +7,38 @@ import {
   type FiatProviderSupportedRegions
 } from '../fiatProviderTypes'
 
-export const RETURN_URL_SUCCESS = 'https://edge.app/redirect/success/'
-export const RETURN_URL_FAIL = 'https://edge.app/redirect/fail/'
-export const RETURN_URL_CANCEL = 'https://edge.app/redirect/cancel/'
-export const RETURN_URL_PAYMENT = 'https://edge.app/redirect/payment/'
+// Ramp redirect base hosts. Derive both the outbound `RETURN_URL_*` constants
+// and the inbound `isReturnUrl` matcher from these two so a host migration is a
+// single edit that can never leave the hand-off and match sides out of sync.
+// `deep.edge.app` is claimed as a universal link on iOS and Android (the apex
+// `edge.app` is NOT), matching the buy redirects. The legacy apex host lingers
+// only on orders created before the switch.
+const CLAIMED_REDIRECT_HOST = 'https://deep.edge.app/redirect/'
+const LEGACY_REDIRECT_HOST = 'https://edge.app/redirect/'
+
+// These URLs are handed to the ramp providers as sell redirect / callback URLs
+// and can be reflected back to the user outside the app (provider email or
+// order-history "complete payment" button), so they must live on the claimed
+// host: an external click opens the app and the deep-link parser routes it
+// (PAYMENT to the pre-filled Send scene; the terminal states to a harmless
+// no-op). We always hand providers the claimed-host URL going forward.
+export const RETURN_URL_SUCCESS = `${CLAIMED_REDIRECT_HOST}success/`
+export const RETURN_URL_FAIL = `${CLAIMED_REDIRECT_HOST}fail/`
+export const RETURN_URL_CANCEL = `${CLAIMED_REDIRECT_HOST}cancel/`
+export const RETURN_URL_PAYMENT = `${CLAIMED_REDIRECT_HOST}payment/`
+
+// Match a ramp redirect URL as it appears inside the provider WebView. Orders
+// created before the host switch still carry the legacy apex `edge.app` host —
+// MoonPay persists the payment redirect server-side and can resurface it days
+// later — so the in-WebView interceptors match BOTH the claimed and legacy hosts
+// to keep catching a legacy order resumed after an app update. Use this in place
+// of a raw `startsWith`/`===` against a single constant.
+const RETURN_URL_REDIRECT_HOSTS = [CLAIMED_REDIRECT_HOST, LEGACY_REDIRECT_HOST]
+export const isReturnUrl = (
+  uri: string,
+  kind: 'payment' | 'success' | 'fail' | 'cancel'
+): boolean =>
+  RETURN_URL_REDIRECT_HOSTS.some(host => uri.startsWith(`${host}${kind}/`))
 
 export const NOT_SUCCESS_TOAST_HIDE_MS = 5000
 
--- a/src/util/DeepLinkParser.ts
+++ b/src/util/DeepLinkParser.ts
@@ -14,6 +14,7 @@ import {
   type PromotionLink
 } from '../types/DeepLinkTypes'
 import type { AppParamList } from '../types/routerTypes'
+import type { UriQueryMap } from '../types/WebTypes'
 import { parseQuery, stringifyQuery } from './WebUtils'
 
 /**
@@ -244,6 +245,17 @@ function parseEdgeProtocol(url: URL<string>): DeepLink {
       }
     }
 
+    case 'redirect': {
+      // Provider ramp redirect (e.g. MoonPay "Send with Edge"). All ramp
+      // redirect URLs now live on the claimed `deep.edge.app` host, which
+      // normalizes to this `edge://` path. parseRedirectSection resolves each
+      // section (payment -> pre-filled Send scene; terminal states and any
+      // malformed payment link -> no-op) identically for the apex host below.
+      const link = parseRedirectSection(pathParts[0], parseQuery(url.query))
+      if (link != null) return link
+      break
+    }
+
     case 'recovery': {
       // The new & improved format stores the token as a fragment:
       if (url.hash != null && url.hash !== '') {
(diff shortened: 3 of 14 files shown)


j0ntz:
## 📸 Test evidence (iOS sim)

**agent proof 1209649066660320 01 wallet picker**

agent proof 1209649066660320 01 wallet picker

**agent proof 1209649066660320 02 send prefilled**

agent proof 1209649066660320 02 send prefilled

src/plugins/gui/providers/common.ts
```diff
@@ -7,10 +7,20 @@ import {
...
+// Android (the apex `edge.app` is not), matching the buy redirects, so an
+// external click opens the app and the deep-link parser routes it (PAYMENT to
+// the pre-filled Send scene; the terminal states to a harmless no-op). In-app
+// interception is host-agnostic: Paybis/MoonPay use `uri.startsWith(...)` and
+// Banxa an exact `===` against the same constant, so swapping the host is
+// symmetric and does not change the in-webview behavior.
+export const RETURN_URL_SUCCESS = 'https://deep.edge.app/redirect/success/'
+export const RETURN_URL_FAIL = 'https://deep.edge.app/redirect/fail/'
+export const RETURN_URL_CANCEL = 'https://deep.edge.app/redirect/cancel/'
+export const RETURN_URL_PAYMENT = 'https://deep.edge.app/redirect/payment/'
```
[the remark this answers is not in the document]
j0ntz: Valid for the transition window (MoonPay persists the payment redirect server-side and can resurface a legacy order days later). Fixed in d2676b2c5: added `isReturnUrl(uri, kind)` in providers/common.ts that matches BOTH the new `deep.edge.app` and the legacy apex `edge.app` hosts, and swapped every in-WebView interceptor to use it (moonpay/paybis/banxa provider + ramp, payment/success/fail/cancel). Providers are still handed the claimed-host URL going forward, so a legacy order resumed after an app update is still intercepted. Covered by src/__tests__/plugins/gui/providers/common.test.ts (dual-host + kind + look-alike-host cases).

src/actions/DeepLinkingActions.tsx
```diff
@@ -233,6 +235,67 @@ async function handleLink(
...
+      // provider disambiguates by network metadata we do not get here), so we
+      // must not exclude token matches when a native one also matches.
+      const symbol = currencyCode.split('_')[0].toUpperCase()
+      const assets: EdgeAsset[] = []
+      for (const pluginId of Object.keys(account.currencyConfig)) {
+        const currencyConfig = account.currencyConfig[pluginId]
+        if (currencyConfig.currencyInfo.currencyCode.toUpperCase() === symbol) {
+          assets.push({ pluginId, tokenId: null })
+        }
+        const { builtinTokens } = currencyConfig
```
j0ntz: **Correctness (high):** This asset scan only enumerates `builtinTokens`, so a user-added custom token whose ticker matches the redirect's `baseCurrencyCode` is invisible here. `pickWallet` itself matches against `wallet.enabledTokenIds` (which includes custom tokens), so a sell of a non-builtin token resolves `assets.length === 0` and shows `alert_deep_link_no_wallet_for_uri` even when the user holds a matching wallet, forcing a manual send instead of the pre-filled Send scene.

Consider iterating `allTokens` (or the union with enabled/custom tokens) rather than `builtinTokens`.

src/util/DeepLinkParser.ts
```diff
@@ -369,6 +406,27 @@ function parseEdgeAppLink(url: URL<string>): DeepLink {
...
+function parsePaymentRedirect(query: UriQueryMap): DeepLink | null {
+  const baseCurrencyCode = query.baseCurrencyCode ?? undefined
+  const depositWalletAddress = query.depositWalletAddress ?? undefined
+  if (baseCurrencyCode == null || depositWalletAddress == null) return null
+
+  return {
+    type: 'paymentRedirect',
+    currencyCode: baseCurrencyCode,
+    depositAddress: depositWalletAddress,
+    amount: query.baseCurrencyAmount ?? undefined,
```
j0ntz: **Correctness (med-high):** `parsePaymentRedirect` accepts `baseCurrencyAmount` without validating it is a non-empty numeric string. `query.baseCurrencyAmount ?? undefined` only guards `null`/absent, not empty-string or garbage:
- `baseCurrencyAmount=` (present but empty) yields `amount === ''`, which flows into `mul('', multiplier) === '0'` at DeepLinkingActions.tsx:283, silently pre-filling a zero send amount.
- `baseCurrencyAmount=abc` throws `Invalid number` from biggystring *after* the user has already completed the wallet picker, surfacing as a generic error toast rather than a clean up-front "invalid link".

Validate that the amount is a positive numeric string here and drop it (leave `undefined`) otherwise.

src/actions/DeepLinkingActions.tsx
```diff
@@ -233,6 +235,67 @@ async function handleLink(
...
+      // open the Send scene pre-filled with the deposit address, amount, and
+      // destination tag / memo so the user can finish the sell order.
+      const { currencyCode, depositAddress, amount, addressTag } = link
+
+      // Collect every native AND token asset that shares the symbol, and let
+      // the user disambiguate via the wallet picker. A provider sell can be a
+      // token whose ticker collides with another chain's native asset (the
+      // provider disambiguates by network metadata we do not get here), so we
+      // must not exclude token matches when a native one also matches.
+      const symbol = currencyCode.split('_')[0].toUpperCase()
```
j0ntz, line 254: **Correctness (med):** Ticker-only matching discards the chain suffix of the provider's currency code (`usdc_polygon`/`usdt_trc20` -> `USDC`/`USDT`). Every same-ticker asset across every enabled chain is then offered in the picker with no network hint. The adjacent comment acknowledges the provider disambiguates "by network metadata we do not get here," but the user is left to guess, and picking the wrong chain sends the deposit to an address that does not match the sell order. Worth surfacing a network label in the picker (or narrowing by chain when the provider does supply it).

src/actions/DeepLinkingActions.tsx
```diff
@@ -233,6 +235,67 @@ async function handleLink(
...
+      if (result?.type !== 'wallet') break
+      const { walletId, tokenId } = result
+      const wallet = account.currencyWallets[walletId]
+      if (wallet == null) break
+
+      const nativeAmount =
+        amount != null
+          ? mul(
+              amount,
+              getExchangeDenom(wallet.currencyConfig, tokenId).multiplier
```
j0ntz, line 302: **Correctness (low-med):** `getExchangeDenom` does not throw on an unresolvable `tokenId`; it returns `emptyEdgeDenomination` with `multiplier: '1'`. If `tokenId` from `pickWallet` ever fails to resolve to a denom (e.g. a token-metadata refresh race), `mul(amount, '1')` treats a decimal-unit amount as already-native and produces a wildly wrong `nativeAmount` in the Send scene, with no error surfaced. Consider guarding for the empty-denom fallback and aborting with a toast instead of computing a silent wrong amount.

src/util/DeepLinkParser.ts
```diff
@@ -361,6 +380,24 @@ function parseEdgeAppLink(url: URL<string>): DeepLink {
     }
   }
 
+  // Handle provider ramp redirects (e.g. MoonPay "Send with Edge"), which
+  // legacy orders may still point at the apex https://edge.app/redirect/...
+  // A `/redirect/payment` carries the deposit details for a pending sell order;
+  // the terminal states (`success`/`fail`/`cancel`) resolve to a no-op.
+  if (firstPath === 'redirect') {
+    const section = pathParts[1]
+    if (section === 'payment') {
```
j0ntz: **Correctness (med):** Malformed/param-less payment links are handled asymmetrically across hosts. In `parseEdgeProtocol` the `edge://redirect/payment` branch uses `parsePaymentRedirect(...) ?? { type: 'noop' }`, but here in `parseEdgeAppLink` (the legacy apex `edge.app` path) a failed parse falls through to `{ type: 'other' }` and opens a browser to `edge.app` (an unclaimed host, likely a dead page). Mirror the `?? { type: 'noop' }` fallback here so a stale/truncated apex payment link degrades the same way on both hosts.

src/util/DeepLinkParser.ts
```diff
@@ -244,6 +245,24 @@ function parseEdgeProtocol(url: URL<string>): DeepLink {
       }
     }
 
+    case 'redirect': {
```
j0ntz, line 248: **Cleanup (simplification):** The redirect-section dispatch (`payment` -> `parsePaymentRedirect`; `success`/`fail`/`cancel` -> noop) is duplicated between this `parseEdgeProtocol` case and the `parseEdgeAppLink` block below, both ultimately calling `parsePaymentRedirect`. A future change to redirect semantics (a new terminal state, or the noop-vs-fallback behavior in the finding above) must be edited in two places in sync. Extracting a single `parseRedirectSection(section, query, fallback)` helper collapses the duplication and prevents the two hosts from drifting.

src/actions/DeepLinkingActions.tsx
```diff
@@ -233,6 +235,67 @@ async function handleLink(
...
+      const { currencyCode, depositAddress, amount, addressTag } = link
+
+      // Collect every native AND token asset that shares the symbol, and let
+      // the user disambiguate via the wallet picker. A provider sell can be a
+      // token whose ticker collides with another chain's native asset (the
+      // provider disambiguates by network metadata we do not get here), so we
+      // must not exclude token matches when a native one also matches.
+      const symbol = currencyCode.split('_')[0].toUpperCase()
+      const assets: EdgeAsset[] = []
+      for (const pluginId of Object.keys(account.currencyConfig)) {
```
j0ntz, line 256: **Cleanup (reuse):** This hand-rolled loop over `account.currencyConfig` + `builtinTokens` to build `EdgeAsset[]` from a currency code duplicates the mechanics of `currencyCodesToEdgeAssets` in `src/util/CurrencyInfoHelpers.ts` (used by LoginActions). The two are not a clean drop-in swap (the shared helper resolves a single pluginId per chain code and splits on `:`, whereas this intentionally collects cross-chain ticker collisions), but the divergence is exactly the kind of thing that belongs as a param/variant on the shared helper rather than a second private copy that can drift from it.

src/plugins/gui/providers/common.ts
```diff
@@ -7,10 +7,34 @@ import {
...
+
+// Match a ramp redirect URL as it appears inside the provider WebView. Orders
+// created before the host switch still carry the legacy apex `edge.app` host —
+// MoonPay persists the payment redirect server-side and can resurface it days
+// later — so the in-WebView interceptors match BOTH the claimed and legacy hosts
+// to keep catching a legacy order resumed after an app update. Use this in place
+// of a raw `startsWith`/`===` against a single constant.
+const RETURN_URL_REDIRECT_HOSTS = [
```
j0ntz: **Cleanup (simplification):** `RETURN_URL_REDIRECT_HOSTS` re-hardcodes the two host+path strings (`https://deep.edge.app/redirect/`, `https://edge.app/redirect/`) already embedded in the `RETURN_URL_*` constants just above. The two are not derived from one source, so a future host migration must update both spots; missing one silently breaks either the outbound URL handed to providers or the inbound `isReturnUrl` matcher, with no compiler error. Define `CLAIMED`/`LEGACY` host constants once and build both the `RETURN_URL_*` values and this array from them.

src/actions/DeepLinkingActions.tsx
j0ntz: Fixed in a654892b. The asset scan now iterates allTokens (builtin plus user-added custom tokens) instead of builtinTokens, so a custom token whose ticker matches the redirect resolves a wallet and opens the pre-filled Send scene rather than showing the no-wallet alert. Verified on-sim: an ETH payment redirect resolved the ETH wallets and opened a populated Send scene.

src/util/DeepLinkParser.ts
j0ntz: Fixed in a654892b. parsePaymentRedirect now carries baseCurrencyAmount only when it is a positive finite number; empty or non-numeric values are dropped so the Send scene opens without an amount instead of pre-filling '0' or throwing from biggystring after the wallet picker. Unit tests added for empty and non-numeric amounts.

src/actions/DeepLinkingActions.tsx
j0ntz: Acknowledged as a known limitation (already noted in the code comment). We do not receive the provider's chain suffix, so we cannot narrow by chain here. The wallet picker does surface each candidate wallet's network (e.g. 'ETH Base', 'ETH Arbitrum One'), so the user disambiguates by the wallet's visible network. A richer network hint tied to the sell order would need chain metadata the provider strips; tracked as a follow-up, not changed in this PR.

src/actions/DeepLinkingActions.tsx
j0ntz: Fixed in a654892b. When the picked tokenId is unresolvable (an allTokens miss), we now abort with the no-wallet toast instead of computing mul(amount, '1') on the empty-denom fallback, which would have produced a wrong native amount.

src/util/DeepLinkParser.ts
j0ntz: Fixed in a654892b. Both hosts now route through a shared parseRedirectSection helper, so a malformed apex edge.app/redirect/payment resolves to a no-op (matching the edge:// path) instead of falling through to type other and opening a dead browser page. Unit test updated to expect the no-op.

src/util/DeepLinkParser.ts
j0ntz: Fixed in a654892b. Extracted parseRedirectSection(section, query), shared by parseEdgeProtocol and parseEdgeAppLink, collapsing the duplicated payment/success/fail/cancel dispatch so the two hosts cannot drift.

src/actions/DeepLinkingActions.tsx
j0ntz: Considered and deliberately kept separate. currencyCodesToEdgeAssets resolves a single pluginId per chain code and splits on ':', which would break the intentional cross-chain ticker-collision collection this path needs (it must offer every same-ticker wallet across chains). Folding a collision-tolerant variant into the shared helper would also touch the login path that consumes it, so it is deferred as an optional refactor rather than risked here.

src/plugins/gui/providers/common.ts
j0ntz: Fixed in a654892b. Added single CLAIMED_REDIRECT_HOST and LEGACY_REDIRECT_HOST constants and derived both the RETURN_URL_* values and the RETURN_URL_REDIRECT_HOSTS matcher array from them, so a host migration is one edit and the hand-off and match sides cannot fall out of sync.

src/util/DeepLinkParser.ts
```diff
@@ -369,6 +391,71 @@ function parseEdgeAppLink(url: URL<string>): DeepLink {
...
+  const amount =
+    Number.isFinite(amountNum) && amountNum > 0 ? rawAmount : undefined
+
+  return {
+    type: 'paymentRedirect',
+    currencyCode: baseCurrencyCode,
+    depositAddress: depositWalletAddress,
+    amount,
+    addressTag: query.depositWalletAddressTag ?? undefined
+  }
```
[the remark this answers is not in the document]
j0ntz: Valid, fixed in fd3bad0b. parsePaymentRedirect now treats a present-but-blank depositWalletAddressTag as absent (rawTag.trim() !== '' check), so a blank tag yields addressTag undefined and the Send scene gets no memo/uniqueIdentifier instead of an empty string. Unit test added for the blank-tag case.

j0ntz:
## 📸 Test evidence — payment-redirect deep link drives a populated Send scene (on-sim, slot UDID 15772779), plus cancel redirect no-ops with no error

**agent proof 1209649066660320 01 payment send scene populated**

agent proof 1209649066660320 01 payment send scene populated

**agent proof 1209649066660320 02 wallet picker eth**

agent proof 1209649066660320 02 wallet picker eth

**agent proof 1209649066660320 03 cancel noop no error**

agent proof 1209649066660320 03 cancel noop no error

peachbits (approved):

brianstohl87-cpu (commented):
What next?

PR #6062, opened 2026-06-26
Fix Use Biometrics toggle reverting after leaving Settings

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Requirements

- [ ] Tested on iOS device

### Description

Asana: https://app.asana.com/0/1215088146871429/1213428787796786

Fixes the Settings "Use Biometrics" toggle reverting to its previous state after the user leaves and re-enters the Settings scene.

Root cause: commit `acc2adbc` moved biometric state out of Redux into a `@tanstack/react-query` query (`['biometricState', account.id]`) plus a local `touchIdEnabled` state that syncs from the query only once per mount. `handleUpdateTouchId` persisted the change via `enableTouchId`/`disableTouchId` but never updated the query cache, and the app-level cache has no `staleTime` and does not refetch on focus. On re-entering Settings the scene remounts, the one-time sync reads the stale cached value, and the toggle shows the old state. Reported on Android; also reproduces on iOS, since the cause is platform-agnostic.

Fix: after the enable/disable persists, update the cached biometric state with `queryClient.setQueryData` so the cache reflects the value the user set. The next mount's sync then reads the correct value.

Testing: added a regression test (`SettingsScene.biometric.test.tsx`) that mounts the scene, toggles biometrics off, and asserts the shared query cache flips to the persisted value across a remount. It fails on the pre-fix code (cache stays `true`) and passes with the fix. Full `tsc`, eslint, and the jest suite pass. The toggle could not be exercised on the simulator because biometric enrollment is unavailable headlessly (no `simctl` biometry command; the Simulator Face ID enrollment menu requires GUI automation that is not reachable in this environment), so the deterministic test is the verification of the changed code path.

---

> [!NOTE]
> **Low Risk**
> Settings UI and React Query cache only; biometric enable/disable still goes through existing edge-login-ui-rn APIs, with regression tests covering the changed path.
> 
> **Overview**
> Fixes the **Use Biometrics** switch in Settings appearing to revert after you leave and come back.
> 
> Biometric on/off is loaded with React Query (`biometricState`), but toggling only called `enableTouchId` / `disableTouchId` and never updated the shared cache. On remount, the scene read stale cached data instead of what the user had set.
> 
> **`SettingsScene`** now drives the switch from query data only (drops the one-time local `touchIdEnabled` sync). **`handleUpdateTouchId`** cancels in-flight refetches, optimistically updates the cache with `setQueryData`, persists via the login UI helpers, and rolls the cache back if persistence fails. A shared **`biometricStateQueryKey`** keeps the query and updates aligned.
> 
> Adds **`SettingsScene.biometric.test.tsx`** to assert cache stays correct across remount and that a late stale refetch cannot overwrite a completed toggle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec579c250d8bdd69e6c38c2bbaf3f0c08c90428a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/components/scenes/SettingsScene.tsx
+++ b/src/components/scenes/SettingsScene.tsx
@@ -1,4 +1,4 @@
-import { useQuery } from '@tanstack/react-query'
+import { useQuery, useQueryClient } from '@tanstack/react-query'
 import { asMaybe } from 'cleaners'
 import type { EdgeLogType } from 'edge-core-js'
 import {
@@ -67,6 +67,13 @@ import { asPrivateNetworkingSetting } from '../themed/MaybePrivateNetworkingSett
 
 type Props = EdgeAppSceneProps<'settingsOverview'>
 
+// Shared query key so the query definition and the optimistic cache update
+// can never drift apart.
+const biometricStateQueryKey = (accountId: string): [string, string] => [
+  'biometricState',
+  accountId
+]
+
 export const SettingsScene: React.FC<Props> = props => {
   const { navigation } = props
   const theme = useTheme()
@@ -84,9 +91,14 @@ export const SettingsScene: React.FC<Props> = props => {
 
   const account = useSelector(state => state.core.account)
 
-  // Load biometric state locally (not from Redux)
+  const queryClient = useQueryClient()
+
+  // Load biometric state locally (not from Redux). The toggle renders
+  // directly off this query so its on-screen value can never diverge from the
+  // cache; handleUpdateTouchId keeps the cache correct via an optimistic
+  // setQueryData.
   const { data: biometricState } = useQuery({
-    queryKey: ['biometricState', account.id],
+    queryKey: biometricStateQueryKey(account.id),
     queryFn: async () => {
       const [touchEnabled, supportedType] = await Promise.all([
         isTouchEnabled(account),
@@ -100,18 +112,6 @@ export const SettingsScene: React.FC<Props> = props => {
     },
     enabled: account != null
   })
-
-  // Local state to track touch ID enabled status (can be toggled by user)
-  const [touchIdEnabled, setTouchIdEnabled] = React.useState<boolean | null>(
-    null
-  )
-
-  // Sync local state with loaded state
-  React.useEffect(() => {
-    if (biometricState != null && touchIdEnabled == null) {
-      setTouchIdEnabled(biometricState.isTouchEnabled)
-    }
-  }, [biometricState, touchIdEnabled])
 
   const supportsTouchId = biometricState?.isTouchSupported ?? false
   const username = useWatch(account, 'username')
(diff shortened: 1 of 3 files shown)


src/components/scenes/SettingsScene.tsx
```diff
@@ -223,6 +225,16 @@ export const SettingsScene: React.FC<Props> = props => {
...
+      // Keep the cached biometric state in sync with the value we just
+      // persisted. Without this, the cached query result stays stale and
+      // re-entering Settings shows the previous toggle state instead of the
+      // one the user set.
+      if (biometricState != null) {
+        queryClient.setQueryData(['biometricState', account.id], {
+          ...biometricState,
+          isTouchEnabled: newValue
+        })
+      }
```
peachbits: The fix is correct, but it compensates for a redundant-state pattern instead of removing it. The toggle renders from a local `touchIdEnabled` mirror of the `biometricState` query, and the sync effect runs only once (`touchIdEnabled == null` guard, line 113). On re-entry `useQuery` serves the stale cached value first (stale-while-revalidate), the one-time guard latches it, and the later refetch can't overwrite the local mirror — that's the revert. This `setQueryData` keeps the cache correct so the latched value ends up right.

Cleaner direction: derive the toggle straight from the query (drop the `touchIdEnabled` mirror + the one-time guard) and keep this `setQueryData` as the optimistic update — that makes the stale-mirror class of bug structurally impossible.

Nit: `['biometricState', account.id]` is now duplicated (also line 91); a shared key constant/factory would prevent drift.

peachbits: suggestion doesn't provide any additional fixing except for cleanup so it isn't necessary. approving

peachbits (approved):

src/components/scenes/SettingsScene.tsx
j0ntz: Done, both points applied in the amended commit (e8924f865):

- Removed the redundant-state pattern: the toggle now renders directly off the `biometricState` query. Dropped the `touchIdEnabled` local mirror and its one-time `touchIdEnabled == null` sync effect, so there is no second source of truth to latch a stale value.
- `handleUpdateTouchId` now does an optimistic `setQueryData` (immediate UI feedback since the toggle derives from the cache), awaits enable/disable, and reverts the cache on error. This makes the stale-mirror class of bug structurally impossible.
- Nit: extracted a module-scope `biometricStateQueryKey(accountId)` factory, used by both the query definition and the optimistic update, so the key can no longer drift.

Regression test and full verify-repo (tsc, eslint, jest) pass on the new commit.

src/components/scenes/SettingsScene.tsx
```diff
@@ -214,18 +214,29 @@ export const SettingsScene: React.FC<Props> = props => {
...
+    const newValue = !biometricState.isTouchEnabled
+    const queryKey = biometricStateQueryKey(account.id)
+    // Optimistically flip the cache so the toggle (which renders off the
+    // query) updates immediately and stays correct when Settings is left and
+    // re-entered. Without this the cached value stays stale and re-entry shows
+    // the previous state.
+    queryClient.setQueryData(queryKey, {
+      ...biometricState,
+      isTouchEnabled: newValue
+    })
```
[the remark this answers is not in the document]
j0ntz: Good catch, fixed in 2b06fa9e6. `handleUpdateTouchId` now calls `await queryClient.cancelQueries({ queryKey })` before the optimistic `setQueryData`, so any refetch started on mount is cancelled and react-query ignores its result. A background fetch that read the pre-toggle keychain value can no longer resolve late and clobber the toggle back.

Added a deterministic regression test (`cancels an in-flight refetch so a stale read cannot overwrite a completed toggle`): it holds the mount refetch in-flight, toggles off, then resolves the stale refetch with the old value and asserts the cache stays `false`. Verified it FAILS without the `cancelQueries` guard (cache flips back to `true`) and passes with it.

PR #6064, opened 2026-06-29
Upgrade to React Native 0.86 / Expo SDK 57 (new architecture)

peachbits:
**iOS blur root cause (found with a physical iPhone X, iOS 16.7):** React Native 0.86
removed `StyleSheet.absoluteFillObject`; spreading the now-undefined value yields an
empty style, so any absolutely positioned layer that spread it collapsed to nothing.
edge-login-ui-rn spread it in six places (its modal blur sheet - the "grey modal" with
the scene sharp behind it - plus EdgeButton, EdgeCard and ChallengeModal backgrounds),
fixed in PR #301 and carried here through the patch. The same collapse hit
`@react-navigation/drawer`'s overlay (side menu no longer dimmed or closed on outside
tap) - fixed via the drawer patch. Two blur libraries failed identically before the
fix, which is how the library was ruled out; iOS now unifies on expo-blur and
`rn-id-blurview` is dropped. Verified: login-ui and GUI modals blur on simulator and
device; side-menu overlay dims and closes.

---
  - https://app.asana.com/0/0/1216132410792890
  - https://app.asana.com/0/0/1218599169188093

---

> [!NOTE]
> **High Risk**
> Full-stack platform upgrade (new architecture, native toolchains, Firebase, and rendering) touches every screen; regressions in blur, navigation overlays, scrolling, and native crashes are plausible despite targeted flag and library mitigations.
> 
> **Overview**
> **Upgrades the app to React Native 0.86 and Expo SDK 57 with the new architecture enabled on Android and iOS**, including Gradle 9 / compileSdk 36, Expo’s unified autolinking, Hermes V1 with React-Core built from source on iOS (min deployment 16.4), and bumped native deps (Firebase 12, Reanimated 4.5, etc.).
> 
> **Android startup** drops `ReactNativeHostWrapper` and, after new-arch `load()`, resets React Native feature flags to enable ShadowTree commit-exhaustion protection and **keep view recycling disabled** so recycled views don’t leave stale dimmed buttons or font sizes. Release lint is turned off to avoid OOM; Gradle JVM heap is increased.
> 
> **UI stack migrations** replace legacy blur/gradient/keyboard libraries with **expo-blur**, **expo-linear-gradient**, and **react-native-keyboard-controller**; **rn-id-blurview** and **react-native-linear-gradient** are removed from pods. Babel now always uses `react-native-worklets/plugin` (no Android-specific Reanimated path).
> 
> The **CHANGELOG** records follow-on fixes for new-arch behavior: side-menu overlay tap-to-close, modal/header blur (solid fallbacks on older Android), keyboard/modal/amount-field/accessibility issues, `minimumFontScale` via a small RN patch, deduped airship/patina installs, and measured Android scroll performance gains in release benchmarks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb41a875061c8a6cfe0540b04e8a7fedb49d32f8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


peachbits:
History cleanup after a review question: commit 2 ("Upgrade to React Native 0.85.3") previously contained a spurious `newArchEnabled true→false` flip that commit 3 flipped back — an artifact of how the upgrade commit was reconstructed during the re-split, not a real setting change. Fixed: the flag now changes exactly once, in commit 1 (`false→true`), and commits 2–5 don't touch it. The Android commit keeps its legitimate `jvmargs` bump.

Audited the whole series for more of the same (blob-level revert detection across all commits + exact-line added-then-removed scan, lockfile excluded): this was the only instance. Final tree is bit-identical to the previously validated one, and all 5 commits re-passed the per-commit precommit gate (521/521 tests on commit 1's RN 0.79 base; 517/517 on 0.85).

peachbits:
🔴→✅ **Regression found in review and fixed**: the PR was accidentally **reverting develop PR #6044** ("Deprecate Botanix to keys-only mode on July 9, 2026") — the `keysOnlyMode` functions in `WalletAndCurrencyConstants.ts`, their test, the keys-only wiring in `WalletListMenuModal`/`WalletListModal`, and the CHANGELOG entries. Root cause: when this branch was originally rebuilt onto a newer develop base, the upgrade commit carried file-wholesale content from an older tree, stomping develop changes in files both change-sets touched.

Fixed by restoring develop's versions of all 5 files into the upgrade commit — the upgrade legitimately changed **zero lines** in them, so they now drop out of the PR diff entirely. Verification:
- `keysOnlyMode` intact at every commit in the chain
- Full sweep of every develop change in the base window vs the PR tip: **no remaining reverts or divergences** (#6060 and the maestro-testID commit were never affected)
- All 5 commits re-passed the per-commit precommit gate — now **521/521 tests at every commit** (the previous 517-vs-521 gap between commits was itself the stomped Botanix test file)

The Botanix July 9 deprecation survives this PR.

peachbits:
🐛 **On-device Android testing found (and fixed) a blocking crash** — new commit `12195ddb1`.

**Symptom:** release build on a physical device (OnePlus 5T) crashed the moment the password login scene mounted (tap "Sign in" from the welcome carousel).

**Root cause:** RN 0.85's Android Fabric text measurement throws `IllegalArgumentException: FontSize should be a positive value` for any `Text` with font size 0 — `TextLayoutManager`'s `letterSpacing.isNaN()` check unconditionally invokes a getter that validates `fontSize > 0` (the old architecture silently tolerated 0, which is why this never crashed pre-upgrade). `FilledTextInput` animates its side-icon sizes to exactly 0 (clear icon on an empty input), and those feed `AnimatedFontIcon`'s `fontSize` — so **any scene with a text input** crashed on Android, most visibly login via edge-login-ui-rn (same component pattern).

**Fix:** clamp animated font sizes to ≥1 (visually identical — the icon containers already collapse to zero width) in the GUI's `AnimatedFontIcon` + `FilledTextInput`, plus a 1-line `patches/edge-login-ui-rn+3.36.0.patch` until the fix lands upstream in edge-login-ui-rn. The 9 snapshot updates in the commit are the fix's signature (`fontSize: 0 → 1` on at-rest icons — the old snapshots had captured the crash-causing zero).

**Verified:** rebuilt release APK on the same device — password login scene renders, inputs + eye icon work, zero fatal exceptions. Commit gated: localize/warnings clean, tsc clean, eslint 0 errors, 521/521 tests.

**Follow-up:** edge-login-ui-rn needs the same clamp in its own repo (`src/components/icons/ThemedIcons.tsx`) so the patch can be dropped.

peachbits:
♻️ **Rebased onto current develop** (was 67 commits behind and conflicting; now conflict-free). Same 6 commits, same content — mechanical notes:

- Two conflicts, both in generated/dependency files: `package.json` (union: develop's `react-native-piratechain ^0.6.1` + this branch's `react-native-reanimated 4.3.1`) and the lockfiles — **regenerated, not hand-merged** (`npm install --package-lock-only` for npm; a fresh `pod install` on the merged tree transplanted into the iOS commit).
- Post-rebase audits all clean: **0 develop-window stomps** (every file develop changed in the 67-commit window verified intact at the tip — includes the piratechain/monero/edge-core/accountbased/exchange-plugins bumps and the maestro testID work), 0 within-series reverts, `newArchEnabled` flips exactly once (commit 1).
- Full per-commit precommit gate re-run: **6/6 PASS at 557/557 tests** (suite grew with develop; every commit green under its own dependency set).

Also: the edge-login-ui-rn fontSize fix now exists as a proper branch in that repo (`matthew/fabric-fontsize-clamp`) pending review/PR — once released, the `patches/edge-login-ui-rn+3.36.0.patch` here can be dropped in favor of a version bump.

peachbits:
🔁 **Retargeted to React Native 0.86.0 / Expo SDK 57** — same 6-commit structure, now a single hop from 0.79.2 to current stable (no intermediate 0.85 in history).

**What changed in the retarget:**
- Upgrade commit now pins RN 0.86.0 + Expo ^57.0.7 with the SDK 57 set (reanimated 4.5.0, worklets 0.10.0, gesture-handler ~2.32.0, keyboard-controller 1.21.9). TypeScript deliberately held at ~5.8.3 (SDK 57 suggests TS 6 — a major bump that belongs in its own PR).
- reanimated 4.5 typing break fixed in `EdgeAnim.tsx` (ComplexAnimationBuilder went generic; presets' static `delay` declares the base-builder return — matched their declared signature with one documented cast).
- Expo 57 factory-delegate change: `customize` now receives the new-arch `UIView` (was `RCTRootView`).
- **Podfile now pins the engine pairing** (`RCT_HERMES_V1_ENABLED=1`, `RCT_USE_PREBUILT_RNCORE=0`): the 0.86.0 prebuilt React-Core binary is built against the non-V1 jsi ABI (`jsi::Runtime` vs V1's `jsi::IRuntime`) and dyld-crashes at launch when paired with the default V1 `hermesvm` — React-Core from source is the coherent pairing for the V1 engine Edge already ships. Deterministic regardless of shell environment.
- ShadowTree patch regenerated as `patches/react-native+0.86.0.patch` (flag still default-off in 0.86; verified present).

**Verification:** audits clean (0 develop-window stomps, 0 within-series reverts, coherent prebuilt-free lock); per-commit gates 6/6 at 557/557 (JS surface; post-sweep amendments were iOS-native only, covered by builds); Android release APK builds; iOS builds from source and **boots to login on the simulator**. On-device Android smoke pending device reconnect.

edge-login-ui-rn's clamp branch (`matthew/fabric-fontsize-clamp`) is unaffected by the retarget.

src/components/common/BlurBackground.tsx
```diff
@@ -1,42 +1,258 @@
...
 
 import { cacheStyles, type Theme, useTheme } from '../services/ThemeContext'
 
+export { BlurTargetView }
+
 const isAndroid = Platform.OS === 'android'
 
-/** A blur background WITH rounded corners, used for most components */
-export const BlurBackground = () => {
+// Android below 12 (API 31) has no working blur under the new architecture:
```
j0ntz: Noting for reference: currently 5% of our Android users fall under this category. Acceptable tradeoff.

(Not a change request)

PR #6069, opened 2026-07-10
Hide Get Raw Keys behind Developer Mode

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

`- changed: Hide the wallet "Get Raw Keys" option behind Developer Mode, while still showing it for wallets that fail to load.`

### Dependencies

none

### Requirements

- [x] Tested on iOS device

### Description

[Asana task](https://app.asana.com/0/0/1208354993842606/f)

Asana: https://app.asana.com/1/9976422036640/project/1201386023359449/task/1208354993842606

Hides the `Get Raw Keys` item in the wallet overflow menu unless the account has Developer Mode enabled. Raw Keys is a power-user tool; gating it behind Developer Mode declutters the menu for normal users while keeping the capability available to those who need it.

The safety fallback is preserved: when a wallet fails to load (`wallet == null`), Raw Keys still appears unconditionally, so keys can always be recovered from a broken wallet. This is handled by a separate early-return branch that is intentionally not gated.

No change was needed to make the Raw Keys JSON read-only: `RawTextModal` already renders the content as read-only text (a `Paragraph`, not an input) and `getRawKeys` passes `disableCopy`.

Implementation: read `state.ui.settings.developerModeOn` in `WalletListMenuModal`, and skip the `getRawKeys` option in the menu loop when Developer Mode is off.

Test evidence (iOS sim, edge-funds) attached below: menu with Developer Mode ON shows Get Raw Keys; menu with Developer Mode OFF hides it.

---

> [!NOTE]
> **Low Risk**
> Small UI gating change in the wallet menu with an explicit recovery exception for wallets that fail to load; no changes to key export APIs or authentication.
> 
> **Overview**
> **Get Raw Keys** in the wallet list overflow menu is now shown only when **Developer Mode** is enabled (`state.ui.settings.developerModeOn`), so normal users see a simpler menu while power users can still reach raw key export from settings.
> 
> The modal reads that flag and skips the `getRawKeys` entry when building options for loaded wallets. **Recovery is unchanged:** if the wallet object is missing (`wallet == null`), the menu still offers Get Raw Keys (and archive) with no Developer Mode check, so keys can be exported from a broken wallet. Existing rules for light accounts (no seed/raw keys when username is null) and the loaded-wallet path are otherwise the same.
> 
> CHANGELOG documents the behavior change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c35e3d644072f4d754d44befcd9182daf2466fef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased (develop)
 
 - changed: Route maestro test builds to a dedicated Zealot channel so they no longer appear in the production release list.
+- changed: Hide the wallet "Get Raw Keys" option behind Developer Mode, while still showing it for wallets that fail to load.
 
 ## 4.49.0 (staging)
 
--- a/src/components/modals/WalletListMenuModal.tsx
+++ b/src/components/modals/WalletListMenuModal.tsx
@@ -147,6 +147,9 @@ export const WalletListMenuModal: React.FC<Props> = props => {
   const pausedWallets = useSelector(
     state => state.ui.settings.userPausedWalletsSet
   )
+  const developerModeOn = useSelector(
+    state => state.ui.settings.developerModeOn
+  )
 
   const wallet = useWatch(account, 'currencyWallets')[walletId]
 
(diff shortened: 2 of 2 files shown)


j0ntz:
## 📸 Test evidence (iOS sim, edge-funds)

**agent proof 1208354993842606 01 devmode on rawkeys visible**

agent proof 1208354993842606 01 devmode on rawkeys visible

**agent proof 1208354993842606 02 devmode off rawkeys hidden**

agent proof 1208354993842606 02 devmode off rawkeys hidden

peachbits (approved):

PR #6098, opened 2026-07-24
Improve unstake scene error handling

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Requirements

- [ ] Tested on iOS device

### Description

Asana: https://app.asana.com/1/9976422036640/project/1200382638405084/task/1211086099095860

Improves the unstake error experience on the StakeModify scene (repro: unstake an asset with no native balance to pay the network fee, e.g. LTC with no LTC).

Per the task and RJ's review note ("do not show the error alert at this scene since we already have an error field; fish out and properly display the error correctly if possible") and Michael's note ("give a message stating that the wallet needs balance to unstake"):

- Removed the `showError` popup (the scary red alert) from the quote-fetch and approve error paths. The scene already has an on-scene error field (`ErrorTile`); it is now the single error surface.
- Replaced the generic "Unknown error occurred" with the real error message, fished out via a small `getDisplayErrorMessage` helper (`src/util/stakeErrorUtils.ts`).
- For an `InsufficientFundsError` during an unstake, the error field now shows a clear message ("This wallet needs a `<currency>` balance to cover the network fee for unstaking.") instead of a bare "Insufficient Funds".

### Testing

- `tsc --noEmit`, eslint, and the full jest suite pass (`verify-repo.sh`).
- New unit test `src/__tests__/stakeErrorUtils.test.ts` (4 cases) covers the fish-out logic: real message passthrough, edge-core-js error subclasses, empty-message fallback, and non-Error fallback.
- iOS sim (edge-funds): the organic trigger could not be driven because the staking-provider RPC is unreachable from this environment ("could not detect network / NETWORK_ERROR"), so no staking position loads. The new error rendering was hack-verified instead: a temporary uncommitted edit forced the real StakeModify scene to render the real `ErrorTile` with the new copy and no popup. See the attached 🩹 hack-forced screenshot. The hack was reverted (clean tree); the string-selection and no-popup logic are also covered by the unit test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased (develop)
 
 - changed: Sign MoonPay buy/sell widget URLs and bind them to the customer's IP via the info server, for MoonPay's on-ramp IP-matching security upgrade.
+- fixed: Improve the unstake error experience by replacing the popup alert and generic "unknown error occurred" with the real error in the scene's error field, and showing a clear message when the wallet lacks the balance to cover the unstaking network fee.
 
 ## 4.50.0 (2026-07-21)
 
--- a/src/__tests__/stakeErrorUtils.test.ts
+++ b/src/__tests__/stakeErrorUtils.test.ts
@@ -1,0 +1,38 @@
+import { describe, expect, test } from '@jest/globals'
+import { DustSpendError, InsufficientFundsError } from 'edge-core-js'
+
+import { lstrings } from '../locales/strings'
+import { getDisplayErrorMessage } from '../util/stakeErrorUtils'
+
+describe('getDisplayErrorMessage', () => {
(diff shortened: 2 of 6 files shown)


j0ntz:
## 📸🩹 Test evidence (hack-forced)

> 🩹 **Some screenshots below are HACK-FORCED.** The marked frames show a state forced by a temporary uncommitted local edit (the natural trigger could not be reproduced on the sim). The hack is not in this PR's diff; the pixels prove the rendering, not the trigger.

**🩹 HACK-FORCED: unstake insufficient funds**

🩹 HACK-FORCED: unstake insufficient funds

j0ntz:
Superseded by #6105, which consolidates this and 11 other small GUI task branches into one PR/branch per the umbrella task's instructions.

PR #6127, opened 2026-07-29
Add app/device attestation for gated info-server requests

paullinator:
Supersedes #6076, which carried the same work across nine `fixup!` commits. Same tree, squashed into one commit so it can be reviewed as a whole.

## Summary

Adds application-level attestation for the info-server endpoints that gate signing (Simplex, Banxa). A background engine keeps a short-lived JWT fresh, using Apple App Attest on iOS and Android Keystore key attestation on Android. Gated callers await the token rather than racing their own handshakes.

The bulk of the work beyond the initial implementation went into the concurrency and quota behaviour, since platform attestation is rate-limited on iOS and a mistake there is expensive across a fleet:

- Single-flight handshakes with a monotonic generation counter, so an attempt the watchdog retired cannot clobber state a live attempt owns.
- Two-tier backoff: only attempts that actually spent a platform attestation grow it. Hangs count as failures, so a device whose native call never answers stops re-attesting on every gated request.
- Escalation to the rate-limited path is reserved for the two things it can fix — a key that cannot sign, and a key the server has judged and rejected. A `5xx`/`429` from the assert endpoint, an unusable `200` body, and a transient native `timeout` all fail into backoff instead, so an info-server outage cannot make the whole fleet discard its keys and re-attest at once.
- iOS: operations are bounded so a hung `attestKey` cannot wedge the serial queue; promises settle exactly once; a late callback neither enrols a key the server never verified nor clears one a newer handshake owns.
- Android: `keystoreLock` is a `ReentrantLock` with a bounded wait, so one wedged Keystore call cannot disable every later operation; all methods stay off the shared native-modules thread.
- `clearKey(keyId)` is scoped to the key the caller meant, on both platforms, since the call can outlive the state it was reasoning about.

`docs/APP_ATTESTATION.md` documents the engine, both native modules, and the escalation rules.

## Test plan

- [x] `src/__tests__/util/attestation.test.ts` — 36 tests, including regressions for every race fixed above
- [x] `src/__tests__/util/attestationNativeBridge.test.ts` — compares the Swift `@objc` selectors against the hand-written Objective-C bridge, which nothing else in the repo verifies (a mismatch is not a build error; React Native only fails at runtime)
- [x] `tsc` and ESLint clean
- [x] `./gradlew :app:compileDebugKotlin` succeeds
- [x] `swiftc -typecheck` succeeds
- [ ] Exercise a real handshake on an iOS device (App Attest needs real hardware) and an Android device

Note: 21 UI suites fail on `develop` independently of this branch (React `act` errors in component tests such as `Card`); verified identical failures at the merge base with none of this code present.

Made with [Cursor](https://cursor.com)


paullinator:
bugbot run

paullinator:
bugbot review

paullinator:
bugbot review

paullinator:
bugbot review

paullinator:
bugbot review

paullinator:
bugbot review

paullinator:
bugbot review

paullinator:
Closing to keep GitHub PR automation off this branch while the follow-up structural work lands. The branch `paul/appAttestationV2` stays open as the head of that work: history is being squashed to the primary commit plus fixups, and a new PR will be opened once the restructure is complete.

PR #6137, opened 2026-08-04
Add app/device attestation for gated info-server requests

paullinator:
## Summary
- Add a background attestation engine (`src/util/attestation.ts`) plus iOS App Attest and Android Keystore native modules that enroll once, refresh via assertion/signature, and cache JWTs on a monotonic clock with lead-ahead refresh and failure backoff.
- Wire Simplex/Banxa jwtSign and createHmac call sites to attach `x-attestation-token` when available; surface gated signing failures in ramp UI instead of silently no-oping.
- Document the client architecture in `docs/APP_ATTESTATION.md`. Depends on info-server lifetime/`expiresIn`/`serverTime` support ([edge-info-server#158](https://github.com/EdgeApp/edge-info-server/pull/158)).

## Test plan
- [ ] `npm test -- src/__tests__/util/attestation.test.ts src/__tests__/util/attestationNativeBridge.test.ts`
- [ ] `tsc` / lint on changed files; `npm run fix-kotlin` / ktlint on attestation Kotlin modules
- [ ] Physical iOS device: boot → challenge → App Attest enroll → token cached; later refresh via `generateAssertion`
- [ ] Physical Android (TEE/StrongBox): enroll via Keystore attestation; refresh via `signChallenge`
- [ ] Point `ENV.INFO_SERVER` at a local info server from [edge-info-server#158](https://github.com/EdgeApp/edge-info-server/pull/158); confirm Banxa/Simplex signing sends `x-attestation-token` and fails loudly on 403
- [ ] Simulator / unsupported device: engine stops cleanly; gated plugins omit the header

---

> [!NOTE]
> **High Risk**
> Touches purchase flows and partner signing endpoints; attestation misconfiguration or token/cache bugs can block buys (403) or weaken gating until server policy is tuned.
> 
> **Overview**
> Adds **app/device attestation** so gated info-server calls (Simplex JWT sign, Banxa HMAC) can prove the real Edge app on real hardware.
> 
> A new **background attestation engine** (`src/util/attestation.ts`) talks to the info server’s `/v1/attest/*` flow: enroll once with platform attestation, then refresh with a cheap assertion/signature, cache JWTs on a **monotonic** deadline, and attach `x-attestation-token` via `attestedJsonHeaders()` when a token is available.
> 
> **Native bridges** on iOS (App Attest + Keychain) and Android (Keystore attestation, StrongBox fallback, serialized lock) implement enrollment, refresh, and scoped `clearKey` on server rejection.
> 
> **Ramp plugins** (Simplex, Banxa) use attested headers for `jwtSign` / `createHmac`; failures surface in the ramp quote flow instead of silent no-ops. Optional `INFO_SERVER` override and clock-skew UX support local testing and misconfigured device time.
> 
> Also ships `docs/APP_ATTESTATION.md`, extensive Jest coverage, signing-cert extraction tooling, and iOS Xcode wiring helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e94ab0aa1fa08c78da762a0aa25c406f1c65a54. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
  - https://app.asana.com/0/0/1217260695616571

### DIFF
--- a/android/app/src/main/java/co/edgesecure/app/EdgeAttestationModule.kt
+++ b/android/app/src/main/java/co/edgesecure/app/EdgeAttestationModule.kt
@@ -1,0 +1,312 @@
+package co.edgesecure.app
+
+import android.os.Build
+import android.security.keystore.KeyGenParameterSpec
+import android.security.keystore.KeyProperties
+import android.security.keystore.StrongBoxUnavailableException
+import android.util.Base64
+import com.facebook.react.bridge.Arguments
+import com.facebook.react.bridge.Promise
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.bridge.ReactContextBaseJavaModule
+import com.facebook.react.bridge.ReactMethod
+import java.security.KeyPairGenerator
+import java.security.KeyStore
+import java.security.MessageDigest
+import java.security.spec.ECGenParameterSpec
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.locks.ReentrantLock
+
+/**
+ * Native bridge for Android Keystore hardware key attestation (device-level
+ * attestation). Generates a fresh EC key in the AndroidKeyStore with the given
+ * attestation challenge and returns the resulting X.509 certificate chain, which
+ * the info server verifies against Google's hardware attestation roots.
+ *
+ * This stays fully open source: it uses only the platform KeyStore APIs, not the
+ * closed-source Play Integrity / SafetyNet SDKs.
+ */
+class EdgeAttestationModule(
+  reactContext: ReactApplicationContext,
+) : ReactContextBaseJavaModule(reactContext) {
+  companion object {
+    // Stable alias: the key is enrolled once via attestation and then reused
+    // to sign challenges (see signChallenge). Cleared only on clearKey or
+    // when re-enrollment is needed.
+    private const val KEY_ALIAS = "edge_attestation_key"
+
+    // Serializes all AndroidKeyStore access to KEY_ALIAS. getAttestation,
+    // signChallenge and clearKey each mutate/read the single shared alias; the
+    // JS engine's watchdog can release its in-flight lock and start a new
+    // handshake while an older native Thread is still running, so without this
+    // lock two overlapping getAttestation calls could delete/regenerate the key
+    // out from under each other and return cross-wired certificate chains.
+    private val keystoreLock = ReentrantLock()
+
+    // Caps how long an operation waits for keystoreLock. Keystore work is local
+    // and synchronous, so a wedged generateKeyPair or sign cannot be interrupted
+    // - but the operations queued behind it can refuse to wait forever, which is
+    // what stops one wedge from taking down every later attestation, refresh and
+    // clear for the life of the process. Well above a slow-but-healthy attested
+    // key generation, and below the JS watchdog so JS gets a real rejection
+    // instead of timing out blind.
+    private const val LOCK_TIMEOUT_SECONDS = 60L
+
+    private enum class AttestCode(
+      val code: String,
+    ) {
+      UNSUPPORTED("unsupported"),
+      LOCK_TIMEOUT("lockTimeout"),
+      NO_KEY("noKey"),
+      ATTESTATION_ERROR("attestationError"),
+      SIGN_CHALLENGE("signChallenge"),
+    }
+  }
+
+  override fun getName(): String = "EdgeAttestation"
+
+  /**
+   * Runs [body] holding [keystoreLock], rejecting with `lockTimeout` if the lock
+   * cannot be acquired within [LOCK_TIMEOUT_SECONDS].
+   *
+   * That rejection code matters twice over. The JS engine reads it as saying
+   * nothing about whether the enrolled key can sign, so it retries the cheap
+   * refresh path instead of escalating to a full attestation. It also reads it as
+   * proof that no attestation was spent - this fires before the lock is held, so
+   * [body] never ran and no key was generated - which keeps a contended lock from
+   * doubling the failure backoff. It is deliberately not the `timeout` iOS
+   * reports: that one fires while waiting on an App Attest callback, so the
+   * platform operation did start and may have counted against the quota.
+   */
+  private fun withKeystoreLock(
+    promise: Promise,
+    body: () -> Unit,
+  ) {
+    try {
+      if (!keystoreLock.tryLock(LOCK_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
+        promise.reject(AttestCode.LOCK_TIMEOUT.code, "Timed out waiting for the Keystore lock")
+        return
+      }
+    } catch (interrupted: InterruptedException) {
+      // Nothing interrupts these threads today: they are plain Threads and no
+      // reference to them is kept. But this call sits outside the try/catch the
+      // callers wrap their own work in, so an escape here would leave the
+      // promise unsettled and take the process down with an uncaught exception
+      // on a background thread. Swapping Thread for an executor or a coroutine,
+      // where interruption is how cancellation arrives, would make that
+      // reachable without anyone touching this file. The same code as the
+      // timeout is right: the lock was never held, so nothing was spent.
+      //
+      // Catching clears the interrupt flag, so restore it: under an executor or
+      // coroutine the layer above reads that flag to see the cancellation, and
+      // swallowing it here would strand a pool thread that should be winding
+      // down.
+      Thread.currentThread().interrupt()
+      promise.reject(AttestCode.LOCK_TIMEOUT.code, "Interrupted waiting for the Keystore lock")
+      return
+    }
+    try {
+      body()
+    } finally {
+      keystoreLock.unlock()
+    }
+  }
+
+  /**
+   * `keyId = base64url(SHA-256(leaf SPKI))` for the enrolled key, matching the
+   * server's derivation. Null when no key is enrolled.
+   */
+  private fun currentKeyId(keyStore: KeyStore): String? {
+    val cert = keyStore.getCertificate(KEY_ALIAS) ?: return null
+    return Base64.encodeToString(
+      MessageDigest.getInstance("SHA-256").digest(cert.publicKey.encoded),
+      Base64.URL_SAFE or Base64.NO_PADDING or Base64.NO_WRAP,
+    )
+  }
+
+  @ReactMethod
+  fun isSupported(promise: Promise) {
+    // Key attestation (setAttestationChallenge) requires API 24+.
+    promise.resolve(Build.VERSION.SDK_INT >= Build.VERSION_CODES.N)
+  }
+
+  @ReactMethod
+  fun getAttestation(
+    challenge: String,
+    promise: Promise,
+  ) {
+    // Key generation can be slow; run off the JS thread. Serialize all Keystore
+    // access so an overlapping handshake cannot corrupt the shared alias.
+    Thread {
+      withKeystoreLock(promise) {
+        val keyAlias = KEY_ALIAS
+        try {
+          if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
+            promise.reject(
+              AttestCode.UNSUPPORTED.code,
+              "Key attestation requires Android 7.0 (API 24) or later",
+            )
+            return@withKeystoreLock
+          }
+
+          // getAttestation is only called when (re-)enrollment is required, so
+          // a leftover key under the stable alias is stale; delete it first.
+          try {
+            val existing = KeyStore.getInstance("AndroidKeyStore")
+            existing.load(null)
+            existing.deleteEntry(keyAlias)
+          } catch (ignored: Exception) {
+            // Best effort.
+          }
+
+          val generator =
+            KeyPairGenerator.getInstance(
+              KeyProperties.KEY_ALGORITHM_EC,
+              "AndroidKeyStore",
+            )
+
+          // The challenge's UTF-8 bytes are bound into the attestation
+          // extension; the server compares them against the challenge it
+          // issued. Builds the spec, optionally requesting a StrongBox-backed
+          // key. A StrongBox (dedicated secure element, e.g. Pixel Titan M) key
+          // attests at `attestationSecurityLevel = strongBox`, which the info
+          // server maps to `secureElement`; a plain TEE key attests as
+          // `trustedEnvironment` -> `hardware`.
+          fun buildSpec(strongBox: Boolean): KeyGenParameterSpec {
+            val builder =
+              KeyGenParameterSpec
+                .Builder(
+                  keyAlias,
+                  KeyProperties.PURPOSE_SIGN or KeyProperties.PURPOSE_VERIFY,
+                ).setAlgorithmParameterSpec(ECGenParameterSpec("secp256r1"))
+                .setDigests(KeyProperties.DIGEST_SHA256)
+                .setAttestationChallenge(challenge.toByteArray(Charsets.UTF_8))
+            if (strongBox && Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+              builder.setIsStrongBoxBacked(true)
+            }
+            return builder.build()
+          }
+
+          // Prefer the highest assurance (StrongBox / secure element) and fall
+          // back to the TEE only when this device has no StrongBox.
+          val wantStrongBox = Build.VERSION.SDK_INT >= Build.VERSION_CODES.P
+          try {
+            generator.initialize(buildSpec(wantStrongBox))
+            generator.generateKeyPair()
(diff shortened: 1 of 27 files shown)


j0ntz:
Review note from an automated pass (sonnet-5 subagents + adversarial verification) run per the Asana task. Two info-server findings are fixed on #158; this one is in your native concurrency design, so I am raising it rather than changing it.

**`getAttestation` can issue a second `attestKey` for a key the first call may already have consumed** (`ios/edge/EdgeAttestation.swift`, the `operationTimeout` path).

Sequence:
1. Handshake A generates key K, stores it as `pendingKeyId`, calls `attestKey(K)`.
2. `attestKey(K)` runs long; `done.wait` hits the 120s `operationTimeout`, A rejects with `timeout`, and the `serialQueue` closure returns. Per the comment there, `pendingKeyId` is deliberately left in place because "the attestation never completed, so the key was never consumed".
3. Handshake B loads `pendingKeyId` K and calls `attestKey(K)` again, while A's call is still outstanding with Apple.
4. If A's call then succeeds, its completion finds `promise.hasSettled`, so it drops the attestation and calls `clearPendingKeyId(ifMatches: K)`. B's `attestKey(K)` fails because a key can only be attested once.

The premise in step 2 is the part worth a second look: after a *timeout* (as opposed to an explicit retryable error) the key may in fact have been consumed. The outcome is bounded and self-healing the way the adjacent races in that file are (B fails, the pending slot clears, the next handshake generates a fresh key and enrolls), so this costs one wasted attestation and one extra round trip, not a stuck device.

Both obvious remedies have real costs you already reasoned about in that file, which is why I did not pick one:
- Reuse the pending key only after an explicit retryable error (not after a timeout): avoids the double-attest, but burns a key per timeout, and Apple treats keys as a limited resource.
- Let the late callback enroll the key anyway: recovers the attestation, but your comment is right that the server never saw it, so the next handshake would spend an assertion round trip and re-attest regardless.

No action needed if you consider it in the same bounded-and-self-healing class as the `ifMatches` windows; in that case a line in the comment saying the timeout path may double-attest would keep the reasoning complete.

paullinator:
Agreed — the timeout-path premise was slightly wrong. After `operationTimeout` the key may already have been consumed by a still-outstanding `attestKey`, so the next handshake can double-attest and waste one round trip. Same bounded/self-healing class as the `ifMatches` windows: B fails, pending clears, the next attempt generates a fresh key.

Pushed a comment-only fixup clarifying that. Not changing the policy — clearing pending on timeout burns keys; enrolling from the late callback doesn’t help because the server never saw that attestation.

src/util/attestation.ts
```diff
@@ -0,0 +1,681 @@
...
+  // produced - while still accepting a late valid JWT when nothing is cached
+  // (the newer attempt may have failed into backoff).
+  const attempt: HandshakeAttempt = {
+    generation: ++handshakeGeneration,
+    usedAttestation: false,
+    countedFailure: false
+  }
+  const handshake: Promise<void> = performHandshake(attempt)
+    .then(freshToken => {
+      if (freshToken == null) {
```
j0ntz, line 590: `unsupported = true` is set here without a generation guard, unlike the branch directly below it and the whole sibling `.catch`. A watchdog-retired attempt can therefore disable the engine permanently.

Sequence: attempt A hangs in `await EdgeAttestation.isSupported()` past `HANDSHAKE_WATCHDOG_MS` (150s). The watchdog bumps `handshakeGeneration` and frees `inFlight`. Attempt B then runs, caches a live token and schedules its refresh. A finally resolves falsy, falls into this branch, and sets `unsupported = true` without checking whether it is still the current attempt. `runHandshake()` returns immediately on that flag, and the only reset is `resetAttestationForTests`, so the engine stays inert for the rest of the session. Once B's token expires, every gated Simplex and Banxa call goes out unattested with nothing logged.

The other `undefined` return, `EdgeAttestation == null`, is synchronous and settles long before the watchdog can fire, so a hung `isSupported()` is the only way in. Narrow, but the failure is silent and session-long, and the fix matches what the neighbouring branches already do:

```ts
if (freshToken == null) {
  if (attempt.generation !== handshakeGeneration) return
  unsupported = true
  return
}
```

android/app/src/main/java/co/edgesecure/app/EdgeAttestationModule.kt
```diff
@@ -0,0 +1,306 @@
...
+  private fun withKeystoreLock(
+    promise: Promise,
+    body: () -> Unit,
+  ) {
+    try {
+      if (!keystoreLock.tryLock(LOCK_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
+        promise.reject(AttestCode.LOCK_TIMEOUT.code, "Timed out waiting for the Keystore lock")
+        return
+      }
+    } catch (interrupted: InterruptedException) {
```
j0ntz, line 90: `InterruptedException` is caught without restoring the flag via `Thread.currentThread().interrupt()`, so the interruption signal is cleared and lost. Nothing reaches this today, as the comment just below notes these are plain unreferenced threads, and the call is correctly rejected as a lock timeout either way. Restoring the flag would make it safe ahead of the executor or coroutine migration that same comment anticipates, where interruption becomes the normal cancellation path.

src/util/attestation.ts
```diff
@@ -0,0 +1,681 @@
...
+ * Validate an attest/assert token response. Both `token` and `expiresIn` are
+ * validated; a malformed response throws and is treated as a failed handshake
+ * (a lifetime that is non-finite, non-positive, or at or under `CLOCK_SKEW_MS`
+ * would otherwise cache a token no caller can use). The parsed token is returned
+ * to the caller rather than cached directly, so `runHandshake` can drop a stale
+ * (watchdog-released) result before it clobbers a fresher token.
+ *
+ * `serverTime` is read only for the skew warning; it never enters `expiresMono`.
+ */
+const parseTokenResponse = (json: unknown): CachedToken => {
```
j0ntz, line 234: `parseTokenResponse` (and `fetchChallenge`) hand-parse the response with `as { token?: unknown; ... }` casts plus manual `typeof` checks. Every other gated response in this PR goes through a cleaner: `asInfoRollup` in network.ts, `asInfoCreateHmacResponse` in banxaProvider, `asInfoJwtSignResponse` in simplexProvider. This module imports `cleaners` nowhere. A cleaner here would give the same validation, plus a derived type other modules could import instead of re-describing the shape.

src/plugins/gui/providers/simplexProvider.ts
```diff
@@ -340,21 +342,30 @@ export const simplexProvider: FiatProviderFactory = {
...
+        const attestationToken = await getAttestationToken()
         const response = await fetchInfo(
           'v1/jwtSign/simplex',
           {
             method: 'POST',
-            headers: { 'Content-Type': 'application/json' },
+            headers: {
+              'Content-Type': 'application/json',
+              ...(attestationToken != null
+                ? { 'x-attestation-token': attestationToken }
```
j0ntz: The `getAttestationToken()` call plus conditional `x-attestation-token` spread is repeated verbatim at 5 sites: here, simplexProvider.ts:458, banxaProvider.ts:1029, banxaRampPlugin.ts:450, and simplexRampPlugin.ts:305. Pulling it into one helper would mean a change to the header name, a second gated header, or the null-handling policy lands once rather than five times. A site missed in that edit silently reverts to unattested requests, which is the kind of regression that does not announce itself.

src/util/attestation.ts
```diff
@@ -0,0 +1,681 @@
...
+  await new Promise(resolve => setTimeout(resolve, ms))
+}
+
+/**
+ * Arm the background timer, replacing whatever was pending. Every path that
+ * declines to start a handshake has to come through here: the engine has no
+ * other clock, so a path that neither starts a handshake nor arms a timer stalls
+ * it until the next gated call (see `runHandshake`).
+ */
+const armTimer = (delayMs: number): void => {
```
j0ntz, line 457: This `armTimer`/`refreshTimer` scheduler covers the same run-then-reschedule-with-a-variable-delay shape as `makePeriodicTask` in `src/util/PeriodicTask.ts`, which already provides start/stop/setDelay and backs FioService, ActionQueueService and LoanManagerService. Reusing it would drop the timer-lifecycle bookkeeping this module currently has to carry and test on its own, which matters given the comment above warns that any path forgetting to call `armTimer` stalls the whole engine.

j0ntz (commented):
Review pass over the client attestation engine. One reachable correctness bug in the handshake completion handler, one latent Android concurrency note, and three structural suggestions. Each was traced through the code before being written up.

src/util/attestation.ts
paullinator: Agreed, and fixed — this was the real one, good catch.

I traced the whole chain to be sure: the watchdog does bump `handshakeGeneration` and free `inFlight`, `runHandshake` returns immediately on the flag, and `resetAttestationForTests` really is the only thing that clears it. Your narrowing holds too — `performHandshake` returns `undefined` in exactly two places, and the `EdgeAttestation == null` one is synchronous, so it settles on a microtask long before the 150s watchdog. A hung `isSupported()` is the only way in.

Took your patch as written. Added a test, `does not retire the engine when a watchdog-retired isSupported finally says no`, which hangs attempt A inside `isSupported()`, lets the watchdog retire it, has B cache a live token, then answers A with `false` and asserts the engine still refreshes. Without the guard it fails on the `isSupported` call count not growing, so the engine really does go inert for the rest of the session.

android/app/src/main/java/co/edgesecure/app/EdgeAttestationModule.kt
paullinator: Agreed, taking it. Added `Thread.currentThread().interrupt()` before the reject, with a short note on why: catching clears the flag, and under an executor or coroutine the layer above reads it to see the cancellation, so swallowing it would strand a pool thread that should be winding down.

You're right that nothing reaches it today, but it's one line and the comment directly below already anticipates the migration that makes it reachable. ktlint is clean on the file.

src/util/attestation.ts
paullinator: Done. `fetchChallenge` and `parseTokenResponse` now go through `asChallengeResponse` and `asTokenResponse`, so this module matches `asInfoRollup`, `asInfoCreateHmacResponse` and `asInfoJwtSignResponse` rather than hand-rolling `typeof` checks.

Two things worth flagging, since they limit how much the cleaner can absorb:

- `asNumber` admits `NaN` and `Infinity`, so the finite/positive check stays hand-written, as does the `lifetimeMs <= CLOCK_SKEW_MS` rejection — that one is business logic rather than shape validation.
- `maybeWarnClockSkew(serverTime)` deliberately runs *before* validation in both functions, so a malformed body still warns about a bad device clock. A whole-object cleaner throws first and would have silently dropped that warning. I pulled `serverTime` into its own `asMaybe(asServerTimeCue)` parse to preserve the ordering.

All 68 existing tests pass, including the malformed-response ones, so the swap is behaviour-preserving.

On the derived type other modules could import: nothing outside `attestation.ts` describes this shape today — `expiresIn` appears nowhere else in `src` — so I kept the cleaners module-local rather than exporting them speculatively.

src/plugins/gui/providers/simplexProvider.ts
paullinator: Done — agreed this was the highest-value of the three structural notes, for exactly the reason you give.

Added `attestedJsonHeaders()` in `attestation.ts` and converted all five sites. It returns the content type plus the conditional `x-attestation-token`, so the header name, a second gated header, or the null-handling policy each change in one place now. I compared the five before converting and the extracted block was byte-identical at every one; what differed (the `3000` timeout arg, the `.catch` wrappers, the body shapes) all sits outside the extraction.

Also added coverage, since grepping showed these five sites had no test touching the header at all: one test asserting the token is attached when the engine has one, and one asserting the header is omitted rather than faked when it doesn't.

src/util/attestation.ts
paullinator: I'm going to decline this one. The shapes do look alike, but `makePeriodicTask` schedules from a different thing than this engine does, and the difference turns out to be load-bearing.

`makePeriodicTask` reschedules when the *task settles*, using whatever `msGap` holds at that moment. `runHandshake` is `(): void` and fire-and-forget — it never awaits `inFlight`. So the task would settle immediately, `resumeWaiting` would arm with the current gap, and the real handshake would finish much later and call `scheduleRefresh` or `scheduleRetryAfterFailure`. Those would call `setDelay` after the timer was already armed, which is a stale-gap race.

`armTimer`'s contract is also "replace whatever is pending with exactly this delay, now". `start()` is a no-op while a timeout is pending, because of the `if (!running && timeout == null)` guard, so every reschedule would become `stop()` followed by `start({ wait })`.

And the delay here is different on every reschedule: refresh lead, doubling backoff, retry floor, and the `waitMs` re-arm. `makePeriodicTask` is built for a fixed gap — all four existing users pass a constant, and exactly one site in the repo (`RampPendingScene.tsx:86`) calls `setDelay` at all.

Net would be more code fighting the abstraction, in the module whose own comment warns that any path failing to re-arm stalls the engine. That's the highest-consequence failure mode in the file, so I'd rather keep the explicit timer. Happy to revisit if you see a shape I'm missing.

j0ntz:
Design doc for the V2 architecture: pushed [8ecca9b38](https://github.com/EdgeApp/edge-react-gui/pull/6137/commits/8ecca9b38) as a fixup on top, replacing docs/APP_ATTESTATION.md with the [App Attestation V2 TDD](https://github.com/EdgeApp/edge-react-gui/blob/8ecca9b38/docs/APP_ATTESTATION.md). It covers both repos: decision record, the engine's timing and backoff invariants, phase history including this week's review round, testing matrix, and glossary, with code quoted at pinned commits so links survive the merge. The previous revision stays linked from the references section, since it keeps the per-file source map and local-testing notes. Generation notes (tdd + no-slop skills) are in the commit message. Drop the commit if you'd rather keep your original doc.

j0ntz (approved):

src/plugins/gui/providers/banxaProvider.ts
```diff
@@ -1019,11 +1020,14 @@ const generateHmac = async (
...
       method: 'POST',
-      headers: { 'Content-Type': 'application/json' },
+      headers: await attestedJsonHeaders(),
       body
     },
     3000
   )
+  // A missing/rejected attestation token returns 403 here; fail loudly rather
+  // than parsing an error body as a signature.
+  if (!response.ok) throw new Error('Banxa failed to create HMAC signature')
```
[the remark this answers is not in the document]
paullinator: The spam loop is real, but it predates this PR, so I'm not treating it as a change introduced here.

On the mechanism, we agree: a sustained failure in the 3s order poll produces a `showError` per tick. The catch falls through to the user-facing branch, `finally` clears `insideInterval`, nothing calls `clearInterval` on the error path, and `showError` has no dedupe — so it's a fresh `AlertDropdown` and a fresh `trackError` every 3 seconds until the webview closes.

Where the report is inaccurate is the claim that `generateHmac` "now throws on any non-OK info-server response." It threw before too. On `develop` there was no `response.ok` check, but the code went straight to `response.json()` and then `asInfoCreateHmacResponse`, which is `asObject({ signature: asString })` — any error body lacking a `signature` string throws a cleaner error. So the poll already spams today for every sustained failure: an info-server 500, network loss, a Banxa outage. Adding the explicit `response.ok` check didn't create the throw; it replaced a late, opaque cleaner `TypeError` with an early, legible message.

What this PR does add is a failure mode that *stays* failed — a device that can't attest, or one in backoff (up to `MAX_BACKOFF_MS`, 30 minutes) — rather than one that's transient. That's a fair concern about which latent bug becomes visible, but it isn't reachable yet: per `docs/APP_ATTESTATION.md`, a provider stored as a bare string is ungated, and providers stay strings until the attesting app is live. Banxa's `createHmac` only begins returning 403 once it's flipped to `{ key, attestationLevel }` in Couch.

Declining the change here and tracking it as a follow-up to land before that rollout flip. Worth noting for whoever picks it up: `src/plugins/ramps/banxa/banxaRampPlugin.ts` has the identical poll-and-`showError` structure, so the fix belongs in both Banxa implementations, and it should be a message-keyed latch rather than a `clearInterval` so transient blips still recover.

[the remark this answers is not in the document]
paullinator: In addition, the threat of a three-second poll that shows an error is only when the user has successfully received a quote and is routed to the web view. In such a case, they would have already successfully made API calls that were successfully attested. So it would be a rare condition under which attestation would fail after being routed to the web view, where it succeeded prior to the web view during quoting time. Such a case is actually worthy of showing an error to the user so that they know something will go wrong in their web flow journey before they continue forward.

PR #6139, opened 2026-08-04
Add per-ABI split APKs to the Android release pipeline

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Requirements

- [ ] Tested on iOS device

No visual changes: build pipeline only.

### Description

Adds per-ABI Android APKs to release builds, distributed through dedicated Zealot channels and driven entirely by deploy-config, with a versionCode scheme that supports uploading them to Google Play as a multi-APK release (Edge uploads self-signed APKs rather than an AAB, since Play App Signing would require sharing the signing key with Google). Direct downloads keep the universal APK.

Measured on a real Jenkins build of this branch (test-gouda cheese build):

| artifact | size | vs universal |
|---|---|---|
| universal APK | 110 MB | |
| arm64-v8a APK | 80.2 MB | -29.8 MB (-27%) |
| armeabi-v7a APK | 77.5 MB | -32.5 MB (-30%) |

The split APKs compress dex the same way the bundletool-derived universal does (AGP stores dex uncompressed at minSdk 28+), which is where most of the saving comes from being preserved; the AAB is hash-identical with and without that packaging change, so Google Play delivery is untouched.

How it works:

- `app/build.gradle` gains a `splits.abi` block gated behind `-PabiSplits`, with `ndk.abiFilters` untouched and always active. A split build emits three APKs: the universal (same size and lib set as a default build, arm ABIs only, verified) plus one per ABI, so the universal stays available across the board. Default invocations (`bundleRelease`, `npm run android:release*`, dev builds) are completely unchanged.
- Branches opt in through an optional `splitArchitectures` list in their android deploy-config block, mapping each ABI to the Zealot channel that receives it:

  ```json
  "android": {
    "master": {
      "splitArchitectures": [
        { "abi": "arm64-v8a", "zealotChannelKey": "<hex>" },
        { "abi": "armeabi-v7a", "zealotChannelKey": "<hex>" }
      ]
    }
  }
  ```

  Branch scoping comes from the config file, so the deploy script itself stays branch-agnostic. Branches without the field, and all maestro builds, skip the feature entirely. Entries without a `zealotChannelKey` are archived but not uploaded.
- For configured branches, `deploy.ts` runs `assembleRelease -PabiSplits` after the existing bundle step (the gradle daemon reuses the compile work, so this only pays for packaging and signing) and archives `<outfile>-arm64-v8a.apk` and `<outfile>-armeabi-v7a.apk` next to the AAB and universal APK, with the same signing config, so cross-updates between any of these artifacts keep working.
- One Zealot channel per ABI means each channel's latest build is always the right architecture. The universal APK keeps uploading to the existing channel and the rsync location for direct downloads.
- Split APKs add per-ABI versionCode offsets (armeabi-v7a +1, arm64-v8a +2) for Google Play multi-APK uploads, which require a unique versionCode per APK. Play serves the highest compatible code, so 64-bit devices get the arm64 APK while the universal keeps the plain build number as the lowest-priority fallback. The build counter advances by 3 per build (gitVersionFile.ts), giving every build a disjoint block of three codes so offsets never collide at any release cadence. Codes stay eight-digit, so the info server's string-compared minBuildNum/maxBuildNum/exactBuildNum rules keep matching, and iOS keeps the raw build number.
- `deploy-config.sample.json` documents the new field and the previously undocumented Zealot fields.

Verified: both gradle configuration modes pass, each split APK contains exactly one `lib/<abi>/` tree (24 .so files each), the split-mode universal matches the default build's size and contains only the two arm ABI trees, versionCodes come out as base/base+1/base+2 on a split build and plain base on a default build (aapt2-checked), `updateVersion.ts` is untouched, so the universal APK and the AAB keep reporting the plain build number, and lint plus the test suite pass.

Ops steps to activate: create one Zealot channel per ABI in the Zealot admin, then add the `splitArchitectures` list to the master block of `deploy-config.json` on the build machine.

---

> [!NOTE]
> **Medium Risk**
> Touches release signing, versionCode allocation, and CI deploy/upload paths; mistakes could break Play uploads or leak credentials in logs, though behavior is opt-in and default builds are unchanged.
> 
> **Overview**
> Adds **optional per-ABI Android release APKs** (arm64-v8a and armeabi-v7a) for sideloading outside Google Play, while the **universal APK** still goes to the main Zealot channel and direct downloads. Branches opt in via **`splitArchitectures`** in deploy-config, each ABI mapped to its own Zealot channel.
> 
> **Gradle** (`android/app/build.gradle`): ABI splits are gated by **`-PabiSplits`** (off by default). Split builds emit universal plus per-ABI APKs, with **per-ABI `versionCode` offsets** (+1 / +2) for Play multi-APK rules. **Dex legacy packaging** is enabled to shrink sideload APKs without changing the AAB.
> 
> **Deploy** (`scripts/deploy.ts`): After the existing bundle/universal step, configured branches run **`assembleRelease -PabiSplits`**, archive split APKs, and upload each to its channel. Zealot uploads use **`callRedacted`** so tokens/keys are masked in CI logs.
> 
> **Build numbers** (`scripts/gitVersionFile.ts`): The counter advances by **3** per build so universal and split `versionCode`s never collide across releases; the universal APK still reports the plain build number for info-server rules.
> 
> CHANGELOG and **`deploy-config.sample.json`** document the new option.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 78adcb7c0183b21100ee2363089622cee5d0c61b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
  - https://app.asana.com/0/0/1217166483159232

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -191,6 +191,13 @@ android {
             useLegacyPackaging true
         }
 
+        // Compress dex, matching how the universal APK that bundletool
+        // derives from the AAB is packaged. AGP stores dex uncompressed
+        // by default at minSdk 28+, which adds ~28 MB to a sideloaded APK:
+        dex {
+            useLegacyPackaging true
+        }
+
         // Edge hacks for zcash and piratechain conflicts:
         resources {
             pickFirst 'compact_formats.proto'
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -105,6 +105,30 @@ android {
         }
     }
 
+    // Edge addition: sideloadable per-ABI APKs for distribution outside
+    // Google Play (Play already serves per-ABI installs from the AAB).
+    // Each one is roughly 31 MB smaller than the universal APK (about
+    // 78-80 MB vs 110 MB as measured). Off by default so normal builds
+    // are unchanged; the deploy script opts in:
+    //   ./gradlew assembleRelease -PabiSplits
+    // A split build emits three APKs: the universal plus one per ABI,
+    // so the universal stays available across the board. The property is
+    // a boolean flag, so -PabiSplits=false (from the command line,
+    // gradle.properties, or CI) disables the feature:
+    def abiSplitsEnabled = project.hasProperty('abiSplits') &&
+        project.property('abiSplits') != 'false'
+    splits {
+        abi {
+            reset()
+            enable abiSplitsEnabled
+            universalApk true
+            // Adding an ABI here also needs an abiVersionOffsets entry
+            // below and a matching supportedAbis entry in
+            // scripts/deploy.ts, which validates the deploy-config:
+            include 'armeabi-v7a', 'arm64-v8a' // Exclude Intel
+        }
+    }
+
     signingConfigs {
         release {
         }
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -129,6 +129,39 @@ android {
         }
     }
 
+    // Edge addition: distinct versionCodes for the split APKs, required
+    // by Google Play multi-APK uploads (each APK in a release needs a
+    // unique versionCode). Play serves the highest compatible code, so
+    // arm64-v8a must outrank armeabi-v7a, and the universal APK keeps
+    // the unmodified base code as the lowest-priority fallback.
+    //
+    // The offsets stay small deliberately. getBuildNumber() returns this
+    // value on Android, and the info server filters promo and info cards
+    // by comparing minBuildNum/maxBuildNum/exactBuildNum against it as
+    // strings, so the universal APK that nearly every user installs must
+    // keep reporting the plain build number. The build counter advances
+    // by 3 per build (see scripts/gitVersionFile.ts), so these offsets
+    // never collide with a neighboring build's codes:
+    def abiVersionOffsets = ['armeabi-v7a': 1, 'arm64-v8a': 2]
+    applicationVariants.all { variant ->
+        variant.outputs.each { output ->
+            def abi = output.getFilter(OutputFile.ABI)
+            if (abi != null) {
+                // Defaulting a missing ABI to 0 would hand that split
+                // APK the universal APK's code, which Play rejects as a
+                // duplicate. Fail the build instead, so an ABI added to
+                // the include list above without an offset here cannot
+                // ship silently:
+                def abiOffset = abiVersionOffsets[abi]
+                if (abiOffset == null) {
+                    throw new GradleException(
+                        "No versionCode offset defined for ABI '${abi}'")
+                }
+                output.versionCodeOverride = variant.versionCode + abiOffset
+            }
+        }
+    }
+
     signingConfigs {
         release {
         }
--- a/scripts/deploy.ts
+++ b/scripts/deploy.ts
@@ -22,6 +22,15 @@ const cutoffDate = new Date()
 cutoffDate.setMonth(now.getMonth() - BUILD_ARCHIVE_MONTHS)
 
 /**
+ * One per-ABI split APK build: the ABI to package, and the Zealot
+ * channel that receives it.
+ */
+interface SplitArchitecture {
+  abi: string
+  zealotChannelKey?: string
+}
+
+/**
  * Things we expect to be set in the config file:
  */
 interface BuildConfigFile {
@@ -46,6 +55,11 @@ interface BuildConfigFile {
   zealotApiToken?: string
   zealotChannelKey?: string
   zealotMaestroChannelKey?: string
+  // Per-ABI split APKs to build, archive, and upload. Android only.
+  // Belongs in a branch block of the android config, so each branch
+  // opts in on its own. Entries without a zealotChannelKey are archived
+  // but not uploaded:
+  splitArchitectures?: SplitArchitecture[]
   hockeyAppId: string
   hockeyAppTags: string
   hockeyAppToken: string
@@ -81,6 +95,7 @@ interface BuildObj extends BuildConfigFile {
   dSymFile: string
   dSymZip: string
   ipaFile: string // Also APK
+  abiApkFiles?: Record<string, string> // Android split APKs, by ABI
 }
 
 interface LatestTestFile {
@@ -519,11 +534,51 @@ function buildAndroid(buildObj: BuildObj): void {
   const universalApk = join(apkPathDir, 'universal.apk')
   buildObj.ipaFile = join(apkPathDir, `${outfile}.apk`)
   fs.renameSync(universalApk, buildObj.ipaFile)
+
+  // Branches configured with splitArchitectures also archive sideloadable
+  // per-ABI APKs for distribution outside Google Play (Play already
+  // serves per-ABI installs from the AAB). Each one is roughly 31 MB
+  // smaller than the universal APK (about 78-80 MB vs 110 MB as
+  // measured). The gradle daemon reuses the compile work from the bundle
+  // task above, so this only pays for packaging and signing. Maestro
+  // builds skip this entirely:
+  const { splitArchitectures } = buildObj
+  if (
+    splitArchitectures != null &&
+    splitArchitectures.length > 0 &&
+    !maestroBuild
+  ) {
+    // The gradle splits block can only emit these ABIs, so anything else
+    // in the config is a typo. This list must match the splits.abi
+    // include list in android/app/build.gradle. Checking before the
+    // build turns a typo into an immediate, clear error rather than a
+    // low-signal ENOENT from the copy below after a multi-minute
+    // assemble, and keeps unvalidated config values out of the path
+    // construction:
+    const supportedAbis = ['arm64-v8a', 'armeabi-v7a']
+    for (const { abi } of splitArchitectures) {
+      if (!supportedAbis.includes(abi)) {
+        throw new Error(`Unsupported abi "${abi}" in splitArchitectures`)
+      }
+    }
+
+    call('./gradlew assembleRelease -PabiSplits')
+    const splitApkDir = join(guiPlatformDir, 'app/build/outputs/apk/release')
+
+    buildObj.abiApkFiles = {}
+    for (const { abi } of splitArchitectures) {
+      const archivedApk = join(archiveDir, `${outfile}-${abi}.apk`)
+      fs.copyFileSync(join(splitApkDir, `app-${abi}-release.apk`), archivedApk)
+      buildObj.abiApkFiles[abi] = archivedApk
+    }
+  }
 }
 
 function buildCommonPost(buildObj: BuildObj): void {
   const {
+    abiApkFiles,
     maestroBuild,
+    splitArchitectures,
     zealotApiToken,
     zealotChannelKey,
     zealotMaestroChannelKey,
@@ -564,14 +619,16 @@ function buildCommonPost(buildObj: BuildObj): void {
     mylog('\nUploaded to HockeyApp')
   }
 
+  // Shared by both Zealot uploads below:
+  const branch = encodeURIComponent(buildObj.repoBranch)
+  const gitCommit = encodeURIComponent(buildObj.guiHash)
+
   // Maestro test builds upload to their own channel so they do not pollute the
   // production channel's release list. Production builds use zealotChannelKey.
   // A maestro build with no zealotMaestroChannelKey configured skips Zealot.
   const channelKey = maestroBuild ? zealotMaestroChannelKey : zealotChannelKey
 
   if (zealotApiToken != null && zealotUrl != null && channelKey != null) {
-    const branch = encodeURIComponent(buildObj.repoBranch)
-    const gitCommit = encodeURIComponent(buildObj.guiHash)
     chdir(buildObj.guiDir)
     const changes = cmd(
       `git diff HEAD^ HEAD CHANGELOG.md | { grep '^+[^+]' || true; }`
@@ -582,10 +639,38 @@ function buildCommonPost(buildObj: BuildObj): void {
       '***********************************************************************\n'
     )
 
-    call(
-      `curl -X POST "${zealotUrl}/api/apps/upload?token=${zealotApiToken}&channel_key=${channelKey}&branch=${branch}&git_commit=${gitCommit}&changelog=${changelog}" -F "file=@${buildObj.ipaFile}"`
+    const token = encodeURIComponent(zealotApiToken)
+    const encodedChannelKey = encodeURIComponent(channelKey)
+    callRedacted(
+      `curl -X POST "${zealotUrl}/api/apps/upload?token=${token}&channel_key=${encodedChannelKey}&branch=${branch}&git_commit=${gitCommit}&changelog=${changelog}" -F "file=@${buildObj.ipaFile}"`,
+      [token, encodedChannelKey]
     )
     mylog('\n*** Upload to Zealot Complete ***')
+  }
+
+  // Architecture-specific APKs go to their own Zealot channels, one per
+  // ABI, so each channel's latest build is always the right architecture.
+  // The universal APK above keeps serving the main channel and direct
+  // downloads. Only branches whose config block lists splitArchitectures
+  // build these at all, and maestro builds never do:
+  if (
+    zealotApiToken != null &&
+    zealotUrl != null &&
+    abiApkFiles != null &&
+    splitArchitectures != null
+  ) {
+    const token = encodeURIComponent(zealotApiToken)
+    for (const { abi, zealotChannelKey: abiChannelKey } of splitArchitectures) {
+      const apkFile = abiApkFiles[abi]
+      if (abiChannelKey == null || apkFile == null) continue
+      const channelKey = encodeURIComponent(abiChannelKey)
+      mylog(`\n\nUploading ${abi} APK to Zealot: ${zealotUrl}`)
+      callRedacted(
+        `curl -X POST "${zealotUrl}/api/apps/upload?token=${token}&channel_key=${channelKey}&branch=${branch}&git_commit=${gitCommit}" -F "file=@${apkFile}"`,
+        [token, channelKey]
+      )
+      mylog(`\n*** Upload of ${abi} APK to Zealot Complete ***`)
+    }
   }
 
   if (buildObj.rsyncLocation != null) {
@@ -713,8 +798,11 @@ function chdir(path: string): void {
   _currentPath = path
 }
 
-function call(cmdstring: string): void {
-  console.log('call: ' + cmdstring)
+/**
+ * Runs a command, inheriting our stdio. Shared by `call` and
+ * `callRedacted` so both stay on the same execution options.
+ */
+function execCommand(cmdstring: string): void {
   childProcess.execSync(cmdstring, {
     encoding: 'utf8',
     timeout: 3600000,
@@ -722,6 +810,35 @@ function call(cmdstring: string): void {
     cwd: _currentPath,
     killSignal: 'SIGKILL'
   })
+}
+
+/**
+ * Like `call`, but logs the command with the given secrets masked, so
+ * tokens and keys do not land in the CI build log. A failed execSync
+ * throws an error whose message embeds the raw command, so the failure
+ * path gets the same masking as the log line.
+ */
+function callRedacted(cmdstring: string, secrets: string[]): void {
+  const redact = (text: string): string => {
+    let out = text
+    for (const secret of secrets) {
+      if (secret !== '') out = out.split(secret).join('<redacted>')
+    }
+    return out
+  }
+
+  console.log('call: ' + redact(cmdstring))
+  try {
+    execCommand(cmdstring)
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error)
+    throw new Error(redact(message))
+  }
+}
+
+function call(cmdstring: string): void {
+  console.log('call: ' + cmdstring)
+  execCommand(cmdstring)
 }
 
 function cmd(cmdstring: string): string {
--- a/scripts/deploy.ts
+++ b/scripts/deploy.ts
@@ -161,8 +161,6 @@ function makeProject(buildObj: BuildObj): void {
     buildObj,
     config[project][buildObj.platformType][buildObj.repoBranch]
   )
-
-  console.log(buildObj)
 }
 
 function makeCommonPost(buildObj: BuildObj): void {
--- a/scripts/gitVersionFile.ts
+++ b/scripts/gitVersionFile.ts
@@ -88,7 +88,28 @@ function updateVersionFile(branch: string, version: string): void {
       const { build: previousBuild } = JSON.parse(result)
       if (typeof previousBuild !== 'number')
         throw new Error(`Invalid previous buildNum ${previousBuild}`)
-      build = Math.max(previousBuild + 1, newBuildNum)
+      // Advance by 3, not 1, so each build owns a block of three
+      // versionCodes: the Android split APKs add per-ABI offsets to the
+      // build number (universal +0, armeabi-v7a +1, arm64-v8a +2, see
+      // app/build.gradle), and Google Play rejects any versionCode it
+      // has ever seen, so consecutive builds must never overlap blocks.
+      // The stride must stay >= the number of APK flavors per build,
+      // which is pinned by the gradle splits include list. Widening the
+      // gap the other way, by giving the splits their own numeric range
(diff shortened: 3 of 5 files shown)


android/app/build.gradle
```diff
@@ -100,8 +100,26 @@ android {
...
+    // default so normal builds are unchanged; the deploy script opts in:
+    //   ./gradlew assembleRelease -PabiSplits
+    splits {
+        abi {
+            reset()
+            enable project.hasProperty('abiSplits')
+            universalApk false
+            include 'armeabi-v7a', 'arm64-v8a' // Exclude Intel
         }
     }
```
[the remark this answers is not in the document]
peachbits: Fixed in cda933aff: the property is now read as a boolean flag, so -PabiSplits=false disables the feature. The abiFilters half of this comment applied to an earlier revision; ndk.abiFilters is unconditional in the current diff and stays active in split builds.

scripts/deploy.ts
```diff
@@ -588,6 +608,31 @@ function buildCommonPost(buildObj: BuildObj): void {
...
+  ) {
+    const branch = encodeURIComponent(buildObj.repoBranch)
+    const gitCommit = encodeURIComponent(buildObj.guiHash)
+    for (const abi of Object.keys(abiApkFiles)) {
+      const abiChannelKey = zealotAbiChannelKeys[abi]
+      if (abiChannelKey == null) continue
+      mylog(`\n\nUploading ${abi} APK to Zealot: ${zealotUrl}`)
+      call(
+        `curl -X POST "${zealotUrl}/api/apps/upload?token=${zealotApiToken}&channel_key=${abiChannelKey}&branch=${branch}&git_commit=${gitCommit}" -F "file=@${abiApkFiles[abi]}"`
+      )
```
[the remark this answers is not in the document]
peachbits: Partially fixed in cda933aff: the token and channel key are now URL-encoded, matching the branch and commit params. Kept the call()/curl pattern for consistency with the existing Zealot upload above, which has the same token-in-logged-command behavior (as do the keystore passwords in the bundletool call). Switching this file to execFileSync with unlogged secrets is worth doing for every secret in deploy.ts at once rather than one call site; leaving that as a follow-up.

scripts/deploy.ts
```diff
@@ -519,11 +534,35 @@ function buildAndroid(buildObj: BuildObj): void {
...
+    !maestroBuild
+  ) {
+    call('./gradlew assembleRelease -PabiSplits')
+    const splitApkDir = join(guiPlatformDir, 'app/build/outputs/apk/release')
+    buildObj.abiApkFiles = {}
+    for (const { abi } of splitArchitectures) {
+      const archivedApk = join(archiveDir, `${outfile}-${abi}.apk`)
+      fs.copyFileSync(join(splitApkDir, `app-${abi}-release.apk`), archivedApk)
+      buildObj.abiApkFiles[abi] = archivedApk
+    }
```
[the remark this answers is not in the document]
peachbits: Fixed: abi values are now validated against the two ABIs the gradle splits block can emit, before any path construction, so a typo or hostile value fails with a clear error naming the bad entry. Folded into the Zealot distribution commit (8c339f2fa). Skipped the separate existsSync check: with the allowlist in place, a missing split APK means the gradle build itself failed, and the loud copyFileSync throw is the right behavior there.

scripts/deploy.ts
```diff
@@ -588,6 +639,32 @@ function buildCommonPost(buildObj: BuildObj): void {
...
+    const branch = encodeURIComponent(buildObj.repoBranch)
+    const gitCommit = encodeURIComponent(buildObj.guiHash)
+    const token = encodeURIComponent(zealotApiToken)
+    for (const { abi, zealotChannelKey: abiChannelKey } of splitArchitectures) {
+      const apkFile = abiApkFiles[abi]
+      if (abiChannelKey == null || apkFile == null) continue
+      const channelKey = encodeURIComponent(abiChannelKey)
+      mylog(`\n\nUploading ${abi} APK to Zealot: ${zealotUrl}`)
+      call(
+        `curl -X POST "${zealotUrl}/api/apps/upload?token=${token}&channel_key=${channelKey}&branch=${branch}&git_commit=${gitCommit}" -F "file=@${apkFile}"`
```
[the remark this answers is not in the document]
peachbits: Fixed in 2a6243239: the per-ABI upload now logs through a callRedacted() helper that masks the token and channel key before printing the command, so neither reaches the CI log. The command itself still runs verbatim.

scripts/updateVersion.ts
```diff
@@ -51,9 +51,17 @@ async function updateAndroid(
...
+  // set in app/build.gradle). Google Play multi-APK uploads require a
+  // unique versionCode per APK, and same-day builds only advance the
+  // build number by 1, so the offsets need this stride to never collide
+  // across releases. The universal APK and the AAB keep the base code,
+  // so every artifact of a release outranks every artifact of the
+  // release before it:
   gradle = gradle.replace(
     /versionCode [0-9]+/g,
-    `versionCode ${versionFile.build}`
+    `versionCode ${versionFile.build * 10}`
```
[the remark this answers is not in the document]
peachbits: Good catch, and confirmed against the code: getBuildNumber() returns the Android versionCode and infoUtils.ts string-compares it to minBuildNum/maxBuildNum/exactBuildNum, so a nine-digit value would have desynced Android from iOS and stopped every existing eight-digit exactBuildNum rule from matching.

Fixed in 248689e1c by dropping the times-10 multiplier entirely. updateVersion.ts is back to zero diff against develop, so the universal APK and the AAB report the plain build number exactly as before and promo/info targeting is untouched. Only the split APK outputs carry offsets, and they stay adjacent and eight-digit: base, base+1, base+2 (aapt2-verified).

Remaining known constraint, now documented in the commit: a Play multi-APK release must not be built from builds fewer than three apart, since consecutive build numbers differ by 1. Zealot and sideload distribution are unaffected.

android/app/build.gradle
```diff
@@ -105,6 +105,48 @@ android {
...
+    // by comparing minBuildNum/maxBuildNum/exactBuildNum against it as
+    // strings, so the universal APK that nearly every user installs must
+    // keep reporting the plain build number:
+    def abiVersionOffsets = ['armeabi-v7a': 1, 'arm64-v8a': 2]
+    applicationVariants.all { variant ->
+        variant.outputs.each { output ->
+            def abi = output.getFilter(OutputFile.ABI)
+            if (abi != null) {
+                output.versionCodeOverride =
+                    variant.versionCode + abiVersionOffsets.get(abi, 0)
```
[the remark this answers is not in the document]
peachbits: Fixed in 50e5e0d38. Rather than reintroducing the times-10 versionCode stride (which broke promo/info build targeting: getBuildNumber() returns the versionCode and the info server string-compares eight-digit rules against it), the build counter itself now advances by 3 per build in gitVersionFile.ts. Every build owns a disjoint block of three codes, so the +1/+2 offsets can never collide with a neighboring build at any release cadence, codes stay eight-digit and date-shaped, and the cross-channel downgrade window is gone. Costs: same-day capacity drops from 99 builds to 33, and iOS build numbers, which share the counter, skip by 3.

scripts/deploy.ts
```diff
@@ -588,6 +639,33 @@ function buildCommonPost(buildObj: BuildObj): void {
     mylog('\n*** Upload to Zealot Complete ***')
   }
 
+  // Architecture-specific APKs go to their own Zealot channels, one per
+  // ABI, so each channel's latest build is always the right architecture.
+  // The universal APK above keeps serving the main channel and direct
+  // downloads. Only branches whose config block lists splitArchitectures
+  // build these at all, and maestro builds never do:
+  if (
```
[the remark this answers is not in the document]
peachbits: Fixed in cd7df354b: the universal upload now goes through callRedacted() with the token and channel key URL-encoded, same as the per-ABI uploads.

scripts/deploy.ts
```diff
@@ -46,6 +55,11 @@ interface BuildConfigFile {
   zealotApiToken?: string
   zealotChannelKey?: string
   zealotMaestroChannelKey?: string
+  // Per-ABI split APKs to build, archive, and upload. Android only.
+  // Belongs in a branch block of the android config, so each branch
+  // opts in on its own. Entries without a zealotChannelKey are archived
+  // but not uploaded:
+  splitArchitectures?: SplitArchitecture[]
```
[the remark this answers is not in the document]
peachbits: Fixed in 6d0862fbe: the console.log(buildObj) dump is removed. This was pre-existing and printed every deploy-config secret (Zealot token and channel keys, keystore password, upload service credentials) into the CI log on each run, so worth noting for ops: historical Jenkins logs contain these values, and rotating the Zealot token and channel keys after this lands would be prudent.

scripts/deploy.ts
```diff
@@ -713,6 +792,25 @@ function chdir(path: string): void {
   _currentPath = path
 }
 
+/**
+ * Like `call`, but logs the command with the given secrets masked, so
+ * tokens and keys do not land in the CI build log.
+ */
+function callRedacted(cmdstring: string, secrets: string[]): void {
```
j0ntz: **Secrets still reach the CI log on failure** (CONFIRMED)

`callRedacted` only masks the secret in the `console.log` line. The raw `cmdstring`, with the plaintext token and channel key, is still passed straight into `childProcess.execSync`, so it does not actually stop secrets from reaching CI logs.

When the Zealot upload curl fails (bad token, network blip, or the 3600000ms timeout), `execSync` throws a Node `Error` whose message is `Command failed: <cmdstring>` with the unredacted URL embedded. That exception is not caught anywhere in `buildCommonPost`, so it propagates to the top level and Node prints the full stack trace, including the plaintext `zealotApiToken` and `zealotChannelKey`, to the CI console.

This defeats the purpose the helper was introduced for, per its own doc comment: "so tokens and keys do not land in the CI build log".

android/app/build.gradle
```diff
@@ -105,6 +105,50 @@ android {
...
+    // strings, so the universal APK that nearly every user installs must
+    // keep reporting the plain build number. The build counter advances
+    // by 3 per build (see scripts/gitVersionFile.ts), so these offsets
+    // never collide with a neighboring build's codes:
+    def abiVersionOffsets = ['armeabi-v7a': 1, 'arm64-v8a': 2]
+    applicationVariants.all { variant ->
+        variant.outputs.each { output ->
+            def abi = output.getFilter(OutputFile.ABI)
+            if (abi != null) {
+                output.versionCodeOverride =
```
j0ntz: **Per-ABI versionCode offset breaks build-number comparisons** (CONFIRMED)

The per-ABI versionCode offset (+1 armeabi-v7a, +2 arm64-v8a) makes `getBuildNumber()` diverge from the actual release build number for users on sideloaded split APKs, breaking every call site that assumes `buildNumber` matches the tracked release build.

Concretely: an ops engineer configures an info/promo card with `exactBuildNum` (or `minBuildNum`/`maxBuildNum`) equal to the release's real build number from `release-version.json`/CHANGELOG/Zealot. `src/util/infoUtils.ts:88` does `if (exactBuildNum != null && exactBuildNum !== buildNumber) continue`, comparing that configured value against react-native-device-info's `getBuildNumber()`.

Universal-APK users match and see the card as intended, but any user who sideloaded the arm64-v8a or armeabi-v7a split APK reports `buildNumber+2` or `+1`, so the exact-match check silently fails and the card (which could be a critical warning or forced-update notice) never renders for them, with no error surfaced anywhere.

android/app/build.gradle
```diff
@@ -134,6 +178,13 @@ android {
             useLegacyPackaging true
         }
 
+        // Compress dex, matching how the universal APK that bundletool
+        // derives from the AAB is packaged. AGP stores dex uncompressed
+        // by default at minSdk 28+, which adds ~28 MB to a sideloaded APK:
```
j0ntz, line 196: **`dex.useLegacyPackaging` applies to every variant, not just ABI-split builds** (CONFIRMED)

This is placed in the top-level `packagingOptions` block, which applies to every build variant, not just the opt-in ABI-split release builds the surrounding comments describe.

Unlike the `splits { abi { enable abiSplitsEnabled } }` block a few lines above (deliberately gated so "normal builds are unchanged" when `-PabiSplits` is not passed), `packagingOptions.dex` is unconditional. Every local debug build, every branch's normal release build, and every maestro test APK now gets legacy (compressed) dex packaging even though nobody asked for the ABI-split feature on those builds, silently changing install-time dex extraction behavior and APK layout repo-wide instead of only for the master branch's opt-in split APKs.

scripts/gitVersionFile.ts
```diff
@@ -88,7 +88,17 @@ function updateVersionFile(branch: string, version: string): void {
       const { build: previousBuild } = JSON.parse(result)
       if (typeof previousBuild !== 'number')
         throw new Error(`Invalid previous buildNum ${previousBuild}`)
-      build = Math.max(previousBuild + 1, newBuildNum)
+      // Advance by 3, not 1, so each build owns a block of three
```
j0ntz, line 91: **The +3 stride applies to every branch and platform, and can roll the day field** (CONFIRMED)

The build-number stride was raised from +1 to +3 for every branch's counter, not just the one branch (master) that actually builds ABI splits. That reduces every branch's same-day build capacity and lets the day-encoded build number roll into the next day's numeric range on high-cadence branches.

`pickBuildNumber()` encodes the day in a fixed 2-digit field (`day*100+counter`). On a branch like develop that can get more than 33 CI builds in one day, incrementing by 3 each time pushes the counter past 99, so `day*100+100` numerically equals `(day+1)*100+0`. A build actually cut on day N gets a build number that reads as day N+1's build 0 anywhere it is decoded or displayed (release notes, Zealot upload notes, CHANGELOG), even though that branch never uses the per-ABI versionCode offsets this stride exists to protect.

Same root cause also at `scripts/gitVersionFile.ts:101`.

scripts/deploy.ts
```diff
@@ -519,11 +532,47 @@ function buildAndroid(buildObj: BuildObj): void {
...
+    splitArchitectures != null &&
+    splitArchitectures.length > 0 &&
+    !maestroBuild
+  ) {
+    call('./gradlew assembleRelease -PabiSplits')
+    const splitApkDir = join(guiPlatformDir, 'app/build/outputs/apk/release')
+
+    // The gradle splits block can only emit these ABIs, so anything else
+    // in the config is a typo. Checking up front turns that into a clear
+    // error instead of a low-signal ENOENT from the copy below, and keeps
```
j0ntz: **ABI validation runs after the gradle build, not before it** (CONFIRMED)

The comment claims an up-front check, but the validation for `splitArchitectures` runs after the build.

With a typo'd ABI in a branch's `deploy-config.json` (e.g. `"arm64"` instead of `"arm64-v8a"`), `buildAndroid()` calls `./gradlew assembleRelease -PabiSplits` first, a multi-minute release build, and only after it completes does the `supportedAbis.includes(abi)` loop run and throw `Unsupported abi "arm64" in splitArchitectures`. The deploy still fails, but only after burning the full build time this check exists to avoid, so a Jenkins release pipeline wastes CI minutes on every abi-typo deploy attempt instead of erroring immediately.

Same root cause also at `scripts/deploy.ts:548`.

android/app/build.gradle
```diff
@@ -105,6 +105,50 @@ android {
...
+    // keep reporting the plain build number. The build counter advances
+    // by 3 per build (see scripts/gitVersionFile.ts), so these offsets
+    // never collide with a neighboring build's codes:
+    def abiVersionOffsets = ['armeabi-v7a': 1, 'arm64-v8a': 2]
+    applicationVariants.all { variant ->
+        variant.outputs.each { output ->
+            def abi = output.getFilter(OutputFile.ABI)
+            if (abi != null) {
+                output.versionCodeOverride =
+                    variant.versionCode + abiVersionOffsets.get(abi, 0)
```
j0ntz: **`Map.get(key, default)` inserts the default into the map** (PLAUSIBLE)

`abiVersionOffsets.get(abi, 0)` uses Groovy's `Map.get(key, defaultValue)` GDK method, whose implementation is `if (!map.containsKey(key)) map.put(key, defaultValue); return map.get(key)`. It mutates the map on a miss, unlike Java's `getOrDefault`.

This is dormant today because the only non-null ABI filters (`armeabi-v7a`, `arm64-v8a`) already exist as keys. But if a future edit adds a third ABI to `splits.abi.include` (e.g. x86_64) without adding a matching `abiVersionOffsets` entry, this call silently writes a 0 offset for it, giving that split APK the same versionCode as the universal APK. Play Console's multi-APK upload (or the Zealot upload of both files) would then reject the release for a duplicate versionCode, with no compile-time or obvious runtime signal pointing here.

scripts/deploy.ts
```diff
@@ -519,11 +532,47 @@ function buildAndroid(buildObj: BuildObj): void {
...
+    // unvalidated config values out of the path construction:
+    const supportedAbis = ['arm64-v8a', 'armeabi-v7a']
```
j0ntz, line 556: **ABI allowlist duplicates the gradle include list** (CONFIRMED, cleanup)

`android/app/build.gradle`'s `splits.abi.include 'armeabi-v7a', 'arm64-v8a'` and this `const supportedAbis = ['arm64-v8a', 'armeabi-v7a']` encode the same set independently, with no shared source of truth.

If a future change adds a third ABI (e.g. x86_64) to the gradle include list but this allowlist is not updated in the same PR, every branch that adds that ABI to `splitArchitectures` gets rejected with "Unsupported abi" even though gradle would happily build it, and diagnosing it means tracing the mismatch across two unrelated files.

scripts/deploy.ts
j0ntz: **`callRedacted` duplicates the entire `execSync` options block from `call`** (CONFIRMED, cleanup)

`call` (line 814) and `callRedacted` both hardcode the same five-key options object (`encoding: 'utf8'`, `timeout: 3600000`, `stdio: 'inherit'`, `cwd: _currentPath`, `killSignal: 'SIGKILL'`).

If the timeout or another execution option ever needs to change (say, raising it for a slower CI runner), only one of the two is likely to get updated, silently reintroducing the old behavior at whichever call site was missed. Extracting a shared `execCommand(cmdstring)` helper called by both, with `callRedacted` adding only its masked `console.log`, would remove the duplication.

scripts/deploy.ts
```diff
@@ -582,12 +631,42 @@ function buildCommonPost(buildObj: BuildObj): void {
...
     mylog('\n*** Upload to Zealot Complete ***')
   }
 
+  // Architecture-specific APKs go to their own Zealot channels, one per
+  // ABI, so each channel's latest build is always the right architecture.
+  // The universal APK above keeps serving the main channel and direct
+  // downloads. Only branches whose config block lists splitArchitectures
+  // build these at all, and maestro builds never do:
+  if (
+    zealotApiToken != null &&
```
j0ntz, line 655: **`branch` and `gitCommit` are re-encoded, duplicating lines 620-621** (CONFIRMED, cleanup)

`const branch = encodeURIComponent(buildObj.repoBranch)` and `const gitCommit = encodeURIComponent(buildObj.guiHash)` recompute the exact expressions already bound at lines 620-621 in the same function.

It costs nothing at runtime, but a future change to how branch/commit are encoded (e.g. adding a fallback for an empty `guiHash`) has to be made in two places, and it is easy to update one and miss the other, leaving the two upload paths silently inconsistent.

j0ntz (commented):
## Automated code review (xhigh)

This PR adds per-ABI split APK builds (gradle splits plus per-ABI versionCode offsets), a +3 build-number stride, and a redacted-command helper for Zealot uploads.

The highest-value findings: a secret-leak gap in `callRedacted` (the unredacted command still reaches `execSync`, so a failed upload prints the token in the CI stack trace), the per-ABI versionCode offset making `getBuildNumber()` diverge from the release build number that info-card `min`/`max`/`exactBuildNum` checks compare against, and two changes that were meant to be opt-in but apply globally: `dex.useLegacyPackaging` in the top-level `packagingOptions` block hits every variant, and the +3 stride cuts same-day build capacity for every branch and platform including iOS.

The rest are lower-severity ordering and duplication cleanups: ABI validation runs after the multi-minute gradle build despite a comment claiming it fails fast, and the ABI allowlist, execSync options, and branch/commit encoding are each duplicated with no shared source of truth.

9 findings below, ranked, each independently verified by a separate agent (8 CONFIRMED, 1 PLAUSIBLE). 0 refuted.

android/app/build.gradle
peachbits: will not fix - the play store doesn't allow separate uploads to share a build number. Card will need to be created using build number min/max ranges or multiple cards targeting different exact build numbers.

android/app/build.gradle
peachbits: The trade off is taking a few extra milliseconds to install the app the first time or add a bunch of complexity to the build.gradle. I opted for simpler build.gradle

scripts/deploy.ts
peachbits: Fixed in 078bc7b87. `callRedacted` now runs the command inside a try/catch and
re-throws with the same masking applied to the error message, so the
`Command failed: <cmd>` text that execSync embeds no longer carries the token or
channel key into the CI stack trace. Verified against a deliberately failing
curl: the raw execSync error prints `token=<the real token>`, the wrapped one
prints `token=<redacted>`, and the secret appears nowhere in the stack.

scripts/gitVersionFile.ts
peachbits: Keeping the +3 stride, with the reasoning now written into the code (891ab6cf0)
instead of left implicit.

The stride can't be traded for a separate numeric range for the splits
(`build * 10 + offset` with +1 here): the next build's universal APK would then
carry a lower versionCode than the previous build's split APKs, which Play treats
as a downgrade for that APK, and the universal has to keep reporting the plain
build number because `getBuildNumber()` returns the versionCode and infoUtils
string-compares it against min/max/exactBuildNum. A block of >= 3 codes per build
is the smallest scheme that satisfies both constraints.

On the day-field roll: the bleed isn't new, +3 only moves the threshold from 99
same-day builds on a branch to 33. When it does happen the numbers stay unique
and strictly increasing (`Math.max(previousBuild + 3, newBuildNum)`), so nothing
that depends on ordering or uniqueness breaks — only the date reading of a
display string is off, and only past 33 builds in one day on one branch. iOS
skipping by 3 is cosmetic for the same reason. Both costs are now spelled out in
the comment.

scripts/deploy.ts
peachbits: Fixed in 078bc7b87: the `supportedAbis` loop now runs before
`./gradlew assembleRelease -PabiSplits`, so an abi typo fails immediately instead
of after a multi-minute assemble. The comment was right about the intent and
wrong about the placement; it now describes what the code actually does.

android/app/build.gradle
peachbits: Fixed in 891ab6cf0. Switched to the subscript form (`abiVersionOffsets[abi]`),
which is a plain get with no insert, and — more to the point — an ABI with no
offset now throws a `GradleException` instead of silently defaulting to 0 and
duplicating the universal APK's versionCode. Verified by temporarily dropping
`arm64-v8a` from the map: `./gradlew :app:tasks -PabiSplits` fails with
`No versionCode offset defined for ABI 'arm64-v8a'`. With the map intact, both
configuration modes still configure cleanly.

scripts/deploy.ts
peachbits: There's no clean shared source of truth between a Groovy build file and a TS
deploy script, and deriving the list at runtime (globbing the output dir) would
give up the up-front validation the ordering finding above asks for. Left as two
lists, with a cross-reference comment added on both sides (2a1001489, 078bc7b87)
so adding an ABI points at the `include` list, the `abiVersionOffsets` map, and
`supportedAbis` together.

scripts/deploy.ts
peachbits: Fixed in 078bc7b87: `call()` and `callRedacted()` now both go through a shared
`execCommand()` that owns the options object. `callRedacted` adds only the masked
log line and the masked error.

scripts/deploy.ts
peachbits: Fixed in 078bc7b87: `branch` and `gitCommit` are computed once above both upload
blocks. `token` stays per-block, since it depends on the `zealotApiToken`
null-narrowing in each condition.

j0ntz (approved):

PR #6141, opened 2026-08-06
Ramps deep link entry and affiliate provider priority

j0ntz:
### Technical Design Document

[ramps-deeplink-provider-priority.md](https://github.com/EdgeApp/edge-react-gui/blob/jon/ramps-deeplink-provider-priority/src/docs/ramps-deeplink-provider-priority.md)

### Description

Gives the ramps buy/sell flow a deep link entry point and provider priority.

Part A, link-scoped flow access. New `edge://buy[/<providerId>[/<paymentType>]]` and `edge://sell[...]` links (plus the `https://deep.edge.app/...` form) open `RampCreateScene` with an optional provider and payment type pinned to the top of the quote results. The pin lives in navigation params only: no `saveAccountReferral` call, and `?af=` attribution keeps its existing independent behavior. It is scoped to the flow the link opened, so stepping to the option list and back keeps it while leaving the tab clears it. A pin that matches no quote logs and falls back to the normal ordering, which is what `/buy/moonpay/cashapp` does while the info server `rampQuoteFilter` disables Cash App. An unrecognized payment type in a link is dropped rather than failing the link, so a stale partner URL still opens the flow.

Part B, affiliate provider priority. `useRampPreferredProviders` reads the existing `promoCards2` `pluginPromotions.preferProviders` through `filterInfoCards` (not `getDisplayInfoCards`), matched against `accountReferral`. Reading the ungated path is what makes silent priority config work: a card with `localeMessages: {}` and no `ctaButton` never renders in the carousel but still carries its behavior payload. No info server changes.

Both inputs feed one comparator (`compareRampQuotes`) shared by `useRampQuotes` and `RampSelectOptionScene`, so the payment-type group order and the provider order inside each group agree. Anything that makes a claim about the RATE stays out of that ordering: the select scene's "Best Rate" badge and the create scene's exchange rate (which also feeds the light-account purchase limit) both go through `getBestRateRampQuote`, so a pinned link changes which option the user lands on first, never what the app reports as the market rate.

Two product decisions shape the ordering rules. The "Best Rate" badge is display-honest: it renders only beside the number it describes, so under a provider pin or an active affiliate preference the badge is deliberately absent rather than relocated (a marker pointing at a competitor works against the promotion the session exists to deliver). And the pins and preferences outrank the has-amounts tier for both links and affiliate configs: an intentional promotion wins outright, so `edge://buy/libertyx` visibly surfaces LibertyX's amount-less placeholder card at the top, while an unpreferred placeholder still sinks below priced quotes.

The design doc above carries both rules with worked examples: section 5.6 and decision 10.8 for the badge, decision 10.7 for the ordering tier.

The design doc above carries the paste-ready CouchDB `promoCards2` blocks and the partner link instructions.

Asana: https://app.asana.com/0/1215088146871429/1217224633446931

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Requirements

- [x] Tested on iOS device

---

> [!NOTE]
> **Medium Risk**
> Touches buy/sell quote ordering and deep-link navigation for a commercially sensitive flow, but only reorders results and does not change quote fetching, payments, or auth.
> 
> **Overview**
> Adds **deep-link entry** into the ramps buy/sell flow via `edge://buy[/<providerId>[/<paymentType>]]` and `edge://sell[...]` (plus `https://deep.edge.app` equivalents). Links can optionally **pin a provider and payment type** to the top of quote results for that visit only; pins live in navigation params, clear on tab blur, and degrade gracefully when nothing matches.
> 
> Also adds **affiliate provider priority** from existing `promoCards2` `pluginPromotions`, via a new `useRampPreferredProviders` hook that reads silent (non-display) cards. Deep-link pins outrank affiliate preferences.
> 
> Both inputs feed a shared `compareRampQuotes` comparator used by `useRampQuotes` and the option list. Rate claims stay honest: the create scene exchange rate and the "Best Rate" badge use `getBestRateRampQuote` so promotions change landing order, not reported market rates. Under a promotion the badge is deliberately absent rather than pointing at a competitor.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6db2b61d6088c9134ec983d5997ada7908c18640. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/actions/DeepLinkingActions.tsx
+++ b/src/actions/DeepLinkingActions.tsx
@@ -51,6 +51,8 @@ const CREATE_WALLET_ASSETS: Record<string, EdgeAsset> = {
  * The app has just received some of link,
  * so try to follow it if possible, or save it for later if not.
  */
+// TODO: Remove NavigationBase dependency. Requires inversion of navigation
+// control for v7 migration.
 export function launchDeepLink(
   navigation: NavigationBase,
   link: DeepLink
@@ -210,7 +212,11 @@ async function handleLink(
 
       if (checkAndShowLightBackupModal(account, navigation)) break
 
-      const address = await wallet.getReceiveAddress({ tokenId: null })
+      const [address] = await wallet.getAddresses({ tokenId: null })
+      if (address == null) {
+        showError(lstrings.alert_deep_link_no_wallet_for_uri)
+        break
+      }
       const response = await fetch(`${link.uri}${address.publicAddress}`)
       if (response.ok) {
         showToast(lstrings.azteco_success)
--- a/src/actions/DeepLinkingActions.tsx
+++ b/src/actions/DeepLinkingActions.tsx
@@ -490,6 +490,26 @@ async function handleLink(
       break
     }
 
+    case 'rampCreate': {
+      // Open the ramps buy/sell flow, optionally pinning a provider and payment
+      // type to the top of the quote results. The pin lives in the navigation
+      // params only: nothing is written to the account referral state, and a
+      // pin that matches no quote degrades to the normal ordering.
+      const { direction, providerId, paymentType } = link
+      if (direction === 'buy') {
+        navigation.navigate('buyTab', {
+          screen: 'pluginListBuy',
+          params: { providerId, paymentType }
+        })
+      } else {
+        navigation.navigate('sellTab', {
+          screen: 'pluginListSell',
+          params: { providerId, paymentType }
+        })
+      }
+      break
+    }
+
     case 'ramp': {
       const result = rampDeeplinkManager.handleDeeplink(link)
       if (!result.success) {
--- a/src/components/scenes/RampCreateScene.tsx
+++ b/src/components/scenes/RampCreateScene.tsx
@@ -1,6 +1,6 @@
 import { useFocusEffect } from '@react-navigation/native'
 import { useQuery } from '@tanstack/react-query'
-import { div, eq, gt, mul, round, toBns } from 'biggystring'
+import { div, gt, mul, round, toBns } from 'biggystring'
 import type {
   EdgeCurrencyWallet,
   EdgeDenomination,
@@ -24,6 +24,7 @@ import { getSpecialCurrencyInfo } from '../../constants/WalletAndCurrencyConstan
 import { useHandler } from '../../hooks/useHandler'
 import { useRampLastCryptoSelection } from '../../hooks/useRampLastCryptoSelection'
 import { useRampPlugins } from '../../hooks/useRampPlugins'
+import { useRampPreferredProviders } from '../../hooks/useRampPreferredProviders'
 import { useRampQuotes } from '../../hooks/useRampQuotes'
 import {
   type SupportedPluginResult,
@@ -31,6 +32,7 @@ import {
 } from '../../hooks/useSupportedPlugins'
 import { useWatch } from '../../hooks/useWatch'
 import { lstrings } from '../../locales/strings'
+import type { FiatPaymentType } from '../../plugins/gui/fiatPluginTypes'
 import type {
   RampPlugin,
   RampQouteAmount,
@@ -39,6 +41,11 @@ import type {
 } from '../../plugins/ramps/rampPluginTypes'
 import { getBestQuoteError } from '../../plugins/ramps/utils/getBestError'
 import { getRateFromRampQuoteResult } from '../../plugins/ramps/utils/getRateFromRampQuoteResult'
+import {
+  getBestRateRampQuote,
+  rampQuoteHasAmounts,
+  type RampQuotePriority
+} from '../../plugins/ramps/utils/rampQuotePriority'
 import { getDefaultFiat } from '../../selectors/SettingsSelectors'
 import { useDispatch, useSelector } from '../../types/reactRedux'
 import type {
@@ -81,6 +88,15 @@ import { RampRegionSelect } from './RampCreateScene/RampRegionSelect'
 export interface RampCreateParams {
   forcedWalletResult?: WalletListWalletResult
   regionCode?: string
+  /**
+   * Ramp provider to pin to the top of the quote results for this navigation
+   * only, from an `edge://buy/<providerId>` style deep link. Nothing is written
+   * to the account referral state, and a provider that returns no quotes falls
+   * back to the normal ordering.
+   */
+  providerId?: string
+  /** Payment type to pin to the top, with the same link-scoped semantics. */
+  paymentType?: FiatPaymentType
 }
 
 type Props = (
@@ -92,8 +108,12 @@ type Props = (
 
 export const RampCreateScene: React.FC<Props> = (props: Props) => {
   const { direction, navigation, route } = props
-  const { regionCode: initialRegionCode, forcedWalletResult } =
-    route?.params ?? {}
+  const {
+    regionCode: initialRegionCode,
+    forcedWalletResult,
+    providerId: pinnedProviderId,
+    paymentType: pinnedPaymentType
+  } = route?.params ?? {}
 
   const theme = useTheme()
   const styles = getStyles(theme)
@@ -369,6 +389,24 @@ export const RampCreateScene: React.FC<Props> = (props: Props) => {
     denomination
   ])
 
+  // Providers this account's affiliation prefers, from the info server
+  const preferredProviderIds = useRampPreferredProviders(direction)
+
+  // A provider pinned by the deep link outranks the affiliate preference
+  const quotePriority: RampQuotePriority = React.useMemo(
+    () => ({
+      preferPluginIds:
+        pinnedProviderId == null
+          ? preferredProviderIds
+          : [
+              pinnedProviderId,
+              ...preferredProviderIds.filter(id => id !== pinnedProviderId)
+            ],
+      preferPaymentType: pinnedPaymentType
+    }),
+    [pinnedProviderId, pinnedPaymentType, preferredProviderIds]
+  )
+
   // Fetch quotes using the custom hook
   const {
     quotes: allQuotes,
@@ -379,11 +417,17 @@ export const RampCreateScene: React.FC<Props> = (props: Props) => {
     rampQuoteRequest,
     plugins: Object.fromEntries(
       supportedPlugins.map(result => [result.plugin.pluginId, result.plugin])
-    )
+    ),
+    priority: quotePriority
   })
 
-  // Get the best quote using .find because we want to preserve undefined in its type
-  const bestQuote = allQuotes.find((_, index) => index === 0)
+  // The scene's exchange rate and light-account purchase limit are claims about
+  // the rate, so they read the best-rate quote rather than the first
+  // prioritized one, which is whatever the link or the affiliate config pinned.
+  const bestQuote = React.useMemo(
+    () => getBestRateRampQuote(allQuotes, direction),
+    [allQuotes, direction]
+  )
 
   // For Max flow, select the quote with the largest supported amount
   const maxQuoteForMaxFlow = React.useMemo(() => {
@@ -553,6 +597,36 @@ export const RampCreateScene: React.FC<Props> = (props: Props) => {
     dispatch(logEvent(direction === 'buy' ? 'Buy_Quote' : 'Sell_Quote'))
   })
 
+  // Drop the deep link pin when the user leaves the buy/sell tab. React
+  // Navigation keeps route params on the tab's route for the whole app
+  // session, so leaving them in place would pin every later visit to the tab,
+  // not just the flow the link opened. The listener is on the TAB, so stepping
+  // forward to the option list and back keeps the pin: only leaving the tab
+  // ends it. Subscribing on `navigation` alone (never on the params) also
+  // keeps a warm deep link that arrives while the tab is focused from being
+  // cleared by a re-subscription.
+  // The pins are read through a ref so the guard below can see them without
+  // entering the dep array, which is what the `navigation`-only subscription
+  // above depends on.
+  const pinsRef = React.useRef({ pinnedProviderId, pinnedPaymentType })
+  pinsRef.current = { pinnedProviderId, pinnedPaymentType }
+
+  React.useEffect(() => {
+    const tabNavigation = navigation.getParent()
+    if (tabNavigation == null) return
+    return tabNavigation.addListener('blur', () => {
+      const { pinnedProviderId, pinnedPaymentType } = pinsRef.current
+      // Nothing to drop: tab switching is the app's most-travelled path, and a
+      // user who never tapped a deep link would otherwise pay a params update
+      // plus a re-render every time they leave the tab.
+      if (pinnedProviderId == null && pinnedPaymentType == null) return
+      navigation.setParams({
+        providerId: undefined,
+        paymentType: undefined
+      })
+    })
+  }, [navigation])
+
   //
   // Handlers
   //
@@ -693,7 +767,9 @@ export const RampCreateScene: React.FC<Props> = (props: Props) => {
     )
 
     navigation.navigate('rampSelectOption', {
-      rampQuoteRequest
+      rampQuoteRequest,
+      providerId: pinnedProviderId,
+      paymentType: pinnedPaymentType
     })
   })
 
@@ -809,12 +885,16 @@ export const RampCreateScene: React.FC<Props> = (props: Props) => {
       !isPickerOpen
     ) {
       navigation.navigate('rampSelectOption', {
-        rampQuoteRequest
+        rampQuoteRequest,
+        providerId: pinnedProviderId,
+        paymentType: pinnedPaymentType
       })
       // Reset transient state to avoid leaving the scene in a max "mode"
       setPendingMaxNav(false)
     }
   }, [
+    pinnedProviderId,
+    pinnedPaymentType,
     pendingMaxNav,
     isPickerOpen,
     maxQuoteForMaxFlow,
@@ -1252,9 +1332,6 @@ async function getMaxSpendExchangeAmount(
   )(maxSpendNativeAmount)
   return maxSpendExchangeAmount
 }
-
-const rampQuoteHasAmounts = (quote: RampQuote): boolean =>
-  !eq(quote.fiatAmount, '0') || !eq(quote.cryptoAmount, '0')
 
 /**
  * Generates the unavailable warning body, optionally appending a suggestion
--- a/src/components/scenes/RampCreateScene.tsx
+++ b/src/components/scenes/RampCreateScene.tsx
@@ -1100,12 +1100,12 @@ export const RampCreateScene: React.FC<Props> = (props: Props) => {
 }
 
 // Export separate components for buy and sell routes
-export const RampCreateBuyScene = (
-  props: BuySellTabSceneProps<'pluginListBuy'>
-): React.ReactElement => <RampCreateScene {...props} direction="buy" />
-export const RampCreateSellScene = (
-  props: BuySellTabSceneProps<'pluginListSell'>
-): React.ReactElement => <RampCreateScene {...props} direction="sell" />
+export const RampCreateBuyScene: React.FC<
+  BuySellTabSceneProps<'pluginListBuy'>
+> = props => <RampCreateScene {...props} direction="buy" />
+export const RampCreateSellScene: React.FC<
+  BuySellTabSceneProps<'pluginListSell'>
+> = props => <RampCreateScene {...props} direction="sell" />
 
 const getStyles = cacheStyles((theme: ReturnType<typeof useTheme>) => ({
   flagIconLarge: {
--- a/src/components/scenes/RampSelectOptionScene.tsx
+++ b/src/components/scenes/RampSelectOptionScene.tsx
@@ -1,17 +1,19 @@
-import { eq } from 'biggystring'
 import * as React from 'react'
 import { ActivityIndicator, Image, View } from 'react-native'
 import { sprintf } from 'sprintf-js'
 
 import paymentTypeLogoApplePay from '../../assets/images/paymentTypes/paymentTypeLogoApplePay.png'
 import { COUNTRY_CODES } from '../../constants/CountryConstants'
+import { ENV } from '../../env'
 import { formatFiatString } from '../../hooks/useFiatText'
 import { useHandler } from '../../hooks/useHandler'
 import { useRampPlugins } from '../../hooks/useRampPlugins'
+import { useRampPreferredProviders } from '../../hooks/useRampPreferredProviders'
 import { useRampQuotes } from '../../hooks/useRampQuotes'
 import { useSupportedPlugins } from '../../hooks/useSupportedPlugins'
 import { formatNumber } from '../../locales/intl'
 import { lstrings } from '../../locales/strings'
+import type { FiatPaymentType } from '../../plugins/gui/fiatPluginTypes'
 import { FiatProviderError } from '../../plugins/gui/fiatProviderTypes'
 import type {
   RampPlugin,
@@ -19,6 +21,13 @@ import type {
   RampQuoteRequest,
   SettlementRange
 } from '../../plugins/ramps/rampPluginTypes'
+import {
+  compareRampQuotes,
+  getBestRateRampQuote,
+  getUnmatchedRampQuotePriority,
+  rampQuoteHasAmounts,
+  type RampQuotePriority
+} from '../../plugins/ramps/utils/rampQuotePriority'
 import { useSelector } from '../../types/reactRedux'
 import type { BuySellTabSceneProps } from '../../types/routerTypes'
 import { getPaymentTypeIcon } from '../../util/paymentTypeIcons'
@@ -38,13 +47,24 @@ import { EdgeText } from '../themed/EdgeText'
 
 export interface RampSelectOptionParams {
   rampQuoteRequest: RampQuoteRequest
+  /**
+   * Ramp provider to pin to the top of the results for this navigation only,
+   * from an `edge://buy/<providerId>` style deep link.
+   */
+  providerId?: string
+  /** Payment type to pin to the top, with the same link-scoped semantics. */
+  paymentType?: FiatPaymentType
 }
 
 interface Props extends BuySellTabSceneProps<'rampSelectOption'> {}
 
 export const RampSelectOptionScene: React.FC<Props> = (props: Props) => {
-  const { route } = props
-  const { rampQuoteRequest } = route.params
+  const { navigation, route } = props
+  const {
+    rampQuoteRequest,
+    providerId: pinnedProviderId,
+    paymentType: pinnedPaymentType
+  } = route.params
   const { direction } = rampQuoteRequest
 
   const theme = useTheme()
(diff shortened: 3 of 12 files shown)


j0ntz:
## 📸 Test evidence

**buy deeplink cold start**

buy deeplink cold start

**venmo pinned first**

venmo pinned first

**best rate badge on true best**

best rate badge on true best

**cashapp graceful fallback**

cashapp graceful fallback

**affiliate priority paybis first**

affiliate priority paybis first

**unaffiliated ordering unchanged**

unaffiliated ordering unchanged

j0ntz:
## 📸 Test evidence (after fix)

**create scene shows true best rate**

create scene shows true best rate

**venmo still pinned after fix**

venmo still pinned after fix

src/components/scenes/RampCreateScene.tsx
```diff
@@ -379,7 +416,8 @@ export const RampCreateScene: React.FC<Props> = (props: Props) => {
     rampQuoteRequest,
     plugins: Object.fromEntries(
       supportedPlugins.map(result => [result.plugin.pluginId, result.plugin])
-    )
+    ),
+    priority: quotePriority
```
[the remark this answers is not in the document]
j0ntz: Fixed. `RampCreateScene` no longer reads `allQuotes[0]`: the exchange rate, the conversion math behind it, and the light-account purchase limit now go through the shared `getBestRateRampQuote(allQuotes, direction)` helper, which sorts a copy with the rate-only comparator. The select scene's "Best Rate" badge uses the same helper. Verified on the iOS sim with `edge://buy/moonpay/venmo`: the create scene shows 1 ETH = 1,985.60 USD (the Paybis best rate) while the option list still opens on the pinned MoonPay Venmo card, screenshots in the "Test evidence (after fix)" comment. Unit coverage added in `rampQuotePriority.test.ts`.

src/components/scenes/RampSelectOptionScene.tsx
```diff
@@ -97,9 +132,25 @@ export const RampSelectOptionScene: React.FC<Props> = (props: Props) => {
...
+  // simply fall back to the unpinned ordering:
+  React.useEffect(() => {
+    if (isLoadingQuotes || allQuotes.length === 0) return
+    const unmatched = getUnmatchedRampQuotePriority(allQuotes, quotePriority)
+    if (unmatched.length > 0) {
+      console.log(
+        `RampSelectOptionScene: no quotes matched ${unmatched.join(
+          ', '
+        )}; showing unpinned results`
+      )
```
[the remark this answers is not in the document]
j0ntz: Fixed. The unmatched-pin log is now gated behind `ENV.DEBUG_VERBOSE_LOGGING`, matching the convention in `infiniteApi.ts` and `translateError.ts`.

j0ntz:
## 📸 Test evidence (pin no longer outlives the visit)

**pin cleared on later visit**

pin cleared on later visit

src/actions/DeepLinkingActions.tsx
```diff
@@ -484,6 +486,26 @@ async function handleLink(
...
+          params: { providerId, paymentType }
+        })
+      } else {
+        navigation.navigate('sellTab', {
+          screen: 'pluginListSell',
+          params: { providerId, paymentType }
+        })
+      }
+      break
+    }
```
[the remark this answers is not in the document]
j0ntz: Fixed. `RampCreateScene` now clears `providerId` / `paymentType` from its route params when the scene loses focus, which happens after they have already been forwarded into the `rampSelectOption` params, so the flow the link opened keeps its pin and nothing later does. Verified on the iOS sim: `edge://buy/moonpay/venmo` lands on the pinned MoonPay Venmo option list; popping back to the Buy tab and requesting quotes again gives the plain best-rate ordering (Paybis ACH first, Best Rate badge), screenshot in the "Test evidence (pin no longer outlives the visit)" comment.

j0ntz:
## 📸 Test evidence (pin lifetime)

**pin kept on in flow back**

pin kept on in flow back

**pin cleared after leaving tab**

pin cleared after leaving tab

src/components/scenes/RampCreateScene.tsx
```diff
@@ -553,6 +597,23 @@ export const RampCreateScene: React.FC<Props> = (props: Props) => {
...
+    React.useCallback(() => {
+      return () => {
+        if (pinnedProviderId == null && pinnedPaymentType == null) return
+        navigation.setParams({
+          providerId: undefined,
+          paymentType: undefined
+        })
+      }
+    }, [navigation, pinnedPaymentType, pinnedProviderId])
+  )
```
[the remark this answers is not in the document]
j0ntz: Both cases fixed by moving the clear off the screen focus effect and onto the TAB. The create scene now subscribes once to the buy/sell tab’s `blur` event (`navigation.getParent()`), with `navigation` as the only dependency:

- Warm deep link on an already-focused tab: the subscription never re-runs on a param change, so nothing tears down and clears the params the link just set.
- Back from `rampSelectOption`: the tab never blurred, so the pin is still there.
- Leaving the tab: the params are cleared, so a later visit to Buy is unpinned.

Verified all three on the iOS sim (screenshots in the "Test evidence (pin lifetime)" comment): `edge://buy/moonpay/venmo` cold start pins MoonPay Venmo; swiping back to the create scene and requesting quotes again keeps it pinned; switching to Assets and back drops it to the plain best-rate ordering (Paybis ACH first).

src/actions/DeepLinkingActions.tsx
```diff
@@ -210,7 +212,7 @@ async function handleLink(
 
       if (checkAndShowLightBackupModal(account, navigation)) break
 
-      const address = await wallet.getReceiveAddress({ tokenId: null })
+      const [address] = await wallet.getAddresses({ tokenId: null })
```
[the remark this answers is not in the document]
j0ntz: Fixed. The Azteco path now guards the destructured address and shows `alert_deep_link_no_wallet_for_uri` instead of throwing when `getAddresses` returns an empty array. (That call site changed in this branch only because `getReceiveAddress` is deprecated; the deprecation fix is unrelated to the ramps work but the guard belongs with it.)

src/hooks/useRampPreferredProviders.ts
```diff
@@ -0,0 +1,64 @@
...
+        if (pluginPromotion.pluginType !== direction) continue
+        for (const providerId of pluginPromotion.preferProviders ?? []) {
+          if (!preferredProviders.includes(providerId)) {
+            preferredProviders.push(providerId)
+          }
+        }
+      }
+    }
+    return preferredProviders
+  }, [accountReferral, countryCode, direction])
```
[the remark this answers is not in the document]
j0ntz: Fixed. `infoServerData.rollup?.promoCards2` is now read during render and included in the memo key, so a scene that mounts before the first info-server fetch lands picks the preferences up on the next render, and a 5-minute refresh that replaces the rollup is picked up too.

src/components/scenes/RampCreateScene.tsx
```diff
@@ -553,6 +597,25 @@ export const RampCreateScene: React.FC<Props> = (props: Props) => {
...
+  React.useEffect(() => {
+    const tabNavigation = navigation.getParent()
+    if (tabNavigation == null) return
+    return tabNavigation.addListener('blur', () => {
+      navigation.setParams({
+        providerId: undefined,
+        paymentType: undefined
+      })
+    })
+  }, [navigation])
```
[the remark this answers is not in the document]
j0ntz: Fixed. `RampSelectOptionScene` now runs the same tab-blur listener on its own route, so the option list clears its copy of the pin when the user leaves the Buy/Sell tab.

Verified on the iOS sim with a before/after pair. Pre-fix: open `edge://buy/moonpay/venmo`, Next to the option list, tap Assets, tap Buy, and MoonPay Venmo is still first at 0.236638 ETH while Apple Pay and Credit quote 0.250262. Post-fix: the same sequence comes back to ACH Bank Transfer (Paybis) first wearing the Best Rate badge at 0.2517196 ETH, with no MoonPay pin. Section 5.3 of the design doc records the second listener.

j0ntz:
## 📸🪓 Test evidence (after fix): option-list pin now clears on tab leave

> 🪓 **Hack-forced evidence:** the fix commit was stashed to capture the pre-fix frame; restored before committing. Temporary uncommitted edit, reverted before commit; the marked frames prove the rendering, not the trigger.

**pin active after fix**

pin active after fix

**pin cleared after tab leave**

pin cleared after tab leave

**🪓 HACK-FORCED: pre fix pin survives tab leave**

🪓 HACK-FORCED: pre fix pin survives tab leave

**tdd mermaid renders on github**

tdd mermaid renders on github

j0ntz:
## 📸 Test evidence: followup test-gap closure (promo priority, expiry, af attribution, warm start, sell)

**control no promo order**

control no promo order

**promo code activated**

promo code activated

**paybis first via activepromotions**

paybis first via activepromotions

**future enddate pin applies**

future enddate pin applies

**expired card reverts order**

expired card reverts order

**expired card prunes activepromotions**

expired card prunes activepromotions

**af attribution moonpay first**

af attribution moonpay first

**visible promo card cta**

visible promo card cta

**cta warm pin venmo first**

cta warm pin venmo first

**warm openurl delivery**

warm openurl delivery

**no af no referral write**

no af no referral write

**sell moonpay pinned**

sell moonpay pinned

j0ntz:
Non-visual evidence from the same run, to go with the screenshots above.

**Section 6 JSON blocks against the real cleaners.** Both blocks were extracted from the committed doc and run through `edge-info-server`'s `asInfoCard` and `asPromoCards2` (`asHealingArray(asInfoCard)`) in node, against `edge-info-server@3.12.0` from the gui's own `node_modules` (byte-identical `asInfoCard` to the repo checkout):

```
PASS  asInfoCard  6.1 visible card with a deep link CTA
      promoId="moonpay" pluginPromotions=[{"pluginType":"buy","preferProviders":["moonpay"]}]
      round-trip stable: yes
PASS  asInfoCard  6.2 silent priority card
      promoId="moonpay" pluginPromotions=[{"pluginType":"buy",...},{"pluginType":"sell",...}]
      localeMessages keys=[] ctaButton=undefined
      round-trip stable: yes
PASS  asPromoCards2 (asHealingArray): 2 in -> 2 out (0 silently dropped required)
PASS  negative control: 3 in (1 missing "background") -> 2 out (drop expected)
```

The negative control matters because the healing array drops a malformed card silently; without it a green result proves nothing.

**AC2, now runtime rather than code reading.** A temporary uncommitted marker inside `saveAccountReferral` wrote one line per call into the app container, and `CreationReason.json` was hashed before and after:

| Launch | saveAccountReferral calls | installerId / activePromotions | CreationReason.json |
|---|---|---|---|
| baseline, no deep link | 1 | `undefined` / `[]` | unchanged |
| `edge://buy/moonpay/venmo`, driven to the option list | 1 | `undefined` / `[]` | unchanged (same sha1) |
| `https://deep.edge.app/buy/moonpay/cashapp?af=moonpay` | 1 | `undefined` / `["moonpay"]` | unchanged |

So the pin path adds no referral write at all, while the `af` form does exactly one, and it lands in `activePromotions` rather than `installerId` because the referral server 404s the code. Both hacks were reverted; `git status` is clean of them.

**One caveat worth recording, not a defect in this PR.** The "Best Rate" badge is attached to `getBestRateRampQuote(allQuotes)`, while each card renders its payment-type group's first quote. When the pinned provider is not the best-rate provider, the best-rate quote stops leading any group and the badge renders nowhere rather than moving. Visible in the Paybis-pinned screenshot. The design doc's testing section now says so.

Testing was done on a throwaway account rather than `edge-funds` for everything that writes account-synced state (`activePromotions`, attribution), so no shared roster account was dirtied. The sell drive used `edge-funds` because it needs a balance.

src/components/scenes/RampSelectOptionScene.tsx
```diff
@@ -126,28 +200,16 @@ export const RampSelectOptionScene: React.FC<Props> = (props: Props) => {
       grouped.set(paymentType, [...existing, quote])
     })
 
-    // Sort quotes within each payment type group
+    // Sort quotes within each payment type group. The group order itself comes
+    // from the insertion order above, which follows the already-prioritized
+    // `allQuotes`, so both levels honor the same preferences.
+    const compare = compareRampQuotes(direction, quotePriority)
```
peachbits, line 228: Sorting **within** each group by the priority means `providerQuotes[0]` — the quote the card actually displays — is now the preferred provider's quote, not the group's best-rate quote. But `isBestOption` (:400-406) still compares the *displayed* quote against `bestQuoteOverall`, so the "Best Rate" badge can vanish from the whole list.

Concrete: Paybis credit @ 100, MoonPay credit @ 110, MoonPay venmo @ 120, `moonpay` pinned or affiliate-preferred.
- `bestQuoteOverall` = `paybis:credit` ✅
- credit card displays `moonpay:credit` → no badge
- venmo card displays `moonpay:venmo` → no badge
- `paybis:credit` is never displayed, so **no card is badged at all**

Before this PR the groups sorted by rate only, so the global best quote was always the displayed quote of its group and the badge always rendered. The earlier bugbot fix corrected the badge's *claim* but moved the badged quote out of view.

Two ways out: badge the group whose **best** quote is `bestQuoteOverall` (rather than its displayed quote), or keep the within-group sort rate-only and let the priority drive group order + which provider the card opens on.

src/plugins/ramps/utils/rampQuotePriority.ts
```diff
@@ -0,0 +1,120 @@
...
+    const [paymentRankA, pluginRankA] = getPriorityRank(a, priority)
+    const [paymentRankB, pluginRankB] = getPriorityRank(b, priority)
+    if (paymentRankA !== paymentRankB) return paymentRankA - paymentRankB
+    if (pluginRankA !== pluginRankB) return pluginRankA - pluginRankB
+
+    const hasAmountA = rampQuoteHasAmounts(a)
+    const hasAmountB = rampQuoteHasAmounts(b)
+    if (hasAmountA && !hasAmountB) return -1
+    if (!hasAmountA && hasAmountB) return 1
+    if (!hasAmountA && !hasAmountB) return 0
```
peachbits, line 70: The priority ranks short-circuit **above** the has-amounts tier, so a preferred provider's amount-less quote outranks every priced quote.

That is reachable in production: `createExternalRampPlugin` always emits `cryptoAmount: '0'` / `fiatAmount: '0'` with `specialQuoteRateMessage` ([createExternalRampPlugin.ts#L150-L159](https://github.com/EdgeApp/edge-react-gui/blob/90ec73b139596d6484dc963241e207d609d9ee51/src/plugins/ramps/utils/createExternalRampPlugin.ts#L150-L159)). So if an affiliate's `preferProviders` (or an `edge://buy/<providerId>` link) names an external provider, its placeholder quote becomes `providerQuotes[0]` and the card renders "Tap to view quote amount and rate" instead of a real total — displacing a priced quote in the same group and floating that group to the top of the list.

Moving the `hasAmountA` / `hasAmountB` comparison above the two rank checks fixes it without weakening the pin: the preferred provider still wins, just within the tier of quotes that actually have amounts.

src/components/scenes/RampSelectOptionScene.tsx
```diff
@@ -97,9 +150,25 @@ export const RampSelectOptionScene: React.FC<Props> = (props: Props) => {
...
+    priority: quotePriority
   })
 
+  // A pin that matched nothing is expected (a provider with no quotes for this
+  // request, or a payment type the info server currently disables). The results
+  // simply fall back to the unpinned ordering:
+  React.useEffect(() => {
+    if (isLoadingQuotes || allQuotes.length === 0) return
+    const unmatched = getUnmatchedRampQuotePriority(allQuotes, quotePriority)
+    if (unmatched.length > 0 && ENV.DEBUG_VERBOSE_LOGGING) {
```
peachbits: `getUnmatchedRampQuotePriority` runs on every quote refresh (every 30s) in production before the `ENV.DEBUG_VERBOSE_LOGGING` check ever happens. Worth gating first:

```ts
if (!ENV.DEBUG_VERBOSE_LOGGING) return
if (isLoadingQuotes || allQuotes.length === 0) return
const unmatched = getUnmatchedRampQuotePriority(allQuotes, quotePriority)
if (unmatched.length > 0) {
  console.log(...)
}
```

peachbits (commented):
Two notes on the ordering precedence in `compareRampQuotes` (the group-sort comment and the amounts-tier comment below are the same root cause: the priority ranks are compared before everything else), plus one nit on the debug-log effect.

src/components/scenes/RampCreateScene.tsx
peachbits, line 628: Nit / optional. The listener has no idea whether a pin exists, so it fires on every tab blur for the whole app session — from this scene and from `RampSelectOptionScene` (:117-126). A user who never taps a deep link still dispatches a `SET_PARAMS` setting `undefined` → `undefined` every time they swipe off the Buy tab, again off the Sell tab, and again from the option list when it's mounted. Each one is a navigation state update plus a re-render of those scenes.

The inline guard (`if (pinnedProviderId == null && pinnedPaymentType == null) return`) isn't available here, since it drags the params back into the dep array — which is exactly what caused the earlier "pin cleared by focus cleanup" finding. A ref keeps the `[navigation]`-only subscription intact:

```ts
const pinsRef = React.useRef({ pinnedProviderId, pinnedPaymentType })
pinsRef.current = { pinnedProviderId, pinnedPaymentType }

React.useEffect(() => {
  const tabNavigation = navigation.getParent()
  if (tabNavigation == null) return
  return tabNavigation.addListener('blur', () => {
    const { pinnedProviderId, pinnedPaymentType } = pinsRef.current
    if (pinnedProviderId == null && pinnedPaymentType == null) return
    navigation.setParams({ providerId: undefined, paymentType: undefined })
  })
}, [navigation])
```

The dispatch is a no-op and the extra render isn't perceptible — the only reason it's worth mentioning is that it's permanent overhead on the app's most-travelled path (tab switching), paid by every user to support a feature that fires in a small fraction of sessions. Reasonable to skip if that reads as noise.

src/components/scenes/RampSelectOptionScene.tsx
j0ntz: Took the first option: \`isBestOption\` now tests the card's whole provider group (\`providerQuotes.some(...)\`) instead of its displayed quote, so the badge lands on the payment-type card that HOLDS \`bestQuoteOverall\` rather than disappearing when a pin moves that quote out of index 0 in every group at once.

The second option would have made the pin stop changing which provider each card shows, which is the feature's headline behavior, so I kept the within-group priority sort. The tradeoff is now stated explicitly: the badge means "the best available rate is on this payment method", and the figure on the card face stays the pinned provider's own. Section 5.6 of the design doc is rewritten around that rule, including why the previous displayed-quote scope drops the badge from the entire list.

Verified on the sim with a before/after pair. Live quotes could not produce the state naturally (the best quote sits in the single-provider ACH group, where a pin cannot displace it), so I temporarily filtered ACH out of the scene to put the best quote inside the multi-provider Apple Pay group, then drove \`edge://buy/paybis\`: pre-fix rule renders no badge anywhere, post-fix rule badges Apple Pay while that card still shows Paybis. Both frames are on the PR, labelled as hack-forced. The hack is reverted.

src/plugins/ramps/utils/rampQuotePriority.ts
j0ntz: Fixed as suggested: the \`hasAmountA\` / \`hasAmountB\` comparison now runs above both rank checks in \`compareRampQuotes\`, so the preference wins within the tier of quotes that have amounts. Two regression tests cover the reachable cases: a preferred provider whose quote has no amounts, and a preferred payment type whose only quote has no amounts, both now sort below the priced quotes.

The reasoning is recorded as decision 10.7 in the design doc, with the \`createExternalRampPlugin\` placeholder as the concrete trigger, and section 5.4 now states the tier order.

src/components/scenes/RampSelectOptionScene.tsx
j0ntz: Gated first: the effect now returns on \`!ENV.DEBUG_VERBOSE_LOGGING\` before \`getUnmatchedRampQuotePriority\` runs, so production does no scan on the 30s refresh.

src/components/scenes/RampCreateScene.tsx
j0ntz: Took the ref, in both scenes: \`RampCreateScene\` and \`RampSelectOptionScene\` each carry their own copy of the listener, so both now read the pins through \`pinsRef\` and return early when neither is set. The \`[navigation]\`-only subscription is unchanged.

Not treating it as noise. Section 5.3 of the design doc now records why the guard has to go through a ref rather than an inline null check, since the inline form is what would drag the params back into the dep array and reintroduce the earlier "pin cleared by focus cleanup" bug.

Re-verified on the sim after the change: \`edge://buy/paybis\` pins Paybis across every payment-type group, and leaving the buy tab and returning still drops the pin, with the list back to Moonpay on Apple Pay and Credit. Both frames are on the PR.

j0ntz:
## 📸🪓 Test evidence (review fixes)

> 🪓 **Hack-forced evidence:** temporarily filtered ACH quotes out of RampSelectOptionScene so the globally-best quote landed inside a multi-provider group (live quotes put it in the single-provider ACH group, where a pin cannot displace it); frame 41 additionally restores the pre-fix isBestOption rule. Temporary uncommitted edit, reverted before commit; the marked frames prove the rendering, not the trigger.

**unpinned control badge on ach**

unpinned control badge on ach

**🪓 HACK-FORCED: before fix no badge anywhere**

🪓 HACK-FORCED: before fix no badge anywhere

**🪓 HACK-FORCED: after fix badge on group holding best**

🪓 HACK-FORCED: after fix badge on group holding best

**paybis pinned badge on ach**

paybis pinned badge on ach

**pin cleared after tab leave**

pin cleared after tab leave

src/components/scenes/RampSelectOptionScene.tsx
```diff
@@ -335,16 +408,22 @@ const QuoteResult: React.FC<{
...
-    providerQuote.pluginId === bestQuoteOverall.pluginId &&
-    providerQuote.paymentType === bestQuoteOverall.paymentType &&
-    providerQuote.fiatAmount === bestQuoteOverall.fiatAmount
+    rampQuoteHasAmounts(bestQuoteOverall) &&
+    providerQuotes.some(
+      quote =>
+        quote.pluginId === bestQuoteOverall.pluginId &&
+        quote.paymentType === bestQuoteOverall.paymentType &&
+        quote.fiatAmount === bestQuoteOverall.fiatAmount
+    )
```
peachbits: The group-scoped `.some()` is the right fix, but dropping `hasSelectedAmounts &&` from the condition looks like it came along for the ride — the fix doesn't need it.

After the has-amounts tier moved above the rank checks in `compareRampQuotes`, the group holding `bestQuoteOverall` always opens on a priced quote, so keeping `hasSelectedAmounts` would never suppress the badge in the default state. It only takes effect once the user picks a provider by hand from the `CardListModal`: choosing an amount-less provider (LibertyX / BitsOfGold) in the badged group leaves a **Best Rate** badge sitting over a card face that reads "Tap to view quote amount and rate".

Putting it back costs nothing and removes that state:

```ts
const isBestOption =
  hasSelectedAmounts &&
  bestQuoteOverall != null &&
  rampQuoteHasAmounts(bestQuoteOverall) &&
  providerQuotes.some(...)
```

src/plugins/ramps/utils/rampQuotePriority.ts
```diff
@@ -0,0 +1,129 @@
...
+export const compareRampQuotes =
+  (
+    direction: 'buy' | 'sell',
+    priority: RampQuotePriority = NO_PRIORITY
+  ): ((a: RampQuote, b: RampQuote) => number) =>
+  (a, b) => {
+    const hasAmountA = rampQuoteHasAmounts(a)
+    const hasAmountB = rampQuoteHasAmounts(b)
+    if (hasAmountA && !hasAmountB) return -1
+    if (!hasAmountA && hasAmountB) return 1
```
peachbits: Question on the scope of this tier, and it's a cost my original comment didn't price in — so treat this as a design question rather than a defect.

The has-amounts tier now sits above the **deep-link pin**, not just the affiliate preference. `libertyx` and `bitsofgold` are the only two `createExternalRampPlugin` providers and they *only* ever emit amount-less placeholders, so `edge://buy/libertyx` now sinks LibertyX to the bottom of its group and the link produces no visible pin at all. `getUnmatchedRampQuotePriority` won't explain it either, since the provider did return quotes — they just have no amounts.

The distinction I was actually reaching for: a deep-link pin is explicit user intent ("show me this provider"), so arguably it belongs *above* the has-amounts tier; an affiliate `preferProviders` is a monetization signal that shouldn't degrade what the card displays, so it belongs *below*. That would mean splitting `RampQuotePriority.preferPluginIds` into a hard pin and a soft preference.

Worth it only if partner links to those two providers are a real use case — if they aren't, the current single tier is the simpler thing and decision 10.7 already documents it. Your call; I don't think it blocks.

peachbits (commented):
Both fixes look right — I checked that the `.some()` scope badges exactly one card (each group is keyed by `quote.paymentType` and the predicate requires `paymentType` equality, so only the best quote's own group can match), and the reordered comparator is still a valid total order. Two follow-ups below: one small thing the badge fix picked up, and a question about the cost of the tier reorder.

j0ntz:
## 📸 Test evidence (product decision round: display-honest badge, promotion outranks has-amounts)

**libertyx pinned placeholder top**

libertyx pinned placeholder top

**libertyx tap opens external flow**

libertyx tap opens external flow

**BTC list, badge on the displayed best quote (ACH · Paybis). ACH is single-provider for BTC, so a provider pin cannot displace this badge; the displacement case is frame "before fix no badge anywhere" in the previous evidence comment, which ran this exact display-honest rule**

BTC list, badge on displayed best quote

src/components/scenes/RampSelectOptionScene.tsx
j0ntz: Superseded by a product decision that resolves this thread and the group-scope question together: the badge reverts to display honesty, i.e. exactly the restored \`hasSelectedAmounts && providerQuote === bestQuoteOverall\` form you quoted. The group-scoped \`.some()\` is gone, so the hand-picked-placeholder state this thread guards against no longer exists.

Rationale (design doc decision 10.8): the pins and affiliate preferences exist to give a deep link full UX control and to intentionally promote a provider. A Best Rate starburst pointing at a competitor one picker-tap from the promoted card works against that, so under a promotion the badge is deliberately absent rather than relocated. The unpinned majority case is unchanged. The review-round frame "before fix no badge anywhere" in the earlier evidence comment shows this exact rule under a displacing pin; it is now the intended rendering.

src/plugins/ramps/utils/rampQuotePriority.ts
j0ntz: Good catch that the tier silently killed \`edge://buy/libertyx\`, and product answered the design question the other way: partner links to the external providers ARE a real use case, and the intent extends to affiliate configs too. Both inputs are intentional promotions, so the priority ranks move back above the has-amounts tier for both. The hard-pin/soft-preference split was considered and declined as complexity the intent does not need (decision 10.7 records this round trip, including your scenario).

The accepted cost is the one your original comment priced: a preferred external provider surfaces its placeholder card on top. That face is the honest surface for a provider whose number only exists on its own site. Three regression tests lock the contract: preferred provider and preferred payment type float their placeholders up; an unpreferred placeholder still sinks.

Verified on the sim: \`edge://buy/libertyx\` with a BTC wallet now lands with "In-person Cash and Debit Card / Powered By LibertyX / Select to view quote / Settlement: Instant" pinned at the top of the option list, and tapping it opens the LibertyX external flow. Frames are on the PR. One env note: the debug build needed \`libertyx\` added to \`RAMP_PLUGIN_INITS\` in env.json (no entry means the plugin never loads and the pin falls back gracefully, which is what an unconfigured production build would do too).

src/components/scenes/RampSelectOptionScene.tsx
j0ntz: Follow-up: the earlier reply ("took the first option") no longer reflects the code. Product reverted the group-scoped badge to display honesty (decision 10.8): the badge renders only beside the number it describes, and its absence under a promotion is intended, since a marker pointing at a competing provider works against the promotion the pin or affiliate preference exists to deliver. The current \`isBestOption\` is byte-wise the pre-review form.

src/plugins/ramps/utils/rampQuotePriority.ts
j0ntz: Follow-up: the earlier reply ("fixed as suggested") no longer reflects the code. Product chose the promotion side for both pins and affiliate preferences, so the ranks are back above the has-amounts tier (decision 10.7): a preferred external provider intentionally surfaces its placeholder card on top, and \`edge://buy/libertyx\` visibly lands. Your original concern survives as three regression tests plus the unpreferred-placeholder-still-sinks case, and the follow-on thread on line 64 has the full decision.

peachbits (approved):

PR #6154, opened 2026-08-14
Wire attestation tokens into edge-core-js context

paullinator:
## Summary
- **WIP / do not merge yet.** This PR depends on edge-core-js [#736](https://github.com/EdgeApp/edge-core-js/pull/736) (`setAttestationToken` / `x-attestation-token`). Merge that first, publish a new `edge-core-js` version, then bump `package.json` here before merging.
- Push info-server attestation tokens into core via `context.setAttestationToken` when the Edge context opens.
- Allow `LOGIN_SERVER` / `INFO_SERVER` env overrides for local E2E stacks.
- Also includes a Jest `NODE_ENV=test` fix so Socket-wrapped `npm test` does not break GestureDetector tests.

## Dependencies
- **Blocked on:** https://github.com/EdgeApp/edge-core-js/pull/736
- Follow-up before merge: bump `edge-core-js` in `package.json` / lockfile to the published release that includes #736.

## Test plan
- [ ] Wait for [edge-core-js #736](https://github.com/EdgeApp/edge-core-js/pull/736) to land and bump the dependency
- [ ] Confirm attestation token is pushed on context open and cleared on logout/stale
- [ ] Login-server requests carry `x-attestation-token` when a token is available
- [ ] Jest suite passes under Socket-wrapped `npm test`

---

> [!NOTE]
> **Medium Risk**
> Touches login-server attestation headers and depends on unreleased edge-core-js; wiring includes unsubscribe on context close, but incorrect token timing could affect CAPTCHA gating on login.
> 
> **Overview**
> **WIP — blocked on edge-core-js `setAttestationToken` (#736) before merge.**
> 
> Adds **`onAttestationToken`** so attestation JWT updates propagate to subscribers. Token cache updates now notify listeners with the current *servable* token (or `undefined` when rejected, expired, or unservable during backoff/hang). **`EdgeCoreManager`** subscribes when the Edge context opens and calls **`context.setAttestationToken`**, unsubscribing and stopping pushes when the context closes.
> 
> **Env overrides:** optional **`LOGIN_SERVER`** and **`INFO_SERVER`** arrays in env config override Maestro test servers or defaults for local E2E stacks.
> 
> **Jest:** the test script sets **`NODE_ENV=test`** so Socket-wrapped `npm test` does not break react-native-gesture-handler in UI tests.
> 
> Unit tests cover subscribe replay, success/rejection notifications, and unsubscribe behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6266f776d9561a7cc59e49cbefc9ee9da62d83c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/components/services/EdgeCoreManager.tsx
+++ b/src/components/services/EdgeCoreManager.tsx
@@ -38,6 +38,7 @@ import { useHandler } from '../../hooks/useHandler'
 import { useIsAppForeground } from '../../hooks/useIsAppForeground'
 import { lstrings } from '../../locales/strings'
 import { addMetadataToContext } from '../../util/addMetadataToContext'
+import { onAttestationToken } from '../../util/attestation'
 import { allPlugins } from '../../util/corePlugins'
 import { fakeUser } from '../../util/fake-user'
 import {
@@ -163,8 +164,20 @@ export const EdgeCoreManager: React.FC<Props> = props => {
 
   const handleContext = useHandler((context: EdgeContext) => {
     console.log('EdgeContext opened')
+    let active = true
+    const pushToken = (token: string | undefined): void => {
+      if (!active) return
+      context.setAttestationToken(token).catch((error: unknown) => {
+        if (!active) return
+        console.warn('[attestation] setAttestationToken failed', error)
+      })
+    }
+    const unsubscribeToken = onAttestationToken(pushToken)
+    // onAttestationToken sync-replays the current servable token on subscribe.
     context.on('close', () => {
       console.log('EdgeContext closed')
+      active = false
+      unsubscribeToken()
       setContext(null)
     })
     ++counter.current
@@ -197,8 +210,8 @@ export const EdgeCoreManager: React.FC<Props> = props => {
     ENV.DEBUG_EXCHANGES ? exchangeDebugUri : exchangeUri
   ]
 
-  let infoServer: string | undefined
-  let loginServer: string | undefined
+  let infoServer: string | string[] | undefined
+  let loginServer: string | string[] | undefined
   let syncServer: string | undefined
 
   if (shouldUseTestServers()) {
@@ -206,6 +219,13 @@ export const EdgeCoreManager: React.FC<Props> = props => {
     infoServer = INFO_TEST_SERVER
     loginServer = LOGIN_TEST_SERVER
     syncServer = SYNC_TEST_SERVER
+  }
+
+  if (ENV.LOGIN_SERVER != null && ENV.LOGIN_SERVER.length > 0) {
+    loginServer = ENV.LOGIN_SERVER
+  }
+  if (ENV.INFO_SERVER != null && ENV.INFO_SERVER.length > 0) {
+    infoServer = ENV.INFO_SERVER
   }
 
   return (
--- a/src/util/attestation.ts
+++ b/src/util/attestation.ts
@@ -119,6 +119,11 @@ const MAX_BACKOFF_MS = 30 * 60 * 1000
 let cachedToken: CachedToken | undefined
 let inFlight: Promise<void> | undefined
 let refreshTimer: ReturnType<typeof setTimeout> | undefined
+// Drop the JWT when it crosses the skew window, so push listeners (edge-core)
+// stop sending a token getAttestationToken would already withhold. The
+// handshake timer is not this clock: it is armed while the token is still
+// servable, and a failed refresh can leave the next tick behind a backoff.
+let serveUntilTimer: ReturnType<typeof setTimeout> | undefined
 // `undefined` means no prior stamp. Initializing these to `0` worked with
 // `Date.now()` (epoch is always far past) but a monotonic clock starts near
 // zero, so `0` would look like "just now" and park every first handshake behind
@@ -142,19 +147,7 @@ let unsupported = false
 // warning per day of app uptime so the refresh cadence cannot re-prompt.
 let lastClockWarnAtMono: number | undefined
 
-/** Test-only: clear module state between Jest cases. */
-export const resetAttestationForTests = (): void => {
-  cachedToken = undefined
-  inFlight = undefined
-  if (refreshTimer != null) clearTimeout(refreshTimer)
-  refreshTimer = undefined
-  lastFailureAt = undefined
-  lastHandshakeAt = undefined
-  consecutiveFailures = 0
-  handshakeGeneration = 0
-  unsupported = false
-  lastClockWarnAtMono = undefined
-}
+const tokenListeners = new Set<(token: string | undefined) => void>()
 
 /**
  * Whether the cached token may be handed to a gated caller right now.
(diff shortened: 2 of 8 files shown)


src/components/services/EdgeCoreManager.tsx
```diff
@@ -163,8 +164,19 @@ export const EdgeCoreManager: React.FC<Props> = props => {
 
   const handleContext = useHandler((context: EdgeContext) => {
     console.log('EdgeContext opened')
+    let active = true
+    const pushToken = (token: string | undefined): void => {
+      if (!active) return
+      context.setAttestationToken(token).catch((error: unknown) => {
```
j0ntz, line 170: `pushToken` checks `active` only at its synchronous entry, so a push already in flight when the context closes still lands after unsubscribe. Worst case is a warn or a set on the discarded context, so minor, but re-checking `active` when the call settles would tighten it.

```mermaid
sequenceDiagram
    participant att as attestation.ts
    participant mgr as EdgeCoreManager
    participant ctx as EdgeContext
    att->>mgr: listener(token)
    mgr->>mgr: active is true, proceed
    mgr-)ctx: setAttestationToken(token) async
    ctx-->>mgr: close event
    mgr->>mgr: active = false, unsubscribe
    ctx-->>mgr: earlier push settles on closed context
```

src/util/attestation.ts
```diff
@@ -457,6 +490,12 @@ const delay = async (ms: number): Promise<void> => {
 const armTimer = (delayMs: number): void => {
   if (refreshTimer != null) clearTimeout(refreshTimer)
   refreshTimer = setTimeout(() => {
+    // If the cached token can no longer be served (expiry), clear it so
+    // onAttestationToken listeners (e.g. EdgeCoreManager → setAttestationToken)
+    // drop the stale JWT before the handshake runs.
+    if (cachedToken != null && !canServeToken()) {
```
j0ntz: Nit: this guard is copy-pasted at three sites (armTimer, the handshake catch, the watchdog); a `dropUnservableToken()` helper keeps a future servability change from missing one path.

src/util/attestation.ts
```diff
@@ -142,9 +142,42 @@ let unsupported = false
...
+/**
+ * Subscribe to attestation token changes. Returns an unsubscribe function.
+ * Immediately invokes the listener with the current servable token (if any).
+ */
+export const onAttestationToken = (
+  listener: (token: string | undefined) => void
+): (() => void) => {
+  tokenListeners.add(listener)
+  try {
+    listener(canServeToken() ? cachedToken?.token : undefined)
```
j0ntz: Nit: the servable-token ternary now lives in three places (here, `setCachedToken`, and `getAttestationToken`'s tail); a `getServableToken()` used by all three keeps subscribers and pollers in lockstep.

j0ntz:
Undrafting to let bugbot run

j0ntz (approved):

src/components/services/EdgeCoreManager.tsx
j0ntz: Still open on 138680b (head unchanged since the review). Related to the Bugbot "stale token left in core" thread on this PR, which I independently confirmed: both are the push bridge lacking a guard the pull path (`getAttestationToken`) gets for free.

src/util/attestation.ts
j0ntz: Still open on 138680b.

src/util/attestation.ts
j0ntz: Still open on 138680b. A shared `getServableToken()` would also give the stale-token issue Bugbot flagged a single place to fix.

src/util/attestation.ts
```diff
@@ -142,9 +142,42 @@ let unsupported = false
...
+  cachedToken = next
+  const token = canServeToken() ? cachedToken?.token : undefined
+  for (const listener of tokenListeners) {
+    try {
+      listener(token)
+    } catch (error) {
+      console.warn('[attestation] token listener threw', error)
+    }
+  }
+}
```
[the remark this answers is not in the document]
j0ntz: Confirmed against 138680b, not a false positive. `canServeToken()` is purely time-based (`monotonicNow() < expiresMono - CLOCK_SKEW_MS`) but listeners only fire from `setCachedToken`, so the servable-to-expired transition emits nothing on its own. `getAttestationToken` re-evaluates on every read and is safe; the push bridge into edge-core is not.

The window opens when a proactive refresh fails, because the next clear then waits on the failure backoff (and RN throttles background timers):

```mermaid
sequenceDiagram
    participant tmr as refresh timer
    participant att as attestation.ts
    participant core as edge-core-js
    participant srv as login server
    att->>core: setAttestationToken(jwt)
    Note over att: scheduleRefresh at expiry minus 5 min
    tmr->>att: handshake attempt
    att--xtmr: handshake fails, arm backoff
    Note over att: token expires, canServeToken false,<br/>no listener fires
    core->>srv: request with expired jwt
    srv->>srv: verify fails on every key, force-refresh, fails again
    srv-->>core: served as unattested
    tmr->>att: backoff fires, setCachedToken(undefined)
    att->>core: setAttestationToken(undefined)
```

Impact is fail-open rather than a security hole (the server re-checks expiry), but it costs the user a CAPTCHA they earned the right to skip, and each such request takes the login server's slow verify path: the double key-loop always, plus a real JWKS fetch up to once per 60s per worker (see my `with-api-key` thread on EdgeApp/edge-login-server#194).

Cheapest fix that also closes my two nit threads: a single `getServableToken()` used by `setCachedToken`, the `onAttestationToken` replay, and `getAttestationToken`, with the refresh timer armed at expiry rather than only at expiry minus the lead, so the clear does not depend on a successful handshake.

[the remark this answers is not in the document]
j0ntz: The finding stands. The push to 6266f77 is a rebase onto newer develop, and `src/util/attestation.ts` is byte-identical across the two heads (blob `8b43905` at both 138680b and 6266f77), as is `EdgeCoreManager.tsx` (`732070e`). The re-run that returned clean and auto-resolved this thread read exactly the code that produced the finding, so I reopened it.

The mechanism is unchanged: `canServeToken()` is time-based, listeners fire only from `setCachedToken`, and after a failed proactive refresh the clear waits on backoff, so core keeps sending an expired JWT in that window.

j0ntz (changes requested):
The push to 6266f77 is a rebase onto newer develop and changes no attestation code. Both files are byte-identical to the reviewed head 138680b:

- `src/util/attestation.ts` blob `8b43905` at both commits
- `src/components/services/EdgeCoreManager.tsx` blob `732070e` at both commits

All four review items are still open:

1. `EdgeCoreManager.tsx` `pushToken` checks `active` only at synchronous entry, so a push in flight when the context closes still lands after unsubscribe.
2. The unservable-token guard is copy-pasted at three sites in `attestation.ts`.
3. The servable-token ternary is re-derived in three places.
4. Bugbot's "stale token left in core", which I independently confirmed. I reopened that thread: it was auto-resolved by the bot's own re-run on this rebase, not by a fix.

Bugbot's check-run was NEUTRAL with findings on 138680b and SUCCESS on 6266f77, over identical code. A green bot check on this PR is not evidence that the finding was addressed.

Travis is red on this head.

My earlier approval predates all of this, hence changes requested. The fix that closes items 1 through 4 together is one `getServableToken()` used by `setCachedToken`, the `onAttestationToken` replay, and `getAttestationToken`, plus arming the refresh timer at expiry rather than only at expiry minus the lead.

src/components/services/EdgeCoreManager.tsx
paullinator: **Fixed** in f20b6afbb.

`pushToken` now re-checks `active` when `setAttestationToken` rejects, so a context close during the in-flight bridge call no longer warns on a discarded context.

src/util/attestation.ts
paullinator: **Fixed** in f20b6afbb.

`dropUnservableToken()` now owns the unservable guard. `armTimer`, the handshake catch, the watchdog, and the new serve-until timer all call it.

src/util/attestation.ts
paullinator: **Fixed** in f20b6afbb.

`getServableToken()` is used by `setCachedToken` / listener notify, the `onAttestationToken` replay, and `getAttestationToken`, so push and pull stay in lockstep.

**Pushback** on using this helper as the stale-token fix: DRY is good, but listeners still only fire from `setCachedToken`. A token can become unservable while the cache still holds it, and without a timer that calls `setCachedToken(undefined)` at the skew window, core keeps the expired JWT. That gap is the dedicated serve-until timer in the Bugbot thread, not this helper by itself.

src/util/attestation.ts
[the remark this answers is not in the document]
paullinator: Agreed this was real, and that the 6266f77 rebase was not a fix (same `attestation.ts` / `EdgeCoreManager.tsx` blobs as 138680b).

**Fixed** in f20b6afbb with a dedicated `serveUntilTimer` at `expiresMono - CLOCK_SKEW_MS`, independent of handshake/backoff. When that fires, `dropUnservableToken()` → `setCachedToken(undefined)` notifies listeners so core stops sending the JWT. Covered by `fires with undefined when a cached token becomes unservable` (failed refresh + backoff longer than the skew window still emits `undefined` at expiry).

**Pushback** on arming the *refresh* timer only at expiry. That would drop the 5-minute early handshake (`REFRESH_LEAD_MS`), which is load-bearing: watchdog + backoff + one handshake still has to finish before the cached token dies. The bug is “no notify at expiry,” not “refresh too early.” Serve-until is the right shape; replacing `REFRESH_LEAD_MS` is not.

Login-server JWKS-on-every-request is already gone (`withAttestation` on EdgeApp/edge-login-server#194). This GUI fix is still worth it because the user would otherwise get a CAPTCHA they should have skipped.

j0ntz (approved):

PR #6159, opened 2026-08-14
Read ramp provider priority from the rampProviderPriority document

j0ntz:
### Technical Design Document

[ramps-deeplink-provider-priority.md](https://github.com/EdgeApp/edge-react-gui/blob/jon/ramp-provider-priority-doc/src/docs/ramps-deeplink-provider-priority.md)

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-info-server/pull/161

**This PR is red until that one merges and publishes.** It bumps `edge-info-server` to an unpublished `^3.14.0`, so `npm install` fails with `ETARGET` and every downstream check with it. The lockfile bump follows the publish.

### Requirements

No visual changes: this reorders an existing list, it does not alter any component.

- [x] Tested on iOS device

### Description

Asana: https://app.asana.com/0/1215088146871429/1217207276378713

Moves the ramps affiliate provider ordering off `promoCards2` and onto the info server's new `rampProviderPriority` document.

The read this replaces went through `filterInfoCards` rather than `getDisplayInfoCards` specifically so that a card with `localeMessages: {}` could configure ordering and never render. That made behavior a display card wearing a disguise, which is why dismissing a visible card hid the ad but left its ordering running.

What changed:

- `useRampPreferredProviders` reads `infoServerData.rollup?.rampProviderPriority`. The `filterInfoCards` read is gone; the function itself stays, since the promo carousel still uses it.
- Matching moved into `src/util/rampProviderPriority.ts` as pure functions, so the rules are unit-testable without a store. An entry matches on `installerId` **or** an active promotion, rather than requiring both to agree when both were present, which had excluded a user who picked up the promotion after installing from somewhere else.
- `getActivePromoIds` now unions the `promoCards2`-derived ids with the `rampProviderPriority` keys that apply. Without this the split would be incomplete: `activePromotions` was derived from cards alone, so a promo code naming a document entry was dropped on entry and again at every login, and the only way to activate one would have been to keep a display card alive purely to name its promo id.
- `src/docs/ramps-deeplink-provider-priority.md` is updated to the new mechanism, with an `edge-info-server` design section and a phase 4 history entry.

### Testing

17 new unit tests in `src/__tests__/util/rampProviderPriority.test.ts` cover the OR match, per-direction scoping, country include/exclude, date windows, an unparseable date, dedupe across entries, and the promo-id derivation. Full suite: 99 files, 720 tests, 107 snapshots, plus `tsc` and `eslint` clean against a locally built copy of the dependency.

On the iOS sim, in a throwaway account, driven end to end with the document hard-coded locally in place of the server response (the deployed info server does not serve it yet), so the real matching and ordering code ran against it. Screenshots below: the option list with no promotion, the code activated in Promotion Settings, and the same list reordered.

The middle screenshot is also the evidence for the `getActivePromoIds` change. Before it, Promotion Settings accepted the code and listed nothing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - added: "-m" tag on the version number in the Help scene for Maestro test builds
 - added: Sign Message option in the wallet list menu for Bitcoin-family wallets, letting users prove self-hosted wallet ownership to exchanges by signing an exchange-provided message.
 - added: `edge://buy` and `edge://sell` deep links (and their `https://deep.edge.app` equivalents) that open the buy/sell flow, optionally pinning a provider and payment method to the top of the quote options for that visit.
-- added: Provider priority in the buy/sell options for affiliated accounts, configured through the info server promo card data.
+- added: Provider priority in the buy/sell options for affiliated accounts, configured through the info server's `rampProviderPriority` document. An account matches an entry by its installer id or by holding the promotion, and entries scope by country and date.
 - changed: Target Android 16 (API level 36), which Google Play requires for app updates submitted after Aug 30, 2026. Predictive back is opted out of for now, since React Native 0.79 cannot handle it, so the back button behaves exactly as it did before.
 - changed: Sign MoonPay buy/sell widget URLs and bind them to the customer's IP via the info server, for MoonPay's on-ramp IP-matching security upgrade.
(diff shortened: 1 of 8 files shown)


j0ntz:
## 📸 Test evidence

**baseline no promo**

baseline no promo

**promo activated**

promo activated

**paybis first**

paybis first

PR #6169, opened 2026-08-21
Apply custom token additions on Manage Tokens save

j0ntz:
### Description

Adding a custom token from Manage Tokens left the screen in two wrong states:
the outer Save stayed greyed out unless a default token had also been toggled,
and the new token did not join the enabled tokens at the top of the list. It was
only findable by searching for it by name.

Both came from the deferred-save state added in "Add deferred save to
ManageTokensScene":

- The effect that syncs externally-enabled tokens added the new token to
  `pendingEnabledTokenIds` **and** to `baselineSet`, so the addition never
  counted as an unsaved change and the Save button stayed disabled.
- `sortingBaselineSet` was a `useMemo` with empty deps, frozen at mount, so a
  token that appeared later sorted with the disabled tokens.

The fix defers the enable to the Manage Tokens Save button, which is what the
task asks for:

- `EditTokenScene` takes a `deferEnable` param. When set, and only for a
  brand-new custom token, it creates the token and leaves the enable to the
  caller. The other entry points (wallet list, scan, create-wallet) do not pass
  it and keep enabling immediately.
- `ManageTokensScene` passes `deferEnable`, spots the new custom token id, and
  puts it in the pending set and the sorting baseline. The outer Save then
  applies it, and it renders with the enabled tokens at the top.
- Tokens enabled from outside the scene now also join the sorting baseline, so
  they show up at the top too instead of only through a name search.

`testID`s were added to the wallet list rows and the Manage Tokens rows so the
maestro flow could target them; the wallet rows had no stable selector and their
accessibility strings are whole-row concatenations.

Asana: https://app.asana.com/0/1215088146871429/1217694458639427

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Requirements

- [x] Tested on iOS device

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@
 - fixed: Improve the unstake error experience by replacing the popup alert and generic "unknown error occurred" with the real error in the scene's error field, and showing a clear message when the wallet lacks the balance to cover the unstaking network fee.
 - fixed: Tapping Max on the Sell scene no longer briefly shows the entered fiat amount in the crypto field while the max is being calculated.
 - fixed: An info card no longer disappears into an empty gap when the carousel's card list shrinks. A card's position comes entirely from an animated transform keyed on its index, and that transform is not re-applied when a surviving card shifts slots, so dropping a card left the ones after it parked a full card-width off-screen. The carousel now remounts a card whose slot changes. Reproduces wherever the list shrinks after mount - most visibly when a `noBalance` card is filtered out as balances finish loading.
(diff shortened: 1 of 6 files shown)


j0ntz:
## 📸 Test evidence (iOS sim, maestro)

**manage tokens before**

manage tokens before

**add token edg**

add token edg

**new token enabled save active**

new token enabled save active

**saved back on assets**

saved back on assets

**persisted after save**

persisted after save

PR #6171, opened 2026-08-21
fix(maestro): persist GraalJS state via output

EdgeClaws:
## Scope

Isolates the GraalJS compatibility repair from the existing Maestro-test fixes.

Cross-command JavaScript state now uses Maestro `output.*` instead of local `var` assignments, covering the shared account/wallet helpers and affected existing flows.

## Verification

- `git diff --check`
- No device run was launched for this commit split; existing test evidence remains attached to the individual test repairs.

Stack base for the clean Maestro-fixes review branch.

### DIFF
--- a/maestro/07-wallets/C000032-archive-and-restore-wallets.yaml
+++ b/maestro/07-wallets/C000032-archive-and-restore-wallets.yaml
@@ -22,38 +22,38 @@ tags:
 
 
 # Archive wallets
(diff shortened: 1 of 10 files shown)


EdgeClaws:
Superseded by #6172, which now contains this unchanged YAML-only GraalJS state-persistence commit plus all current Maestro test repairs in one direct develop-based PR.

PR #6172, opened 2026-08-21
fix(maestro): repair recent existing test flows

EdgeClaws:
QA owns the Maestro YAML in this PR. **Developer review is limited to the three TypeScript files and their matching changelog entries below.**

## Review scope

Four `testID` additions:

- `WalletSettingsModal.tsx` — `walletSettingsNameInput`, `walletSettingsDoneButton`
- `CreateWalletEditNameScene.tsx` — `splitWalletConfirmButton`
- `ManageTokensRow.tsx` — `manageTokensRow.<currencyCode>`
- `CHANGELOG.md` — three `changed:` entries covering those hooks

These expose existing controls to Maestro. There are no visible UI, behavior, or business-logic changes.

## Dependency context

`develop` now supplies published `edge-login-ui-rn@^3.37.2`. The temporary GitHub pin used for build validation is gone; 3.37.2 contains the exact spinner-ID source validated in matched Android/iOS build `26083001`.

QA verified all four app hooks on iPhone 17 and iPhone 17e. [Asana handoff](https://app.asana.com/1/9976422036640/project/1213843652804305/task/1217893556844537)

### DIFF
--- a/src/components/themed/ManageTokensRow.tsx
+++ b/src/components/themed/ManageTokensRow.tsx
@@ -83,7 +83,11 @@ export const ManageTokensRowComponent: React.FC<Props> = props => {
   })
 
   return (
-    <Pressable style={styles.row} onPress={handleToggle}>
+    <Pressable
+      style={styles.row}
+      testID={`manageTokensRow.${token.currencyCode}.${token.displayName}`}
+      onPress={handleToggle}
+    >
       <CryptoIcon
         marginRem={[0, 0.5, 0, 0]} // We don't need left margins because there's no border. This component effectively is the left "border"
         sizeRem={2}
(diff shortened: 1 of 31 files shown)


src/components/themed/ManageTokensRow.tsx
```diff
@@ -83,7 +83,11 @@ export const ManageTokensRowComponent: React.FC<Props> = props => {
   })
 
   return (
-    <Pressable style={styles.row} onPress={handleToggle}>
+    <Pressable
+      style={styles.row}
+      testID={`manageTokensRow.${token.currencyCode}`}
```
peachbits: use a more unique string like concatenating the currency code and displayname

peachbits (changes requested):

src/components/themed/ManageTokensRow.tsx
JP0P: We did consider this but didn't want to complicate it too much. Making that exact change now.

peachbits: I took another look and there's a better option and that's to use the pluginId tokenId together. I didn't realize those were in scope during the first review. It's more correct since collisions are impossible. Currently, collisions are very unlikely and since it's for the testid field it's not as important to get it right. It also complicate test writing since you'll need to know how to derive the tokenId which can vary depending on plugin. I'll leave it up to you to determine if the extra headache of identifying assets is worth the cost for correctness.

peachbits (approved):

PR #6183, opened 2026-08-29
Native HMAC apiSigner and signed infoRollup appKeys

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

- https://github.com/EdgeApp/edge-core-js/pull/739
- https://github.com/EdgeApp/edge-info-server/pull/163

### Requirements

- [x] Tested on iOS device

### Description

Splits runtime `env.json` into non-secret `config.json` and secret `keys.json`, with four plugin maps (`corePlugins`, `swapPlugins`, `guiApiKeys`, `rampPlugins`) plus `globalKeys`. WalletConnect is `globalKeys.WALLETCONNECT_PROJECT_ID`. Plugin maps are opaque objects (no field-by-field `*_INIT` / `*_API_KEY` flattening).

Fetches remote secrets from signed `GET /v1/infoRollup/:appId` as sibling `appKeys` (partner id `config.appId ?? 'edge'`), with DeviceSettings `getKeysCache` and baked-in `keys.json` as fallbacks. `pluginApiKeys.posthog` is never served.

Adds a native Edge API HMAC signer (`edgeKey.json` + XOR-split C shards) so login-server requests can be signed outside the JS bundle via `apiSigner`, with JS `KEYS.EDGE_API_*` remaining as a fallback.

Rebased onto current `develop` (keeps Swapter and marketing-push tracking).

---

> [!NOTE]
> **High Risk**
> Changes authentication (native HMAC + infoRollup signing), secret storage, and cold-start boot gating before edge-core initializes—any regression affects login, partner plugins, or first-launch behavior.
> 
> **Overview**
> Replaces the monolithic **`env.json`** setup with **`config.json`** (non-secret flags and plugin enablement), **`keys.json`** (secrets and plugin init objects), and build-only **`edgeKey.json`**. Runtime code reads **`CONFIG`**, **`KEYS` / `globalKeys`**, and resolved **`pluginMaps`** instead of a flat **`ENV`**, and deploy/CI now deep-merge **`configJson` / `keysJson`** branch overrides (legacy **`envJson`** is ignored with a migration error when needed).
> 
> Adds a **native Edge API HMAC signer** on iOS and Android: Gradle/Xcode generate **`edge_api_secret.c`** from XOR-split shards, expose **`EdgeApiSigner`** to React Native, and pass **`apiSigner`** into edge-core for login-server signing while keeping **`KEYS.EDGE_API_*`** as a JS fallback.
> 
> At cold start, the app **blocks on signed `GET /v1/infoRollup/:appId`** (or **`DeviceSettings` `keysCache`**) before building plugins, merges **`appKeys`** over baked-in secrets without hot-swapping a running core, and caches overlays for the next launch. Migration tooling (**`split-env-json`**, **`split-baked-and-server-keys`**) and architecture docs describe the new shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d2229ddd7805afa71df3a23c1f4a42da314ce9b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/hooks/useRampPlugins.ts
```diff
@@ -51,7 +51,7 @@ export function useRampPlugins({ account }: UseRampPluginsOptions): {
             const store = createStore(storeId, account.dataStore)
 
             // Create a minimal config for the plugin
-            const initOptions = ENV.RAMP_PLUGIN_INITS[pluginId]
+            const initOptions = pluginMaps.rampPlugins[pluginId]
 
             // If there is no init option defined for the plugin, simply skip over it
             if (initOptions == null) {
```
j0ntz, line 57: This guard still only rejects `null`, but the new merge can hand back a bare `true`.

`mergePluginInit(true, undefined)` returns `true` (`src/configKeysMerge.ts:88`), which is the "enabled in config.json, no credentials on the keys side yet" state. That is the shipped state on any build where `slimKeysJson` stripped the plugin maps and the signed appKeys fetch has not landed. `true` is not `null`, so it flows into `factory(config)` as `initOptions`, and `asInitOptions(true)` throws inside the plugin's `asObject` cleaner (e.g. `banxaRampPlugin.ts:596`, `paybisRampPlugin.ts:470`). The per-plugin `catch` only `console.warn`s, so the ramp option disappears from buy/sell for the whole session, and the effect deps are `[account, navigation, dispatch]` so it never retries.

`initializeProviders.ts:47` in this same PR added exactly this guard for `guiApiKeys`, with a comment explaining the case. The same test belongs here:

```ts
if (initOptions == null || typeof initOptions === 'boolean') {
  continue
}
```

(The boolean case only bites GUI-side loaders. Core normalizes `true` to `{}` at `plugins-actions.ts:82`, so `corePlugins.ts` is fine as written.)

src/util/edgeApiSigner.ts
```diff
@@ -0,0 +1,155 @@
...
+export function makeNativeApiSigner(): EdgeApiSigner {
+  const module = getNativeModule()
+  if (module == null) {
+    throw new Error('EdgeApiSigner native module is not available')
+  }
+  return {
+    async signMessage(message: string) {
+      const signed = asSignedMessage(await module.signMessage(message))
+      // Stub builds embed a placeholder that is not a valid Authorization value.
+      if (!isUsableApiKey(signed.apiKey) || signed.signature === '') {
```
j0ntz, line 73: `isUsableApiKey` rejects whitespace in `apiKey`, but `signature` is only checked for emptiness, and it goes straight into `Authorization: HMAC ${apiKey} ${signature}` in `keysServer.ts`. A native binding that returned a signature with a stray space, tab, or CR/LF would produce a header the server cannot split into three parts, and the failure would look like a signing mismatch rather than a malformed header.

The JS path cannot produce this (it is `base64.stringify(...)`), so the native delegate is the first place an unvalidated string reaches the header. Cheap to close:

```ts
if (!isUsableApiKey(signed.apiKey) || !isUsableApiKey(signed.signature)) {
```

edge-core-js#739 has the same asymmetry at `login-fetch.ts:34`.

src/components/services/EdgeCoreManager.tsx
```diff
@@ -139,6 +184,51 @@ export const EdgeCoreManager: React.FC<Props> = props => {
...
+      } catch (error: unknown) {
+        // initializeKeys itself never rejects, but buildContextOptions can.
+        // Without a fallback, contextOptions stays null, Providers/Airship never
+        // mount, and native BootSplash never hides.
+        console.warn(
+          'EdgeCoreManager: keys boot failed; using baked-in plugins',
+          String(error)
+        )
+        try {
+          setContextOptions(await buildContextOptions())
```
j0ntz, line 212: This fallback re-runs the identical call with identical inputs.

The comment says `initializeKeys` itself never rejects, which leaves `buildContextOptions()` as the only thing in the `try` that can throw. Calling it a second time with nothing changed in between will throw again for the same reason, so the catch adds a duplicate `[apiSigner]` log line and a second `warmNativeApiKey` round trip and then lands in the inner catch anyway.

If the goal is "boot with baked-in plugins when the keys store fails", the retry needs to differ from the first attempt (skip the native signer, for instance). Otherwise the outer try/catch can collapse to the inner one.

src/components/services/EdgeCoreManager.tsx
```diff
@@ -221,19 +302,31 @@ export const EdgeCoreManager: React.FC<Props> = props => {
...
+    loginServer = CONFIG.LOGIN_SERVER
+  }
+  if (CONFIG.INFO_SERVER != null && CONFIG.INFO_SERVER.length > 0) {
+    infoServer = CONFIG.INFO_SERVER
+  }
+
+  if (bootFatalError != null) {
+    return (
+      <View style={{ flex: 1, justifyContent: 'center', padding: 24 }}>
+        <Text>Edge failed to start: {bootFatalError}</Text>
```
j0ntz, line 315: The string is hardcoded English rather than `lstrings.*`, and the style is inline instead of `cacheStyles`/`getStyles`.

The modal this replaced at least pulled `lstrings.string_ok_cap`. `lstrings` is a plain import with no dependency on Providers or Airship, so it is available here. A `boot_failed_message_1s` key plus a `getStyles` entry keeps this consistent with the rest of the app.

src/components/services/EdgeCoreManager.tsx
```diff
@@ -125,12 +119,63 @@ const crashReporter: EdgeCrashReporter = {
...
+async function buildContextOptions(): Promise<EdgeContextOptions> {
+  const { EDGE_API_KEY: apiKey, EDGE_API_SECRET: apiSecret } = KEYS
+  const nativeKey = hasNativeApiSigner() ? await warmNativeApiKey() : ''
+  const nativeApiSigner = nativeKey !== '' ? makeNativeApiSigner() : undefined
+  const jsPair =
+    isUsableApiKey(apiKey) && apiSecret != null && apiSecret.byteLength > 0
+      ? { apiKey, apiSecret }
+      : undefined
+  console.log(
+    `[apiSigner] native=${nativeApiSigner != null} keysFallback=${
```
j0ntz, line 131: Unguarded `console.log` in production, same as `[keys] tier=...` at `keysStore.ts:428`. Both are useful when checking a device run, so rather than dropping them, route them through the category logger the way Phaze does:

```ts
debugLog('keys', `tier=${keysTier} assurance=${assuranceLevel ?? 'none'} ...`)
debugLog('keys', `apiSigner native=${nativeApiSigner != null} keysFallback=${jsPair != null}`)
```

`debugLog` (`src/util/logger.ts`) is silent unless `keys` is in `LOG_CONFIG.enabledCategories` in `config.json`, and `enableLogCategory('keys')` turns it on at runtime. The `console.error` below reports a real misconfiguration and should stay as is.

j0ntz (commented):
Review of this PR together with its two dependencies, edge-core-js#739 and edge-info-server#163. The HMAC design itself checks out: I compiled `native/edge-api-signer/edge_hmac.c` standalone and it matches SHA-256("abc"), SHA-256(""), SHA-256(1e6 x 'a'), and RFC 4231 HMAC cases 1, 2, 3, 6 and 7, including the >block-size key path, clean under `-Wall -Wextra`. The signed string also lines up end to end: `keysServer.signPath` builds `/v1/infoRollup/...` and the server verifies `req.originalUrl` under the `/v1` mount (`src/indexInfo.ts:44`), with a shared test vector on both sides.

Two repo-wide items with no single line to hang them on:

1. **Em dashes (U+2014) in committed code and comments.** 68 added lines carry one, 28 of them outside `docs/`: `scripts/makeApiSigner.ts` (6, including the `auto-generated ... do not edit` banner that ends up in the generated native C), `scripts/splitEnvJson.ts` (4), `src/util/keysStore.ts` (4), `src/keys.ts` (2), `src/util/edgeApiSigner.ts` (2), `scripts/splitBakedAndServerKeys.js` (2), plus one each in `configKeysMerge.ts`, `configKeysSchema.ts`, `initializeProviders.ts`, `types/types.ts`, `network.ts`, `tracking.ts`, `makeNativeHeaders.ts`, and 41 in `docs/`. A comma, colon, or parentheses reads the same. Ruleset: https://github.com/EdgeApp/edge-dev-agents/blob/main/.cursor/skills/no-slop/SKILL.md

2. **Commit hygiene.** `5707c51c0` has a 68-character subject ("Log appKeys LAYER sentinels and native HMAC signer status on launch."); the cap is 50. Something like "Log keys tier and signer status on launch" fits. `568056f7c` also carries a `Co-Authored-By: Claude Opus 5` trailer.

src/actions/NotificationActions.ts
```diff
@@ -354,12 +367,16 @@ export const fetchLegacySettings = async (
   )
 }
 
-async function legacyGet(path: string) {
+async function legacyGet(path: string): Promise<any> {
```
j0ntz, line 370: New explicit `any` return type. `fetchPush` returns an `EdgeFetchResponse`, so this is `Promise<unknown>` at worst, and the one caller (`fetchLegacySettings`) already declares the concrete shape `{ '1': boolean; '24': boolean; fallbackSettings?: boolean }`. Returning `Promise<unknown>` and cleaning at the call site, or hoisting that shape into a cleaner, keeps the type checker on.

src/configKeysMerge.ts
```diff
@@ -0,0 +1,225 @@
...
+      asUnknownMap(keys.rampPlugins ?? {})
+    )
+  }
+}
+
+/**
+ * Truncate a single secret to its first 8 characters so it can be shown for
+ * debugging without leaking the full value. Non-strings are returned as-is.
+ */
+export function redactKey(value: unknown): unknown {
```
j0ntz, line 204: `redactKey` has no callers outside `src/__tests__/configKeysMerge.test.ts`, and its body is the same scalar branch `redactValue` already implements two functions down. Worth deleting along with its test case, so a future change to the truncation length has one place to land.

src/hooks/useRampPlugins.ts
j0ntz: Addressed in f152e784d: the guard skips a boolean entry as well as null. A related case (ramp inits stored as `{}`) is in this review.

src/util/edgeApiSigner.ts
j0ntz: Addressed in 2fe55f93e: the signature now goes through `isUsableApiKey` too.

src/components/services/EdgeCoreManager.tsx
j0ntz: Addressed: at fac86e736 the boot path is a single try/catch around `initializeKeys` and `buildContextOptions`, with no retry.

src/components/services/EdgeCoreManager.tsx
j0ntz: Addressed in 48a56d8b7 and 61d62149e: the message comes from `lstrings.boot_failed_message_1s` and the styles from `cacheStyles`.

src/components/services/EdgeCoreManager.tsx
j0ntz: Addressed in 16544a6e7: both lines now go through `debugLog('keys', ...)`.

src/actions/NotificationActions.ts
j0ntz: Addressed in b5ce9231c: `legacyGet` returns `unknown`, and the callers clean the responses with `asV1Settings` and `asLegacySettings`.

src/configKeysMerge.ts
j0ntz: Addressed in 67ee6545c: `redactKey` and its test are removed.

src/util/corePlugins.ts
```diff
@@ -1,141 +1,192 @@
...
+    binancesmartchain: coreInit('binancesmartchain'),
+    bobevm: true,
+    botanix: coreInit('botanix'),
+    cardano: coreInit('cardano'),
+    cardanotestnet: coreInit('cardanotestnet'),
+    mayachain: coreInit('mayachain'),
+    celo: coreInit('celo'),
+    coreum: coreInit('coreum'),
+    cosmoshub: coreInit('cosmoshub'),
+    ecash: coreInit('ecash'),
```
j0ntz: **Warning.** `ecash` moved out of the UTXO list into `buildAccountbasedPlugins`. This came from the rebase: develop's ebf40024a ("Add slow-sync card for large UTXO wallets") split the table into `buildAccountbasedPlugins` / `buildUtxoPlugins` and put `ecash: ENV.ECASH_INIT` in the UTXO half (`corePlugins.ts:85` at 0c4d2e96f). The rebased commit adopted the split but kept `ecash` in the accountbased half, and it is the only id that left the UTXO set (diffed the key lists). `getUtxoPluginIds()` is `Object.keys(utxoPlugins)`, so `isUtxoPluginId('ecash')` is now false and `TransactionListTop.tsx:178` stops showing the large-UTXO slow-sync banner for eCash wallets. eCash is a UTXO plugin (edge-currency-plugins `utxobased/info/ecash`). Move `ecash: coreInit('ecash')` back into `buildUtxoPlugins`.

src/configKeysMerge.ts
```diff
@@ -0,0 +1,278 @@
...
+export function mergePluginInit(
+  configValue: unknown,
+  keysValue: unknown
+): unknown {
+  if (configValue === false) return false
+  // Keys are never an off switch: only config.json can disable a plugin. A
+  // remote/cache overlay that says `false`, or hands back `{}`, contributes
+  // nothing and leaves the config-side value exactly as it was. `deepMerge`
+  // applies the same rule one level down.
+  if (keysValue === false || keysValue === null || isEmptyObject(keysValue)) {
```
j0ntz: **Warning.** `mergePluginInit(true, {})` now returns `true` (asserted at `configKeysMerge.test.ts:144`), and `useRampPlugins.ts:62` skips booleans. `bitsofgold` and `libertyx` take an all-optional `asInitOptions`, so a legacy `{}` meant "enabled with defaults"; `scripts/splitEnvJson.ts:256` still writes `{}` to keys for exactly that reason ("so merge(true, {}) === {}"). Any ramp plugin configured that way now drops out of buy/sell silently. Keeping `{}` as a real value in `mergePluginInit` (the "no opinion" rule belongs in `mergeKeysOverlay`, which already has it) would restore it. I have not checked this against the production secrets repo.

src/locales/en_US.ts
```diff
@@ -1480,6 +1483,8 @@ const strings = {
   price_change_notification: 'Price Notification',
   price_change_buy_sell_trade: 'Would you like to buy, sell, or exchange %1$s?',
   // Update notices
+  plugins_unavailable_message_s:
```
j0ntz: `plugins_unavailable_message_s` uses a bare `%s`. The string convention is numbered suffixes and positional args: `plugins_unavailable_message_1s` with `%1$s`, as this PR already does for `boot_failed_message_1s`.

src/util/keysStore.ts
```diff
@@ -0,0 +1,500 @@
...
+ * won so a runtime check can prove the remote path was exercised.
+ *
+ * Warm start (any mergeable cache) unblocks on the cache and refreshes in the
+ * background for the *next* launch, so keys are never hot-swapped underneath a
+ * running core and later launches do not stall on the network. The cache does
+ * not expire. Cold start (first launch / no usable cache) blocks for at most
+ * `COLD_TOTAL_TIMEOUT_MS`, then falls through to the baked-in file.
+ */
+async function doInitializeKeys(): Promise<void> {
+  let settingsTimer: ReturnType<typeof setTimeout> | undefined
```
j0ntz: The timer + `new Promise` + `Promise.race` + `clearTimeout` block is written out four times in this file (settings gate, cold gate, salvage wait, background refresh) and twice more in `DeviceSettingsActions.ts`. A small `raceTimeout(promise, ms, onTimeout)` helper would keep the `clearTimeout` handling in one place.

src/plugins/stake-plugins/generic/pluginInfo/cardanoKilnPool.ts
```diff
@@ -18,8 +22,12 @@ const kilnPolicyConfig: Array<
       type: 'cardano-pooled-kiln',
       pluginId: 'cardano',
 
-      accountId: ENV.KILN_MAINNET_ACCOUNT_ID,
-      apiKey: ENV.KILN_MAINNET_API_KEY,
+      get accountId() {
```
j0ntz: The `get accountId()` / `get apiKey()` pair is repeated on all 10 Kiln entries (8 here, 2 in `ethereumKilnPool.ts`). One shared getter object spread into each `adapterConfig` would make a future key rename a one-line change.

src/components/services/EdgeCoreManager.tsx
```diff
@@ -125,19 +123,117 @@ const crashReporter: EdgeCrashReporter = {
...
+        // Providers/Airship never mount, and native BootSplash never hides.
+        hideSplash()
+        setBootFatalError(String(error))
+      }
+    },
+    [],
+    'EdgeCoreManager'
+  )
+
+  // Cache the public API key from native for push / notification callers:
```
j0ntz: The "Cache the public API key from native" effect repeats work that is already done before it can matter: `buildContextOptions` (line 128) awaits `warmNativeApiKey()`, and the module-scope `initializeKeys()` does the same through `fetchKeysInner`. The single-flight makes it harmless, but the three separate "native signer present and key usable" checks can drift; consider one helper that returns the signer or undefined.

j0ntz (approved):
Approving on the condition that the two Warning comments are fixed before merge: eCash belongs back in the UTXO plugin list, and `{}` ramp inits need to stay enabled. The rest are suggestions.

PR #6189, opened 2026-09-02
Explain an empty Manage Tokens search result

j0ntz:
### Description

Manage Tokens now says when a search matches nothing.

Searching Manage Tokens for a contract address the wallet's token list does not carry rendered a bare scene: no rows, no message, and the "Add Custom" button hidden behind the keyboard. That frame is pixel-identical to a search that is not matching at all, so nobody can tell "this token is not in the list" from "the search is broken".

That ambiguity is what this PR's task is about. QA filed [1217587221308437](https://app.asana.com/0/1215088146871429/1217587221308437) against the Manage Tokens contract-address search, that search was fixed in [EdgeApp/edge-react-gui#6170](https://github.com/EdgeApp/edge-react-gui/pull/6170), and a later QA pass attached another blank-scene screenshot. This run reproduced the QA steps on current `develop` and the search works: the full MOG contract address `0xaaeE1A9723aaDB7afA2810263653A34bA2C21C7a` returns the Mog row, and so does UNI's `0x1f9840a85d5af5bf1d1762f925bdaddc4201f984`. What the blank frame cannot show is which of the two causes produced it.

MOG makes the gap easy to hit. It is not a built-in token: it is absent from `edge-currency-accountbased`'s compiled Ethereum list and from the info-server token payload, so it only exists in an account after someone adds it through "Add Custom". On an account that never did, an address search for it correctly returns nothing, and says nothing.

The fix renders a centered message in the list's empty slot while a search is active. A search that matches nothing now says so, and points at the control that adds the token.

Asana: https://app.asana.com/0/1215088146871429/1218117826000799

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Requirements

- [x] Tested on iOS device

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased (develop)
 
+- added: A "no tokens match your search" message on the Manage Tokens search. An address for a token the wallet's list does not carry used to render an empty scene, which is indistinguishable from a search that is not working.
+
 ## 4.51.0 (staging)
 
 - added: Push info-server attestation tokens into edge-core-js via `setAttestationToken` so the login server can skip CAPTCHA for attested devices, and allow `LOGIN_SERVER` / `INFO_SERVER` env overrides for local E2E stacks.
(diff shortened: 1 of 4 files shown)


j0ntz:
## 📸 Test evidence

**mog full address finds mog**

mog full address finds mog

**unknown address says no match**

unknown address says no match

PR #6190, opened 2026-09-03
Lock the send slider once a broadcast has been attempted

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Requirements

- [x] Tested on iOS device

### Description

GUI half of the "send failed in the UI but the money moved" incident. Asana: https://app.asana.com/1/9976422036640/project/1213880789473005/task/1217135300337949

Engine companion (independent, neither blocks the other): EdgeApp/edge-currency-plugins#458

One intended Bitcoin send became five real payments. Every broadcast succeeded over the HTTP fallback, the engine's `saveTx` then threw, and `handleSliderComplete` treated that throw like any other failure: a generic network error card plus a `finally` that re-armed the slider. Because `saveTx` marks the inputs spent before it throws, each retry re-quoted on the remaining UTXOs, so every slide was a fresh real payment.

A failure reported by `broadcastTx` does not prove the transaction is absent from the network. A server can relay it and still fail to answer, and the network conditions that broke the broadcast also make an immediate "did it land?" check unreliable. Polling was considered and rejected: it blocks the flow and any timeout is arbitrary. So the scene treats the broadcast call as a one-way boundary instead.

**The slider is now idempotent.** A ref is set immediately before `broadcastTx` is called. Once set, the `finally` block never re-arms the slider, on success or failure, and the slider renders locked as "Send Attempted" for the life of the scene. The ref rather than state is what the handler reads, so the FIO no-bundled retry recursion cannot re-arm it either.

**Honest messaging replaces the generic error.** Failures at or after the boundary show "Transaction Status Unknown", telling the user the transaction may have reached the network, to check a block explorer or wait for a confirmation email, and that sending again could produce a duplicate payment. Separate copy distinguishes a broadcast that reported failure from an error after a successful one. There is no fake happy path and no silent success.

**The card is sticky and the scene does not call `onDone` with the error.** Once a broadcast has been attempted the quote effect is frozen, the address-reset handler no longer clears `error`, and the quote-expiry callback is a no-op, so editing the amount or address, or the quote timing out, cannot remove the only explanation the user has for the locked slider. For gift cards `onExpired` navigates back, so without that guard expiry would also have popped the scene. The locked branch deliberately does not call `onDone(error)`: the ramp launchers (paybis, moonpay) `navigation.pop()` the send scene when that promise rejects and show a generic failure, which would hide the card and return the user to a flow that can start another real payment. Those flows still terminate through the existing `onBack`-on-unmount path when the user leaves, as on develop.

**Pre-broadcast failures are unchanged.** PIN, the `beforeTransaction` hook, the FIO fee check and signing all still show their existing errors and still re-arm the slider, because nothing can have been sent. The hook failure also now calls `resetSlider()` instead of leaving the slider spinning, which was a pre-existing bug.

`SafeSlider` gains an explicit `lockedText` prop for this. Overloading the existing `disabled` prop was the first attempt and was wrong: a completed slider that a parent then disables would swap its spinner for `disabledText`, which defaults to "Enter an Amount". That would have affected the other fourteen scenes using `SafeSlider` whenever a parent disables mid-callback, which `SendScene2` itself can do via `processingAmountChanged` and `hasPendingTx`.

The locked-state card is longer than a normal error and the slider floats over the bottom of the scroll view, so the existing `needsScrollToEnd` mechanism scrolls it into view. That is a one-line reuse of the pattern four other handlers already use.

### Testing

Driven on the iOS simulator against the real send scene, with the broadcast stubbed so nothing could reach the network and using a zero-balance wallet. Run twice: once on the original commit, and again on an iPhone 17 Pro Max after the review fixups.

| Scenario | Result |
|---|---|
| Broadcast succeeds, then `saveTx` throws "No addresses to process" (the incident) | Slider locks, post-broadcast copy shown, card auto-scrolled fully above the slider |
| Broadcast itself throws (ambiguous) | Slider locks, ambiguous-failure copy shown |
| Broadcast throws an engine-named rejection (`ErrorEosInsufficientCpu`) | Slider locks, the EOS-specific copy shows instead of "status unknown" |
| Signing throws before any broadcast | Slider re-arms to "Slide to Confirm"; the same scene then locks on the next attempt |
| Editing the amount on a locked scene | Card stays, lock stays, fee tile stays static (a fee-tile spinner found here is fixed in the last fixup) |
| Dragging the locked slider across the full track, both directions | No movement, no second send |

Existing jest suite: 722 passing, including `SendScene2.ui.test.tsx`. `tsc` and `eslint` clean.

### Known gap

The lock is per scene instance. Backing out to the wallet and tapping Send again gives a fresh slider. That matches the direction agreed for this task, where the friction of re-entering the flow is the point, but it is not a hard guarantee against a determined second send.

---

> [!NOTE]
> **Medium Risk**
> Changes core send UX and error handling where mistaken retries can cause duplicate payments, though scope is mostly SendScene2 plus an optional SafeSlider prop.
> 
> **Overview**
> Prevents duplicate sends when the UI reports failure but a transaction may already be on the network (e.g. broadcast succeeds then `saveTx` throws).
> 
> **Send confirmation is one-shot per scene visit.** After signing, the scene marks a broadcast boundary before calling `broadcastTx`. The confirm slider stays completed and shows **"Send Attempted"** via a new `SafeSlider` **`lockedText`** prop—it no longer re-arms on success or ambiguous failure. Pre-broadcast errors (PIN, hooks, signing, deterministic node rejections like EOS CPU) still reset the slider as before; `beforeTransaction` hook failures now also reset the slider instead of leaving it spinning.
> 
> **Messaging and sticky state.** Ambiguous failures at or after broadcast use new **"Transaction Status Unknown"** copy (explorer / confirmation email, duplicate-payment warning), with separate strings for broadcast-time vs post-broadcast errors. Named engine rejections keep their specific messages. The scene freezes re-quoting, ignores quote expiry, and does not clear the error when resetting address fields; it avoids calling ramp **`onDone(error)`** so launchers do not pop the scene and hide the warning.
> 
> New locale strings and a CHANGELOG entry document the behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 908d8a910c9d7bd6a3b1b2793fbc445e87259e51. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/components/scenes/SendScene2.tsx
```diff
@@ -1324,12 +1336,20 @@ const SendComponent: React.FC<Props> = props => {
         }
 
         const signedTx = await coreWallet.signTx(edgeTransaction)
+
+        // From this point on the transaction may reach the network, so lock
+        // the slider for the life of this scene no matter what happens next.
+        // The render-side flag is set in the finally block, so the slider
+        // keeps its spinner while the attempt is in flight.
+        broadcastAttemptedRef.current = true
```
j0ntz, line 1344: Setting the ref here is right for the lock. Using the same ref at line 1530 to choose the *message* is what causes trouble: deterministic rejections thrown from inside `broadcastTx` now read as ambiguous.

`EosEngine.broadcastTx` maps `tx_cpu_usage_exceeded` / `tx_net_usage_exceeded` / `ram_usage_exceeded` onto `ErrorEosInsufficientCpu` / `Net` / `Ram` **after the node has explicitly refused the transaction**. Those now hit the early return, so an EOS user short on CPU is told the transaction "may still have gone through" and that "sending again could result in a duplicate payment", and the slider locks for the life of the scene. Nothing was accepted, and the actual fix (stake more CPU) is not reachable without backing out of the whole send flow.

Splitting the two decisions keeps the design intact: lock unconditionally, as you do now, but let the existing `errorCasted.name` branches supply the copy when the engine has named a deterministic rejection, and fall through to Transaction Status Unknown for everything else.

For the record on the neighbouring branches: `ErrorAlgoRecipientNotActivated` is unaffected (Algorand throws it from `signTx`, before this line), and the 504 branch is correctly superseded, since a 504 during a broadcast is genuinely ambiguous.

j0ntz (commented):
Reviewed alongside EdgeApp/edge-currency-plugins#458; the findings are inline.

Nit: the commit subject `Lock the send slider once a broadcast has been attempted` is 56 characters, over Edge's 50-character limit. `Lock the send slider after a broadcast attempt` is 46.

src/components/scenes/SendScene2.tsx
```diff
@@ -1507,6 +1527,33 @@ const SendComponent: React.FC<Props> = props => {
...
         let error = err
 
+        if (broadcastAttemptedRef.current) {
+          // No happy path and no generic "network error": the broadcast was
+          // attempted, so tell the user exactly what we know and point them
+          // at the explorer before they consider sending again.
+          logActivity(
+            `Error ${
+              broadcastSucceeded ? 'after' : 'during'
+            } broadcastTx (txid ${edgeTransaction.txid}): ${String(err)}`
```
j0ntz, line 1537: `edgeTransaction.txid` is empty here for UTXO sends, which is the incident this PR exists for. `makeSpend` returns `txid: ''` (UtxoEngine.ts), and the real txid is assigned during `signTx` inside the engine, on the far side of the core bridge, so the scene's copy never receives it. `signedTx` holds it but is `const`-scoped inside the `try`.

The log line the error card sends the user hunting for therefore records no txid at all. Hoisting `let signedTx: EdgeTransaction | undefined` above the `try` and logging `signedTx?.txid ?? edgeTransaction.txid` fixes it.

src/components/scenes/SendScene2.tsx
```diff
@@ -1507,6 +1527,33 @@ const SendComponent: React.FC<Props> = props => {
...
+                  : lstrings.send_broadcast_failure_message_s,
+                errorCasted.message
+              )
+            )
+          )
+          // The locked-state card is longer than a normal error, and the
+          // slider floats over the bottom of the scroll view. Scroll it into
+          // view so the whole message is readable without scrolling by hand.
+          needsScrollToEnd.current = true
+          return
```
j0ntz, line 1554: This path returns without calling `onDone`, so a send launched from a ramp sell or a gift-card purchase ends with no completion callback, and the locked slider means the user cannot reach one by retrying either. The flow is simply stranded on the send scene.

`onDone` already takes an error first (`moonpayRampPlugin` rethrows it, `banxaRampPlugin` and `GiftCardPurchaseScene` also accept it), so `onDone?.(errorCasted)` here would let the plugin flow terminate on its own terms while the scene keeps its honest messaging.

Not a regression against develop, where the generic error card skipped `onDone` too, but locking the slider is what makes it terminal.

j0ntz (changes requested):
Correcting the verdict on my earlier review: those findings are blocking, not advisory.

The EOS path and the stranded `onDone` both change what a user sees after a send that failed, which is the behavior this PR exists to get right.

src/components/scenes/SendScene2.tsx
peachbits: Confirmed against EosEngine.broadcastTx and fixed in 7dc86322d. The lock is unchanged; the message decision now runs after the existing errorCasted.name branches, keyed on a small DETERMINISTIC_REJECTIONS set (the three EOS names plus Algorand). Those keep their specific copy; everything else at or after the boundary falls through to Transaction Status Unknown, including the 504 branch as you noted. One follow-up worth a separate decision: for these named rejections nothing was accepted, so re-arming the slider would be safe too. I have kept the lock for now to match the agreed design.

src/components/scenes/SendScene2.tsx
peachbits: Confirmed, makeSpend returns txid: '' for UTXO. Fixed in 7dc86322d: signedTx is hoisted above the try as EdgeTransaction | undefined and the log line uses signedTx?.txid ?? edgeTransaction.txid.

src/components/scenes/SendScene2.tsx
peachbits: Confirmed against the four callers (moonpay, banxa, fiatPlugin, GiftCardPurchaseScene). Fixed in 7dc86322d: the locked branch now calls onDone(errorCasted) and only logs the callback's own rejection, since the scene already shows the error. The scene itself stays put so the honest message remains visible.

peachbits:
Subject reworded to `Lock the send slider after a broadcast attempt` (46 chars). The three inline findings are addressed in fixup 7dc86322d; details on each thread.

src/components/scenes/SendScene2.tsx
```diff
@@ -1559,10 +1593,63 @@ const SendComponent: React.FC<Props> = props => {
...
+          // the success callback by retrying, so hand them the error now and
+          // let them wind down on their own terms while this scene keeps its
+          // message. The scene already shows the error, so a rejection from
+          // the callback is only logged.
+          if (onDone != null) {
+            const p = onDone(errorCasted)
+            p?.catch((e: unknown) => {
+              console.log(e)
+            })
+          }
```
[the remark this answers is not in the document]
peachbits: Confirmed and fixed in 39644cc54. paybis and moonpay both navigation.pop() the send scene when the onDone promise rejects, then reopen the provider webview with a generic showError, so calling onDone(error) here would have hidden the Transaction Status Unknown card and put the user back in a flow that can start another payment. The locked branch no longer calls onDone. The flows still terminate through the existing onBack-on-unmount path when the user leaves, which is develop's behavior today.

src/components/scenes/SendScene2.tsx
```diff
@@ -1559,10 +1593,63 @@ const SendComponent: React.FC<Props> = props => {
...
+              lstrings.send_broadcast_failure_title,
+              sprintf(
+                broadcastSucceeded
+                  ? lstrings.send_broadcast_post_error_message_s
+                  : lstrings.send_broadcast_failure_message_s,
+                errorCasted.message
+              )
+            )
+          }
+          setError(error)
```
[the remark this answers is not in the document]
peachbits: Confirmed and fixed in 39644cc54. Once a broadcast has been attempted: the quote useAsyncEffect returns early (a fresh quote is meaningless when the transaction may already be on the network, and its success path was the setError(undefined) you found), handleResetSendTransaction no longer clears the error, and handleTimeoutDone is a no-op. That last one also mattered for gift cards, whose onExpired navigates back and would have popped the locked scene on quote expiry.

src/components/scenes/SendScene2.tsx
peachbits: Reversing this in 39644cc54, and re-opening the thread so you see why. Bugbot flagged it after the fixup and it checks out: paybis and moonpay both navigation.pop() the send scene when their onDone promise rejects, then reopen the provider webview with a generic showError. So onDone(error) here dismisses the Transaction Status Unknown card and drops the user back into the sell flow, free to start another payment, which is the MoonPay case from the incident notes. Your underlying point stands, the success callback is unreachable after a lock, but the flows are not stranded: backing out of the scene fires onBack on unmount, the launchers reject with SendErrorBackPressed and wind down quietly, exactly as on develop. A cleaner ending would be for the launchers to receive the real error when the user leaves rather than BackPressed, but that changes each launcher's post-pop behavior and needs per-flow testing, so I have left it as a follow-up rather than fold it in here.

src/components/scenes/SendScene2.tsx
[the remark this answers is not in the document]
peachbits: Addendum from the on-device run: freezing the quote effect left the Network Fee tile on "Calculating Fee" forever after an amount edit on a locked scene, because the amount handler sets processingAmountChanged before the effect runs and the early return never cleared it. Fixed in c6d526c42 by clearing it in the early return. Verified on an iPhone 17 Pro Max simulator: editing the amount on a locked scene now keeps the card, keeps the lock, and leaves the fee tile static. The PR description's Testing table has the full re-run.

j0ntz (approved):

src/components/scenes/SendScene2.tsx
```diff
@@ -1588,6 +1674,16 @@ const SendComponent: React.FC<Props> = props => {
...
+      // Once a broadcast has been attempted the transaction may already be on
+      // the network. Re-quoting would be meaningless, and the success path
+      // below clears `error`, which is the only explanation the user has for
+      // the locked slider. Freeze the quote instead. The amount handler has
+      // already flagged "calculating" by the time we get here, so clear it
+      // or the fee tile spins forever.
+      if (broadcastAttemptedRef.current) {
+        setProcessingAmountChanged(false)
+        return
+      }
```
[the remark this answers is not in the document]
peachbits: Confirmed the mechanism, and we are shipping with it as a known, accepted gap rather than another fixup.

For this to fire, two things have to line up. First, the user has to edit the amount or address after sliding but before broadcastAttemptedRef is set: for a plain UTXO send that window is signTx, tens of milliseconds; only a FIO-address send (checkRecordSendFee is a network call) opens it to seconds. Second, that quote has to finish after the failed broadcast has already written the card. With the two-wave fan-out a failure normally takes seconds to 30 seconds and a UTXO quote finishes in milliseconds, so the usual ordering is quote first, card second, which is harmless. The bad ordering needs a slow network-bound quote plus a fast explicit rejection, on top of the user fidgeting with inputs mid-send.

If it does hit, the slider stays locked with no card. The lock is ref-based and untouched, so there is no path to a repeat payment; only the explanation is lost. Fix, when we take it, is to re-check broadcastAttemptedRef at the existing makeSpendCounter staleness bail-out after the await and in the catch path. Logged as a follow-up.

PR #6199, opened 2026-09-08
Add exchange deep links with promo attribution

j0ntz:
### Technical Design Document

[exchange-deep-links.md](https://github.com/EdgeApp/edge-react-gui/blob/jon/deeplink-new-format/src/docs/exchange-deep-links.md)

### Description

[Asana task](https://app.asana.com/0/0/1210180856778864/f)

📄 [Exchange deep links: pre-selected assets and link-scoped promo attribution](https://github.com/EdgeApp/edge-react-gui/blob/jon/deeplink-new-format/src/docs/exchange-deep-links.md)

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Requirements

- [x] Tested on iOS device

### Test evidence

<table>
<tr>
<td rowspan="4" width="170" valign="top"><code>(earlier)</code><br><sub>Add exchange deep links with promo attribution</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260907-172602-agent-proof-1210180856778864-01-buy-deeplink-btc-wallet-filter.png"></a><br><sub><b>1.</b> buy deeplink btc wallet filter</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260907-172602-agent-proof-1210180856778864-02-swap-deeplink-both-assets.png"></a><br><sub><b>2.</b> swap deeplink both assets</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260907-175609-agent-proof-1210180856778864-03-token-deeplink-usdc.png"></a><br><sub><b>3.</b> token deeplink usdc</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260907-181103-agent-proof-1210180856778864-04-retest-after-fixes.png"></a><br><sub><b>4.</b> retest after fixes</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260916-021336-agent-proof-1210180856778864-01-swap-link-asks-source-wallet-first.png"></a><br><sub><b>1.</b> swap link asks source wallet first</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260916-021336-agent-proof-1210180856778864-02-then-asks-receiving-wallet.png"></a><br><sub><b>2.</b> then asks receiving wallet</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260916-021336-agent-proof-1210180856778864-03-swap-scene-both-sides-set.png"></a><br><sub><b>3.</b> swap scene both sides set</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260916-021336-agent-proof-1210180856778864-04-buy-link-forces-picked-btc-wallet.png"></a><br><sub><b>4.</b> buy link forces picked btc wallet</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260916-021336-agent-proof-1210180856778864-05-dropdown-pick-replaces-forced-wallet.png"></a><br><sub><b>5.</b> dropdown pick replaces forced wallet</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260916-021336-agent-proof-1210180856778864-06-unresolvable-asset-opens-buy-with-no-picker.png"></a><br><sub><b>6.</b> unresolvable asset opens buy with no picker</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260916-031951-agent-proof-1210180856778864-07-swap-link-quote-eth-to-btc.png"></a><br><sub><b>7.</b> swap link quote eth to btc</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260916-031951-agent-proof-1210180856778864-08-promo-swap-executed-success.png"></a><br><sub><b>8.</b> promo swap executed success</sub></td>
</tr>
<tr>
<td rowspan="3" width="170" valign="top"><a href="https://github.com/EdgeApp/edge-react-gui/pull/6199/commits/0fb32a201eb4745267d8651d8363d0781c93ae90"><code>0fb32a2</code></a><br><sub>fixup! Add exchange deep links with promo attribution</sub><br><sub>2026-09-18</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260917-173329-agent-proof-1210180856778864-01-promo-held-buy-tab.png"></a><br><sub><b>1.</b> promo held buy tab</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260917-173329-agent-proof-1210180856778864-02-send2-above-tabs-promo-held.png"></a><br><sub><b>2.</b> send2 above tabs promo held</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260917-173329-agent-proof-1210180856778864-03-tab-switch-releases-promo.png"></a><br><sub><b>3.</b> tab switch releases promo</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260917-173329-agent-proof-1210180856778864-04-swap-preselected.png"></a><br><sub><b>4.</b> swap preselected</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260917-173329-agent-proof-1210180856778864-05-bare-swap-clears-selection.png"></a><br><sub><b>5.</b> bare swap clears selection</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260917-183825-agent-proof-1210180856778864-06-sell-amount-entered-btc.png"></a><br><sub><b>6.</b> sell amount entered btc</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-react-gui/pr-6199/20260917-183825-agent-proof-1210180856778864-07-eth-link-clears-sell-amount.png"></a><br><sub><b>7.</b> eth link clears sell amount</sub></td>
<td width="210"></td>
<td width="210"></td>
</tr>
</table>

---

> [!NOTE]
> **Medium Risk**
> Touches deep-link entry, navigation concurrency, and analytics attribution across buy/sell/swap; mistakes could mis-credit campaigns or drop links, but behavior is heavily tested and scoped away from auth or custody.
> 
> **Overview**
> Adds **`edge://exchange/buy|sell|swap`** (and `https://deep.edge.app/exchange/...`) so campaigns can open buy, sell, or swap with optional **`buyAsset` / `sellAsset`** specs and a session-only **`promoId`**. Parsing normalizes checksummed `0x` token ids and degrades bad asset specs instead of failing the whole link.
> 
> **Deep link handling** now waits for wallets when a picker is needed, resolves assets via **`pickLinkedWallet`**, pre-fills ramp/swap navigation (including **`forcedWalletResult`** on ramp scenes), and returns **`false`** when the user dismisses a picker so callers (e.g. notification center) can keep the promo unread. Concurrency guards drop overlapping **exchange** links while a picker is open but still allow links like WalletConnect.
> 
> **Promo attribution** lives in a new **`linkPromo`** Redux slice (promo id + tab). It is set when a link or promo card opens a flow, overrides **`promoIds`** in **`logEvent`** until **`Buy_Success` / `Sell_Success` / `Exchange_Shift_Success`**, and is released on tab switch via **`useTabBlur`** / **`useLinkPromoRelease`** (without clearing during in-flow blurs like sell **`send2`**). Promo cards and notification CTAs stamp the card id through **`withCardPromoId`** unless the URL already has **`promoId`**.
> 
> Ramp/swap scenes clear link-scoped params and forced wallets when leaving the tab or picking a different wallet; sell amounts reset when the linked asset changes. Extensive unit tests and a design doc were added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8b28c419402c1a91c8cca48e2b1534450e0ad917. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/actions/WalletListActions.tsx
+++ b/src/actions/WalletListActions.tsx
@@ -24,7 +24,7 @@ export function updateWalletsSort(walletsSort: SortOption): ThunkAction<void> {
       type: 'UI/SETTINGS/SET_WALLETS_SORT',
       data: { walletsSort }
     })
-    writeWalletsSort(account, walletsSort).catch(error => {
+    writeWalletsSort(account, walletsSort).catch((error: unknown) => {
       showError(error)
     })
   }
--- a/src/components/scenes/NotificationCenterScene.tsx
+++ b/src/components/scenes/NotificationCenterScene.tsx
@@ -28,7 +28,7 @@ import { EdgeText } from '../themed/EdgeText.tsx'
 
 type Props = EdgeAppSceneProps<'notificationCenter'>
 
-export const NotificationCenterScene = (props: Props) => {
+export const NotificationCenterScene: React.FC<Props> = props => {
   const { navigation } = props
   const theme = useTheme()
   const styles = getStyles(theme)
@@ -204,7 +204,7 @@ export const NotificationCenterScene = (props: Props) => {
             }
             const { name, currencyInfo } = wallets[walletId]
 
-            const handleCloseNewToken = async () => {
+            const handleCloseNewToken = async (): Promise<void> => {
               // Since this isn't a priority notification, we can just fully
               // complete it here
               await completeNotif(key)()
@@ -213,7 +213,7 @@ export const NotificationCenterScene = (props: Props) => {
                 data: { walletId }
               })
             }
-            const handlePressNewToken = async () => {
+            const handlePressNewToken = async (): Promise<void> => {
               await handleCloseNewToken()
               navigation.navigate('manageTokens', {
                 walletId,
@@ -257,7 +257,7 @@ export const NotificationCenterScene = (props: Props) => {
             )
               return null
 
-            const handlePromoPress = async () => {
+            const handlePromoPress = async (): Promise<void> => {
               try {
                 // If it's already marked as expired or if validation fails, just open the URL
                 // The URL could be a download link, a web page, etc.
--- a/src/docs/exchange-deep-links.md
+++ b/src/docs/exchange-deep-links.md
@@ -1,0 +1,520 @@
+# Exchange deep links: pre-selected assets and link-scoped promo attribution
+
+| | |
+|---|---|
+| Status | Implemented |
+| Author | Jon Tzeng |
+| Reviewer | - |
+| Last updated | 2026-09-08 |
+| Repos | [edge-react-gui](https://github.com/EdgeApp/edge-react-gui) |
+| Implementation | [EdgeApp/edge-react-gui#6199](https://github.com/EdgeApp/edge-react-gui/pull/6199) |
+| Supersedes | - |
+| Related | [Asana: Deeplink - New format](https://app.asana.com/0/1215088146871429/1210180856778864) |
+
+<!-- tdd-code-fingerprint: 537152b2b6e63a206c9fee52791270a7fdccfbfa -->
+
+File and symbol references point at the `jon/deeplink-new-format` branch of edge-react-gui, pinned per code block. Direction came from the task description (growth's campaign-tracking request) and the operator's ruling on it.
+
+## Contents
+
+1. [Problem](#1-problem)
+2. [Prior art](#2-prior-art)
+3. [Goals and non-goals](#3-goals-and-non-goals)
+4. [Design overview](#4-design-overview)
+   - [4.1 Link grammar and parsing](#41-link-grammar-and-parsing)
+   - [4.2 Wallet resolution and navigation](#42-wallet-resolution-and-navigation)
+   - [4.3 The link-scoped promo id](#43-the-link-scoped-promo-id)
+   - [4.4 Promo cards as link sources](#44-promo-cards-as-link-sources)
+   - [4.5 Outcomes](#45-outcomes)
+5. [Testing](#5-testing)
+6. [Phase history](#6-phase-history)
+7. [Decisions](#7-decisions)
+8. [Glossary](#8-glossary)
+9. [References](#9-references)
+10. [Post-implementation retrospective](#10-post-implementation-retrospective)
+
+## 1. Problem
+
+Growth runs campaigns that push users toward a specific buy, sell or swap, and wants to measure how much of the resulting volume each campaign produced.
+
+A campaign link could open a flow but not choose an asset. `edge://buy`, `edge://sell` and `edge://swap` navigate to the scene and stop, so a "swap into bitcoin" campaign landed the user on an empty picker and asked them to redo the selection the ad had already made.
+
+Attribution could not follow one entry into a flow either. Conversion events read `params.promoIds` from [`accountReferral`](#accountreferral)'s `activePromotions`, which is durable account state written by `activatePromotion`. An in-app promo card had a [`promoId`](#promoid), but `getDisplayInfoCards` dropped it before the card reached the render path, so a card could not attribute its own conversion; and an external link had no way to name a promo id at all short of the `?af=` affiliate form, which permanently writes the account referral state.
+
+## 2. Prior art
+
+`edge://buy[/<providerId>[/<paymentType>]]` already pins a [ramp](#ramp) provider and payment type for one navigation. The pin lives in navigation params, nothing is written to the account, and a pin that matches no quote degrades to the normal ordering. That link-scoped shape is the model this design follows for both the asset and the promo id.
+
+The `?af=<installerId>` query on the `https://deep.edge.app` form parses as an `affiliate` wrapper, calls `activatePromotion`, and writes the installer id into account referral state. That is durable attribution for the lifetime of the account, and the wrong instrument for "credit this one visit to the scene".
+
+`edge://swap` navigated to `swapCreate` with no params and a code comment reserving query parameters for later. This design is that later.
+
+Extending the three existing hosts was rejected; see [decision 7.1](#71-a-new-exchange-host-rather-than-extending-the-existing-ones).
+
+## 3. Goals and non-goals
+
+Goals:
+
+- Parse `https://deep.edge.app/exchange/<buy|sell|swap>?buyAsset=&sellAsset=&promoId=` and its `edge://exchange/...` equivalent.
+- Pre-select the named asset's wallet, offering wallet creation when the account holds none.
+- Attribute the next conversion to the link's promo id, overriding the account's own promotions for that one entry, then retire it.
+- Let promo cards in the home carousel and the notification center launch these links in-app, carrying the card's own promo id.
+
+Non-goals:
+
+- Currency-code asset specs (`USDC`, `ETH`). The format takes a [pluginId](#pluginid) and a [tokenId](#tokenid). See [follow-up 10.2](#102-where-this-document-was-wrong-or-silent).
+- Persisting a link promo id into [accountReferral](#accountreferral). The id is deliberately session-only and per-entry.
+- Pinning a [ramp](#ramp) provider or payment type from an `exchange` link. `edge://buy/<providerId>/<paymentType>` still owns that.
+- Pre-selecting more than the two sides a swap has.
+
+## 4. Design overview
+
+A link travels through four stages: the parser turns a URL into a typed link, the handler resolves each named asset to a wallet, navigation carries the wallet into the scene, and a redux slice carries the promo id to whichever [conversion event](#conversion-event) fires next.
+
+```mermaid
+sequenceDiagram
+    participant Source as Deep link or promo card
+    participant Parser as DeepLinkParser
+    participant Handler as DeepLinkingActions
+    participant Picker as pickWallet
+    participant Store as Redux store
+    participant Scene as Ramp / swap scene
+    participant Tracker as tracking.logEvent
+
+    Source->>Parser: parseDeepLink(url)
+    Parser-->>Handler: rampCreate | swap { asset(s), promoId }
+    alt link named an asset
+        Handler->>Picker: pickWallet({ assets, showCreateWallet })
+        alt user dismissed the picker
+            Picker-->>Handler: null
+            Handler-->>Source: false (no navigation, promo id untouched)
+        else resolved
+            Picker-->>Handler: { walletId, tokenId }
+        end
+    end
+    Handler->>Store: LINK_PROMO_ID/SET { promoId }
+    Handler->>Scene: navigate with the resolved wallet
+    Scene->>Tracker: logEvent('Buy_Quote' | 'Buy_Success' | ...)
+    Tracker->>Store: read linkPromoId, emit promoIds
+    opt event is a conversion
+        Tracker->>Store: LINK_PROMO_ID/SET { undefined }
+    end
+```
+
+### 4.1 Link grammar and parsing
+
+The `exchange` host takes the direction as its single path segment and everything else as query parameters:
+
+```
+edge://exchange/buy?buyAsset=<spec>[&promoId=<id>]
+edge://exchange/sell?sellAsset=<spec>[&promoId=<id>]
+edge://exchange/swap[?buyAsset=<spec>][&sellAsset=<spec>][&promoId=<id>]
+```
+
+An asset `<spec>` is `<pluginId>[_<tokenId>]`: `bitcoin`, `arbitrum`, `ethereum_0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48`. Both URL forms share one parser, so the `https://deep.edge.app` links growth authors and the `edge://` scheme the app registers behave identically.
+
+`buy` and `sell` produce the existing `rampCreate` link type with two new optional fields; `swap` produces the existing `swap` type with three. Each direction reads only the asset on its own side, because the other side of a [ramp](#ramp) is the user's fiat and has no wallet to select.
+
+[`src/types/DeepLinkTypes.ts`](https://github.com/EdgeApp/edge-react-gui/blob/7f55f8a98d2f3193e0cc92c8d4a28608938a0c4a/src/types/DeepLinkTypes.ts)
+```typescript
+export interface RampCreateLink {
+  type: 'rampCreate'
+  direction: FiatDirection
+  providerId?: string
+  paymentType?: FiatPaymentType
+  asset?: EdgeAsset
+  promoId?: string
+}
+
+export interface SwapLink {
+  type: 'swap'
+  buyAsset?: EdgeAsset
+  sellAsset?: EdgeAsset
+  promoId?: string
+}
+```
+
+Asset resolution splits on the first underscore, and normalizes only the token-id form that needs it:
+
+[`src/util/DeepLinkParser.ts`](https://github.com/EdgeApp/edge-react-gui/blob/7f55f8a98d2f3193e0cc92c8d4a28608938a0c4a/src/util/DeepLinkParser.ts)
+```typescript
+function parseOptionalAsset(spec: string | null): EdgeAsset | undefined {
+  if (spec == null || spec === '') return undefined
+
+  const separatorIndex = spec.indexOf('_')
+  if (separatorIndex < 0) return { pluginId: spec, tokenId: null }
+
+  const pluginId = spec.slice(0, separatorIndex)
+  const rawTokenId = spec.slice(separatorIndex + 1)
+  if (pluginId === '' || rawTokenId === '') {
+    console.warn(`Ignoring malformed deep link asset: ${spec}`)
+    return undefined
+  }
+  return { pluginId, tokenId: normalizeTokenId(rawTokenId) }
+}
+
+function normalizeTokenId(tokenId: string): string {
+  return /^0x[0-9a-fA-F]+$/.test(tokenId)
+    ? tokenId.slice(2).toLowerCase()
+    : tokenId
+}
+```
+
+An unknown direction is rejected by `asFiatDirection`, which throws and leaves the URL unhandled: there is no flow to open, so degrading would land the user somewhere they did not ask for. A malformed asset spec degrades instead, dropping the pre-selection while still opening the flow. Both calls are covered by [decision 7.4](#74-malformed-assets-degrade-unknown-directions-do-not).
+
+### 4.2 Wallet resolution and navigation
+
+`getDeepLinkReadiness` decides how much account state a link needs before it can run. An `exchange` link that names an asset opens the wallet picker, so it waits for `'wallets'`; one that names none only navigates, so `'account'` is enough. Getting this wrong would fire the picker against a half-loaded wallet list right after login.
+
+One helper covers every asset a link can name, on both link types:
+
+[`src/actions/DeepLinkingActions.tsx`](https://github.com/EdgeApp/edge-react-gui/blob/7f55f8a98d2f3193e0cc92c8d4a28608938a0c4a/src/actions/DeepLinkingActions.tsx)
+```typescript
+async function pickLinkedWallet(
+  account: EdgeAccount,
+  navigation: Parameters<typeof pickWallet>[0]['navigation'],
+  asset: EdgeAsset | undefined
+): Promise<WalletListWalletResult | undefined | null> {
+  if (asset == null) return undefined
+
+  const result = await pickWallet({
+    account,
+    assets: [asset],
+    navigation,
+    showCreateWallet: true
+  })
+  return result?.type === 'wallet' ? result : null
+}
+```
+
+The three-valued return is what lets one call site distinguish "the link named nothing" from "the user backed out". `pickWallet` auto-picks when exactly one of the account's wallets matches the asset and shows the modal otherwise, including the zero-match case, where `showCreateWallet` turns the modal into the create-wallet path. This matches how the `rewards` link already resolves assets.
+
+A `null` propagates out of `handleLink` as `false`, which is why `handleLink` and `launchDeepLink` now return a boolean rather than `void`. Nothing navigates and no promo id is stored, so a dismissed picker leaves no trace. The one caller that acts on `false` is the notification center, in [section 4.4](#44-promo-cards-as-link-sources).
+
+For `rampCreate`, the resolved wallet rides into the scene as the `forcedWalletResult` navigation param that `RampCreateScene` already reads, where it takes precedence over the user's last selection:
+
+[`src/components/scenes/RampCreateScene.tsx`](https://github.com/EdgeApp/edge-react-gui/blob/7f55f8a98d2f3193e0cc92c8d4a28608938a0c4a/src/components/scenes/RampCreateScene.tsx)
+```typescript
+const selectedCrypto = forcedWalletResult ?? rampLastCryptoSelection
+```
+
+Because it wins over the last selection, leaving it set would override the user's own wallet choice for the rest of the session. The scene's existing tab-blur listener already cleared the provider and payment-type pins; `forcedWalletResult` joins them, and the listener's early-return guard grows a third condition so a user who never tapped a deep link still pays no params update on tab switches.
+
+For `swap`, both sides are resolved before either is used, and the navigation params are omitted entirely when the link named no asset. Passing `undefined` wallet ids would blank a selection the user had already made on the swap scene, which a bare `edge://swap` used to preserve.
+
+### 4.3 The link-scoped promo id
+
+The id lives in its own root reducer slice rather than in [accountReferral](#accountreferral):
+
+[`src/reducers/RootReducer.ts`](https://github.com/EdgeApp/edge-react-gui/blob/7f55f8a98d2f3193e0cc92c8d4a28608938a0c4a/src/reducers/RootReducer.ts)
+```typescript
+linkPromoId: (state: string | null = null, action: Action): string | null => {
+  switch (action.type) {
+    case 'LINK_PROMO_ID/SET':
+      return action.data.promoId ?? null
+    case 'LOGOUT':
+      return null
+    default:
+      return state
+  }
+}
+```
+
+`LINK_PROMO_ID/SET` carries `string | undefined` and the reducer coalesces to `null`, so one action both sets and clears. Every `exchange` navigation dispatches it, which means a link with no promo id actively clears whatever a previous link left behind. `LOGOUT` clears it too, so an id cannot cross an account switch.
+
+```mermaid
+stateDiagram-v2
+    [*] --> Empty: initial state
+    Empty --> Held: LINK_PROMO_ID/SET with an id<br/>(exchange link or promo card)
+    Held --> Held: quote and other non-conversion events<br/>log promoIds [id]
+    Held --> Empty: Buy_Success / Sell_Success / Exchange_Shift_Success
+    Held --> Empty: exchange link with no promoId
+    Held --> Empty: LOGOUT
+```
+
+Reading it inside `logEvent` is what makes one write cover every conversion path. Seven ramp providers plus `SwapConfirmationScene` call `onLogEvent` directly, and all of them funnel through `logEvent`, so the override and the retirement each need exactly one edit:
+
+[`src/util/tracking.ts`](https://github.com/EdgeApp/edge-react-gui/blob/7f55f8a98d2f3193e0cc92c8d4a28608938a0c4a/src/util/tracking.ts)
+```typescript
+const { linkPromoId } = state
+params.promoIds =
+  linkPromoId == null ? accountReferral.activePromotions : [linkPromoId]
+```
+
+Retirement happens after the event is dispatched to the backends, gated on the event being one of the three that mean the campaign has been credited:
+
+[`src/util/tracking.ts`](https://github.com/EdgeApp/edge-react-gui/blob/7f55f8a98d2f3193e0cc92c8d4a28608938a0c4a/src/util/tracking.ts)
+```typescript
+const CONVERSION_EVENTS: TrackingEventName[] = [
+  'Buy_Success',
+  'Sell_Success',
+  'Exchange_Shift_Success'
+]
+```
+
+Non-conversion events (`Buy_Quote` and the rest) read the id without consuming it, so an abandoned quote does not spend the attribution.
+
+### 4.4 Promo cards as link sources
+
+A card's [`promoId`](#promoid) reaches the launch path through `DisplayInfoCard`. `filterInfoCards` already read the field to gate a card to affiliated accounts, but `getDisplayInfoCards` did not copy it onto the display object, so the render path could not see it. Adding it there makes it available to both card surfaces, and `addPromoCardToNotifications` carries it into the stored `NotifInfo` so the notification-center copy of a card attributes the same campaign as the carousel copy. The cleaner gained `promoId: asMaybe(asString)`, which keeps older stored notifications parsing.
+
+`linkReferralWithCurrencies` takes the card's promo id as a third argument and stamps it onto the parsed link before launch:
+
+[`src/actions/WalletListActions.tsx`](https://github.com/EdgeApp/edge-react-gui/blob/7f55f8a98d2f3193e0cc92c8d4a28608938a0c4a/src/actions/WalletListActions.tsx)
+```typescript
+function withCardPromoId(
+  link: DeepLink,
+  promoId: string | undefined
+): DeepLink {
+  if (promoId == null || promoId === '') return link
+
+  switch (link.type) {
+    case 'rampCreate':
+    case 'swap':
+      return link.promoId == null ? { ...link, promoId } : link
+    case 'affiliate':
+      return { ...link, link: withCardPromoId(link.link, promoId) }
+    case 'marketing':
+      return link.link == null
+        ? link
+        : { ...link, link: withCardPromoId(link.link, promoId) }
+    default:
+      return link
+  }
+}
+```
+
+The `link.promoId == null` guard is the precedence rule from [decision 7.3](#73-a-url-promo-id-beats-the-cards-own): a promo id already in the URL survives. The `affiliate` and `marketing` cases exist because a card whose call-to-action carries `?af=` or a campaign wrapper parses as a wrapper around the link that actually opens the flow, and the stamp has to reach through it.
+
+The notification center previously sent every card call-to-action straight to `openBrowserUri`, which meant an `edge://exchange/...` card in that surface opened a browser. It now routes through `linkReferralWithCurrencies` like the carousel does; a non-Edge URL still ends up at `Linking.openURL` through that function's own `type === 'other'` fallback. The scene also completes the notification only when the launch returned `true`, so backing out of the wallet picker no longer retires the promo unread.
+
+### 4.5 Outcomes
+
+Reachable entry configurations and what each produces. Rows follow from [4.1](#41-link-grammar-and-parsing) through [4.4](#44-promo-cards-as-link-sources); where this table and those sections disagree, the sections are correct.
+
+| Entry | Wallet picker | Lands on | `linkPromoId` after | `promoIds` on the next conversion |
+|---|---|---|---|---|
+| `exchange/buy?buyAsset=bitcoin`, several bitcoin wallets | modal | buy flow, wallet forced | `null` | account promotions |
+| `exchange/buy?buyAsset=bitcoin`, exactly one bitcoin wallet | auto-picked, no modal | buy flow, wallet forced | `null` | account promotions |
+| `exchange/buy?buyAsset=bitcoin`, no bitcoin wallet | modal in create-wallet mode | buy flow after creation | `null` | account promotions |
+| `exchange/buy?buyAsset=bitcoin&promoId=bob` | as above | buy flow, wallet forced | `"bob"` | `["bob"]` |
+| `exchange/buy` (no asset) | none | buy flow, no wallet forced | `null` | account promotions |
+| `exchange/buy?sellAsset=ethereum` | none | buy flow, no wallet forced | `null` | account promotions |
+| `exchange/swap?buyAsset=X&sellAsset=Y` | one per side | swap scene, both sides set | `null` | account promotions |
+| `exchange/swap?sellAsset=Y` | sell side only | swap scene, buy side cleared | `null` | account promotions |
+| `exchange/swap` (no assets) | none | swap scene, selection preserved | `null` | account promotions |
+| Card call-to-action `exchange/buy?buyAsset=X`, card `promoId=card1` | as above | buy flow, wallet forced | `"card1"` | `["card1"]` |
+| Card call-to-action `exchange/buy?...&promoId=url1`, card `promoId=card1` | as above | buy flow, wallet forced | `"url1"` | `["url1"]` |
+| Card call-to-action to a non-Edge URL | none | external browser | unchanged | unchanged |
+| Any of the above, picker dismissed | modal, dismissed | nothing, caller unchanged | unchanged | unchanged |
+| `exchange/swap?sellAsset=ethereum_` (malformed) | none | swap scene, no pre-selection | `null` | account promotions |
+| `exchange/lend?buyAsset=bitcoin` (unknown direction) | none | link unhandled | unchanged | unchanged |
+| Legacy `edge://swap`, `edge://buy/<provider>/<type>` | none | unchanged behavior | unchanged | unchanged |
+
+## 5. Testing
+
+Automated coverage is 128 cases in `src/__tests__/DeepLink.test.ts` (17 of them new for `exchange`) plus 4 in `src/__tests__/reducers/RootReducer.test.ts`.
+
+1. Both URL forms parse to the same link. `https://deep.edge.app/exchange/buy?buyAsset=bitcoin` and `edge://exchange/buy?buyAsset=bitcoin` each produce `rampCreate` with `asset: { pluginId: 'bitcoin', tokenId: null }`.
+2. Mainnet assets on token chains. `exchange/sell?sellAsset=ethereum` and `?sellAsset=arbitrum` give [`tokenId](#tokenid): null` on their own plugin id.
+3. Swap carries two sides. `exchange/swap?buyAsset=bitcoin&sellAsset=ethereum_0xA0b8...` fills both.
+4. Checksummed token ids normalize. `0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48` becomes `a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48`, the form `enabledTokenIds` holds.
+5. Non-hex token ids pass through. A Solana mint keeps its base58 case.
+6. Direction reads its own side only. `exchange/buy?sellAsset=ethereum` yields no asset.
+7. Empty parameters are absent, not empty. `exchange/buy?buyAsset=&promoId=` yields `undefined` for both.
+8. Malformed specs degrade. `sellAsset=_0xA0b8...` and `sellAsset=ethereum_` drop the asset and still return a `swap` link.
+9. Unknown directions reject. `exchange/lend?buyAsset=bitcoin` throws rather than opening a flow.
+10. The affiliate wrapper survives. `https://deep.edge.app/exchange/buy?buyAsset=bitcoin&af=bob` parses as `affiliate` wrapping the `rampCreate`, so `?af=` attribution and asset pre-selection compose.
+11. `linkPromoId` starts `null`, holds a set id, clears on `{ promoId: undefined }`, and clears on `LOGOUT`.
+
+Device verification on the iOS simulator, recorded in the run reports on the task:
+
+12. `edge://exchange/buy?buyAsset=bitcoin` opened the wallet picker filtered to bitcoin.
+13. `edge://exchange/buy?buyAsset=ethereum_0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48` opened the buy flow with the USD Coin token resolved, the wallet auto-selected, and a live quote rendered. Before case 4's normalization the same link opened an empty picker.
+14. `edge://exchange/swap?sellAsset=litecoin&buyAsset=dash&promoId=bob` landed on the swap scene with both sides populated.
+15. The promo-id override was read at runtime by instrumenting `logEvent`: with no link, `linkPromoId: null` and `promoIds: []`; after the [`promoId](#promoid)=bob` link, three `Buy_Quote` events logged `promoIds: ["bob"]`.
+
+Not covered by a device drive: retirement on a [conversion event](#conversion-event) needs a completed funded purchase, so cases 11 and the `CONVERSION_EVENTS` read stand in for it; and the tab-blur release of `forcedWalletResult` has no distinct visual state when the forced wallet and the last-selection wallet are the same.
+
+## 6. Phase history
+
+### Phase 1: plan only
+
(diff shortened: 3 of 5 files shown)


j0ntz:
## 📸 Test evidence

**buy deeplink btc wallet filter**

buy deeplink btc wallet filter

**swap deeplink both assets**

swap deeplink both assets

src/util/DeepLinkParser.ts
```diff
@@ -356,6 +386,33 @@ function parseOptionalPaymentType(
...
+  if (separatorIndex < 0) return { pluginId: spec, tokenId: null }
+
+  const pluginId = spec.slice(0, separatorIndex)
+  const tokenId = spec.slice(separatorIndex + 1)
+  if (pluginId === '' || tokenId === '') {
+    console.warn(`Ignoring malformed deep link asset: ${spec}`)
+    return undefined
+  }
+  return { pluginId, tokenId }
+}
```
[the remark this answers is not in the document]
j0ntz: Fixed. `parseOptionalAsset` now runs the token id through `normalizeTokenId`, which strips the `0x` and lowercases when the id matches `/^0x[0-9a-fA-F]+$/`. Only that prefixed form is rewritten, since Solana mints and Cardano policy ids are case-sensitive base58 and must pass through untouched. Two tests cover both branches: the checksummed USDC address normalizes to the core form, and a base58 mint is unchanged.

src/actions/WalletListActions.tsx
```diff
@@ -139,3 +152,17 @@ const getFirstCurrencyAddress = async (params: {
...
+ */
+function withCardPromoId(
+  link: DeepLink,
+  promoId: string | undefined
+): DeepLink {
+  if (promoId == null || promoId === '') return link
+  if (link.type !== 'rampCreate' && link.type !== 'swap') return link
+  if (link.promoId != null) return link
+  return { ...link, promoId }
+}
```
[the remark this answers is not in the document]
j0ntz: Fixed. `withCardPromoId` is now a switch that recurses through the `affiliate` wrapper (the `?af=` form) and the `marketing` wrapper, stamping the card's promo id on the inner link that actually opens the flow. A promo id already present on that inner link still wins.

src/util/promoCardUtils.ts
```diff
@@ -61,7 +61,8 @@ export const addPromoCardToNotifications = async (
         messageId,
         title,
         body,
-        ctaUrl
+        ctaUrl,
+        promoId
```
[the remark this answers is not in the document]
j0ntz: Fixed. `checkAndAddExpiredPromos` now writes `promoId: card.promoId` alongside `ctaUrl`, so an expired promo tapped in the notification center attributes the same as a dismissed one.

j0ntz:
## 📸 Test evidence (after fix)

**token deeplink usdc**

token deeplink usdc

src/actions/DeepLinkingActions.tsx
```diff
@@ -601,16 +628,29 @@ async function handleLink(
...
         navigation.navigate('buyTab', {
           screen: 'pluginListBuy',
-          params: { providerId, paymentType }
+          params: { providerId, paymentType, forcedWalletResult }
         })
       } else {
         navigation.navigate('sellTab', {
           screen: 'pluginListSell',
-          params: { providerId, paymentType }
+          params: { providerId, paymentType, forcedWalletResult }
```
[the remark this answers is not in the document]
j0ntz: Fixed. The tab-blur cleanup in `RampCreateScene` now clears `forcedWalletResult` alongside `providerId` and `paymentType`, and `pinsRef` tracks it so the early-return guard still skips the params update for users who never tapped a deep link. The forced wallet is link-scoped like the other pins, so leaving it set would have overridden the user's own wallet choice for the rest of the session.

src/components/scenes/NotificationCenterScene.tsx
```diff
@@ -257,11 +258,25 @@ export const NotificationCenterScene = (props: Props) => {
...
+                    // Same flat-navigation cast the other shared consumers in
+                    // this scene need, written against the callee's own
+                    // parameter type so it follows the v7 migration.
+                    navigation as Parameters<
+                      typeof linkReferralWithCurrencies
+                    >[0],
+                    ctaUrl,
+                    promoId
+                  )
+                )
```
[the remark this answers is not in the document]
j0ntz: Fixed. `handleLink` now returns false when the user dismisses a wallet picker, `launchDeepLink` and `linkReferralWithCurrencies` pass that through, and the notification center completes the notice only when the link actually opened. A cancelled picker leaves the promo in place.

j0ntz:
## 📸 Test evidence (after review fixes)

**retest after fixes**

retest after fixes

src/actions/DeepLinkingActions.tsx
```diff
@@ -284,9 +295,26 @@ async function handleLink(
       await doRequestAddress(navigation, state.core.account, dispatch, link)
       break
 
-    case 'swap':
-      navigation.navigate('swapTab', { screen: 'swapCreate' })
```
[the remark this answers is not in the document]
j0ntz: Fixed. Both wrappers now return the inner call: `affiliate` returns `handleLink(...)` directly after activating the promotion, and `marketing` returns it when it carries a link. A dismissed picker behind an `?af=` or campaign wrapper now propagates false to the notification center.

src/actions/DeepLinkingActions.tsx
```diff
@@ -284,9 +295,26 @@ async function handleLink(
...
+
+      navigation.navigate('swapTab', {
+        screen: 'swapCreate',
+        params: {
+          fromWalletId: fromResult?.walletId,
+          fromTokenId: fromResult?.tokenId,
+          toWalletId: toResult?.walletId,
+          toTokenId: toResult?.tokenId
+        }
+      })
```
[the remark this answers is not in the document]
j0ntz: Fixed. The swap handler now passes `params: undefined` when neither side resolved, so a bare `edge://swap` navigates exactly as it did before this branch and leaves an existing selection alone. A one-sided link still sends only the side it named, leaving the other undefined for the scene to keep.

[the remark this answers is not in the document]
j0ntz: Correcting my previous reply: the second sentence was wrong. `SwapCreateScene` derives both wallets straight from `route.params`, so a one-sided link does clear the side it does not name, and react-navigation's nested `navigate` has no merge option to change that. That is the behavior I am keeping: a swap link fully specifies the swap it opens, so `?sellAsset=litecoin` gives LTC to an empty destination rather than inheriting whatever happened to be selected. The fix in this thread stands for the case it was raised about: a link naming no asset at all now passes `params: undefined` and leaves an existing selection untouched.

src/util/tracking.ts
```diff
@@ -262,7 +273,11 @@ export function logEvent(
         const { accountReferral } = account
         params.refDeviceInstallerId = deviceReferral.installerId
         params.refDeviceCurrencyCodes = deviceReferral.currencyCodes
-        params.promoIds = accountReferral.activePromotions
+        // A deep link or promo card that opened this flow attributes its own
+        // conversion, overriding the account's promo ids for this one entry.
+        const { linkPromoId } = state
+        params.promoIds =
+          linkPromoId == null ? accountReferral.activePromotions : [linkPromoId]
```
peachbits: **Critical:** nothing retires `linkPromoId` when the linked flow is abandoned. Open `edge://exchange/buy?buyAsset=bitcoin&promoId=bob`, back out of the quote, and the next conversion later in the same login session (a swap from the swap tab, a buy started from a price alert) is logged with `promoIds: ['bob']`. Only then is the id cleared. Every event logged in between also sends `['bob']` instead of `accountReferral.activePromotions`.

The task scopes the id to one entry into the flow (quoted in decision 7.2), and the `RootReducer` comment says it never leaks into a later conversion. Both only hold when the linked flow converts, so every drop-off inflates the campaign's volume. The id needs a lifetime tied to that entry. One option is to clear it in the tab-blur listener that already drops the link's pins in `RampCreateScene`. The provider webview and bank form are registered inside the buy/sell stacks, so handing off to a provider doesn't blur the tab. The swap tab would need the same listener; it has none today.

Separately, the retirement at L425-427 uses the `linkPromoId` read at L278, before `await checkNotifications()`, and clears the slice unconditionally. If a new link sets its id during that await, the id is wiped. Checking that `getState().linkPromoId` still matches before dispatching avoids that.

src/util/promoCardUtils.ts
peachbits, line 65: **Warning:** the bugbot thread about expired promos is resolved as fixed, but `checkAndAddExpiredPromos` (L180-186) still writes `promoCard: { messageId, title: ctaTitle, body: title, ctaUrl }` with no `promoId`. The fix isn't in this head, and it isn't in the earlier commits bugbot reviewed (547d61b, 7e80b6f, 429fef2). An expired promo tapped in the notification center still won't be attributed to its card.

src/components/scenes/RampCreateScene.tsx
```diff
@@ -608,21 +608,35 @@ export const RampCreateScene: React.FC<Props> = (props: Props) => {
...
+        return
+      // The forced wallet is link-scoped like the pins: leaving it set would
+      // keep overriding the user's own wallet choice for the rest of the
+      // session, since it always wins over the last-selection setting.
       navigation.setParams({
         providerId: undefined,
-        paymentType: undefined
+        paymentType: undefined,
+        forcedWalletResult: undefined
       })
```
peachbits: **Warning:** this handles leaving the tab. While the user stays on the tab, though, the forced wallet still can't be changed. `handleCryptDropdown` (L694-728) only calls `setRampCryptoSelection`, and `selectedCrypto = forcedWalletResult ?? rampLastCryptoSelection` keeps showing the forced wallet. So after `exchange/buy?buyAsset=bitcoin`, picking another wallet or asset from the dropdown does nothing. The early return there also compares against `rampLastCryptoSelection` rather than the wallet on screen.

This predates the PR for the wallet-details Buy/Sell entries, but this link makes it a main path. Clearing `forcedWalletResult` with `setParams` when the user picks a wallet would fix both.

Also, `TransactionListTop`, `BuyCrypto` and `CoinRankingDetailsScene` pass `forcedWalletResult` too. So it isn't only link-scoped, and this blur handler now resets their pre-selection as well.

src/actions/DeepLinkingActions.tsx
```diff
@@ -145,24 +155,25 @@ export function getDeepLinkReadiness(link: DeepLink): DeepLinkReadiness {
...
 }
 
 /**
- * Launches a link if it app is able to do so.
- * @returns true if the link is handled,
- * or false if the app is in the wrong state to handle this link.
+ * Follow a link.
+ * @returns false when the user backed out of a prompt the link raised (a
+ * wallet picker they dismissed), so callers can leave whatever surface sent
+ * them here untouched. True otherwise.
```
peachbits, line 304: **Warning:** only the `exchange` paths and the PIX warning actually return false. `azteco`, `paymentRedirect`, `rewards` and `other` still `break` after a dismissed `pickWallet` and return true. The notification center now sends every CTA the parser recognizes (anything but `other`) through here and completes the promo whenever this returns true. So a promo whose CTA is a `rewards` link (`https://edge.app/rewards/?data={{REWARDS:...}}`), an `azteco` link or a `paymentRedirect` link disappears when its picker is cancelled. That's the same case bugbot raised for exchange links. (`other` only reaches here behind an `?af=` wrapper, since `linkReferralWithCurrencies` opens bare `other` links itself.) Returning false from those branches too would fix it.

src/actions/DeepLinkingActions.tsx
```diff
@@ -277,16 +288,38 @@ async function handleLink(
...
 
-    case 'swap':
-      navigation.navigate('swapTab', { screen: 'swapCreate' })
+    case 'swap': {
+      const { buyAsset, sellAsset, promoId } = link
+
+      const toResult = await pickLinkedWallet(account, navigation, buyAsset)
+      if (toResult === null) return false
+      const fromResult = await pickLinkedWallet(account, navigation, sellAsset)
+      if (fromResult === null) return false
```
peachbits: **Warning:** when both sides need a picker, the user gets two identical "Select Wallet" modals in a row, receive side first, with nothing saying which is which. `pickWallet` accepts `headerTitle`, and `select_src_wallet` / `select_recv_wallet` already exist. `pickLinkedWallet` could take a title, and this could ask for the sell side first.

src/actions/DeepLinkingActions.tsx
```diff
@@ -654,4 +700,29 @@ async function handleLink(
...
+): Promise<WalletListWalletResult | undefined | null> {
+  if (asset == null) return undefined
+
+  const result = await pickWallet({
+    account,
+    assets: [asset],
+    navigation,
+    showCreateWallet: true
+  })
+  return result?.type === 'wallet' ? result : null
```
peachbits, line 898: **Warning:** when the account can't resolve the asset, this opens an empty picker. That covers a typo'd plugin id, a contract that isn't in `allTokens`, a chain this build doesn't include, or a Sui coin type pasted as `0x…::usdc::USDC` (Edge's Sui token ids run the address, module and name together). Dismissing the picker returns false, so the flow never opens and the partner link dead-ends instead of degrading the way decision 7.4 intends. Returning `undefined` when `account.currencyConfig[asset.pluginId]` or its `allTokens[asset.tokenId]` is missing would open the flow with nothing pre-selected.

src/util/DeepLinkParser.ts
```diff
@@ -356,6 +386,49 @@ function parseOptionalPaymentType(
...
 }
 
+/**
+ * Resolve an `exchange` link's asset from its `<pluginId>[_<tokenId>]` spec:
+ *
+ *   bitcoin
+ *   ethereum_0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48
+ *
+ * The split is on the FIRST underscore so a plugin id containing one still
+ * parses; no underscore means the chain's mainnet asset. Like the payment-type
```
peachbits: **Nit:** this reason is backwards. Splitting on the first underscore keeps any later underscores in the token id; a plugin id containing one would be cut short. No current plugin id has an underscore, but token ids can: Edge's Sui token ids are the coin type's address, module and name run together, and Move module names allow underscores. The PR description repeats the same reasoning.

src/docs/exchange-deep-links.md
```diff
@@ -0,0 +1,520 @@
...
+edge://exchange/buy?buyAsset=<spec>[&promoId=<id>]
+edge://exchange/sell?sellAsset=<spec>[&promoId=<id>]
+edge://exchange/swap[?buyAsset=<spec>][&sellAsset=<spec>][&promoId=<id>]
+```
+
+An asset `<spec>` is `<pluginId>[_<tokenId>]`: `bitcoin`, `arbitrum`, `ethereum_0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48`. Both URL forms share one parser, so the `https://deep.edge.app` links growth authors and the `edge://` scheme the app registers behave identically.
+
+`buy` and `sell` produce the existing `rampCreate` link type with two new optional fields; `swap` produces the existing `swap` type with three. Each direction reads only the asset on its own side, because the other side of a [ramp](#ramp) is the user's fiat and has no wallet to select.
+
+[`src/types/DeepLinkTypes.ts`](https://github.com/EdgeApp/edge-react-gui/blob/7f55f8a98d2f3193e0cc92c8d4a28608938a0c4a/src/types/DeepLinkTypes.ts)
```
peachbits: **Suggestion:** all eight code links (L118-263) are pinned to `7f55f8a98d2f`, which doesn't exist on GitHub (`No commit found for SHA`), so they all 404. Re-pin them to a pushed commit, or link to `develop` the way the glossary does. Also, the doc landed in 6cf8033 ("Fix lint warnings in deeplink promo files") rather than in the feature commit it documents.

src/actions/WalletListActions.tsx
```diff
@@ -139,3 +158,32 @@ const getFirstCurrencyAddress = async (params: {
...
+ * already carry one of its own. Every other link type is returned untouched.
+ *
+ * A card whose call-to-action carries `?af=` parses as an `affiliate` wrapper,
+ * and a campaign link as a `marketing` wrapper, so both are followed to the
+ * link that actually opens the flow.
+ */
+function withCardPromoId(
+  link: DeepLink,
+  promoId: string | undefined
+): DeepLink {
```
peachbits: **Warning:** bugbot already found one bug in this helper (the `?af=` unwrap), and it has no test. It's pure, so exporting it and covering URL-over-card precedence plus the `affiliate`/`marketing` unwrap would be cheap. The `logEvent` side (the override, and retirement on `CONVERSION_EVENTS`) has no test either; the reducer tests only cover the slice.

peachbits (changes requested):
Blocking: the link promo id outlives an abandoned flow (`tracking.ts`). The comments marked Warning should also be fixed before merge. The rest are suggestions.

src/util/tracking.ts
j0ntz: Fixed. The slice is now `linkPromo: { promoId, tab }`, and `useLinkPromoRelease` on the buy, sell and swap scenes releases it when its own tab blurs. The release compares the live slice's `tab`, because a link dispatches before it navigates and the tab it leaves blurs with the new promo already set. The conversion retirement now re-reads `getState().linkPromo` and clears only its own promo. On the sim, leaving the Buy tab after `promoId=bob` cleared it and the next `Buy_Quote` logged `promoIds: []`; `trackingLinkPromo.test.ts` covers the await race.

src/util/promoCardUtils.ts
j0ntz: Fixed: `checkAndAddExpiredPromos` now writes `promoId: card.promoId` onto the stored notification.

src/components/scenes/RampCreateScene.tsx
j0ntz: Fixed: `handleCryptDropdown` clears `forcedWalletResult` on a pick and its early return compares against `selectedCrypto`. On the sim, after a link forced My Bitcoin, picking My Ether 4 switched the scene (evidence frame 05). The blur reset does now also end the pre-selection from `TransactionListTop`, `BuyCrypto` and `CoinRankingDetailsScene`; each of those is a one-visit pre-selection too, so leaving the tab ending it matches.

src/actions/DeepLinkingActions.tsx
j0ntz: Fixed: `azteco`, `paymentRedirect`, `rewards` and `other` now return false on a dismissed picker, so the notification center keeps those promo cards too.

src/actions/DeepLinkingActions.tsx
j0ntz: Fixed: `pickLinkedWallet` takes a title, and a swap link resolves the sell side first with `select_src_wallet`, then the buy side with `select_recv_wallet`. Verified on the sim (evidence frames 01 and 02).

src/actions/DeepLinkingActions.tsx
j0ntz: Fixed: `pickLinkedWallet` returns undefined when `currencyConfig[asset.pluginId]` or its `allTokens[asset.tokenId]` is missing, so the flow opens with nothing pre-selected. On the sim, `edge://exchange/buy?buyAsset=notachain` opened the Buy scene with no picker (evidence frame 06).

src/util/DeepLinkParser.ts
j0ntz: Fixed: the comment now gives the actual reason, that later underscores stay in the token id, which Sui coin types need. The PR description no longer restates the design; it links the design doc, which carries the corrected wording.

src/docs/exchange-deep-links.md
j0ntz: Fixed: the code captions now cite the branch instead of a sha. The doc is a fixup of the first commit, so every push re-slots it ahead of the code it quotes and rewrites that code's commits, which is how the `7f55f8a98` pins were orphaned; a branch ref resolves for the life of the review. The doc stays in the first commit by our design-doc convention, and it can move into the feature commit at the final squash.

src/actions/WalletListActions.tsx
j0ntz: Added. `withCardPromoId` is exported and `LinkPromoActions.test.ts` covers URL-over-card precedence, the `affiliate` and `marketing` unwrap and the no-op cases, plus `releaseLinkPromo`. `trackingLinkPromo.test.ts` covers `logEvent`: the override, the account fallback, no retirement on a quote, retirement on `Buy_Success`, and a newer promo surviving the awaited notification check (that case fails when the retirement clears unconditionally).

j0ntz:
Addressed. The blocking item is fixed: the link promo now carries its tab and is released when the user leaves that tab without converting, verified on the simulator. Every Warning is fixed and replied inline, and the doc links and parser comment are corrected. The re-test frames are in the evidence table.

src/util/tracking.ts
```diff
@@ -262,7 +273,13 @@ export function logEvent(
...
         params.refDeviceInstallerId = deviceReferral.installerId
         params.refDeviceCurrencyCodes = deviceReferral.currencyCodes
-        params.promoIds = accountReferral.activePromotions
+        // A deep link or promo card that opened this flow attributes its own
+        // conversion, overriding the account's promo ids for this one entry.
+        const { linkPromo } = state
+        params.promoIds =
+          linkPromo == null
+            ? accountReferral.activePromotions
+            : [linkPromo.promoId]
```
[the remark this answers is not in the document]
j0ntz: Fixed: `logEvent` now reads `linkPromo` synchronously when the event is dispatched, before its awaits, and the retirement compares that snapshot against the live slice. `SwapConfirmationScene` logs `Exchange_Shift_Success` before `updateSwapCount`, which can wait on a review prompt. `trackingLinkPromo.test.ts` covers a conversion whose tab is left before its params are built, and that case fails when the promo is read late.

src/__tests__/util/trackingLinkPromo.test.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+  const dispatch = (action: Action): Action => {
+    dispatched.push(action)
+    return action
+  }
+  return {
+    dispatched,
+    state,
+    run: event => {
+      logEvent(event)(dispatch as any, (() => state) as any, undefined as any)
```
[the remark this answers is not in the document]
j0ntz: Fixed: the harness now calls the thunk with `(dispatch, getState)` only.

src/util/tracking.ts
```diff
@@ -403,6 +425,19 @@ export function logEvent(
...
+        // keeps a later, unrelated buy or swap on the account's own promo ids.
+        //
+        // Re-read the slice instead of trusting the value captured at dispatch:
+        // the event was built across several awaits, and a link that arrived
+        // meanwhile owns the attribution now. Clearing unconditionally would
+        // throw that newer promo away.
+        if (linkPromo != null && CONVERSION_EVENTS.includes(event)) {
+          if (getState().linkPromo?.promoId === linkPromo.promoId) {
+            dispatch({ type: 'LINK_PROMO/SET', data: { linkPromo: null } })
+          }
```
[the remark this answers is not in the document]
j0ntz: Fixed: the retirement now clears only when `getState().linkPromo === linkPromo`. Every link stores a fresh `LinkPromo`, so a newer link for the same campaign and tab is a different object and keeps its credit. `trackingLinkPromo.test.ts` covers it, and that case fails under a `promoId` comparison.

j0ntz:
Funded conversion through an exchange link, on the simulator at head `18ee679ff` (code unchanged since): `edge://exchange/swap?sellAsset=ethereum&buyAsset=bitcoin&promoId=bob` swapped 0.0049995 ETH ($11.99) for 0.00014972 BTC through Rango and reached the success scene. `Exchange_Shift_Quote`, `Exchange_Shift_Start` and `Exchange_Shift_Success` each logged `promoIds: ["bob"]`.

`swapSuccess` is registered in the root stack, outside the swap tab, so pushing it blurs the tab and released the promo in the same moment `Exchange_Shift_Success` was dispatched. That is the case the dispatch-time read covers; with the earlier late read this conversion would have logged the account promotions instead.

These two frames are not in the evidence table above: the table's start marker was lost from the body, so the attach step could not add a row.

| Quote | Success |
|---|---|
|  |  |

src/actions/WalletListActions.tsx
```diff
@@ -53,14 +67,19 @@ export function linkReferralWithCurrencies(
...
     const parsed = parseDeepLink(uri)
-    if (parsed.type === 'other') await Linking.openURL(uri)
-    else await dispatch(launchDeepLink(navigation, parsed))
+    if (parsed.type === 'other') {
+      await Linking.openURL(uri)
+      return true
+    }
+    return await dispatch(
+      launchDeepLink(navigation, withCardPromoId(parsed, promoId))
+    )
```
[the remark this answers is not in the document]
j0ntz: Fixed in `launchDeepLink` itself: a link whose readiness is `wallets` now waits, through `waitForWallets`, for every active wallet to load or fail before it is handled. Promo card taps and any other direct caller get the same guarantee as the manager queue, and the manager's own calls return at once. `LinkPromoActions.test.ts` covers the wait, and a cold-start `edge://exchange/buy?buyAsset=bitcoin` still raised the picker on the sim.

src/actions/DeepLinkingActions.tsx
```diff
@@ -145,24 +155,62 @@ export function getDeepLinkReadiness(link: DeepLink): DeepLinkReadiness {
...
+    // DeepLinkingManager only launches a link once the app has the state it
+    // needs, but promo card taps call in directly. A link that opens a wallet
+    // picker must see the full wallet list, so wait for it here too; for the
+    // manager's own calls this returns at once.
+    if (getDeepLinkReadiness(link) === 'wallets') {
+      await waitForWallets(getState().core.account)
+    }
     const state = getState()
-    await handleLink(navigation, dispatch, state, link)
+    return await handleLink(navigation, dispatch, state, link)
```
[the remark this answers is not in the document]
j0ntz: Fixed: links that arrive during a wallet wait share one wait, and only the latest of them is followed; the earlier ones return false without opening a picker or setting a promo. `LinkPromoActions.test.ts` covers two links started during one wait, and the case fails without the latest-link check.

src/actions/DeepLinkingActions.tsx
```diff
@@ -145,24 +155,62 @@ export function getDeepLinkReadiness(link: DeepLink): DeepLinkReadiness {
...
+    }
+    cleanups.push(
+      account.watch('activeWalletIds', check),
+      account.watch('currencyWallets', check),
+      account.watch('currencyWalletErrors', check)
+    )
+    // A wallet may have finished between the first check and subscribing:
+    check()
+  })
+}
```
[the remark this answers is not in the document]
j0ntz: Fixed for logout: the wait also watches `loggedIn`, resolves false and unsubscribes when the account logs out, and concurrent waits share one set of watchers, so repeated taps cannot accumulate them. A wallet that neither loads nor errors keeps that single wait until logout, the same hold `DeepLinkingManager` already places on its own queue; a timeout would instead open the picker against the partial list this wait exists to avoid.

src/actions/DeepLinkingActions.tsx
```diff
@@ -601,16 +721,33 @@ async function handleLink(
...
         navigation.navigate('buyTab', {
           screen: 'pluginListBuy',
-          params: { providerId, paymentType }
+          params: { providerId, paymentType, forcedWalletResult }
         })
       } else {
         navigation.navigate('sellTab', {
           screen: 'pluginListSell',
-          params: { providerId, paymentType }
+          params: { providerId, paymentType, forcedWalletResult }
```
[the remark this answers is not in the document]
j0ntz: Leaving this as is. On develop, `edge://buy` and `edge://sell` already navigate with `params: { providerId, paymentType }`, and that navigate replaces the route params, so a legacy link has always dropped a Coin Ranking or earlier forced wallet; a bare `exchange/buy` matches it. Nothing the user picked is lost: without `forcedWalletResult` the scene shows `rampLastCryptoSelection`. The swap path differs because its params are the live selection itself. Keeping a previous forced wallet would need merge semantics, which would also keep a previous link's provider and payment-type pins.

src/actions/DeepLinkingActions.tsx
```diff
@@ -145,24 +155,94 @@ export function getDeepLinkReadiness(link: DeepLink): DeepLinkReadiness {
...
+      // Only the latest link to start waiting is followed, like the manager's
+      // single pending slot. Repeated taps during the wait would otherwise
+      // stack pickers and let a later card's promo overwrite an earlier one.
+      const thisLink = ++latestWalletLink
+      const loggedIn = await waitForWallets(getState().core.account)
+      if (!loggedIn || thisLink !== latestWalletLink) return false
+    }
     const state = getState()
-    await handleLink(navigation, dispatch, state, link)
+    return await handleLink(navigation, dispatch, state, link)
```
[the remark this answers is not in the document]
j0ntz: Fixed: once a `wallets` link is being followed, `launchDeepLink` drops any newer one until it finishes, so a second tap after wallets have loaded cannot stack a picker or overwrite the first link's promo and navigation. During the wallet wait, when nothing is on screen yet, the latest link still wins. `LinkPromoActions.test.ts` covers a second link arriving while the first picker is open, and that case fails without the guard.

src/components/scenes/SwapCreateScene.tsx
```diff
@@ -88,6 +89,10 @@ export const SwapCreateScene: React.FC<Props> = props => {
   const theme = useTheme()
   const dispatch = useDispatch()
 
+  // An `edge://exchange/swap` link credits its promo to this one entry into
+  // the swap flow, so leaving the tab without swapping gives the credit back.
+  useLinkPromoRelease(navigation, 'swapTab')
+
```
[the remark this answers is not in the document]
j0ntz: Leaving this as is. The swap scene already behaves this way for every caller that navigates into it with wallet ids: on develop, `CoinRankingDetailsScene`, `TransactionListTop`, `InsufficientFeesModal` and `ScanActions` all set `fromWalletId`/`toWalletId` on the mounted scene without touching the entered amount. The amount stays on screen, and nothing quotes until the user taps Next, so the new pair is never quoted silently. Resetting amounts on a param change would be a swap scene change for all of those entries, outside this PR.

src/actions/DeepLinkingActions.tsx
```diff
@@ -145,24 +155,110 @@ export function getDeepLinkReadiness(link: DeepLink): DeepLinkReadiness {
...
+}
+
+/** Counts links that waited for wallets; the latest one is the one followed. */
+let latestWalletLink = 0
+
+/** Whether a link that opens a wallet picker is being followed right now. */
+let walletLinkInProgress = false
+
+/** The wallet wait in progress, shared by every link that arrives during it. */
+let walletWait: { account: EdgeAccount; loaded: Promise<boolean> } | undefined
```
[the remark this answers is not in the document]
j0ntz: Fixed the one that could stick: the in-progress guard is now keyed to the account (`walletLinkAccount`), so a picker a logout leaves unsettled cannot drop the next account's links; `LinkPromoActions.test.ts` covers it and fails with an account-agnostic flag. The other two were already safe across logout: `walletWait` is keyed by account and resolves false on logout, and `latestWalletLink` only ever increases, so a new session's links always compare as newest.

src/hooks/useLinkPromoRelease.ts
```diff
@@ -0,0 +1,45 @@
...
+  tab: LinkPromoTab
+): void {
+  const dispatch = useDispatch()
+
+  React.useEffect(() => {
+    const tabNavigation = navigation.getParent()
+    if (tabNavigation == null) return
+    return tabNavigation.addListener('blur', () => {
+      dispatch(releaseLinkPromo(tab))
+    })
```
peachbits: **Critical:** this releases the promo in the middle of a sell. The Banxa, MoonPay and Paybis sell flows push `send2` for the deposit, and `send2` is an `AppStack` screen above the tabs (`Main.tsx:1091`). react-navigation re-emits a parent's blur to its focused child (`useFocusEvents` in core 6.4.17), so opening the send scene fires this listener and clears the promo. `Sell_Success` is only logged after the send completes (`banxaRampPlugin.ts:1503`, `moonpayRampPlugin.ts:1119`, `paybisRampPlugin.ts:1391`), so a sell opened from a campaign link is never attributed. The comment above (L28-30) holds for the webview and bank form but not for the send step; my earlier suggestion missed it too.

One fix: skip the release while this tab is still the tab navigator's active route, so only switching tabs counts as leaving, e.g. `const { routes, index } = tabNavigation.getState(); if (routes[index].name === tab) return`.

src/actions/DeepLinkingActions.tsx
```diff
@@ -145,24 +155,116 @@ export function getDeepLinkReadiness(link: DeepLink): DeepLinkReadiness {
...
+    // needs, but promo card taps call in directly. A link that opens a wallet
+    // picker must see the full wallet list, so wait for it here too; for the
+    // manager's own calls this returns at once.
+    if (getDeepLinkReadiness(link) === 'wallets') {
+      const { account } = getState().core
+
+      // Once a link is being followed its picker may already be on screen, so
+      // a newer one is dropped rather than stacking a second picker and then
+      // overwriting the first link's promo and navigation.
+      if (walletLinkAccount === account) return false
```
peachbits: **Warning:** this drop covers every `wallets`-readiness link, not just the exchange pickers it was added for. While any of them is mid-flow (a wallet picker, a `requestAddress` or `modal` prompt), every other one returns false here. That's the intent for card taps, but `DeepLinkingManager` clears its pending slot before launching and ignores the result (`DeepLinkingManager.tsx:77-78`). So a WalletConnect or payment link opened from another app while a promo's wallet picker is up is silently lost; before this PR it was handled. Scoping the drop to `rampCreate`/`swap` links, or holding the newer link until the current one finishes, would keep them.

src/actions/DeepLinkingActions.tsx
```diff
@@ -277,16 +379,56 @@ async function handleLink(
...
+      })
+
+      // Navigate with no params at all when the link named no asset: passing
+      // undefined wallet ids would blank a selection the user already made on
+      // the swap scene, which is what a bare `edge://swap` used to preserve.
+      navigation.navigate('swapTab', {
+        screen: 'swapCreate',
+        params:
+          fromResult == null && toResult == null
+            ? undefined
```
peachbits: **Suggestion:** this branch doesn't preserve anything. A nested `navigate` rebuilds the child action from `name`, `params` and `path` only, with no `merge` (`useNavigationBuilder.tsx:501-506`, core 6.4.17). `StackRouter` then replaces an existing route's params with the action's, unless the screen declares `initialParams` (`StackRouter.tsx:441-449`, routers 6.1.9), and `swapCreate` declares none (`Main.tsx:616-622`). So the child's params become `undefined` and the wallet selection is cleared either way, exactly as `navigate('swapTab', { screen: 'swapCreate' })` already does on develop. The special case is a no-op, and the comment above, decision 7.6 and the outcomes row `selection preserved` (`exchange-deep-links.md:350`) describe behavior the code does not have. Either drop the branch and correct the doc, or navigate to the tab with no nested payload (`navigation.navigate('swapTab')`), which does leave the swap stack untouched, at the cost of landing wherever that tab was.

peachbits (changes requested):
Re-review at 3c02e5d: all 9 earlier comments are addressed in the code, with tests. Two new issues in the fixes: the tab-blur release clears the promo before a sell's send step finishes (blocking), and the new link guard drops unrelated links that arrive from outside the app. One smaller note on the swap navigation.

src/hooks/useLinkPromoRelease.ts
j0ntz: Fixed in 20fdb0738. The release runs through a new shared hook, `src/hooks/useTabBlur.ts`, which reads the tab navigator's own state inside the blur listener and returns while that navigator is still on the flow's tab:

```ts
return tabNavigation.addListener('blur', () => {
  const { index, routes } = tabNavigation.getState()
  if (routes[index].name === tab) return

  handleLeave()
})
```

Only a real tab switch ends the entry now, so an `AppStack` scene opening above the tabs (`send2`, `swapSuccess`) leaves the promo held. `RampCreateScene` had the identical defect in its own blur listener, which cleared the pinned `providerId`, `paymentType` and `forcedWalletResult` on the same event; it delegates to the same hook.

Driven on the simulator with the three dispatch sites logged through the app's file logger. `edge://exchange/buy?buyAsset=bitcoin&promoId=agent-proof` set the promo, a `edge://redirect/payment/...` link then pushed `send2` above the tabs (the same `AppStack` push the sell deposit performs) and the promo survived it, and switching to the Home tab released it:

```
[LINKPROMO] slice set to {"promoId":"agent-proof","tab":"buyTab"}
[LINKPROMO] blur tab=buyTab activeRoute=buyTab skip=true
[LINKPROMO] blur tab=buyTab activeRoute=buyTab skip=true
[LINKPROMO] blur tab=buyTab activeRoute=home skip=false
[LINKPROMO] release called tab=buyTab held={"promoId":"agent-proof","tab":"buyTab"}
[LINKPROMO] slice set to null
```

The `skip=true` pair is the send2 push: two lines because `RampCreateScene` subscribes twice, once for the promo and once for its own params. Unit coverage is in `src/__tests__/hooks/useTabBlur.test.tsx`.

src/actions/DeepLinkingActions.tsx
j0ntz: Scoped in 20fdb0738. The drop now applies only to the two link types that open the promo pickers:

```ts
const isExchangeLink = link.type === 'rampCreate' || link.type === 'swap'
```

Every other `wallets`-readiness link (`walletConnect`, `paymentRedirect`, `requestAddress`, `plugin`, `fiatPlugin`) is followed while a picker is open, the way it was before this PR. Covered by `follows a walletConnect link that arrives while a picker is open` in `src/__tests__/actions/LinkPromoActions.test.ts`.

The simulator drive above exercises the same path: the `edge://redirect/payment/...` link arrived while the buy flow's promo was held and routed through to the Send scene instead of being dropped.

src/actions/DeepLinkingActions.tsx
j0ntz: Removed in 20fdb0738. `case 'swap'` takes the single `navigate('swapTab', { screen: 'swapCreate', params })` path, with no branch on `params === undefined`.

The doc is corrected in 4ceb64f6c: the outcomes row in `exchange-deep-links.md` now reads `swap scene, selection cleared`, decision 7.6 and section 4.2 describe the clearing, and 7.10 / 7.11 record the two scoping fixes above.

Confirmed on the simulator: `edge://exchange/swap?sellAsset=bitcoin&buyAsset=ethereum` pre-selected My Bitcoin and My Ether 4, and a bare `edge://swap` afterwards left both pickers empty.

j0ntz:
Both new issues and the smaller note are fixed and replied to in their threads, at 20fdb0738 (code) and 4ceb64f6c (doc).

- Tab-blur release during a sell: a new `src/hooks/useTabBlur.ts` reads the tab navigator's own state and returns while it is still on the flow's tab, so `send2` and `swapSuccess` opening above the tabs no longer end the entry. `RampCreateScene` had the same defect on its pinned provider, payment type and forced wallet, and uses the hook now.
- Link guard dropping unrelated links: the drop is scoped to `rampCreate` and `swap`, so `walletConnect`, `paymentRedirect`, `requestAddress`, `plugin` and `fiatPlugin` links are followed while a picker is open, as before.
- Swap navigation note: the no-op `params === undefined` branch is gone, and the doc's outcomes row, decision 7.6 and section 4.2 now say the selection is cleared.

Verified on the iOS simulator with the promo dispatch sites logged: a promo-carrying buy link held the promo across a `send2` push and released it only on a tab switch, and a bare `edge://swap` cleared a prior pre-selection. Unit coverage is in `src/__tests__/hooks/useTabBlur.test.tsx` and `src/__tests__/actions/LinkPromoActions.test.ts`.

src/actions/DeepLinkingActions.tsx
```diff
@@ -145,24 +155,125 @@ export function getDeepLinkReadiness(link: DeepLink): DeepLinkReadiness {
...
+      // and then overwriting the first link's promo and navigation.
+      if (isExchangeLink && walletLinkAccount === account) return false
+
+      // While wallets load nothing is on screen yet, so the latest link to
+      // start waiting is the one followed, like the manager's single pending
+      // slot. Repeated taps during the wait would otherwise stack pickers.
+      const thisLink = ++latestWalletLink
+      const loggedIn = await waitForWallets(account)
+      if (!loggedIn || thisLink !== latestWalletLink) return false
+      if (isExchangeLink && walletLinkAccount === account) return false
```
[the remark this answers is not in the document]
j0ntz: Valid, both halves. Fixed in 68bd1216c.

`isExchangeLink` now reads the type through a new `unwrapDeepLink` helper, so an `exchange` link an `affiliate` or `marketing` wrapper carries takes the drop: the wrapper's `handleLink` case hands its inner link straight back to the same handler, so the type that matters is the inner one, exactly as `getDeepLinkReadiness` already unwraps for readiness.

The guard also moved above the readiness branch and now runs on the account-readiness path too, so an `exchange` link that names no asset is dropped while a picker is up. It raises no picker of its own, but it does dispatch `LINK_PROMO/SET` and navigate with fresh ramp params, which is the overwrite the guard exists to prevent. It does not claim the slot itself, since there is nothing to hold.

Two regression tests in `src/__tests__/actions/LinkPromoActions.test.ts` cover the pair, and both fail against the previous commit.

src/components/scenes/SwapConfirmationScene.tsx
```diff
@@ -343,6 +344,8 @@ export const SwapConfirmationScene: React.FC<Props> = (props: Props) => {
             }
           })
         )
+
+        await dispatch(updateSwapCount())
```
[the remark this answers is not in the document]
j0ntz: Fixed in 7dac6d4d9.

The review-count update is now dropped rather than awaited, so a rejection from it cannot fall into the shift-failure handler and log a completed swap as failed:

```ts
dispatch(updateSwapCount()).catch(() => {})
```

`updateSwapCount` reads review-trigger data off the account, which is IO that can reject, and it is not part of the swap. The ramp flow already drops its own count update the same way in `src/plugins/gui/fiatPlugin.tsx`. `Exchange_Shift_Success` keeps its position ahead of it, so the promo release still happens before anything that can wait on a review prompt.

src/actions/DeepLinkingActions.tsx
```diff
@@ -145,24 +172,135 @@ export function getDeepLinkReadiness(link: DeepLink): DeepLinkReadiness {
...
     const state = getState()
-    await handleLink(navigation, dispatch, state, link)
+
+    // An exchange link that names no asset opens no picker of its own, so it
+    // never claims the slot above, but it does set the promo and the ramp
+    // params that a picker already on screen is about to write. Drop it while
+    // one is up, for the same reason the waiting path drops its own.
+    if (isExchangeLink && walletLinkAccount === state.core.account) return false
+
+    return await handleLink(navigation, dispatch, state, link)
```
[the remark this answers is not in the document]
j0ntz: Fixed in 0fb32a201. An exchange link that names no asset now bumps `latestWalletLink` when it navigates, so an exchange link still waiting for wallets returns `false` when the wait ends instead of opening its picker and overwriting the newer link. A regression test in `LinkPromoActions.test.ts` covers the interleaving and fails without the change.

src/components/scenes/RampCreateScene.tsx
```diff
@@ -30,6 +31,7 @@ import {
   type SupportedPluginResult,
   useSupportedPlugins
 } from '../../hooks/useSupportedPlugins'
+import { useTabBlur } from '../../hooks/useTabBlur'
```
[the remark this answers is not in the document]
j0ntz: Fixed in 0fb32a201. `RampCreateScene` now clears the sell amount and any pending max whenever the selected asset changes, keyed off the selection itself rather than the dropdown, so a forced wallet a deep link sets on a mounted scene (and the tab switch that drops it) resets the amount the same way a dropdown pick does.

peachbits (approved):
Approved. All three asks from the last round are fixed in the code: the tab-state guard in the new `useTabBlur` (shared by the promo release and the ramp pins), the exchange-only scoping of the link drop, and the swap params branch along with its doc row. I re-ran the suite at this head: `tsc` is clean and jest passes 775 tests; the 5 snapshot failures are the asset-path artifact that fails on develop too.

One non-blocking note: the design doc's ten code links pin `79464bb4`, which GitHub no longer has, because the re-squash orphaned it the same way `7f55f8a9` went. Any squash orphans a branch commit, so linking `develop` as the glossary does, or re-pinning right before merge, is what survives.

j0ntz:
Repinned the design doc's ten code links to `develop`, matching the glossary rows, so a future squash cannot orphan them.

